PR #36, opened 2014-12-25
Trim backslash from commands instead of adding space

akoeplinger:
This solves a problem on Mono where the path would be created as
specified (i.e. including the spaces).

The spaces are only necessary to avoid an ending backslash escaping
the quote char on Windows. Trimming the backslash has the same effect.


weshaggard:
LTGM

weshaggard:
@dotnet-bot okay to test

FiveTimesTheFun:
LGTM - thank you.

Petermarcu:
@dotnet-bot add to whitelist

PR #38, opened 2014-12-30
Skip xunit.runner.tdnet.dll when resolving NuGet package references

akoeplinger:
The assembly isn't strong name signed and causes a compile error when it is referenced from a signed assembly (such as those built in corefx) with the Mono mcs compiler.

csc.exe seems to ignore this as the reference isn't actually used anywhere.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/ResolveNuGetPackages.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/ResolveNuGetPackages.cs
@@ -227,6 +227,14 @@ namespace Microsoft.Build.Tasks
 
                             case WellKnownAssetType.ReferenceOnly:
                                 {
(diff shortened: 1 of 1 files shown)


weshaggard:
@akoeplinger Is this change still needed? If so @FiveTimesTheFun can you please investigate further? I would prefer we not hack this in the ResolveNugetPackages task.

FiveTimesTheFun:
@dotnet-bot okay to test

akoeplinger:
I just checked the new "xunit.core.netcore" nupkg from @FiveTimesTheFun PR https://github.com/dotnet/corefx/pull/358 and it doesn't contain the problematic xunit.runner.tdnet.dll anymore, so I think this shouldn't be needed anymore once that PR is in.

PR #39, opened 2014-12-30
Run OpenSourceSign target after Compile target

akoeplinger:
It used to run before 'CopyFilesToOutputDirectory', but that target has a different name on Mono's xbuild (namely 'DeployOutputFiles'), so it doesn't run there.

Using AfterTargets="Compile" shouldn't make a difference and works on Mono as well.

Note: I'm not that well versed in the intricacies of MSBuild targets, so if this causes problems on Windows please let me know.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
@@ -11,7 +11,7 @@
   </PropertyGroup>
   
   <Target Name="OpenSourceSign" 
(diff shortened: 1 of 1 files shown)


weshaggard:
@nguerrera Are you aware of any reason we cannot do the OpenSourceSign task after Compile?

weshaggard:
@dotnet-bot okay to test

nguerrera:
LGTM. I'm not aware of any issue with that.

weshaggard:
Thanks @akoeplinger.

PR #45, opened 2015-01-12
Official builds in VSO override the $(OutDir) value so the build agent k...

jhendrixMSFT:
Unblock official builds in VSO.

Official builds in VSO override the $(OutDir) value so the build agent
knows where to find the build output. The FindTestDirectories target
uses $(BaseOutputPathWithConfig) instead of $(OutDir), so when $(OutDir)
is overridden the test binaries aren't found and the build fails.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -9,11 +9,11 @@
 
   <Target Name="FindTestDirectories">
     <ItemGroup>
-      <AllTestAssemblies Include="$(BaseOutputPathWithConfig)*Tests\*Tests.dll" />
+      <AllTestAssemblies Include="$(OutDir)*Tests\*Tests.dll" />
       <AllTestAssembliesWithParent Include="@(AllTestAssemblies)">
         <ParentDir>$([System.IO.Path]::GetDirectoryName(%(AllTestAssemblies.Identity)))</ParentDir>
       </AllTestAssembliesWithParent>
-      <TestDirectories Include="@(AllTestAssembliesWithParent -> '%(ParentDir)')" Exclude="@(_SkipTestAssemblies -> '$(BaseOutputPathWithConfig)%(Identity)')" />
+      <TestDirectories Include="@(AllTestAssembliesWithParent -> '%(ParentDir)')" Exclude="@(_SkipTestAssemblies -> '$(OutDir)%(Identity)')" />
     </ItemGroup>
   </Target>
 


nguerrera:
Looks good to me. 

In the future, please format commit messages as described here: 
https://github.com/dotnet/corefx/wiki/Contributing#commits

jhendrixMSFT:
Thanks for the feedback, I'll update the commit message.

weshaggard:
Unless I'm missing something this would break the running the tests as I don't think OutDir is set. Can you please verify that all the tests aren't just skipped now? Also I believe this will conflict with @FiveTimesTheFun change he is making (or made not sure if it was committed yet).

weshaggard:
Also as another question is OutDir the only option for this? I think we generally have an OutputRootPath that is more appropriate to set so people can build on that.

terrajobst:
Public MSBuild keys off everything of `OutDir`. So if we use other variables we should probably base them on the value passed in for `OutDir`.

weshaggard:
The problem with using OutDir directly is that it doesn't take into consideration all the different project names and such. OutDir ends up getting set correctly via OutputPath which we set using OutputRootPath appended with different things like the flavor and project name. We need the flexibility to set this differently per project (i.e. bin\debug\System.Collections.Immutable) and not have it set to bin\debug for all the projects.

jhendrixMSFT:
The tests are executed (I verified this yesterday before sending the PR).  In this context $(OutDir) is explicitly set in src\dirs.proj to $(BaseOutputPathWithConfig) before importing this target so the net effect is the same (note there will be a slight tweak to this as well in a future PR).

jhendrixMSFT:
Also, from combing through build logs yesterday I see that $(OutDir) is properly set per project, i.e. it isn't bin\debug for everything.

weshaggard:
@jhendrixMSFT Can you help me understand how that works if you are explicitly setting it via a property being passed to build.cmd? Perhaps what I'm missing is how OutDir is getting set.

jhendrixMSFT:
By default $(OutDir) is set to $(OutputPath) in Microsoft.Common.CurrentVersion.targets and we build the correct $(OutputPath) per project in our top-level dir.props file.  src\dirs.proj is a special case as it defines $(OutDir) explicitly to $(BaseOutputPathWithConfig) (packages.targets needs the root output directory); this breaks the VSO build scenario as $(OutDir) is specified on the command line (I have a fix ready for this to conditionally set $(OutDir) but the first step is to get tests.targets updated).

Also FWIW packages.targets correctly keys on $(OutDir) so making this change keeps our targets consistent.

FiveTimesTheFun:
Hi @jhendrixMSFT , it looks like this does conflict with my change.  Since 0db3add1d605df33ca4e0ca929062c84ef26fb4b we no longer used  $(BaseOutputPathWithConfig), but instead $(TestWorkingDir) which gets set in dir.props like this:

  <PropertyGroup>
    <ProjectDir>$(MSBuildThisFileDirectory)</ProjectDir>
    <SourceDir>$(ProjectDir)src\</SourceDir>
    <BinDir>$(ProjectDir)bin\</BinDir>
    <ToolsDir>$(BinDir)tools\</ToolsDir>
    <TestWorkingDir>$(BinDir)tests\</TestWorkingDir>
    <PackagesDir>$(SourceDir)packages\</PackagesDir>
  </PropertyGroup>


Here's the conflict:

<<<<<<< HEAD
      <TestAssemblies Include="$(TestWorkingDir)*.Tests\**\*.Tests.dll" Exclude="@(SkipTestAssemblies)" />
      <TestDirectories Include="@(TestAssemblies -> '%(RootDir)%(Directory)')" KeepDuplicates="false" />
=======
      <AllTestAssemblies Include="$(OutDir)*Tests\*Tests.dll" />
      <AllTestAssembliesWithParent Include="@(AllTestAssemblies)">
        <ParentDir>$([System.IO.Path]::GetDirectoryName(%(AllTestAssemblies.Identity)))</ParentDir>
      </AllTestAssembliesWithParent>
      <TestDirectories Include="@(AllTestAssembliesWithParent -> '%(ParentDir)')" Exclude="@(_SkipTestAssemblies -> '$(OutDir)%(Identity)')" />
>>>>>>> 0faaa202d1a2cb524e580c1a9f29955c50126ff6

jhendrixMSFT:
I see.  I'm going to close this PR and will look at the new changes.

PR #46, opened 2015-01-13
Enabling running a subset of tests through build

Priya91:


weshaggard:
How do you expect people to call this? Can you provide the command line you expect people to use?

Priya91:
build.cmd /p:IncludeTestAssemblies="Microsoft.Win32.Registry.Tests"

terrajobst:
Works for me. LGTM.

weshaggard:
What happen to the idea of having a RunTests.cmd that just takes the test assembly? Personally I think this would be useful because it also allows us to do other things like start the tests under a debugger and such without needing to pass a bunch more options that we have to remember to Build.cmd.

If we do still go with a property I also don't know that IncludeTestAssemblies really conveys what we are doing here. It seems like this is just additional tests to include as opposed to just the tests to run. Perhaps something like "TestsToRun" or something like that.

Priya91:
If we need to have a RunTests.cmd, this is what is expected:
1. Build all product assemblies
2. Build just the test assembly specified
3. RunTest

This is the same with the current build.cmd, and here only the input to build.cmd is changing, the runtests.cmd will have all the functionality equivalent to build, getting the correct msbuild version, downloading buildtools, verifying packages, etc.. 

So from what i see, the runtests.cmd will just be a wrapper over build.cmd, 

1. Runtests "Assembly1" "Assembly2"
2. In Runtests.cmd, parse the input and call build.cmd /p:TestToRun="Assembly1;Assembly2"

But this change is on the corefx repo, in test.targets from buildtools, this is all is needed to filter out the tests needed to run.

weshaggard:
Yes this is the buildtools part of the change which looks good to me. Thanks @Priya91 we can consider a runtests.cmd wrapper on the CoreFX if necessary. My guess is that RunTests.cmd would be a wrapper that we could use to iterate on the test runs without need to rebuild everything else so it would build the library and tests you need and that is it. I'm not sure if that is fully necessary however we can stick with just the build.cmd option for now.

weshaggard:
By the way you will need to update this slightly based on @FiveTimesTheFun recent changes.

PR #47, opened 2015-01-13
Correct package dependency version to xunit.run..dep..netcore

FiveTimesTheFun:
xunit.console.netcore project had an incorrect package dependency
version on xunit.runner.dependencies.netcore, which failed appvoyer.
Correct the version to 1.0.0-prerelease


weshaggard:
Assuming CI passes the change looks good to me.

FiveTimesTheFun:
Jenkins is still pending (appears down) but AppVeyor passes, so bringing this in.  Thanks @weshaggard .

PR #50, opened 2015-01-16
Clean up depending.targets file to make it more general purpose

weshaggard:
Renamed the target to make it more descriptive to RestoreAllPackageDependencies
while keeping the old EnsureDependencies target as a wrapper for now.

Added inputs/outputs to make incremental build work without requiring the
restore of all the packages on every build.

Added an empty PackageConfigsToExclude so people can override the wildcard Include
if needed.

Started outputing a message whenever the target starts as it tends to take a little
while and it is useful for people to know that restore is happening.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/depending.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/depending.targets
@@ -1,18 +1,32 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <Target Name="EnsureDependencies">
-    <ItemGroup>
-      <PackageConfigs Include="$(SourceDir)**\packages.config" />
-    </ItemGroup>
+  <PropertyGroup>
+    <RestoreAllPackageDependenciesSemaphoreFile>$(PackagesDir)\RestoreAllPackageDependencies.complete</RestoreAllPackageDependenciesSemaphoreFile>
+  </PropertyGroup>
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/depending.targets
```diff
@@ -1,18 +1,34 @@
...
+    <Error Condition="'$(PackagesDir)' == ''" Text="PackagesDir property needs to be set before calling RestoreAllPackageDependencies targets" />
 
     <!-- Restore Packages -->
     <Exec
-      Condition="'@(PackageConfigs)'!=''"
       StandardOutputImportance="Low"
       Command="&quot;$(NuGetToolPath)&quot; restore &quot;%(PackageConfigs.FullPath)&quot; -PackagesDirectory &quot;$(PackagesDir.TrimEnd('\'))&quot; $(NuGetConfigCommandLine)" 
       />
 
+    <Touch Files="$(RestoreAllPackageDependenciesSemaphoreFile)" AlwaysCreate="true" />
```
terrajobst: I still don't understand why we need this. I've seen this in other targets as well.

weshaggard: The reason for that semaphore file is to support incremental building. In some targets like this one there aren't any clear outputs that we can pick before hand so we have to make one, and the semaphore file is a way to handle that. We don't want msbuild to do any work here if the pacakge.config files aren't newer than the semaphore file so we skip the target all together.

terrajobst: I see, that makes sense.

PR #54, opened 2015-01-23
Add wait option to xunit.console.netcore

weshaggard:
This adds support for the -wait option to our xunit.console.netcore runner so that we can use that option while using CTRL+F5 in VS and see the results.


nguerrera:
There's a typo in the first commit message : "projecte" if you care to rewrite it away.

LGTM otherwise.

weshaggard:
Fixed the commit message typo thanks.

PR #56, opened 2015-01-29
Add script to bump package build numbers and use it

nguerrera:
We're getting in to a state where building is restoring 3 copies of every package. We should keep the build numbers bumped until we have either a wildcard (latest build) mechanism or we have RTM'ed most packages and most dependencies are stable.

This bumps *.config, *.nuspec, and *.csproj to 22512 in the buildtools repo.

A script is included to do this again next time and I will do the same in the corefx repo.

### DIFF
--- a/src/Scripts/set-package-build.cmd
+++ b/src/Scripts/set-package-build.cmd
@@ -1,0 +1,26 @@
+@echo off
+if "%1" == ""   goto :usage
+if "%2" == ""   goto :usage
+goto :main
+
+:usage
+echo Usage: %~n0 [directory] [build]
+echo.
+echo Recursively replaces all occurrences of 4.0.XX-beta-YYYYY in
+echo [directory]\*.config;*.csproj;*.nuspec with 4.0.XX-beta-[build]
+goto :eof
+
+:main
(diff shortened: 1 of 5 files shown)


nguerrera:
cc: @mellinoe @ericstj

ericstj:
LGTM

src/Scripts/set-package-build.cmd
```diff
@@ -0,0 +1,26 @@
...
+goto :eof
+
+:main
+pushd %1 && git clean -f packages && call :replace %2 && call :cleanup && popd
+goto :eof
+
+:replace
+for /r %%f in (*.config *.csproj *.nuspec) do (
+   perl -i.tmp_spb -p -e "s/(4\.0\.[0-9]+-beta)-([0-9]+)/$1-%1/" %%f || goto :eof
+)
```
weshaggard: I would have preferred powershell I guess I can live with perl :)

nguerrera: Pardon my cmd/perl. My powershell is weak and perl is oh so good at regex replace. :)

ericstj: Was going to say the same, but PS isn't friendly to our x-plat devs.

PR #57, opened 2015-01-30
Reference in-box v4 assemblies for msbuild v12 and v14 compat

nguerrera:
Fixes building in VS 2015 without the 2013 build tools installed.


davkean:
LGTM, but I'm going to raise this with MSBuild team, this seems broken behavior.

nguerrera:
Yes. It does seem strange.

PR #58, opened 2015-01-30
Fix build.cmd /p:Configuration=Release and use msbuild from path

nguerrera:


davkean:
LGTM.

dir.traversal.targets
```diff
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
-  <Import Project="$(MSBuildToolsPath)\Microsoft.Common.targets" />
-
```
weshaggard: This was fixed differently in the corefx repo (see https://github.com/dotnet/corefx/pull/406) should we instead fix it that way or do you think it is better to remove the Microsoft.Common.targets import?

nguerrera: Unless I'm missing something, I like what I have here better. I don't see the point of that import, and even less the point of moving things around to serve it.

weshaggard: I kind of like this approach as well, I don't actually remember why I ever did that import :) However I was just pointing out that it was different and thus inconsistent in the CoreFX repo, we could of course do the same fix over there.

PR #59, opened 2015-02-06
Feature work to support official daily builds from VSO.

jhendrixMSFT:
Add necessary metadata to enable Authenticode signing in VSO builds.

Nuget packages can now automatically be pushed to a nuget server
specified via MSBuild var PublishPackageSource.

Added a daily build number in version.prop to be used in file and
nuget packages.  By setting MSBuild var DailyBuildNumGenEnabled to
true when building the build will increment the value before building
and commit and push to origin after package publishing.

Updated MSBuild task GetPackageVersion to generate a nuget package
verison by combining the embedded package version in the nuspec file
with the daily build number.

Added a dependency on the build tools nuget package so this repo can
take advantage of this new functionality.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -1,18 +1,23 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Build Tools Version -->
+  <PropertyGroup>
+    <BuildToolsVersion>1.0.21-prerelease</BuildToolsVersion>
+  </PropertyGroup>
+
   <!-- Common repo directories -->
   <PropertyGroup>
     <ProjectDir>$(MSBuildThisFileDirectory)</ProjectDir>
     <SourceDir>$(ProjectDir)src\</SourceDir>
     <BinDir>$(ProjectDir)bin\</BinDir>
-    <ToolsDir>$(BinDir)tools\</ToolsDir>
     <TestWorkingDir>$(BinDir)tests\</TestWorkingDir>
     <PackagesDir>$(SourceDir)packages\</PackagesDir>
+    <ToolsDir>$(PackagesDir)Microsoft.DotNet.BuildTools.$(BuildToolsVersion)\lib\</ToolsDir>
   </PropertyGroup>
 
   <!-- Common nuget properties -->
   <PropertyGroup>
     <NuGetToolPath>$(ToolsDir)NuGet.exe</NuGetToolPath>
-    <NuGetConfigFile>$(SourceDir)nuget\NuGet.Config</NuGetConfigFile>
+    <NuGetConfigFile>$(SourceDir)NuGet.Config</NuGetConfigFile>
     <NuGetConfigCommandLine
       Condition="Exists('$(NuGetConfigFile)')">-ConfigFile &quot;$(NuGetConfigFile)&quot;</NuGetConfigCommandLine>
   </PropertyGroup>
--- a/src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
@@ -5,20 +5,19 @@ using LibGit2Sharp;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using System;
+using System.Diagnostics;
 using System.Linq;
+using System.Xml.Linq;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
     public class GetPackageVersion : Task
     {
         [Required]
-        public string RepositoryRoot { get; set; }
+        public string DailyBuildNumber { get; set; }
 
         [Required]
-        public string PackageName { get; set; }
-
-        [Required]
-        public bool RequireExplicitVersions { get; set; }
+        public string NuSpecFile { get; set; }
 
         [Output]
         public string PackageVersion { get; set; }
@@ -27,42 +26,19 @@ namespace Microsoft.DotNet.Build.Tasks
         {
             Log.LogMessage("Starting GetPackageVersion");
 
-            using (var repository = new Repository(RepositoryRoot))
-            {
-                string tipSha = repository.Head.Tip.Sha;
-                string prefix = PackageName + "-";
+            XDocument nuspecFile = XDocument.Load(NuSpecFile);
+            var ns = nuspecFile.Root.GetDefaultNamespace();
 
-                // Get tags that begin with the package name, on the tip
-                string[] matchingTags = repository.Tags.Where(t =>
-                        t.Target.Sha == tipSha &&
-                        t.Name.StartsWith(prefix)).Select(t => t.Name).ToArray();
+            var embeddedVerNode =
+                (from el in nuspecFile.Descendants(ns + "version")
+                select el).FirstOrDefault();
 
-                // Verify we didn't have multiple matching tags
-                Action<string> throwTagException = (message) => { throw new InvalidOperationException(string.Format(message, tipSha, prefix)); };
-                if (matchingTags.Count() > 1)
-                    throwTagException("Multiple tags found on commit {0} beginning with {1} (expected one)");
-
-                if (matchingTags.Count() == 1)
-                {
-                    PackageVersion = matchingTags[0].Substring(prefix.Length);
-                }
-                else
-                {
-                    if (RequireExplicitVersions == true)
-                        throwTagException("No tag found on commit {0} beginning with {1} (expected one)");
-
-                    PackageVersion = GetImplicitVersion(repository.Head.Tip.Sha);
-                }
-            }
+            Debug.Assert(embeddedVerNode != null, "embeddedVerNode shouldn't be null");
+            PackageVersion = string.Format("{0}-{1}", embeddedVerNode.Value, DailyBuildNumber);
 
             Log.LogMessage(string.Format("GetPackageVersion completed successfully - chose version {0}", PackageVersion));
 
             return true;
-        }
-
-        private string GetImplicitVersion(string sha)
-        {
-            return string.Format("999.999.999-sha{0}", sha.Substring(0, 10));
         }
     }
 }
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/UpdateBuildNumber.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/UpdateBuildNumber.targets
@@ -1,0 +1,41 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="GetNextDailyBuildNumber" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll" />
+
+  <Target Name="GetNextDailyBuildNumber"
+    BeforeTargets="Build"
+    Condition="'$(UpdateDailyBuildNumber)' == 'true'"
+    >
+    <GetNextDailyBuildNumber
+      VersionPropsFile="$(SourceDir)BuildNumber.props">
+      <Output PropertyName="DailyBuildNumber" TaskParameter="DailyBuildNumber" />
+    </GetNextDailyBuildNumber>
+  </Target>
+
+  <Target Name="CommitDailyBuildNumber"
+    AfterTargets="BuildPackages"
+    Condition="'$(UpdateDailyBuildNumber)' == 'true'"
+    >
+    <!-- configure the commit to show up as the dotnet bot -->
+    <Exec
+      WorkingDirectory="$(SourceDir)"
+      StandardOutputImportance="Low"
+      Command="git config user.name &quot;dotnet-bot&quot;" />
+
+    <Exec
+      WorkingDirectory="$(SourceDir)"
+      StandardOutputImportance="Low"
+      Command="git config user.email &quot;dotnet-bot@microsoft.com&quot;" />
+
+    <!-- commit and push to origin -->
+    <Exec
+      WorkingDirectory="$(SourceDir)"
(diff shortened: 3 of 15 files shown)


dir.props
```diff
@@ -1,18 +1,23 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Build Tools Version -->
+  <PropertyGroup>
+    <BuildToolsVersion>1.0.21-prerelease</BuildToolsVersion>
+  </PropertyGroup>
```
weshaggard: Just for clarity it may be worth adding a comment here that we need a version already on the nuget feed to build the buildtools repo and it cannot be the currently built one.

src/Microsoft.DotNet.Build.Tasks/GetDailyBuildNumber.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            XDocument versionProps = XDocument.Load(VersionPropsFile);
+            var defaultNS = versionProps.Root.GetDefaultNamespace();
+
+            var dailyBuildNode =
+                (from el in versionProps.Descendants(defaultNS + "DailyBuildNumber")
+                 select el).FirstOrDefault();
+
+            Debug.Assert(dailyBuildNode != null, "dailyBuildNode shouldn't be null");
+            var incDailyBuildNum = int.Parse(dailyBuildNode.Value) + 1;
```
weshaggard: Should we use int.TryParse to be more resilient to failure?

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
```diff
@@ -5,20 +5,19 @@
...
         [Required]
-        public string RepositoryRoot { get; set; }
+        public string DailyBuildNumber { get; set; }
 
         [Required]
-        public string PackageName { get; set; }
-
-        [Required]
-        public bool RequireExplicitVersions { get; set; }
+        public string NuspecFile { get; set; }
```
weshaggard: nit: NuSpecFile

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -37,12 +32,22 @@
       Condition="'@(PackagesNuSpecFiles)'!=''"
       IgnoreStandardErrorWarningFormat="true"
       StandardOutputImportance="Low"
-      Command="&quot;$(NuGetToolPath)&quot; pack &quot;%(PackagesNuSpecFiles.FullPath)&quot; -BasePath &quot;$(PackagesBasePath.TrimEnd('\'))&quot; -OutputDirectory &quot;$(PackagesOutDir.TrimEnd('\'))&quot; $(VersionCmdLine)" />
+      Command="&quot;$(NuGetToolPath)&quot; pack &quot;%(PackagesNuSpecFiles.FullPath)&quot; -BasePath &quot;$(PackagesBasePath.TrimEnd('\'))&quot; -OutputDirectory &quot;$(PackagesOutDir.TrimEnd('\'))&quot; -Version %(PackagesNuSpecFiles.PackageVersion)" />
```
weshaggard: This line is getting way too long. Would you mind splitting it up into a set of Property concats?

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
```diff
@@ -27,42 +26,19 @@ public override bool Execute()
         {
             Log.LogMessage("Starting GetPackageVersion");
 
-            using (var repository = new Repository(RepositoryRoot))
```
weshaggard: With this change I assume we can remove the dependency on the git library now correct?

src/version.props
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
```
weshaggard: I think I would call this file BuildNumber.Props to be more specific on what it actually contains.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -0,0 +1,43 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="GetDailyBuildNumber"
```
weshaggard: At some point we will have a versioning.targets that will generate the AssemblyVersion.cs file. Do you see this information going into that file (I can see both directions)? If not then can you name it for its purpose, which is to generate the build number?

src/dir.targets
```diff
@@ -1,3 +1,9 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!-- metadata used for Authenticode signing in official VSO builds -->
+  <ItemGroup>
+    <FilesToSign Include="$(OutDir)$(TargetFileName)">
+      <Authenticode>Microsoft</Authenticode>
```
weshaggard: Should this include the more general $(TargetPath)?

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -0,0 +1,43 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="GetDailyBuildNumber" 
+    AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"
+    Condition="Exists('$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll')" />
+
+  <Target Name="GetDailyBuildNumber"
+    BeforeTargets="Build"
+    Condition="'$(DailyBuildNumGenEnabled)' == 'true'"
+    >
+    <GetDailyBuildNumber
```
weshaggard: Just for clarity I think it would be worth calling this Target GetNextDailyBuildNumber.

src/Microsoft.DotNet.Build.Tasks/GetDailyBuildNumber.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Xml.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetDailyBuildNumber : Task
```
weshaggard: Probably worth calling this GetNextDailyBuildNumber.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -0,0 +1,43 @@
...
+      VersionPropsFile="$(SourceDir)version.props">
+      <Output PropertyName="DailyBuildNumber" TaskParameter="DailyBuildNumber" />
+    </GetDailyBuildNumber>
+  </Target>
+
+  <Target Name="CommitDailyBuildNumber"
+    AfterTargets="BuildPackages"
+    Condition="'$(DailyBuildNumGenEnabled)' == 'true'"
+    >
```
weshaggard: Do you expect DailyBuildNumGenEnabled to only be set via the official build? Perhaps we can call it something better like UpdateDailyBuildNumber or something that indicates that the number will be incremented.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -0,0 +1,43 @@
...
+    <!-- configure the commit to show up as the dotnet bot -->
+    <Exec
+      WorkingDirectory="$(SourceDir)"
```
weshaggard: It may be worth just using the git APIs in the build task to actually do this work instead of doing multiple Exec tasks in a  row.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
weshaggard: Now that I see its usage I don't think this would be the same file that would generate the AssemblyVersion.cs so I think we should probably rename the file to something more specific to its task. Perhaps UpdateBuildNumber.targets  or something similar.

src/Microsoft.DotNet.Build.Tasks/GetDailyBuildNumber.cs
jhendrixMSFT: Good idea, I'll make this change.

src/Microsoft.DotNet.Build.Tasks/GetDailyBuildNumber.cs
jhendrixMSFT: If the data is malformed how do you suggest we recover?  I'd prefer to fail-fast in the event of malformed data so people know it's broken.  I will add a comment to the props file indicating that the value must be numeric.

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
jhendrixMSFT: Fixed.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
jhendrixMSFT: I thought about that however I feel that it obscures the implementation and makes it a bit more difficult to change (this implementation is completely transparent).  That said if there are other advantages to using the git API over this method that I'm missing I'm all for changing it (one thing that comes to mind is retry logic if the push fails, hmm...).

src/version.props
jhendrixMSFT: Good idea, fixed.

dir.props
jhendrixMSFT: This is just a place-holder value.  My plan was to update this after the changes have been checked in and a new package uploaded to the feed.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
jhendrixMSFT: I like it, fixed.

jhendrixMSFT: And to answer your question, yeah I envision that this would be set via the official build but as it is right now there's nothing to prevent anybody from setting it.  We might want to guard against such a case, while not hard to overcome it would prevent some from publishing a package by accident (I don't want to gate getting this checked in on this issue though).

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
jhendrixMSFT: I've trimmed it down a bit.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
jhendrixMSFT: My original thinking was to also perform the updates to AssemblyVersion.cs from this targets file however I didn't flush it out very far.  I will rename this per your suggestion for now, we can always rename it later as the need arises.

src/dir.targets
jhendrixMSFT: Good catch, fixed.

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
jhendrixMSFT: We can but I suspect we're going to need it in the future, or if we want to use it for committing and pushing the build number changes.

src/Microsoft.DotNet.Build.Tasks/Targets/UpdateBuildNumber.targets
```diff
@@ -0,0 +1,43 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="GetNextDailyBuildNumber"
+    AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"
+    Condition="Exists('$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll')" />
```
weshaggard: Are there any cases where this will not exist?

src/dir.targets
```diff
@@ -1,3 +1,9 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!-- metadata used for Authenticode signing in official VSO builds -->
+  <ItemGroup>
```
weshaggard: Should this be conditioned? Or do we expect it is OK for it to always be set?

weshaggard:
Overall LGTM I added a couple more minor comments.

src/Microsoft.DotNet.Build.Tasks/Targets/UpdateBuildNumber.targets
jhendrixMSFT: No, this is from a copy/paste.  I'll update this to be inline with the other targets files in this project that use the build tasks dll.

src/dir.targets
jhendrixMSFT: I would prefer to always include it.  A little extra metadata that goes unused for dev builds shouldn't have any adverse affects.  Also, if for some reason this does become a problem we should have an early read instead of the VSO builds suddenly breaking.

PR #62, opened 2015-02-10
Add xml report generation support to xunit runner.

alexperovich:
This change adds the ability to create an xml report when running xunit.
The package changes are required because System.IO.FileStream has a dependency on System.Threading.ThreadPool


eatdrinksleepcode:
LGTM

weshaggard:
LGTM
cc @bartonjs as an FYI

PR #66, opened 2015-02-12
Split the Microsoft.DotNet.TestHost into 2 packages.

AlexGhiondea:
One that will match the package produced from CoreCLR and one that will contain the additional assemblies required but not produced from the CoreCLR repo


AlexGhiondea:
Note: This change should be merged only after @jhendrixMSFT change.

weshaggard:
@dotnet-bot test this please

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -2,7 +2,7 @@
 <package>
   <metadata minClientVersion="2.8.1">
     <id>Microsoft.DotNet.BuildTools</id>
-    <version>1.0.25-prerelease</version>
+    <version>1.0.26-prerelease</version>
```
weshaggard: Please revert this part of the change as it isn't needed any longer.

PR #69, opened 2015-02-12
Fix publishing of NuGet packages.

jhendrixMSFT:
The original testing performed was against a private feed which didn't
require an API key.  For public feeds we need to provide an API key.
I've added a new property SourceAPIKey where this is specified.


weshaggard:
Looks like our CI is still failing with the same issue you already fixed. It does work for me locally so perhaps there is something else wrong or we need to reset the CI somehow.

As for the change it looks good to me. I'm assuming that is one of the parameters you need to pass at the command line to publish an updated package, but I will way for those instructions offline.

jhendrixMSFT:
test this please

PR #71, opened 2015-02-12
Build log filename variable

jthelin-microsoft:
- Set name of build log file and build project file into env variable so
they are more easily discoverable in the log file, and only defined
once.
- Fix URL for .NET Developer Guide and minor typo in comments.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -3,21 +3,23 @@ setlocal
 
 :: Note: We've disabled node reuse because it causes file locking issues.
 ::       The issue is that we extend the build with our own targets which
-::       means that that rebuilding cannot successully delete the task
+::       means that that rebuilding cannot successfully delete the task
 ::       assembly. 
 
 :: Check prerequisites
 if not defined VS120COMNTOOLS (
     if not defined VS140COMNTOOLS (
         echo Error: build.cmd should be run from a Visual Studio 2013 or 2015 Command Prompt.  
-        echo        Please see https://github.com/dotnet/corefx/wiki/Developer%%20Guide for build instructions.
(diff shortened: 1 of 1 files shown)


weshaggard:
Thanks @jthelin-microsoft while the change looks good to me I'm curious what prompted you to make the env variable changes? How does this make them more discoverable in the log file? Don't you already need to know the log file to discover what is in it :)

jthelin-microsoft:
If your build is running on a remote build server somewhere in the cloud, you will often see the contents of the build log (for example Jenkins-CI web page) without opening the log file directly. 

If [or more likely when!] you have to log in to the build server to diagnose a tricky build failure, it is very useful to be able to quickly find that log file and associated build artifacts on disk.

This is especially true if your build server handles multiple build definitions from multiple projects, all with very similar but subtly different directory path locations, like our internal tfs servers currently do.

jthelin-microsoft:
And the "only defined once" bit is quite useful too ;)

weshaggard:
More interesting to me is the actual build command line which we are also putting into the log which is why we call the build target twice once with echo and once without. However I'm also perfectly fine with having them be in environment variables assuming your CI system or logs are capturing those. As for the duplication it isn't really bad in this build.cmd it is a little worse in the build.cmd in our corefx repo so feel free to submit a PR with an update for that one as well (assuming you are doing work with that repo as well, which I assume you are otherwise I'm not sure why you are using our build tools :))

jthelin-microsoft:
Yes, seeing the full exact command line that is being executed makes things so much easier too!

Equivalent change submitted to fxcore as well, as you suggested: https://github.com/dotnet/corefx/pull/762

PR #73, opened 2015-02-12
Remove the SourceAPIKey property.

jhendrixMSFT:
Providing the API key as a property exposes the key via the build
definition and build logs which is a security risk.  We will configure
the build machine with the appropriate API key instead.


FiveTimesTheFun:
Thanks @jhendrixMSFT , looks good to me.

weshaggard:
LGTM

PR #74, opened 2015-02-13
Centralize authoring for real signing.

jhendrixMSFT:
Moved metadata for real signing in VSO builds to sign.targets file;
this way all projects that reference build tools will automatically
have real signing enabled.

Added metadata to enable strong name signing with a new MSBuild
property UseOpenSourceSign to toggle between OSS and real signing.

Updated build tools dependency to latest version.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
@@ -1,6 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <!-- metadata used for Authenticode and strong-name signing in official VSO builds -->
+  <ItemGroup>
+    <FilesToSign Include="$(TargetPath)">
+      <Authenticode>Microsoft</Authenticode>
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
```diff
@@ -1,6 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <!-- metadata used for Authenticode and strong-name signing in official VSO builds -->
+  <ItemGroup>
+    <FilesToSign Include="$(TargetPath)">
+      <Authenticode>Microsoft</Authenticode>
+      <StrongName Condition="'$(SkipSigning)' != 'true'">StrongName</StrongName>
```
weshaggard: Given that you are redoing this should we change the condition to be more specific? $(SkipStrongNameSigning)? In case we need to have an opt out for the Authenticode as well.

weshaggard: Never mind I think keeping this as is fine after looking more at its usage.

weshaggard:
LGTM

PR #76, opened 2015-02-13
Add InvariantCulture to xml output.

alexperovich:
This change sets the IFormatProvider for number formatting in the xml output to CultureInfo.InvariantCulture

The test results were output with mixed british and regular decimal places. This resulted in multiplying durations by 1000 when the results were processed.

### DIFF
--- a/src/common/XmlTestExecutionVisitor.cs
+++ b/src/common/XmlTestExecutionVisitor.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Globalization;
 using System.Linq;
 using System.Xml.Linq;
 using Xunit.Abstractions;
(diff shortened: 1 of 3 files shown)


alexperovich:
CC: @FiveTimesTheFun

bartonjs:
LGTM.

I was concerned that the integer values (like the number of tests) was going to be affected by this, too; but the XAttribute constructor's conversion writes down all the primitive types in a way that is indistinguishable from invariant.   So only custom ToString calls are areas for concern.

src/common/XmlTestExecutionVisitor.cs
```diff
@@ -40,7 +41,7 @@ XElement CreateTestResultElement(ITestResultMessage testResult, string resultTex
                     new XAttribute("name", XmlEscape(testResult.Test.DisplayName)),
                     new XAttribute("type", testResult.TestCase.TestMethod.TestClass.Class.Name),
                     new XAttribute("method", testResult.TestCase.TestMethod.Method.Name),
-                    new XAttribute("time", testResult.ExecutionTime.ToString()),
+                    new XAttribute("time", testResult.ExecutionTime.ToString(CultureInfo.InvariantCulture)),
```
bartonjs: Actually, since they're not specifying a custom format here, maybe it would be better to just not call ToString() here at all... but maybe that'd look too inconsistent when compared to the other fixes.

XAttribute will format decimals using InvariantCulture.

FiveTimesTheFun:
Looks good to me too... thanks Alex :)

PR #78, opened 2015-02-16
Update daily build number to follow SemVer v1 rules.

jhendrixMSFT:
Per SemVer v1 the pre-release version number is sorted by lexicographic
order.  To ensure that build-10 is greater than build-2 we must pad the
build number with the appropriate number of preceeding zeroes.  In this
cas the value is an unsigned 16-bit value in base 10 making a total of
five characters (e.g. build-00010 > build-00002).

Renamed GetNextDailyBuildNumber to GetNextBuildNumber and updated all
references.


src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
```diff
@@ -34,7 +34,7 @@ public override bool Execute()
                 select el).FirstOrDefault();
 
             Debug.Assert(embeddedVerNode != null, "embeddedVerNode shouldn't be null");
-            PackageVersion = string.Format("{0}-{1}", embeddedVerNode.Value, DailyBuildNumber);
+            PackageVersion = string.Format("{0}-{1}", embeddedVerNode.Value, BuildNumber);
```
weshaggard: Does this work in the version in the nuspec is just a version without a text name? something like 1.0.0-00005 with that work? I know we hit the error when it was 0.0.0-4 not sure if that was just because of the 0 version or if it buildnumer requires a spec.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -61,7 +61,7 @@
 
     <GetPackageVersion
       Condition="Exists('$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll')"
-      DailyBuildNumber="$(DailyBuildNumber)"
+      BuildNumber="$(BuildNumber)"
```
weshaggard: Should we default this to something if it isn't set? To avoid errors when people try to use this without having set it?

weshaggard:
LGTM just had a couple minor questions/comments.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
jhendrixMSFT: I thought about this but I couldn't think of a scenario where this should legitimately not be set.  If there is I can certainly add a default value.

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
jhendrixMSFT: I just tried it, still fails.  According to the SemVer spec the value must be alphanumeric.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
weshaggard: What about repos that don't have a BuildNumber.props file setup? Is it going to be a new requirement that anyone using buildtools will need that file? If so that may be OK just want to be explicit about it. I guess they don't actually need the props file they just need to define that value itself.

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
weshaggard: That seems like it is going to be a problem for shipping stable releases. How do people ship a version of a package with just 1.0.0? Currently GetPackageVersion is used for all the places that produce the packages and it seems like we need a way to specify whether or not it is a prerelease or not. Perhaps we need another property or something in the NuSpec that indicates that.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
jhendrixMSFT: Thinking about this a bit more, we can condition the GetNuGetPackageVersions target to run only if BuildNumber is defined.  I think this is a bit more clean than picking some arbitrary default value.

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
jhendrixMSFT: We can tell if this is a prerelease package by parsing the version string and if it isn't then omit the prerelease build number.

PR #80, opened 2015-02-17
A few fixes for handling of BuildNumber.

jhendrixMSFT:
If BuildNumber is not defined then skip target GetNuGetPackageVersions
and use the package version embedded in the NuSpec file.

For task GetPackageVersion only append the build number to prerelease
packages to comply with the SemVer v1.0 spec.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
@@ -7,6 +7,7 @@ using Microsoft.Build.Utilities;
 using System;
 using System.Diagnostics;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Xml.Linq;
 
 namespace Microsoft.DotNet.Build.Tasks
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
```diff
@@ -34,7 +35,13 @@ public override bool Execute()
                 select el).FirstOrDefault();
 
             Debug.Assert(embeddedVerNode != null, "embeddedVerNode shouldn't be null");
-            PackageVersion = string.Format("{0}-{1}", embeddedVerNode.Value, BuildNumber);
+
+            // if this is a prerelease version then append the build number to the end
+            Regex preReleaseMatch = new Regex("-[0-9A-Za-z]+$");
```
weshaggard: I assume you don't actually need to "0-9" as it should only be chars in this spot correct? While I think this solves our issue for now I wonder if we need to explicitly pass in the "prerelease" string or not instead of having that in the nuspec.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/GetPackageVersion.cs
jhendrixMSFT: Per the SemVer spec any prerelease identifier is allowed to be alphanumeric but you are correct that today our prerelease identifiers don't contain any numeric characters.  As for moving the prerelease identifier out of the NuSpec file, I think this makes sense as long as every package built from a repo is in the same bucket (i.e. they are either all prerelease or not).

weshaggard: Sounds good I do think we will need to either keep the value in the nuspec or in the nuproj if we start using those to build the packages.

PR #81, opened 2015-02-17
Create the infrastructure to allow testing changes across repos.

AlexGhiondea:
This is going to be driven by the localpublish.props file at the source root
of the repo. By default this file does not exist and it is not tracked by git.

You can create the file with contents similar to this to enable the repo that
is currently built from publishing to a local path. This enables the scenario
when one package has a dependency to the other and you want to test your
changes locally.

If the file is present the following are going to happen:
- The package folder in the target repo is going to be removed
- A packages.config file is going to be created for the package that we are
installing. This is required to bypass the resolving of dependencies that is
triggered by installing a packaged referenced by name.
- A nuget install command is issued to install the package in the target repo

Structure of localpublish.props file:
<Project ToolsVersion="12.0" DefaultTargets="Build"
         xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <Packages Include="<directory of the package created by the repo>">
      <PackageName><name of the package></PackageName>
      <PackageVersion><version, including -prerelease></PackageVersion>
      <InstallLocation><Location in the other repo where the package
      					should be installed</InstallLocation>
    </Packages>
</Project>

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
@@ -18,6 +18,10 @@
     <NuGetOutputDirectory>&quot;$(PackagesOutDir.TrimEnd('\'))&quot;</NuGetOutputDirectory>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <LocalPublishPropsFile Condition="'$(LocalPublishPropsFile)'==''">$(PackagesDir)\localpublish.props</LocalPublishPropsFile>
+  </PropertyGroup>
+
   <Target Name="BuildPackages"
     DependsOnTargets="GetNuGetPackageVersions"
     Condition="'$(SkipBuildPackages)' != 'true'">
@@ -75,5 +79,45 @@
     <PropertyGroup>
       <_TempPackageVersion />
     </PropertyGroup>
+  </Target>
+
+  <Import Condition="Exists('$(LocalPublishPropsFile)')"  Project="$(LocalPublishPropsFile)"/>
+  
+  <!-- Publish the packages locally -->
+  <Target Name="LocalPackagePublish" Condition="'@(LocalPackages)'!=''"
+          Inputs="%(LocalPackages.Identity)\%(LocalPackages.PackageName).%(LocalPackages.PackageVersion).nupkg;%(LocalPackages.InstallLocation)\%(LocalPackages.PackageName).%(LocalPackages.PackageVersion)" 
+          Outputs="%(LocalPackages.InstallLocation)\%(LocalPackages.PackageName).%(LocalPackages.PackageVersion)\MODIFIED.txt" >
+    
+    <!-- Generate the packages.config -->
+    <ItemGroup>
+      <LocalPackageConfigLine Include="&lt;?xml version=&quot;1.0&quot; encoding=&quot;utf-8&quot;?&gt;" />
+      <LocalPackageConfigLine Include="&lt;packages&gt;" />
+      <LocalPackageConfigLine Include="&lt;package id=&quot;%(LocalPackages.PackageName)&quot; version=&quot;%(LocalPackages.PackageVersion)&quot; /&gt;"/>
+      <LocalPackageConfigLine Include="&lt;/packages&gt;" />
+    </ItemGroup>
+
+    <PropertyGroup>
+      <RestoreConfig>%(LocalPackages.InstallLocation)\packages.%(LocalPackages.PackageName).config</RestoreConfig>
+    </PropertyGroup>
+    
+    <WriteLinesToFile File="$(RestoreConfig)" Lines="@(LocalPackageConfigLine)" Overwrite="true" />
+    <Message Text="Generated $(RestoreConfig) for '%(LocalPackages.PackageName).%(LocalPackages.PackageVersion).nupkg'" />
+    
+    <PropertyGroup>
+      <LocalPackagesSources>@(LocalPackages, ';')</LocalPackagesSources>
+    </PropertyGroup>
+    <Message Text="Generated to package source to be '$(LocalPackages)'"/>
+
+    <!-- Remove existing directories -->
+    <RemoveDir Directories="%(LocalPackages.InstallLocation)\%(LocalPackages.PackageName).%(LocalPackages.PackageVersion)" />
+
+    <!-- Restoring the packages -->
+    <Exec Command="$(NuGetExe) install $(RestoreConfig) -Source $(LocalPackagesSources) -NoCache -Prerelease -OutputDirectory %(LocalPackages.InstallLocation)" />
+
+    <!-- Write a file in the packages folder to let the user know this is not the original package -->
+    <ItemGroup>
+      <NotOriginalPackages Include="The package was restored from $(LocalPackagesSources)" />
+    </ItemGroup>
+    <WriteLinesToFile File="%(LocalPackages.InstallLocation)\%(LocalPackages.PackageName).%(LocalPackages.PackageVersion)\MODIFIED.txt" Lines="@(NotOriginalPackages)" Overwrite="true" />
   </Target>
 </Project>


src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +76,41 @@
...
+  
+  <!-- Publish the packages locally -->
+  <Target Name="LocalPackagePublish" Condition="'@(Packages)'!=''"
+          Inputs="%(Packages.Identity)\%(Packages.PackageName).%(Packages.PackageVersion).nupkg;%(Packages.InstallLocation)\%(Packages.PackageName).%(Packages.PackageVersion)" 
+          Outputs="%(Packages.InstallLocation)\localpublish.%(Packages.PackageName).%(Packages.PackageVersion).txt" >
+    <!--AfterTargets="BuildNuGetPackages"-->
+    
+    <!-- Generate the packages.config -->
+    <ItemGroup>
+      <Line Include="&lt;?xml version=&quot;1.0&quot; encoding=&quot;utf-8&quot;?&gt;" />
```
weshaggard: You should make this more specific. Something like LocalPackageConfigLine

AlexGhiondea: Done.

AlexGhiondea:
I will rebase before I commit.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -18,6 +18,10 @@
     <NuGetOutputDirectory>&quot;$(PackagesOutDir.TrimEnd('\'))&quot;</NuGetOutputDirectory>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <LocalPublishPropsFile Condition="'$(LocalPublishPropsFile)'==''">$(SourceDir)\localpublish.props</LocalPublishPropsFile>
```
AlexGhiondea: If we don't want this to be $(SourceDir), what should it be? $(BinDir)? Something else?

weshaggard: $(PackagesDir) is what I would use in corefx. if that is not set in coreclr we can probably just set it there to be consistent.

.gitignore
```diff
@@ -14,4 +14,5 @@ packages
 Session.vim
 .netrwhist
 *~
+localpublish.props
```
weshaggard: Once you switch to something other than $(SourcesDir) you can remove this.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +80,47 @@
...
+    <WriteLinesToFile File="$(RestoreConfig)" Lines="@(LocalPackageConfigLine)" Overwrite="true" />
+    <Message Text="Generated $(RestoreConfig) for '%(Packages.PackageName).%(Packages.PackageVersion).nupkg'" />
+    
+    <PropertyGroup>
+      <PackageSources>@(Packages, ';')</PackageSources>
+    </PropertyGroup>
+    <Message Text="Generated to package source to be '$(PackageSources)'"/>
+
+    <!-- Remove existing directories -->
+    <RemoveDir Directories="%(Packages.InstallLocation)\%(Packages.PackageName).%(Packages.PackageVersion)" />
```
weshaggard: We may want to make this more general at some point perhaps adding a * at the end. However lets stick with this one for now.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +80,47 @@
...
+    <Exec Command="$(NuGetExe) install $(RestoreConfig) -Source $(PackageSources) -NoCache -Prerelease -OutputDirectory %(Packages.InstallLocation)" />
+
+    <!-- Write a file in the packages folder to let the user know this is not the original package -->
+    <ItemGroup>
+      <NotOriginalPackages Include="The package was restored from $(PackageSources)" />
+    </ItemGroup>
+    <WriteLinesToFile File="%(Packages.InstallLocation)\%(Packages.PackageName).%(Packages.PackageVersion)\MODIFIED.txt" Lines="@(NotOriginalPackages)" Overwrite="true" />
+
+
+    <Touch Files="%(Packages.InstallLocation)\localpublish.%(Packages.PackageName).%(Packages.PackageVersion).txt" AlwaysCreate="true" />
```
weshaggard: Why do we need to touch this file why not use the modified.txt instead?

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +80,47 @@
       <_TempPackageVersion />
     </PropertyGroup>
   </Target>
+
+  <Import Condition="Exists('$(LocalPublishPropsFile)')"  Project="$(LocalPublishPropsFile)"/>
+  
+  <!-- Publish the packages locally -->
+  <Target Name="LocalPackagePublish" Condition="'@(Packages)'!=''"
```
weshaggard: I actually think that Packages may conflict with other things we should consider a mores specific name. Perhaps LocalPackages.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
AlexGhiondea: I will make this $(PackagesDir) and will update the CoreCLR repo to have this variable

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
AlexGhiondea: Makes sense. Updated.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
AlexGhiondea: I added the modified.txt later -- I will get rid of the other file.

.gitignore
```diff
@@ -14,4 +14,3 @@ packages
 Session.vim
 .netrwhist
 *~
-
```
weshaggard: I assume you will undo the changes to this file when you squash.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +80,44 @@
...
+      <RestoreConfig>%(LocalPackages.InstallLocation)\packages.%(LocalPackages.PackageName).config</RestoreConfig>
+    </PropertyGroup>
+    
+    <WriteLinesToFile File="$(RestoreConfig)" Lines="@(LocalPackageConfigLine)" Overwrite="true" />
+    <Message Text="Generated $(RestoreConfig) for '%(LocalPackages.PackageName).%(LocalPackages.PackageVersion).nupkg'" />
+    
+    <PropertyGroup>
+      <PackageSources>@(LocalPackages, ';')</PackageSources>
+    </PropertyGroup>
+    <Message Text="Generated to package source to be '$(LocalPackagesources)'"/>
```
weshaggard: You updated a few places to LocalPackagesources, I assume from a bad find/replace but the property group above is still called PackageSources.

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +80,44 @@
...
+
+    <!-- Remove existing directories -->
+    <RemoveDir Directories="%(LocalPackages.InstallLocation)\%(LocalPackages.PackageName).%(LocalPackages.PackageVersion)" />
+
+    <!-- Restoring the packages -->
+    <Exec Command="$(NuGetExe) install $(RestoreConfig) -Source $(LocalPackagesources) -NoCache -Prerelease -OutputDirectory %(LocalPackages.InstallLocation)" />
```
weshaggard: Should be PackageSources

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +80,44 @@
...
+
+    <!-- Write a file in the packages folder to let the user know this is not the original package -->
+    <ItemGroup>
+      <NotOriginalPackages Include="The package was restored from $(LocalPackagesources)" />
```
weshaggard: Should be PackageSources

weshaggard:
@AlexGhiondea other than the last few comments and a squash the changes look good to me.

jhendrixMSFT:
test this please

src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -76,4 +80,44 @@
...
+      <RestoreConfig>%(LocalPackages.InstallLocation)\packages.%(LocalPackages.PackageName).config</RestoreConfig>
+    </PropertyGroup>
+    
+    <WriteLinesToFile File="$(RestoreConfig)" Lines="@(LocalPackageConfigLine)" Overwrite="true" />
+    <Message Text="Generated $(RestoreConfig) for '%(LocalPackages.PackageName).%(LocalPackages.PackageVersion).nupkg'" />
+    
+    <PropertyGroup>
+      <LocalPackagesSources>@(LocalPackages, ';')</LocalPackagesSources>
+    </PropertyGroup>
+    <Message Text="Generated to package source to be '$(LocalPackages)'"/>
```
weshaggard: Reading this message doesn't seem to make since any longer.

PR #84, opened 2015-02-19
Bump the version number for the CoreCLR package used

AlexGhiondea:
There is a new package for the CoreCLR package on the myget feed that matches the version that we produce from the CoreCLR repo.

This change updates the version of the CoreCLR package used by tests to the latest one available.


weshaggard:
LGTM

PR #85, opened 2015-02-19
Disable official signing for test projects.

jhendrixMSFT:
Real signing is time consuming so disable it for test projects.


weshaggard:
LGTM We need to ensure it gets set early enough before it is used.

jhendrixMSFT:
Yep I will fix this up on the corefx side as required.

PR #86, opened 2015-02-20
Use PackagesConfig in the packageresolve targets.

weshaggard:
While doing other work I noticed we use PackagesConfigFile and
PackagesConfig in various places. The most common is PackagesConfig
so switch the targets to use that consistently.


weshaggard:
@jhendrixMSFT I assume you will be producing a new build tools package soon with your other change can you please also review/merge this change to get it included in that update.

jhendrixMSFT:
New build is ready.

PR #87, opened 2015-02-20
Fix issues with BuildNumber property.

jhendrixMSFT:
The VSO official build defines global property BuildNumber which is
colliding with our property with the same name.  I have renamed the
property to RevisionNumber to avoid the collision.

Rename BuildNumber.props to the more generic BuildValues.props as it's
likely we'll have other build values in the in the future.


weshaggard:
LGTM. However I do wonder of BuildRevisionNumber or BuildMinorVersion would be better names, but I'm fine either way.

PR #89, opened 2015-02-21
Include the target files in the build tasks project for easier browsing/...

JeremyKuhne:
Simple change so the targets will show up in the Solution Explorer in VS.

### DIFF
--- a/src/BuildTools.sln
+++ b/src/BuildTools.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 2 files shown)


weshaggard:
Thanks @JeremyKuhne we originally did have these listed in VS but the list kept getting stale as we added more so I ended up removing them. Perhaps we can just wildcard them to avoid that issue and still have them easily available in VS.

JeremyKuhne:
Good idea. I'll mull it over and perhaps make that change later. If the IDE supported it better it would be an absolute no brainer. Unfortunately it always breaks the wildcard into single item entries if you add/remove within the IDE...

weshaggard:
While considering that we should also consider using wildcards to specify the list in the nuspec https://github.com/dotnet/buildtools/blob/master/src/nuget/Microsoft.DotNet.BuildTools.nuspec. The less explicit lists we have to maintain the less likely we are to cause issues.

JeremyKuhne:
Going to update to wildcards.

PR #93, opened 2015-02-25
Make value of PackagesOutDir consistent with corefx repo.

jhendrixMSFT:
The corefx repro ends PackagesOutDir with a trailing '\' character while
build tools does not.  In packages.targets property PackagesForPublishing
adds a backslash to PackagesOutDir which works for build tools but fails
for corefx.  This change removes the inconsistency.

Updated build tools package dependency to latest build with this fix.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
@@ -48,7 +48,7 @@
       Text="%(PackagesNuSpecFiles.Filename) NuGet Package -> $(PackagesOutDir)%(PackagesNuSpecFiles.Filename).%(PackagesNuSpecFiles.PackageVersion).nupkg" />
 
     <ItemGroup>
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/packages.targets
```diff
@@ -48,7 +48,7 @@
       Text="%(PackagesNuSpecFiles.Filename) NuGet Package -> $(PackagesOutDir)%(PackagesNuSpecFiles.Filename).%(PackagesNuSpecFiles.PackageVersion).nupkg" />
 
     <ItemGroup>
-      <PackagesForPublishing Include="$(PackagesOutDir)\*.nupkg" />
+      <PackagesForPublishing Include="$(PackagesOutDir)*.nupkg" />
     </ItemGroup>
```
weshaggard: One common approach to avoid these things is to just check for the trailing slash and if it doesn't exist add it.

weshaggard:
LGTM

PR #95, opened 2015-02-27
Include build files

JeremyKuhne:
For easier browsing. Submitted before, but now using wildcards and also including nuspec files. Didn't update nuspec file content to wildcards as they aren't consistent- will investigate further later.

### DIFF
--- a/src/BuildTools.sln
+++ b/src/BuildTools.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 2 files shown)


src/BuildTools.sln
```diff
@@ -12,7 +12,9 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "xunit.console.netcore", "xu
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".nuget", ".nuget", "{8B5A7D0F-2BD5-4892-9662-8FDC3589D938}"
 	ProjectSection(SolutionItems) = preProject
+		nuget\Microsoft.DotNet.BuildTools.nuspec = nuget\Microsoft.DotNet.BuildTools.nuspec
```
weshaggard: Do you know if adding these nuspec's cause them to build in a different way then what we are doing from the command line?

weshaggard:
@JeremyKuhne could you squash these commits?

PR #98, opened 2015-02-27
Intellisense hits the Compile target directly-- pushing the signing to b...

JeremyKuhne:
...efore deployment to the output directory avoids colliding with it.

Submitting this again so I could clean up the bad history state I got myself into.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
@@ -18,9 +18,13 @@
     <DefineConstants>$(DefineConstants);SIGNED</DefineConstants>
     <UseOpenSourceSign Condition="'$(UseOpenSourceSign)' == ''">true</UseOpenSourceSign>
   </PropertyGroup>
-  
-  <Target Name="OpenSourceSign" 
-          AfterTargets="Compile"
+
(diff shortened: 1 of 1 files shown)


nguerrera:
Thanks. LGTM.

nguerrera:
@JeremyKhune @akoeplinger /cc @weshaggard @tarekgh 

This breaks open-source signing. I'm reverting it.

msbuild is evaluating the PrepareForRunDependsOn before this BeforeTarget. 

What I had originally, before CopyFilesToOutputDirectory was most correct for Microsoft's build targets, but didn't work for Mono. I could prepend to $(PrepareForRunDependsOn) instead, but I don't know if that will work for Mono...

akoeplinger:
@nguerrera PrepareForRunDependsOn [seems to do the same thing](https://github.com/mono/mono/blob/6733010353a6024a2b437a6cd5c5c30ae6e99218/mcs/tools/xbuild/data/12.0/Microsoft.Common.targets#L463-L465) on Mono, so that should work.

tarekgh:
LGTM

JeremyKuhne:
@nguerrera PrepareForRunDependsOn sounds good.

weshaggard:
That seems reasonable to me as well.

PR #100, opened 2015-02-27
Support resources code generation

tarekgh:
This change is adding build Task to generate the resources code from resx files. also updating and creating targets files to automatically handle the resources without adding anything to the project files


tarekgh:
@weshaggard 
Wes, do I need to change the package version here too with my changes?

src/Microsoft.DotNet.Build.Tasks/Targets/ResourcesIncludes.targets
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <Compile Include="$(ResourceSRPath)">
+    <Visible>true</Visible>
```
weshaggard: nit:Indent level

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
```diff
@@ -1,59 +1,38 @@
...
-    <_ResIntermediateOutputPath>$(BaseIntermediateOutputPath)$(Configuration)\Res\</_ResIntermediateOutputPath>
-    <_ResPriIntermediateOutputFilePath>$(_ResIntermediateOutputPath)$(AssemblyName).pri</_ResPriIntermediateOutputFilePath>
-    <_ResJsonIntermediateOutputFilePath>$(_ResIntermediateOutputPath)$(AssemblyName).resources.resjson</_ResJsonIntermediateOutputFilePath>
+    <ResourcesSourceOutputDirectory Condition="'$(ResourcesSourceOutputDirectory)' == ''">$(MSBuildProjectDirectory)\Resources</ResourcesSourceOutputDirectory>
+    <StringResourcesPath Condition="'$(StringResourcesPath)'=='' And Exists('$(ResourcesSourceOutputDirectory)\Strings.resx')">$(ResourcesSourceOutputDirectory)\Strings.resx</StringResourcesPath>
+    <ResourceSRPath Condition="'$(ResourceSRPath)'==''">$(ResourcesSourceOutputDirectory)\SR.cs</ResourceSRPath>
+    <IntermediateResOutputFileFullPath Condition="'$(MSBuildProjectExtension)' == '.csproj'">$(IntermediateOutputPath)SR.cs</IntermediateResOutputFileFullPath>
+    <IntermediateResOutputFileFullPath Condition="'$(MSBuildProjectExtension)' == '.vbproj'">$(IntermediateOutputPath)SR.vb</IntermediateResOutputFileFullPath>
```
weshaggard: Why is this the only file that supports the .vb extension?

src/Microsoft.DotNet.Build.Tasks/Targets/ResourcesIncludes.targets
tarekgh: will fix that

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+            catch (Exception e)
+            {
+                Close();
+
+                Log.LogMessage(e.Message);
+
+                if (e is System.UnauthorizedAccessException)
+                {
+                    Log.LogMessage("The generated {0} file needs to be updated but the file is read-only. Please checkout this file so it can be updated.", OutputSourceFilePath);
```
weshaggard: The comment about checking out the file doesn't make sense in a git based world perhaps we just remove the second sentence.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -0,0 +1,239 @@
...
+                {
+                    _targetLanguage = TargetLanguage.VB;
+                }
+
+                _keys = new Dictionary<string, int>();
+                WriteClassHeader();
+                RunOnResFile();
+                WriteDebugCode();
+                WriteClassEnd();
+                Close();
```
weshaggard: Can you put these in a using statement instead of calling Close directly?

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -0,0 +1,239 @@
...
+
+                if (intermediateContent == srContent)
+                    return; // nothing need to get updated
+            }
+
+            File.WriteAllText(OutputSourceFilePath, intermediateContent);
+        }
+
+        private TargetLanguage _targetLanguage = TargetLanguage.CSharp;
+        private ResXResourceReader _resxReader;
```
weshaggard: Generally we put fields at the top of the class.

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
tarekgh: Actually you caught some issue which ResourceSRPath should check for cs or vb. so I need to add extra line there
but in general I think this is the only file we need to support VB. in VB projects we may manually need to add compile item to the SR.vb (I am trying to not create more extra targets file for VB as we don't have many VB projects)

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -0,0 +1,239 @@
...
+                _targetStream.Close();
+                _targetStream = null;
+            }
+        }
+
+        private void WriteClassHeader()
+        {
+            string commentPrefix = _targetLanguage == TargetLanguage.CSharp ? "// " : "' ";
+            _targetStream.WriteLine(commentPrefix + "This is auto generated file. Please don\u2019t modify manually.");
+            _targetStream.WriteLine(commentPrefix + "The file is generated as part of the build through the ResourceGenerator tool ");
```
weshaggard: This comment should be updated now with this model.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
FiveTimesTheFun: I also get UnauthorizedAccessException when a handle is opened by someone else with a restrictive sharing mode - I'd change the message to say it may be read only.

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
```diff
@@ -1,59 +1,38 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
FiveTimesTheFun: Don't think a targets file should have a DefaultTargets attribute.  Also is there a reason to change ToolsVersion?

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: will fix that

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: Wow, this will require touching all SR files when we update the buildtool with this change. I'll fix the comment

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
```diff
@@ -1,59 +1,38 @@
...
-        <ManifestResourceName>%(FileName)</ManifestResourceName>
-        <LogicalName>System.SR.resources</LogicalName>
-      </EmbeddedResource>
-    </ItemGroup>
-  </Target>
-
-</Project>
\ No newline at end of file
+  
+  <Import Project="ResourcesIncludes.targets" Condition="Exists('$(StringResourcesPath)') And Exists('$(ResourceSRPath)') And'$(ExcludeResourcesTarget)'==''"/>
```
FiveTimesTheFun: I see you just ported it from corefx\src\dir.targets but it's the only msbuild property named Exclude* I could find to suppress a target.  The rest are called Skip* (SkipBuildPackages, SkipSigning, SkipTests)

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
weshaggard: We may want to generalize this message a little to make have it make sense on both TFS and github.

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
weshaggard: Agreed and you should leave the ToolsVersion="12.0."

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: yes we still keeping the part saying the file is read only. also I am displaying the exception message which also helpful

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: here is what I have. please update/correct as you think:

"This is auto generated file. Please don't modify manually.
The file is generated as part of the build which takes the project resx resource file
and generated this source code file.
By default the file Resources\\Strings.resx will be used but projects can customize
that by overriding the StringResourcesPath property group.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: will do that

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
tarekgh: I'll remove the DefaultTargets attribute and reverted back the version to 12.0

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
tarekgh: Is there any guidelines for naming here? Skip make sense too but I am wondering if we have standard naming we are using

tarekgh: I'll call it SkipResourcesIncludesTarget as now the target we skip is called ResourcesIncludes

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
weshaggard: I think I would simplify it to:
>Do not edit this file manually it is auto-generated during the build based on the .resx file for this project.

weshaggard:
>Wes, do I need to change the package version here too with my changes? 
Don't update the package version as part of these changes.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: ok. will use that

src/Microsoft.DotNet.Build.Tasks/Targets/ResourcesIncludes.targets
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
akoeplinger: This should be ToolsVersion="12.0" as well like the rest of the targets.

tarekgh: ok fixed. thanks

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
akoeplinger: There's also a space missing in the condition `And'$(ExcludeResourcesTarget)'`, which might cause problems.

tarekgh:
Thanks all for your comments. I am going to merge the pull request

PR #102, opened 2015-03-04
Use Console.ResetColor() to restore color settings

jkotas:
Reading current foreground color is not supported on Unix

### DIFF
--- a/src/xunit.console.netcore/Program.cs
+++ b/src/xunit.console.netcore/Program.cs
@@ -18,8 +18,6 @@ namespace Xunit.ConsoleClient
         [STAThread]
         public static int Main(string[] args)
         {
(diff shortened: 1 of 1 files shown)


jkotas:
@FiveTimesTheFun @stephentoub  Could you please take a look?

src/xunit.console.netcore/Program.cs
```diff
@@ -86,7 +88,9 @@ public static int Main(string[] args)
             }
             finally
             {
+#if !NETCORE
```
stephentoub: Consider calling Console.ResetColor() here instead; that'll have the same effect, and is supported.

stephentoub:
One comment, otherwise LGTM.

FiveTimesTheFun:
Thank you - LGTM too.

PR #103, opened 2015-03-05
Fix the end of line mark in generated code

tarekgh:
This fix is fixing the issue of generating resource code lines
ended with consistent marks (CRLF on Windows and LF on Linux)
Also fixed the task to fail the build when there is any code
generation failure

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
@@ -38,8 +38,6 @@ namespace Microsoft.DotNet.Build.Tasks
 
         public override bool Execute()
         {
-            bool result = true;
-
             try
             {
                 using (_resxReader = new ResXResourceReader(ResxFilePath))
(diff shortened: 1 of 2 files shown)


tarekgh:
@weshaggard  could you have a look?

weshaggard:
I think you may need to also improve your diffing such that it works with different line endings. I don't think a default string compare will work.

tarekgh:
@weshaggard  do you want improve the comparison now or we do it as separate change?

weshaggard:
@tarekgh I think it maybe required now because I suspect the mix of TFS and git will mess around with the line endings and we may likely cause more breaks when mirroring to TFS.

tarekgh:
@weshaggard although I don't think this will cause a problem now as we have consistent end of line marks which should work in TFS and git. but anyway I'll look at changing the comparisons.

tarekgh:
@weshaggard please have a look when you have a chance. Thanks.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -36,10 +36,12 @@ public class GenerateResourcesCode : Task
 
         public bool DebugOnly { get; set; }
 
+        [Output]
+        public string Message { get; set; }
```
weshaggard: Why are you passing out the Message?

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
```diff
@@ -27,12 +27,21 @@
         ResxFilePath="$(StringResourcesPath)" 
         IntermediateFilePath="$(IntermediateResOutputFileFullPath)" 
         OutputSourceFilePath="$(ResourceSRPath)" 
-        AssemblyName="$(AssemblyName)" >
+        AssemblyName="$(AssemblyName)" 
+        ContinueOnError="ErrorAndStop" >
+        
+        <Output PropertyName="GenerateResourcesCodeOutputMessage" TaskParameter="Message" />
```
weshaggard: I wouldn't expect you to need to pick up the message as an out parameter now that you are calling Log.Error I would expect it to already do what you appear to be doing with that message.

weshaggard:
I think we should just let the task handle the message writing on error but the other parts looks good.

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
tarekgh: @weshaggard  Log.LogError didn't work at least when compiled on TFS side. I am not sure if we have Log.Error. when used Log.LogError I am still getting the behavior we have seen before that the build break with some other errors.

tarekgh: @weshaggard maybe we can look at it together on Monday.

weshaggard: We can walk through it on Monday but from my local testing it works as I expect and you shouldn't need anything but the Log.LogError.

tarekgh:
Thanks Wes for your comments. I have fixed all issues and I am going to merge the changes

PR #104, opened 2015-03-05
Add initial set of perf tools.

jhendrixMSFT:
This adds project and NuGet package infrastructure for bring up perf
tooling.  At present the tooling is split into a platform-specific
collection mechanism and a platform-agnostic model of the data used for
reporting purposes.

Created a new NuGet package Microsoft.DotNet.PerfTools to contain all of
the perf tools.

EventTracer
This is a console app used to start and stop collection of ETW data and
can optionally generate an XML report of the data collected.  Since it
relies on ETW it is applicable to Windows only.

PerfEventsData
This library contains a platform agnostic model of performance data.  At
present it only contains performance data types applicable to our
"consumptive" metrics.

### DIFF
--- a/src/ComparePerfEventsData/ComparePerfEventsData.csproj
+++ b/src/ComparePerfEventsData/ComparePerfEventsData.csproj
@@ -1,0 +1,46 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{580DB26D-AEB5-48AE-9FC2-467FF8B91D86}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>ComparePerfEventsData</RootNamespace>
+    <AssemblyName>ComparePerfEventsData</AssemblyName>
+    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <ResolveNuGetPackages>false</ResolveNuGetPackages>
+  </PropertyGroup>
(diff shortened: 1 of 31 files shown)


src/ComparePerfEventsData/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
weshaggard: Shouldn't this project use the common versioning.targets file from buildtools?

src/ComparePerfEventsData/ComparePerfEventsData.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <FileAlignment>512</FileAlignment>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <UseVSHostingProcess>false</UseVSHostingProcess>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="PowerArgs">
+      <HintPath>..\packages\PowerArgs.2.6.0.1\lib\net40\PowerArgs.dll</HintPath>
```
weshaggard: I assume with our ResolveNugetPackages this reference isn't actually needed correct?

weshaggard:
@jhendrixMSFT my comments would apply to all the project files in this initial check-in. What is the plan for using these?

src/ComparePerfEventsData/Properties/AssemblyInfo.cs
jhendrixMSFT: I've fixed up all of the AssemblyInfo.cs files for the new projects I've added.

src/ComparePerfEventsData/ComparePerfEventsData.csproj
jhendrixMSFT: The reference is required for the build to resolve the dependency.  I see this same pattern in other csproj files in this repo but not in corefx.  Apparently whatever mechanism is used to resolve references in corefx hasn't been moved to buildtools?

weshaggard: Yes I have another change that will address this once my current PR for the Build.Common.targets gets merged and we produce a new drop.

jhendrixMSFT:
@weshaggard Are the changes to Build.Common.targets ready and are there any other concerns?  I'd like to get this merged today.

weshaggard:
@jhendrixMSFT see PR https://github.com/dotnet/buildtools/pull/111 once that is merged then this repo will support the same model as corefx.

weshaggard:
I've merged #111 so please feel free to update this PR based on that and you can use the versioning and the packaging resolution if you want. If necessary  well. I t outat fne or the other if you need as well. I did that for some of the projects.

PR #107, opened 2015-03-08
Remove JSon2Txt and ResourceGenerator from BuildTools

weshaggard:
Json2Text and ResourceGenerator were for an old resource model which we no longer plan to use and so they aren't useful any longer so this commit deletes them.


weshaggard:
cc @tarekgh @FiveTimesTheFun

tarekgh:
LGTM. I was planning to do this work but Wes is faster :-)

PR #108, opened 2015-03-09
Bump xunit.console.netcore package version

alexperovich:


weshaggard:
As we chatted please update to 00015 instead.

Issue #112, opened 2015-03-12
Debugging tests broken in VS


ellismg:
a385cfb seems to have broken the ability to run Unit Tests inside VS.  The transformation for {XunitTraitOptions} doesn't seem to happen when loading the project inside VS, so the command line we pass when the test project is set as the startup project is incorrect and causes the test runner to exit with an error.

Do we need to do something specific here when running under VS?

weshaggard:
I wonder if this was broken with change https://github.com/dotnet/corefx/commit/85bd8348750dc65e55d625a2e20552d64fb076bc#diff-aad6833a46296c5e45aba8de08674fa0. However at any case you are correct the replacement piece is broken in VS. It becomes somewhat tricky to do the computation of the traits outside of the targets but we could make the VS arguments ignore the traits all together.

ellismg:
Very possible that it was broken before your refactoring, I haven't tried it in a few weeks.

PR #115, opened 2015-03-14
Fix VS command line arguments used for debugging

weshaggard:
We don't have a way to set the xunit traits for debugging in VS as they have to be computed via a target. We are just going to exclude the trait options while launching in VS by default.

This should fix https://github.com/dotnet/buildtools/issues/112, although I don't have a way to test right now so if anyone would like to test this that would be helpful.

cc @ellismg @nguerrera


nguerrera:
LGTM

Issue #92, opened 2015-02-25
Add @(AllItemsFullPathWithTargetPath) to TestCopyLocal list in publishtest.targets


nslottow:
@weshaggard Thanks for helping figure this out.

This will enable managed test projects that use project references to native projects to publish the native output to the test drop.  This would fix the issue I mentioned in [#1013](https://github.com/dotnet/corefx/pull/1013)

eatdrinksleepcode:
This can be closed (via #116).

weshaggard:
Yes thanks.

PR #116, opened 2015-03-15
Only run tests (and copy test files) if test results are out of date

### LINKED ISSUE #60: Support incremental test runs
(from @eatdrinksleepcode in dotnet/corefx#678)

Executing a build twice in a row should ideally be as close to a no-op as possible. In CoreFX, unit tests are always executed during the second run, even when the both the main and test project have not changed. This could be fixed by specifying the correct Inputs and Outputs to the RunTests task. I would be happy to submit a PR with this change to the BuildTools project.

eatdrinksleepcode:
Fixes #60 

Uses a marker file and Inputs/Outputs on the RunTestsForProject target to skip running tests when nothing has changed since the last successful test run. Also moves copying of the test files into the RunTestsForProject target so that they are only copied if the tests are going to run.

On my machine, this change cuts the build time for CoreFX for an incremental build with no changes from 45-55 seconds to 5 seconds*.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
@@ -37,7 +37,7 @@
       <PackagesConfigs Include="$(ProjectPackagesConfigFile)" />
       <PackagesConfigs Include="$(TestRuntimePackageConfig)" />
     </ItemGroup>
-    
+
     <ResolveNuGetPackages PackagesConfigs="@(PackagesConfigs)"
                           PackageRoot="$(PackagesDir)"
                           Platform="$(PlatformTarget)"
@@ -48,14 +48,15 @@
 
       <Output TaskParameter="ResolvedCopyLocal" ItemName="TestCopyLocal" />
     </ResolveNuGetPackages>
-    
+
     <ItemGroup>
       <TestCopyLocal Include="@(ReferenceCopyLocalPaths)" />
       <TestCopyLocal Include="@(Content)" />
       <TestCopyLocal Include="@(IntermediateAssembly)" />
       <TestCopyLocal Include="@(_DebugSymbolsIntermediatePath)" />
+      <TestCopyLocal Include="@(AllItemsFullPathWithTargetPath)" />
     </ItemGroup>
-    
+
     <Copy
       SourceFiles="@(TestCopyLocal)"
       DestinationFolder="$(TestPath)%(TestTargetFramework.Folder)"
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
@@ -23,20 +23,9 @@
   <PropertyGroup>
     <CopyTestToTestDirectory Condition="'$(CopyTestToTestDirectory)'==''">$(IsTestProject)</CopyTestToTestDirectory>
   </PropertyGroup>
-  
-  <ItemGroup>
-    <TestTargetFramework Include="ASP.NetCore, version=v5.0">
-      <Folder>aspnetcore50</Folder>
-    </TestTargetFramework>
-  </ItemGroup>
 
   <Target Name="CopyTestToTestDirectory"
     Condition="'$(CopyTestToTestDirectory)'=='true'">
-    
-    <ItemGroup>
-      <PackagesConfigs Include="$(ProjectPackagesConfigFile)" />
-      <PackagesConfigs Include="$(TestRuntimePackageConfig)" />
-    </ItemGroup>
 
     <ResolveNuGetPackages PackagesConfigs="@(PackagesConfigs)"
                           PackageRoot="$(PackagesDir)"
@@ -50,11 +39,7 @@
     </ResolveNuGetPackages>
 
     <ItemGroup>
-      <TestCopyLocal Include="@(ReferenceCopyLocalPaths)" />
-      <TestCopyLocal Include="@(Content)" />
-      <TestCopyLocal Include="@(IntermediateAssembly)" />
-      <TestCopyLocal Include="@(_DebugSymbolsIntermediatePath)" />
-      <TestCopyLocal Include="@(AllItemsFullPathWithTargetPath)" />
+      <TestCopyLocal Include="@(RunTestsForProjectInputs)" Exclude="@(PackagesConfigs)" />
     </ItemGroup>
 
     <Copy
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
@@ -21,11 +21,6 @@
   </Target>
 
   <PropertyGroup>
-    <PrepareForRunDependsOn>
-      $(PrepareForRunDependsOn);
-      CopyTestToTestDirectory;
-    </PrepareForRunDependsOn>
-
     <CopyTestToTestDirectory Condition="'$(CopyTestToTestDirectory)'==''">$(IsTestProject)</CopyTestToTestDirectory>
   </PropertyGroup>
   
@@ -36,8 +31,6 @@
   </ItemGroup>
 
   <Target Name="CopyTestToTestDirectory"
-    Inputs="@(PackagesConfig);$(TestRuntimePackageConfig);@(ReferenceCopyLocalPaths);@(Content);@(IntermediateAssembly);@(_DebugSymbolsIntermediatePath)"
-    Outputs="$(TestPath)%(TestTargetFramework.Folder)\*.*"
     Condition="'$(CopyTestToTestDirectory)'=='true'">
     
     <ItemGroup>
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -11,6 +11,8 @@
 
     <TestDisabled>false</TestDisabled>
     <TestDisabled Condition="'$(IsTestProject)'!='true' Or '$(SkipTests)'=='true' Or '$(RunTestsForProject)'=='false'">true</TestDisabled>
+
+    <TestsSuccessfulSemaphore>$(MSBuildProjectFile).tests.lastsucceeded</TestsSuccessfulSemaphore>
   </PropertyGroup>
 
   <ItemGroup>
@@ -50,15 +52,34 @@
     <StartArguments Condition="'$(StartArguments)'==''">$(VSStartArguments)</StartArguments>
     <DebugEngines>{2E36F1D4-B23C-435D-AB41-18E608940038}</DebugEngines>
   </PropertyGroup>
+  
+  <ItemGroup>
+    <TestTargetFramework Include="ASP.NetCore, version=v5.0">
+      <Folder>aspnetcore50</Folder>
+    </TestTargetFramework>
+  </ItemGroup>
 
-  <PropertyGroup>
-    <TestsSuccessfulSemaphore>$(MSBuildProjectFile).tests.lastsucceeded</TestsSuccessfulSemaphore>
-  </PropertyGroup>
+  <ItemGroup>
+    <PackagesConfigs Include="$(ProjectPackagesConfigFile)" />
+    <PackagesConfigs Include="$(TestRuntimePackageConfig)" />
+  </ItemGroup>
+
+  <Target Name="DiscoverTestInputs">
+    <ItemGroup>
+      <RunTestsForProjectInputs Include="@(PackagesConfigs)" />
+      <RunTestsForProjectInputs Include="@(ReferenceCopyLocalPaths)" />
+      <RunTestsForProjectInputs Include="@(Content)" />
+      <RunTestsForProjectInputs Include="@(IntermediateAssembly)" />
+      <RunTestsForProjectInputs Include="@(_DebugSymbolsIntermediatePath)" />
+      <RunTestsForProjectInputs Include="@(AllItemsFullPathWithTargetPath)" />
+    </ItemGroup>
+  </Target>
 
   <!-- On command line run unit tests on CoreCLR after the Test target -->
   <Target Name="RunTestsForProject"
+          DependsOnTargets="DiscoverTestInputs"
           AfterTargets="CheckTestCategories"
-          Inputs="@(PackagesConfig);$(TestRuntimePackageConfig);@(ReferenceCopyLocalPaths);@(Content);@(IntermediateAssembly);@(_DebugSymbolsIntermediatePath);@(AllItemsFullPathWithTargetPath)"
+          Inputs="@(RunTestsForProjectInputs)"
           Outputs="$(TestPath)%(TestTargetFramework.Folder)\$(TestsSuccessfulSemaphore);$(TestPath)%(TestTargetFramework.Folder)\$(XunitResultsFileName)"
           Condition="'$(TestDisabled)'!='true'">
 
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -58,7 +58,7 @@
   <!-- On command line run unit tests on CoreCLR after the Test target -->
   <Target Name="RunTestsForProject"
           AfterTargets="CheckTestCategories"
-          Inputs="@(PackagesConfig);$(TestRuntimePackageConfig);@(ReferenceCopyLocalPaths);@(Content);@(IntermediateAssembly);@(_DebugSymbolsIntermediatePath)"
+          Inputs="@(PackagesConfig);$(TestRuntimePackageConfig);@(ReferenceCopyLocalPaths);@(Content);@(IntermediateAssembly);@(_DebugSymbolsIntermediatePath);@(AllItemsFullPathWithTargetPath)"
           Outputs="$(TestPath)%(TestTargetFramework.Folder)\$(TestsSuccessfulSemaphore);$(TestPath)%(TestTargetFramework.Folder)\$(XunitResultsFileName)"
           Condition="'$(TestDisabled)'!='true'">
 
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -24,8 +24,9 @@
   <PropertyGroup>
     <XunitHost>corerun.exe</XunitHost>
     <TestHost>$(XunitHost)</TestHost>
+    <XunitResultsFileName>testResults.xml</XunitResultsFileName>
     <XunitCommandLine>xunit.console.netcore.exe $(TargetFileName)</XunitCommandLine>
-    <XunitOptions>$(XunitOptions) -xml .\testResults.xml</XunitOptions>
+    <XunitOptions>$(XunitOptions) -xml $(XunitResultsFileName)</XunitOptions>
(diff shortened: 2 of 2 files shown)


eatdrinksleepcode:
*In trying to determine how much time this change cut off the build for CoreFX, I was initially surprised to find that it actually _increased_ the build time; in fact, the build time doubled with this change. After running around in circles for a while, I eventually narrowed it down to the diagnostic output being written to the msbuild.log file. Reducing the log verbosity to normal reduces the build time from over 2 minutes down to 5 seconds. I will file an issue in CoreFX to reduce and/or create an option for controlling the verbosity.

However, the excessive verbosity would explain why the build time didn't go down; but it doesn't explain why the build time actually went _up_. My best guess is that the enormous build output in such a short time is overloading log batching in MSBuild; longer build times when running tests results in more sensibly sized log batches, resulting in no significant impact to the build time. I wonder if someone who is more familiar with the internals of MSBuild than I am can corroborate that theory.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -48,10 +49,20 @@
...
+  <ItemGroup>
+    <TestTargetFrameworkPath Include="$(TestPath)%(TestTargetFramework.Folder)"/>
+    <TestsSuccessfulSemaphore Include="@(TestTargetFrameworkPath->'%(FullPath)\$(MSBuildProjectFile).tests.lastsucceeded')"/>
+    <XunitResultsFile Include="@(TestTargetFrameworkPath->'%(FullPath)\$(XunitResultsFileName)')"/>
+  </ItemGroup>
+
   <!-- On command line run unit tests on CoreCLR after the Test target -->
   <Target Name="RunTestsForProject"
           AfterTargets="CheckTestCategories"
+          Inputs="@(PackagesConfig);$(TestRuntimePackageConfig);@(ReferenceCopyLocalPaths);@(Content);@(IntermediateAssembly);@(_DebugSymbolsIntermediatePath)"
```
weshaggard: Can you please capture these inputs into an item group so they can be shared between this and the Copy target? Perhaps capture these in the TestCopyLocal item group. While you are at this could you handle https://github.com/dotnet/buildtools/issues/92 as well?

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -48,10 +49,20 @@
...
+          Outputs="@(TestsSuccessfulSemaphore);@(XunitResultsFile)"
           Condition="'$(TestDisabled)'!='true'">
+		  
+    <CallTarget Targets="CopyTestToTestDirectory"/>
```
weshaggard: Why not just add a DependsOnTarget="CopyTestToTestDirectory"? Does that still force it to run even if the Inputs/Outputs are in-sync for this target?

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -48,10 +49,20 @@
     <DebugEngines>{2E36F1D4-B23C-435D-AB41-18E608940038}</DebugEngines>
   </PropertyGroup>
 
+  <ItemGroup>
+    <TestTargetFrameworkPath Include="$(TestPath)%(TestTargetFramework.Folder)"/>
+    <TestsSuccessfulSemaphore Include="@(TestTargetFrameworkPath->'%(FullPath)\$(MSBuildProjectFile).tests.lastsucceeded')"/>
```
weshaggard: Why isn't the test results file enough for a semaphore? Is it because it also exists when the a failure occurs?

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -36,8 +31,6 @@
   </ItemGroup>
 
   <Target Name="CopyTestToTestDirectory"
-    Inputs="@(PackagesConfig);$(TestRuntimePackageConfig);@(ReferenceCopyLocalPaths);@(Content);@(IntermediateAssembly);@(_DebugSymbolsIntermediatePath)"
```
weshaggard: I think it is probably worth actually using he same Inputs/Outputs the copy as you are doing for the Run.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -48,10 +49,20 @@
     <DebugEngines>{2E36F1D4-B23C-435D-AB41-18E608940038}</DebugEngines>
   </PropertyGroup>
 
+  <ItemGroup>
+    <TestTargetFrameworkPath Include="$(TestPath)%(TestTargetFramework.Folder)"/>
```
weshaggard: While right now we only have one TestTrargeFramework the goal is to have more in he future and by moving this outside the target you have broken the batching that as occurring in the target. You may need to actually do the file path building inline in the target (even if it is duplicated).

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -66,12 +77,13 @@
     <MakeDir Condition="'$(CoverageEnabledForProject)'=='true'" Directories="$(CoverageReportDir)" />
 
     <Exec Command="$(TestHost) $(TestCommandLine)"
-          WorkingDirectory="$(TestPath)%(TestTargetFramework.Folder)"
+          WorkingDirectory="@(TestTargetFrameworkPath)"
```
weshaggard: This "%" he is what made this target batch on all TestTargetFrameworks but by removing from the target it we are no longer batching.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -22,8 +22,9 @@
 
   <!-- General xunit options -->
   <PropertyGroup>
+    <XunitResultsFileName>testResults.xml</XunitResultsFileName>
     <XunitCommandLine>xunit.console.netcore.exe $(TargetFileName)</XunitCommandLine>
-    <XunitOptions>$(XunitOptions) -xml .\testResults.xml</XunitOptions>
+    <XunitOptions>$(XunitOptions) -xml $(XunitResultsFileName)</XunitOptions>
```
weshaggard: Based on your other work you made me realize that we are going to have a conflict with the testResults.xml file as well when we start batching. We are going to need to figure out a work around for that.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -66,12 +77,13 @@
...
     <Exec Command="$(TestHost) $(TestCommandLine)"
-          WorkingDirectory="$(TestPath)%(TestTargetFramework.Folder)"
+          WorkingDirectory="@(TestTargetFrameworkPath)"
           ContinueOnError="true">
       <Output PropertyName="TestRunExitCode" TaskParameter="ExitCode" />
     </Exec>
 
     <Error Condition="'$(TestRunExitCode)' != '0'" Text="One or more tests failed while running tests from '$(MSBuildProjectName)' please check log for details!" />
+    <Touch Condition="'$(TestRunExitCode)' == '0'" Files="@(TestsSuccessfulSemaphore)" AlwaysCreate="true" />
   </Target>
```
weshaggard: Perhaps we should just rename/move/copy the testresults.xml file to a testResults.failed.xml or testResults.succeeded.xml and use that as the semaphore.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
eatdrinksleepcode: What conflict will we have? If I understand correctly, batching will occur based on the target framework, which is also part of the test working directory.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
eatdrinksleepcode: Correct.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
weshaggard: Your correct as the tests will be in a different directory so they won't conflict.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
eatdrinksleepcode: Correct, dependencies are executed prior to evaluating the Inputs/Outputs of a target. This is true even if the dependency is expressed using Before/AfterTargets instead of DependsOnTargets. (At first this was annoying, but it makes sense: what if you wanted to intentionally affect the inputs of a task using a dependency?)

I am not generally a fan of using CallTarget, but it was the only reliable way I could find to achieve all of the objectives of #60. Another option would be to inline CopyTestToTestDirectory into RunTestsForProject, which is not crazy, but it would make the RunTestsForProject task quite large.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
eatdrinksleepcode: I will take another shot at this while maintaining the batching.

How would the multiple target frameworks be expressed? Multiple Folder nodes under the TestTargetFramework item? Or multiple TestTargetFramework items?

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
eatdrinksleepcode: We could do that. It would slightly complicate any tools that might want to access the file afterwards (saving at the end of a CI run for example).

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
eatdrinksleepcode: The problem with using Outputs on this task - and the reason I ultimately gave up on making this target independently incremental and invoked it from RunTestsForProject instead - is that MSBuild requires Outputs to be statically identified; it will not evaluate wildcards. The reasons are explained [here](http://blogs.msdn.com/b/msbuild/archive/2006/02/22/537150.aspx) (the reasons for not using wildcards are sound, but I would have preferred the MSBuild team to allow the developer to make that decision rather than enforcing it in the tool).

Outputs for the copy cannot easily be statically determined since they include packages references. The package resolution could be moved into its own task that runs prior to the copy, which would allow the outputs for the copy to be statically determined; and in fact I have a local commit that does just that. However, the result was increased complexity in the build files, as well as the running build (timestamps on the entire test directory now have to be compared), for what I felt was little value; executing the copy from inside RunTestsForProject felt like a cleaner solution.

If you would like, I can push my local commit to a branch so you can see what it would look like.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
weshaggard: Yes you are probably right and perhaps a specific semaphore files is still best.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
weshaggard: We haven't fully defined any others yet but I would expect there will be multiple TestTargetFramework items each with their own folder.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
weshaggard: For the Outputs here I was kind of thinking about a similar Semaphore file. However given that you are going to still use the CallTarget on this the inputs/outputs aren't really necessary.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
weshaggard: This seems reasonable to me, at least for now.

eatdrinksleepcode:
I restored the batching; the duplication is annoying, but apparently necessary. I didn't extract the inputs because they won't be re-used; let me know if you still want me to do that. I also implemented #92, and squashed and rebased.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -47,11 +48,19 @@
...
   </PropertyGroup>
 
+  <PropertyGroup>
+    <TestsSuccessfulSemaphore>$(MSBuildProjectFile).tests.lastsucceeded</TestsSuccessfulSemaphore>
+  </PropertyGroup>
+
   <!-- On command line run unit tests on CoreCLR after the Test target -->
   <Target Name="RunTestsForProject"
           AfterTargets="CheckTestCategories"
+          Inputs="@(PackagesConfig);$(TestRuntimePackageConfig);@(ReferenceCopyLocalPaths);@(Content);@(IntermediateAssembly);@(_DebugSymbolsIntermediatePath);@(AllItemsFullPathWithTargetPath)"
```
weshaggard: I would still like these inputs to be pulled out into their own item group as they are shared with the CopyTestToTestDirectory. They aren't going to be used an Inputs to that target but they are used as part of the copy step (i.e. TestCopyLocal item group) and I would perfer we only have one place to keep these sets in sync. For example when you added AllItemsFullPathWithTargetPath you add to add it in both places and if we add others in the future I suspect people will miss one or the other locations.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -47,11 +48,19 @@
...
+          Outputs="$(TestPath)%(TestTargetFramework.Folder)\$(TestsSuccessfulSemaphore);$(TestPath)%(TestTargetFramework.Folder)\$(XunitResultsFileName)"
           Condition="'$(TestDisabled)'!='true'">
 
+    <CallTarget Targets="CopyTestToTestDirectory"/>
```
weshaggard: At some future point we will need to figure out how to pass the batch parameters to CopyTestToTestDirectory as well but for now I'm fine with ignoring that.

weshaggard:
Other than sharing the TestCopyLocal item group the change looks good.

eatdrinksleepcode:
I extracted the inputs of RunTestsForProject into TestInputs, which is then used to seed TestCopyLocal in CopyTestToTestDirectory (removing the package configs and adding the results of resolving the packages). Because several of the input items do not have values until compilation is complete, TestInputs has to be created after compilation, but before the start of RunTestsForProject. It therefore required its own target.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -35,16 +30,25 @@
...
+  <ItemGroup>
+    <PackagesConfigs Include="$(ProjectPackagesConfigFile)" />
+    <PackagesConfigs Include="$(TestRuntimePackageConfig)" />
+  </ItemGroup>
+
+  <Target Name="DiscoverTestInputs" BeforeTargets="RunTestsForProject">
     <ItemGroup>
-      <PackagesConfigs Include="$(ProjectPackagesConfigFile)" />
-      <PackagesConfigs Include="$(TestRuntimePackageConfig)" />
+      <TestInputs Include="@(PackagesConfigs)" />
```
weshaggard: Two more comments\questions:
1) Is it worth putting this target with RunTestsForProject or just keeping it here? Just a question I'm not sure which I like better.
2) We should probably rename this to be more specific. Something like RunTestsForProjectTargetInputs to make sure it doesn't class with other things that maybe TestInputs.

eatdrinksleepcode: 1) If you are asking if the DiscoverTestInputs target should be moved to tests.targets...given the even stronger coupling between RunTestsForProject and CopyTestToTestDirectory, I am not comfortable with the separation between tests.targets and publishtest.targets. It made sense when they were happening at two different points in the lifecycle, but now it just feels strange. I have been dancing around combining them.
2) Agreed, I will rename it.

weshaggard: For #1 I kind of agree with you I'm fine with combining both into tests.targets

eatdrinksleepcode:
I renamed TestInputs to RunTestsForProjectInputs. I also moved DiscoverTestInputs and some associated Items from publishtest.targets to tests.targets. I did not fully collapse those two targets files; I tried it out, but it makes the resulting file very large and messy. I still think some refactoring needs to be done there, but what exactly I am not sure, and I don't want figuring that out to hold up this change any longer. Let me know if you disagree.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -49,12 +52,39 @@
...
+      <RunTestsForProjectInputs Include="@(Content)" />
+      <RunTestsForProjectInputs Include="@(IntermediateAssembly)" />
+      <RunTestsForProjectInputs Include="@(_DebugSymbolsIntermediatePath)" />
+      <RunTestsForProjectInputs Include="@(AllItemsFullPathWithTargetPath)" />
+    </ItemGroup>
+  </Target>
 
   <!-- On command line run unit tests on CoreCLR after the Test target -->
   <Target Name="RunTestsForProject"
+          DependsOnTargets="DiscoverTestInputs"
```
weshaggard: It is my understanding that you cannot use a DependsOnTargets value to compute the Inputs/Outputs/Conditions for a target as all those are evaluated at the same time. Can you please double check that this is doing what you expect?

weshaggard:
Thanks @eatdrinksleepcode I really appreciate your efforts in making this better. I just had the one comment that I would like a little extra verification on and once that is verified then we can merge. As for other refactoring of the test targets I will look more into that at a later point.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
eatdrinksleepcode: It works as intended. From [MSDN](https://msdn.microsoft.com/en-us/library/ee216359.aspx):

> MSBuild determines the target build order as follows:
> ...
> 4. Before a target is executed, its DependsOnTargets targets are run.
> 5. Before a target is executed, any target that lists it in a BeforeTargets attribute is run.
> 6. Before a target is executed, its Inputs attribute and Outputs attribute are compared. If MSBuild determines that any output files are out of date with respect to the corresponding input file or files, then MSBuild executes the target. Otherwise, MSBuild skips the target.
> 7. After a target is executed or skipped, any target that lists it in an AfterTargets attribute is run.

weshaggard: What is unclear to me from that is when the property values in Inputs/Outputs are actually evaluated. It was my impression (could be wrong) that all those properties are evaluated at the same time and in which case the properties for inputs wouldn't have the correct values if they were computed by the DependsOnTargets targets. However that should be easy to test and that is all I was asking for verification on.

eatdrinksleepcode: Ah, I see. If I remove DependsOnTargets="DiscoverTestInputs", then RunTestsForProject is skipped because it has no inputs. If I put it back, the target is executed as expected on the first run, and not executed on the second run, as expected. This tells me that the inputs for RunTestsForProject are not being evaluated until after DiscoverTestInputs has executed; otherwise, executing DiscoverTestInputs would have no effect on whether RunTestsForProject is executed.

weshaggard: Thanks for verifying. That invalidated one of my MSBuild assumptions that I'm going to need to play around with more later to better understand.

weshaggard:
Thanks for all your work on this @eatdrinksleepcode.

PR #119, opened 2015-03-17
Update buildtools to allow generate resources code in intermediate folder.

tarekgh:
We are going to remove all checked in SR.cs in the corefx tree and we'll always generate such files on demand. this change will make handling the resources easier as no need to add any source files to the projects


tarekgh:
@weshaggard  @davkean could you have a quick look?

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -58,25 +53,13 @@ public override bool Execute()
...
-                string message = "Resources code generation is failed\n" + e.Message;
-
-                if (e is System.UnauthorizedAccessException)
-                {
-                    message = message + string.Format("\nThe generated {0} file needs to be updated but the file is read-only.", OutputSourceFilePath);
-                }
-
-                Log.LogError(message);
-
+                Log.LogError("Resources code generation is failed\n" + e.Message);
```
weshaggard: "Failed to generate the resource code with error:\n"

weshaggard:
Other than the error message change it looks good to me.

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
```diff
@@ -23,16 +21,35 @@
...
         ResxFilePath="$(StringResourcesPath)" 
-        IntermediateFilePath="$(IntermediateResOutputFileFullPath)" 
-        OutputSourceFilePath="$(ResourceSRPath)" 
+        OutputSourceFilePath="$(IntermediateResOutputFileFullPath)" 
         AssemblyName="$(AssemblyName)" />
 
     <ItemGroup>
+      <!-- The following Compile element has to be included dynamically inside the Target otherwise intellisense will not work -->
+      <Compile Include="$(IntermediateResOutputFileFullPath)">
+          <Visible>true</Visible>
```
davkean: <Visible> is not going to work. The visual tree is calculated statically at project load, you cannot control the visibility after that.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: I'll fix the message.

src/Microsoft.DotNet.Build.Tasks/Targets/resources.targets
tarekgh: I think you are right. I re-checked that and is not visible. I'll remove <Visible> element

tarekgh:
Thanks Wes and David. I'll merge it.

PR #121, opened 2015-03-20
Initial integration of perf testing tools into the build system.

jhendrixMSFT:
Add PerfTesting.targets file defining properties and targets used to
enable perf testing during the build.  Testing is enabled via msbuild
properties.  At present there are two properties, TestConsumptiveMetrics
and BaselineConsumptiveMetrics.  The former is used to compare live perf
data against a checked in baseline and the latter is used to generate
baseline data.

Added a reference to the perf tools nuget package.

Updated tests.targets to start/stop collection of perf data.  Also added
functionality allowing this target to be called for multiple iterations
(see comments for details).

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -37,6 +37,9 @@
   <!-- The Code Coverage targets will override TestHost and TestCommandLine if coverage is enabled -->
   <Import Project="$(MSBuildThisFileDirectory)CodeCoverage.targets" />
 
+  <!-- import settings for perf testing -->
+  <Import Project="$(MSBuildThisFileDirectory)PerfTesting.targets" />
+
   <!-- In VS (2015 Preview or later currently required): Debug to run unit tests on CoreCLR. -->
   <PropertyGroup Condition="'$(IsTestProject)'=='true'">
     <DebugTestFrameworkFolder>aspnetcore50</DebugTestFrameworkFolder>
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Please ignore, this will be retooled.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -60,16 +82,28 @@
     <!-- Append the xunit trait options to the end of the TestCommandLine -->
     <PropertyGroup>
       <TestCommandLine>$(TestCommandLine.Replace('{XunitTraitOptions}', '$(XunitTraitOptions)'))</TestCommandLine>
+      <EventTracerDataFile>$(TargetFileName)%(TestTargetFrameworkBatch.Count).etl</EventTracerDataFile>
+      <EventTracerMetrics>-c $(ClrMetrics) -k $(KernelMetrics)</EventTracerMetrics>
+      <EventTracerReport>Perf.$(MSBuildProjectName)%(TestTargetFrameworkBatch.Count).xml</EventTracerReport>
```
weshaggard: Can we try to keep most of these in the perf testing targets? It looks like you are putting it here for the batching parameters but in general I think the directory structure will be able to keep these files from clashing.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -47,10 +50,29 @@
...
+    </ItemGroup>
+
+  </Target>
+
   <!-- On command line run unit tests on CoreCLR after the Test target -->
   <Target Name="RunTestsForProject"
           AfterTargets="CheckTestCategories"
-          Condition="'$(TestDisabled)'!='true'">
+          Condition="'$(TestDisabled)'!='true'"
+          Outputs="%(TestTargetFrameworkBatch.Folder) %(TestTargetFrameworkBatch.Count)">
```
weshaggard: You are going to clash with @eatdrinksleepcode current PR https://github.com/dotnet/buildtools/pull/116

PR #128, opened 2015-03-27
Add support for building VB projects.

AlekseyTs:


dir.props
```diff
@@ -69,13 +69,15 @@
     <DebugSymbols Condition="'$(DebugSymbols)' == ''">true</DebugSymbols>
     <Optimize Condition="'$(Optimize)' == ''">false</Optimize>
     <DebugType Condition="'$(DebugType)' == ''">full</DebugType>
-    <DefineConstants>$(DefineConstants);DEBUG;TRACE</DefineConstants>
+    <DefineConstants Condition="'$(MSBuildProjectExtension)' != '.vbproj'">$(DefineConstants);DEBUG;TRACE</DefineConstants>
```
weshaggard: I would prefer if we didn't need to do this in everyone's dir.props and we only had one way to have DefineConstances which is with semi-colon's and then all in one place you translate them to comma delimited.

src/Microsoft.DotNet.Build.Tasks/Targets/FrameworkTargeting.targets
```diff
@@ -52,9 +52,35 @@
...
+            Condition="'%(ReferencePath.Filename)%(ReferencePath.Extension)' == 'System.Runtime.dll'">
+      </FrameworkPathOverrideCandidate>
+    </ItemGroup>
+
+    <PropertyGroup Condition="'@(FrameworkPathOverrideCandidate->Count())' == '1'">
+      <FrameworkPathOverride>@(FrameworkPathOverrideCandidate)</FrameworkPathOverride>
+    </PropertyGroup>
+  </Target>
 
 </Project>
```
weshaggard: This is the place I would translate the DefineConstants to be "comma" delimited.

weshaggard:
Other than the defines this looks good to me.

dir.props
eatdrinksleepcode: What would have to be done in everyone's dir.props? Doesn't this allow each project type to continue following its normal conventions?

weshaggard: I was hoping that we could keep the general defines non-duplicated by just having one set that is semi-colon delimited and we could split that and then comma delimited it in a single location. Any VB projects should choose to be semi-colon or comma delimited as I think either would still work.

AlekseyTs: @weshaggard I am not sure what you mean by saying: "Any VB projects should choose to be semi-colon or comma delimited as I think either would still work." VB compiler doesn't accept semi-colon delimited defines. Replacing semi-colons with commas is error prone because semi-colons could be within values.

src/Microsoft.DotNet.Build.Tasks/Targets/FrameworkTargeting.targets
AlekseyTs: I do not believe that doing this properly is as simple as you think. Not worth the trouble in my opinion.

dir.props
AlekseyTs: It looks like C# compiler handles comma separated defines, I'll switch to using commas for both languages instead.

AlekseyTs:
Made a change to use comma separated list of defines for both languages.

dir.props
```diff
@@ -69,13 +69,13 @@
     <DebugSymbols Condition="'$(DebugSymbols)' == ''">true</DebugSymbols>
     <Optimize Condition="'$(Optimize)' == ''">false</Optimize>
     <DebugType Condition="'$(DebugType)' == ''">full</DebugType>
-    <DefineConstants>$(DefineConstants);DEBUG;TRACE</DefineConstants>
+    <DefineConstants>$(DefineConstants),DEBUG,TRACE</DefineConstants>
```
weshaggard: Have you verified that both the native and Roslyn C# compiler supports comma? We still support both.

AlekseyTs: I tested this change in corefx repository in VS2013 and VS2015 command prompt. There were no build failures.

weshaggard: Thanks @AlekseyTs.

PR #130, opened 2015-03-30
Make incremental build conditional on selected test categories

eatdrinksleepcode:
This fixes dotnet/corefx#1275.

The semaphore file now includes the test categories that were selected to run or not run; selecting new categories will re-run the test if that set of category inclusions/exclusions is out of date.


eatdrinksleepcode:
Note: this will fail if a selected test category contains a character that is illegal in a filename. Is that a scenario we need to handle?

weshaggard:
I'm not worried about test categories that contain illegal file characters at this point. LGTM thanks for the update.

PR #131, opened 2015-03-31
Ensure required properties are set prior to copying files for testing

eatdrinksleepcode:
Fixes dotnet/corefx#1286

ReferenceCopyLocalPaths is populated by ResolveAssemblyReferences, which requires (but does not depend on) ResolveProjectReferences. AllItemsFullPathWithTargetPath is populated by GetCopyToOutputDirectoryItems. These targets are executed during Build; executing Test without Build was leaving these items unpopulated, causing them to not be copied to the test directory.


weshaggard:
LGTM thanks for fixing.

PR #135, opened 2015-04-08
Change properties with paths to use forward slash to be friendly with bo...

JeremyKuhne:
...th Unix and Windows.

This will enable updating the projects depending on .net buildtools so that they can build with MSBuild on Unix. This change is only necessary with properties as includes will be "Windows normalized" by MSBuild.

Note that Windows can care less about slash direction outside of root locations (C:\, \\Server\Share, \\?\). Unix only allows forward slash as a separator.

I tested this out on Unix with additional pending changes on CoreFx. I expect I'll probably find more issues in the build tools, but this is an important first step.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
@@ -4,7 +4,7 @@
   <UsingTask TaskName="ResolveNuGetPackages" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
-    <ProjectPackagesConfigFile Condition="'$(ProjectPackagesConfigFile)'=='' and Exists('$(MSBuildProjectDirectory)\packages.config')">$(MSBuildProjectDirectory)\packages.config</ProjectPackagesConfigFile>
+    <ProjectPackagesConfigFile Condition="'$(ProjectPackagesConfigFile)'=='' and Exists('$(MSBuildProjectDirectory)/packages.config')">$(MSBuildProjectDirectory)/packages.config</ProjectPackagesConfigFile>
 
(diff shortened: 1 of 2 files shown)


JeremyKuhne:
@akoeplinger, @stephentoub, @ellismg - I tried a number of other things and I'm pretty convinced now that the best course of action here is to set a best practice of using forward slashes in our build file properties.

akoeplinger:
LGTM :+1:

xbuild somehow manages to replace backslashes in properties as well - maybe just it doesn't care that `\` is a valid path char on unix?

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -4,7 +4,7 @@
   <UsingTask TaskName="ResolveNuGetPackages" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
-    <ProjectPackagesConfigFile Condition="'$(ProjectPackagesConfigFile)'=='' and Exists('$(MSBuildProjectDirectory)\packages.config')">$(MSBuildProjectDirectory)\packages.config</ProjectPackagesConfigFile>
+    <ProjectPackagesConfigFile Condition="'$(ProjectPackagesConfigFile)'=='' and Exists('$(MSBuildProjectDirectory)/packages.config')">$(MSBuildProjectDirectory)/packages.config</ProjectPackagesConfigFile>
```
tarekgh: are we sure msbuildprojectdirectory never have '\'? I am asking because if we get a combination of '\' and '/' will cause a problem I think

tarekgh:
LGTM as long as we have tested this carefully and make sure we'll never run into situation having a combination of '/' and '\' in same path. Thanks.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
JeremyKuhne: I didn't see any issues with it when I ran this on Unix. MSBuild seems to be doing the right thing.

PR #137, opened 2015-04-13
Embed the resources with different names

tarekgh:
This change is to embed the resources with different name for each assembly
to avoid the resource merge conflict in the store app scenario as
the resources get merged into the app PRI file.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
@@ -20,6 +20,7 @@ namespace Microsoft.DotNet.Build.Tasks
         private StreamWriter _targetStream;
         private StringBuilder _debugCode = new StringBuilder();
         private Dictionary<string, int> _keys;
+        private String _resourcesName;
 
         [Required]
         public string ResxFilePath { get; set; }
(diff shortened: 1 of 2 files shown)


tarekgh:
@weshaggard please have a look

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -169,6 +174,18 @@ private void WriteDebugCode()
             }
         }
 
+        private void WriteGetTypeProperty()
+        {
+            if (_targetLanguage == TargetLanguage.CSharp)
+            {
+                _targetStream.WriteLine("        internal static Type ResourceType {1}{3}              get {1} return typeof({0}); {2}{3}        {2}", _resourcesName + ".SR", "{", "}", Environment.NewLine);
```
weshaggard: Where does this get used? I don't see where you are generating the code that calls this.

weshaggard:
In general I believe I understand the change and it looks good. I just had the one question.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
tarekgh: This property is used in the SR.cs file checked under common which is creating ResourceManager instance. you'll see this in other code review I sent to you yesterday. 

thanks for your review. I'll merge this change after you sign off on the other code review just in case we need to adjust anything else.

PR #138, opened 2015-04-14
Add project.json support to build tools

ericstj:
This change enables buildtools to restore packages referenced by project.json.  New "nuget v3" style packages must be referenced via project.json. /cc @mellinoe @weshaggard @jasonmalinowski @JeremyKuhne

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/MSBuildDependencyResolver.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/MSBuildDependencyResolver.cs
@@ -18,10 +18,13 @@ namespace Microsoft.NuGet.Build.Tasks
     internal sealed class MSBuildDependencyProvider : IDependencyProvider
     {
         private readonly string _msbuildProjectFilePath;
+        private readonly string[] _additionalProjectJsonPaths;
+        
 
-        public MSBuildDependencyProvider(string projectFilePath)
+        public MSBuildDependencyProvider(string projectFilePath, IEnumerable<string> additionalProjectJsonPaths)
         {
             _msbuildProjectFilePath = projectFilePath;
+            _additionalProjectJsonPaths = additionalProjectJsonPaths == null ? new string[0] : additionalProjectJsonPaths.ToArray();
         }
 
         public Library GetDescription(LibraryRange libraryRange, NuGetFramework targetFramework)
@@ -35,15 +38,15 @@ namespace Microsoft.NuGet.Build.Tasks
 
             var projectJsonPath = Path.Combine(Path.GetDirectoryName(_msbuildProjectFilePath), "project.json");
 
-            using (var fileStream = new FileStream(projectJsonPath, FileMode.Open, FileAccess.Read, FileShare.Read))
+            if (File.Exists(projectJsonPath))
             {
-                var packageSpec = JsonPackageSpecReader.GetPackageSpec(fileStream, projectJsonPath, projectJsonPath);
+                // may be running against a project with only a packages.config
+                AddFromProjectJsonFile(dependencies, projectJsonPath, targetFramework);
+            }
 
-                // Grab dependencies from here too
-                var targetFrameworkInfo = packageSpec.GetTargetFramework(targetFramework);
-
-                dependencies.AddRange(packageSpec.Dependencies);
-                dependencies.AddRange(targetFrameworkInfo.Dependencies);
+            foreach(string additionalProjectJsonPath in _additionalProjectJsonPaths)
+            {
+                AddFromProjectJsonFile(dependencies, additionalProjectJsonPath, targetFramework);
             }
 
             var description = new Library
@@ -60,6 +63,20 @@ namespace Microsoft.NuGet.Build.Tasks
             };
 
             return description;
+        }
+
+        private void AddFromProjectJsonFile(List<LibraryDependency> dependencies, string projectJsonPath, NuGetFramework targetFramework)
+        {
+            using (var fileStream = new FileStream(projectJsonPath, FileMode.Open, FileAccess.Read, FileShare.Read))
+            {
+                var packageSpec = JsonPackageSpecReader.GetPackageSpec(fileStream, projectJsonPath, projectJsonPath);
+
+                // Grab dependencies from here too
+                var targetFrameworkInfo = packageSpec.GetTargetFramework(targetFramework);
+
+                dependencies.AddRange(packageSpec.Dependencies);
+                dependencies.AddRange(targetFrameworkInfo.Dependencies);
+            }
         }
 
         public bool SupportsType(string libraryType)
--- a/src/Microsoft.DotNet.Build.Tasks/MSBuildDependencyResolver.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/MSBuildDependencyResolver.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using NuGet.DependencyResolver;
+using NuGet.Frameworks;
+using NuGet.LibraryModel;
+using NuGet.ProjectModel;
+using NuGet.Versioning;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.NuGet.Build.Tasks
+{
+    internal sealed class MSBuildDependencyProvider : IDependencyProvider
+    {
+        private readonly string _msbuildProjectFilePath;
+
+        public MSBuildDependencyProvider(string projectFilePath)
+        {
+            _msbuildProjectFilePath = projectFilePath;
+        }
+
+        public Library GetDescription(LibraryRange libraryRange, NuGetFramework targetFramework)
+        {
+            if (libraryRange.Name != _msbuildProjectFilePath)
+            {
+                return null;
+            }
+
+            var dependencies = new List<LibraryDependency>();
+
+            var projectJsonPath = Path.Combine(Path.GetDirectoryName(_msbuildProjectFilePath), "project.json");
+
+            using (var fileStream = new FileStream(projectJsonPath, FileMode.Open, FileAccess.Read, FileShare.Read))
+            {
+                var packageSpec = JsonPackageSpecReader.GetPackageSpec(fileStream, projectJsonPath, projectJsonPath);
+
+                // Grab dependencies from here too
+                var targetFrameworkInfo = packageSpec.GetTargetFramework(targetFramework);
+
+                dependencies.AddRange(packageSpec.Dependencies);
+                dependencies.AddRange(targetFrameworkInfo.Dependencies);
+            }
+
+            var description = new Library
+            {
+                LibraryRange = libraryRange,
+                Identity = new LibraryIdentity
+                {
+                    Name = libraryRange.Name,
+                    Version = new NuGetVersion(new Version(1, 0)), // TODO: Make up something better
+                    Type = LibraryTypes.MSBuild,
+                },
+                Path = _msbuildProjectFilePath,
+                Dependencies = dependencies
+            };
+
+            return description;
+        }
+
+        public bool SupportsType(string libraryType)
+        {
+            return libraryType == LibraryTypes.MSBuild || libraryType == null;
+        }
+    }
+}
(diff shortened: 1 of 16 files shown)


src/Microsoft.DotNet.Build.Tasks/MSBuildDependencyResolver.cs
```diff
@@ -0,0 +1,87 @@
...
+using NuGet.ProjectModel;
+using NuGet.Versioning;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.NuGet.Build.Tasks
```
JeremyKuhne: Is the code in the Microsoft.Nuget.Build.Tasks namespace coming from somewhere else?

ericstj: Yes, it's coming from Jason's work that's going on in VS.  If we care we can rename to our own namespace and then do the same every
time we take a drop of the code, but I'd prefer to just leave as-is.

JeremyKuhne: Just wanted to be clear that this was a code drop. I don't think we should change the namespace.

JeremyKuhne:
Lgtm.

ericstj:
Thanks!

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -34,23 +42,73 @@
...
+                          TargetFramework="$(NugetTargetFrameworkMoniker)"
                           TargetPlatformMoniker="$(TargetPlatformMoniker)">
 
       <Output TaskParameter="ResolvedAnalyzers" ItemName="Analyzer" />
       <Output TaskParameter="ResolvedReferences" ItemName="Reference" />
       <Output TaskParameter="ResolvedCopyLocal" ItemName="None" />
     </ResolveNuGetPackages>
+
+    <PropertyGroup>
+      <!-- Temporary workaround -->
```
weshaggard: What is this working around? Can you make sure there is an issue tracking this if it is just temporary.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -2,31 +2,39 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <UsingTask TaskName="ResolveNuGetPackages" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
     <ProjectPackagesConfigFile Condition="'$(ProjectPackagesConfigFile)'=='' and Exists('$(MSBuildProjectDirectory)/packages.config')">$(MSBuildProjectDirectory)/packages.config</ProjectPackagesConfigFile>
+    <ProjectJson Condition="'$(ProjectJson)'=='' and Exists('$(MSBuildProjectDirectory)/project.json')">$(MSBuildProjectDirectory)/project.json</ProjectJson>
+    <ProjectLockJson>$(MSBuildProjectDirectory)/project.lock.json</ProjectLockJson>
```
weshaggard: Does this lock file have to be in the root directory or can it be put into the intermediate directory?

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
ericstj: It's working around the fact that ResolveNuGetPackageAssets doesn't currently handle PCL profiles.  Additionally, we should probably change the TFM being used by all our projects to something that actually makes sense for what they are targeting.  Probably the new Dev14 PCL TFM (once it exists) I'll file an issue.

ericstj: Issue filed: https://github.com/dotnet/buildtools/issues/139

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -34,23 +42,73 @@
...
+      <!-- Intersect project-refs with package-refs -->
+      <_ReferenceFileNamesToRemove Include="@(_ReferencesFileNames->'%(OriginalIdentity)')" Condition="'@(_ProjectReferenceFilenames)' == '@(_ReferencesFileNames)' and '%(Identity)' != ''"/>
+      <_NoneFileNamesToRemove Include="@(_NoneFileNames->'%(OriginalIdentity)')" Condition="'@(_ProjectReferenceFilenames)' == '@(_NoneFileNames)' and '%(Identity)' != ''"/>
+
+      <Reference Remove="@(_ReferenceFileNamesToRemove)" />
+      <None Remove="@(_NoneFileNamesToRemove)"/>
+    </ItemGroup>
+    
+    <Message Text="Excluding @(_ReferenceFileNamesToRemove);@(_NoneFileNamesToRemove) from package references since the same file is provided by a project refrence."
+             Condition="'@(_ReferenceFileNamesToRemove)' != '' AND '@(_NoneFileNamesToRemove)' != ''"/>
```
weshaggard: I assume you wanted this condition to be an OR alternatively you could concat like '@(_ReferenceFileNamesToRemove)@(NoneFileNamesToRemove)' != ''"

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
```diff
@@ -26,12 +26,14 @@
...
         ProjectPackagesConfigFile - If not set defaults to $(MSBuildProjectDirectory)\packages.config
-        RestorePackages - If not set defaults to the existence of the file $(ProjectPackagesConfigFile)
-        ResolveNuGetPackages - If not set defaults to to the existence of the file $(ProjectPackagesConfigFile)
+        RestorePackages - If not set defaults to the existence of the file $(ProjectPackagesConfigFile) or $(ProjectJson)
+        ResolveNuGetPackages - If not set defaults to to the existence of the file $(ProjectPackagesConfigFile) or $(ProjectJson)
       
       Depends on properties:
-        NugetRestoreCommand - Used to restore the project packages
+        NugetRestoreCommand - Used to restore the project packages from packages.config
+        DnuRestoreCommand - Used to restore the project packages from project.json
```
weshaggard: Is Dnu the new nuget.exe I guess I will have to wait to see that once you start consuming these changes.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
ericstj: It's actually created by dnu.cmd restore next to the project.  The design of their system is to always put it next to the project.json.  The idea is that it is a cache of the asset applicability and can be checked in.  A scenario I'd like to use is switch to * versioned dependencies, resolve on dev box, generate .lock.json, check in.  Occasionally delete all .lock.jsons and regenerate to pick up new builds.

The property itself is no longer used.  It's left behind from a previous version of this change where I deleted the lock file after it was created (since it's currently not working and using it causes issues, as described in the change descriptions), but instead I switched to just exposing an argument to tell the package resolve task to ignore the file.

I do copy the project.json to intermediate and restore from there for the tests, that's something we could do across all of our projects if we wanted, but I think I'd like to actually check in and consume the lock file in the future.

For now I can remove the property.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
ericstj: Good catch.  I'll fix this.

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
ericstj: Yep, until they rename it

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -38,6 +52,40 @@
...
+      </_ProjectReferenceFilenames>
+
+      <_TestCopyLocalFileNames Include="@(Reference->'%(FileName)%(Extension)')">
+        <OriginalIdentity>%(Identity)</OriginalIdentity>
+      </_TestCopyLocalFileNames>
+
+      <!-- Intersect project-refs with package-refs -->
+      <_TestCopyLocalFileNamesToRemove Include="@(_ReferencesFileNames->'%(OriginalIdentity)')" Condition="'@(_ProjectReferenceFilenames)' == '@(_ReferencesFileNames)' and '%(Identity)' != ''"/>
+
+      <TestCopyLocal Remove="@(_TestCopyLocalFileNamesToRemove)" />
```
weshaggard: Do you actually have to filter these out here? We were not doing it before and presumably with CopyIfNewer will handle this because the project reference should be newer than that the downloaded package reference.

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime-project.json
```diff
@@ -0,0 +1,31 @@
+{
```
weshaggard: I assume you still have a goal of removing this and use the core packages correct?

ericstj: Yeah, I have a comment about it in the change description.  There are a few things blocking me from doing so at the moment.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
ericstj: I'd prefer to not take the chance/rely on ordering of the copy item group.  I had to filter for refs because CSC complains, so I applied the pattern to everything.

I actually plan on killing most of this target once I can move to a meta-package for tests, so my care-factor is low.

weshaggard:
Overall looks good to me. Interesting enough I don't actually read the individual commit messages when reviewing changes in the github UI. Perhaps I need to start looking at those but they aren't easily available in the aggregate code diff view.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -38,6 +52,40 @@
...
+    </ResolveNuGetPackageAssets>
+
+    <!-- We may have an indirect package reference that we want to replace with a project reference -->
+    <ItemGroup>
+      <!-- Convert to filenames so that we can intersect -->
+      <_ProjectReferenceFilenames Include="@(_ResolvedProjectReferencePaths->'%(FileName)%(Extension)')">
+        <OriginalIdentity>%(Identity)</OriginalIdentity>
+      </_ProjectReferenceFilenames>
+
+      <_TestCopyLocalFileNames Include="@(Reference->'%(FileName)%(Extension)')">
```
AlfredoMS: I know this was already merged, but _TestCopyLocalFileNames seems to be unused

ericstj: Ah, that was a typo.

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime-project.json
davidfowl: You can remove the closure and just put the top level deps now

ericstj: Yep, or I could just get rid of this completely, which is my plan.

As far as just using direct dependencies: yes I am well aware. I am planning to do that using a tool across all  of corefx in a future change.  Right now I am just trying to get this working and extra dependencies aren't getting in the way. Trying to lump that work in just further delays this which folks have been needing for months now.

davidfowl: Not sure you want to tool that but sure :smile:

ericstj: We have over 100 projects with the closure burned in. We can't know which packages to drop/keep unless we look at the actual ref's. Tool will look at the refs and spit out minimum deps.  Bonus if I can get it to spit out min version of the deps, despite the refs being at a higher version.

PR #141, opened 2015-04-15
Add support for building partial facades

mellinoe:
This adds the basic support needed for building partial facades. This adds a dependency on a new tools package, "Microsoft.DotNet.BuildTools.ApiTools", produced by our internal build. ApiTools contains GenFacades.exe and ApiCompat.exe. This change only uses GenFacades to fill partial facade assemblies, but both will eventually be used to generate full facade and contract assemblies and verify them against contracts, etc.

This also introduces the concept of a "post-process-assembly" step, which currently contains only two transformations: FillPartialFacade, and OpenSourceSign. Individual targets can register themselves onto the "PostProcessAssemblySteps" ItemGroup to be included in this target.

The biggest addition is in partialfacades.targets. A good chunk of the logic is dealing with resolving mscorlib and the contract assembly via NuGet packages. In the future, much of this special logic can be eliminated when we deliver a true mscorlib reference assembly, and when we build the contract assemblies from source in the repository.

Unfortunately, due to the nature of project.json files and the DNU restore command, it seems like we will need to include a separate project.json file in a special subdirectory in each project which declares the nuget package in which the contract assembly resides. This allows us to restore and resolve those nuget packages for the proper contract assembly as a separate process from resolving the project's regular nuget references.

@weshaggard , @ericstj , @FiveTimesTheFun


src/Microsoft.DotNet.Build.Tasks/Targets/post-process-assembly.targets
```diff
@@ -0,0 +1,22 @@
...
+    <IntermediateAssemblyPath>$(IntermediateOutputPath)$(TargetName)$(TargetExt)</IntermediateAssemblyPath>
+    <PostProcessAssemblyMarker>$(IntermediateAssemblyPath).postprocessed</PostProcessAssemblyMarker>
+  </PropertyGroup>
+
+  <Target Name="PostProcessAssembly"
+    AfterTargets="CoreCompile"
+    Inputs="$(IntermediateAssemblyPath)"
+    Outputs="$(PostProcessAssemblyMarker)">
+
+    <CallTarget Targets="@(PostProcessAssemblySteps)" />
```
ericstj: Why call target and not just have the targets sequence themselves where they need to be? I'm missing what value this target provides, other than maybe the incremental behavior (if that is the main reason, then please write up some comments).  If  you need a sequencing point you could just use dependson property which folks append to as is common practice.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -0,0 +1,141 @@
...
+    Condition="'$(PostFilterNugetReferences)' == 'true'"
+    AfterTargets="ResolveNugetPackages">
+
+    <!-- Well-known default removals -->
+    <ItemGroup>
+      <!-- Only take mscorlib from Microsoft.DotNet.CoreCLR. -->
+      <ReferenceRemovals Include="@(Reference)" Condition="('%(Reference.NuGetPackageName)' == 'Microsoft.DotNet.CoreCLR' or '%(Reference.NuGetPackageId)' == 'Microsoft.DotNet.CoreCLR') and '%(FileName)' != 'mscorlib'"/>
+      <!-- We don't want any of these from Microsoft.DotNet.CoreCLR, either. -->
+      <NoneRemovals Include="@(Reference)" Condition="'%(Reference.NuGetPackageName)' == 'Microsoft.DotNet.CoreCLR'"/>
+    </ItemGroup>
```
ericstj: Do you need to incude NugetPackageId here?

FWIW we currently still use the old task for CoreCLR since the package hasn't been updated.  I'll switch to the new one soon and at that point I'll probably update this.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -0,0 +1,141 @@
...
+         * Seed assemblies
+
+       Fills the "input assembly" by finding types in the contract assembly that are missing from it, and adding type-forwards
+         to those matching types in the seed assemblies. -->
+
+  <!-- Inputs and outputs of FillPartialFacade -->
+  <PropertyGroup Condition="'$(IsPartialFacadeAssembly)' == 'true'">
+    <PartialFacadeAssemblyPath>$(IntermediateOutputPath)$(TargetName)$(TargetExt)</PartialFacadeAssemblyPath>
+    <PartialFacadeSymbols>$(IntermediateOutputPath)$(TargetName).pdb</PartialFacadeSymbols>
+    <GenFacadesInputPath>$(IntermediateOutputPath)PreGenFacades\</GenFacadesInputPath>
```
ericstj: Some recent changes were replacing \ with / to be friendly to unix builds.  Not sure if that applies here.

src/Microsoft.DotNet.Build.Tasks/Targets/post-process-assembly.targets
mellinoe: Yeah, I went with this approach because, basically, FillPartialFacade and OpenSourceSign were conflicting with each other, especially in incremental builds. They essentially have the same inputs and outputs, so each would trigger the other in turn. Additionally, OpenSourceSign fails if the assembly is already signed, so another target modifying the assembly but not "unsigning" it (as FillPartialFacade does) would cause the signing target to fail. I tried to generalize the solution here.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
mellinoe: Right, I think this one was using the target that specified "NuGetPackageId" rather than "-Name", so I included both in this condition. Not sure if its necessary.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
mellinoe: I can change this, although partial facades won't build on Unix for a long time given the state of our tools.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
ericstj: I'd be consistent and include Id in both conditions then.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
ericstj: In that case you can leave it out

src/Microsoft.DotNet.Build.Tasks/Targets/post-process-assembly.targets
ericstj: I see what you mean. It's a side effect of these steps happening in-place and the output being a marker file instead of a real output.  If you didn't do these in-place that might fix the problem.

One problem I see with this now is that incremental behavior of post process assembly only depends on the intermediateassemblypath, but if the seed or contract changes you also need to rerun it. Another thing I am not sure about is the behavior of CallTarget when one target fails, will you continue to execute and touch the marker?

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -0,0 +1,141 @@
...
+    <ItemGroup>
+      <Reference Remove="@(ReferenceRemovals)" />
+      <None Remove="@(NoneRemovals)" />
+    </ItemGroup> 
+  </Target>
+
+  <!-- FindPartialFacadeContractReference
+         Uses the special partial-facade project.json file to restore the NuGet package containing the partial facade's contract assembly.
+         Filters the list of references returned by the ResolveNuGetPackageAssets task to just the contract assembly and stores it.
+  -->
```
weshaggard: I'm not sure I understand why we need this separate project.json resolution. Why can you not just include the contract in the root project.json file and just filter that out of the references?

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -0,0 +1,141 @@
...
+    <GenFacadesOutputPath>$(IntermediateOutputPath)</GenFacadesOutputPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PostProcessAssemblySteps Include="FillPartialFacade" />
+  </ItemGroup>
+
+  <Target Name="FillPartialFacade"
+      Condition="'$(IsPartialFacadeAssembly)' == 'true'"
+      BeforeTargets="OpenSourceSign">
```
weshaggard: Given the way you are doing CallTargets I'm not sure that this BeforeTargets actually does what you want.

src/Microsoft.DotNet.Build.Tasks/Targets/post-process-assembly.targets
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <IntermediateAssemblyPath>$(IntermediateOutputPath)$(TargetName)$(TargetExt)</IntermediateAssemblyPath>
+    <PostProcessAssemblyMarker>$(IntermediateAssemblyPath).postprocessed</PostProcessAssemblyMarker>
+  </PropertyGroup>
+
+  <Target Name="PostProcessAssembly"
+    AfterTargets="CoreCompile"
```
weshaggard: This will break other things. Have a look at the comment in the signing.targets.

src/Microsoft.DotNet.Build.Tasks/Targets/sign.targets
```diff
@@ -26,20 +26,14 @@
         * AfterTargets=PrepareForRun -> hit after @(IntermediateAssembly) has already been copied to output.
```
weshaggard: Have a look at this comment when trying to decide how to hook in the post-processing targets.

src/Microsoft.DotNet.Build.Tasks/Targets/post-process-assembly.targets
weshaggard: Yes I have some concerns with how we are doing this as well I don't like having a single set of inputs/outputs to control all post-processing steps.

weshaggard: It is probably worth having a look at TargetsTriggeredByCompilation to see if we can use that to hook in our post-processing pipeline and then essentially eliminate the inputs/outputs on the post-processing targets and let that be driven by CoreCompile.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
mellinoe: Including the contract in the root project.json will pull in all of that NuGet package's dependencies, which would be more complicated to filter, I think

weshaggard:
cc @JeremyKuhne

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
mellinoe: It seems to work to me, but you may be thinking of something I'm not. With this, FillPartialFacade is always called before OpenSourceSign, regardless of when it is imported, and only runs once.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
weshaggard: Unless project.json is different then project.config I don't believe it pulls in the dependency closure only the packages listed in the file.

mellinoe: According to @ericstj, it is different and pulls in the dependency closure of the packages. Is that correct?

src/Microsoft.DotNet.Build.Tasks/Targets/post-process-assembly.targets
mellinoe: > but if the seed or contract changes you also need to rerun it. 

At least for partial facades, you are going to need to start over from scratch if you want to re-direct the type forwards that have been added. There's no logic in GenFacades for modifying existing type-forwards right now. I think this might indicate that we actually need to include the seed/contract assembly as inputs to CoreCompile if we want truly correct incremental builds, as we need to rebuild the whole assembly to re-do type forwards.

In that case, it might make sense to just hook FillPartialFacade directly after Corecompile with no Inputs/Outputs (basically what Wes said) and undo my changes to sign.targets.

weshaggard: Yes my suggestion, at least for now would be to undo your changes to sign.targets and delete your new post-processing target and just hook the FillParitialFacade target up into TargetsTriggeredByCompilation without inputs/outputs.

mellinoe:
I updated the PR with an additional commit. I did the following:

* Removed the "post-process-assembly" target
* Hooked "FillPartialFacade" directly into the "TargetsTriggeredByCompilation" property, which gets invoked at the end of CoreCompile. I removed the Inputs/Outputs from the FillPartialFacade target.
* Reverted all of the changes to sign.targets

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
weshaggard: Yes based on @ericstj mail that would appear to be the case. In which case can we try to hide this? I assume this project.json file only contains the contract name/version correct? Can we generate that into the intermediates directory as part of the build and then restore from there instead of requiring all these projects to have this extra set of stuff?

mellinoe: Yeah, that would be possible. I'll try it out.

src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
```diff
@@ -115,7 +115,25 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="packages.config" />
-    <None Include="Targets\*.*" />
```
mellinoe: I think I accidentally saved this file in VS. I'll revert this.

mellinoe:
I made an additional commit:

* Removed the need for extra project.json files. The contract NuGet package is now restored via a generated json file in the intermediate directory, constructed with the assembly's name and version number. From my testing, the packages listed in this file are only being downloaded once (I turned off my internet to make sure), so this SHOULD play nicely with incremental build.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
JeremyKuhne: FYI: There is a change coming in MSBuild that fixes file paths in properties assuming the property is (1) just a path and (2) the first directory segment exists. That will allow us not to have to use /.

mellinoe:
Pinging @ericst and @weshaggard, let me know how the latest version looks.

weshaggard:
The changes look good to me.

PR #143, opened 2015-04-16
Override project file to resolve during ResolveNuGetPackages

ericstj:
This is necessary to support some internal build scenarios where we import a CSProj in order to override some build settings.  We still need to use the other CSProj when resolving packages.


ellismg:
LGTM

PR #144, opened 2015-04-17
Support WithCategories and WithoutCategories msbuild properties.

Priya91:
With this change, the csproj need not specify the available categories in the project with TestCategories property. RunTestsWithCategories has been renamed to WithCategories, RunTestsWithoutCategories to WithoutCategories.

By default outerloop and failing tests are excluded. They will only be run if specified in WithCategories.
xunit.console.netcore is modified to return warning if it does not find any tests in the project.
cc @krwq @weshaggard @joshfree

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -6,20 +6,15 @@
 
   <!-- Which categories of tests to run by default -->
   <PropertyGroup>
-    <RunTestsWithCategories Condition="'$(RunTestsWithCategories)' == '' And '$(RunTestsWithoutCategories)' == ''">InnerLoop</RunTestsWithCategories>
-    <TestCategories Condition="'$(TestCategories)'==''">InnerLoop</TestCategories>
-
     <TestDisabled>false</TestDisabled>
     <TestDisabled Condition="'$(IsTestProject)'!='true' Or '$(SkipTests)'=='true' Or '$(RunTestsForProject)'=='false'">true</TestDisabled>
-
     <TestsSuccessfulSemaphore>$(MSBuildProjectFile).tests.lastsucceeded</TestsSuccessfulSemaphore>
   </PropertyGroup>
 
   <ItemGroup>
     <!-- Split semicolon separated lists -->
-    <TestCategoriesItems Include="$(TestCategories)" />
-    <RunTestsWithCategoriesItems Include="$(RunTestsWithCategories)" />
-    <RunTestsWithoutCategoriesItems Include="$(RunTestsWithoutCategories)" />
+    <WithCategoriesItems Include="$(WithCategories)" />
+    <WithoutCategoriesItems Include="$(WithoutCategories)" />
     <UnsupportedPlatformsItems Include="$(UnsupportedPlatforms)"/>
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


krwq:
I'm ok with this as is as soon as it is working for simple cases (Inner and OuterLoop).

I believe the original concept was to support any categories of tests but run only InnerLoop.

Right now adding Perf tests will require changing build tools as they will be run by default. (we exclude only outerloop tests)

I'm fine with that although I heard different opinions on that and hard coding test categories when I was implementing the original feature.

Please make sure @bartonjs and @weshaggard are fine with that too.

Priya91:
@krwq : The concept of innerloop and outerloop have to be defined properly. Any test that is not innerloop (for ex, perf) has to be marked with OuterLoop attribute, if the whole test project is perf, then the whole test project is to be excluded by using the RunTestsForProject=false msbuild property. I dont understand why we are adding extra complexity in defining what is "InnerLoop".

weshaggard:
@Priya91 can you please update the information on the wiki https://github.com/dotnet/corefx/wiki/Developer-Guide and also I would suggest moving that large section on testing into its own guide and just link to it from the developer-guide. I think the new testing guide would be a great place to provide your design doc (i.e. requirements/scenarios) for testing so that folks understand the basics of the infrastructure.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -127,30 +121,19 @@
...
-      <RunWithTraits Condition="'@(RunWithInnerLoopItems)'==''" Include="@(TestCategoriesToRun)" />
-      <RunWithoutTraits Condition="'@(RunWithInnerLoopItems)'!=''" Include="@(TestCategoriesItems)" Exclude="@(TestCategoriesToRun)" />
+      <RunWithTraits Condition="'@(WithCategoriesItems)'!=''" Include="@(WithCategoriesItems)" />
+      <RunWithoutTraits Condition="'@(WithoutCategoriesItemsDistinct)'!=''" Include="@(WithoutCategoriesItemsDistinct)" />
     </ItemGroup>
 
-    <!-- If there is nothing to run, then disable the test -->
-    <PropertyGroup>
-      <TestDisabled Condition="'@(TestCategoriesToRun)'==''">true</TestDisabled>
-    </PropertyGroup>
```
weshaggard: Why did your remove this? We shouldn't be trying to run xunit on test projects that don't actually have any tests to run.

src/xunit.console.netcore/Program.cs
```diff
@@ -298,8 +298,8 @@ static XElement ExecuteAssembly(object consoleLock, string defaultDirectory, Xun
                     {
                         lock (consoleLock)
                         {
-                            Console.ForegroundColor = ConsoleColor.Red;
-                            Console.WriteLine("ERROR:       {0} has no tests to run", Path.GetFileNameWithoutExtension(assembly.AssemblyFilename));
+                            Console.ForegroundColor = ConsoleColor.Yellow;
+                            Console.WriteLine("Warning:       {0} has no tests to run", Path.GetFileNameWithoutExtension(assembly.AssemblyFilename));
                             Console.ForegroundColor = ConsoleColor.Gray;
```
weshaggard: What cases do you see as being valid for not having any tests to run? If a test project doesn't have any tests to run we should avoid running it at the project level.

weshaggard:
RunTestsForProject=false is only to be used to disable the tests for all test runs but that doesn't cover cases were you don't want to have any tests with of a particular category to run.

PR #148, opened 2015-04-21
Expose DefaultNoCategories property to enable specifying the categories(like perf) that are not in innerloop.

Priya91:
This feature will enable extensibility as more categories are added, and need to be excluded from a dev test run (plain build.cmd). So if there are more categories in future, like perf, stress, etc, these can be excluded by specifying <DefaultNoCategories>Perf;Stress</DefaultNoCategories> or /p:DefaultNoCategories=Perf;Stress

CC @krwq

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -15,6 +15,7 @@
     <!-- Split semicolon separated lists -->
     <WithCategoriesItems Include="$(WithCategories)" />
     <WithoutCategoriesItems Include="$(WithoutCategories)" />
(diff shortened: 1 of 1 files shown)


Priya91:
retest this please

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -15,6 +15,7 @@
     <!-- Split semicolon separated lists -->
     <WithCategoriesItems Include="$(WithCategories)" />
     <WithoutCategoriesItems Include="$(WithoutCategories)" />
+    <DefaultNoCategories Include="$(DefaultNoCategories)" />
```
weshaggard: nit: I realize this item is already used below but from a consistency point of view I would have called this DefaultNoCategoriesItems like the other places.

PR #149, opened 2015-04-21
Prevent concurrent restore of same project.json

ericstj:
DNX has an issue with concurrent restore: https://github.com/aspnet/dnx/issues/1682

Previously I would workaround by copying to intermediate, but this breaks in official build where intermediate is not under the scope of our nuget.config, so custom package sources are ignored.

Fix this instead by wrapping the call to dnu restore with a mutex unique to the project.json name.  We only need to do this for the test project.json because that is the only one we'll restore concurrently.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <UsingTask TaskName="ExecWithMutex" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="ResolveNuGetPackages" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="ResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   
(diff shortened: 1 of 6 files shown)


ericstj:
@dotnet-bot test this please

ericstj:
Moreover the nature of the break is a race condition that folks have been noticing in private and CI builds. Essentially if the test project is the first to try to restore a particular package it will fail to find it since it's not using the right feeds.  I'm merging this now to get folks unblocked.

src/NuGet.Config
```diff
@@ -7,7 +7,7 @@
     <add key="myget.org dotnet-coreclr" value="https://www.myget.org/F/dotnet-coreclr/" />
     <add key="myget.org dotnet-corefx" value="https://www.myget.org/F/dotnet-corefx/" />
     <add key="myget.org dotnet-buildtools" value="https://www.myget.org/F/dotnet-buildtools/" />
-    <add key="myget.org aspnetvnext" value="https://www.myget.org/F/aspnetvnext/" />
+    <add key="myget.org nugetbuild" value="https://www.myget.org/F/nugetbuild/" />
```
weshaggard: This this a new feed that we are publishing the dnu packages to?

ericstj: This is a new feed that the nuget team is publish their packages to (the nuget.*.dlls).  Previously they were on aspnetvnext, but that was incorrect.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -20,18 +21,11 @@
...
 
-    <!-- DNU will emit a project.lock.json.  When multiple tests are building for the first time they may race on 
-         running this target and generating the lock.json.  This is causing IOExceptions in dnu for writing lock.json.
-         We should really be handling these test dependencies in a different way (as a package themselves) but for now
-         just restore the project.json from intermediate to avoid the race.  -->
-    <Copy SourceFiles="$(TestRuntimeProjectJson)" DestinationFolder="$(IntermediateOutputPath)">
-      <Output TaskParameter="CopiedFiles" PropertyName="TempTestRuntimeProjectJson"/>
-    </Copy>
-    <Exec Command="$(DnuRestoreCommand) &quot;$(TempTestRuntimeProjectJson)&quot;" StandardOutputImportance="Low" CustomErrorRegularExpression="^Unable to locate .*" />
+    <ExecWithMutex Command="$(DnuRestoreCommand) &quot;$(TestRuntimeProjectJson)&quot;" MutexName="$(TestRuntimeProjectJson)" StandardOutputImportance="Low" CustomErrorRegularExpression="^Unable to locate .*" />
```
weshaggard: This seems like a pretty big hammer do you know if this actually causes any noticeable performance hits? I also assume this is temporary so it maybe worth a comment here about that.

weshaggard:
Overall seems reasonable although I would prefer the actually race condition to be fixed in dnu.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
ericstj: The only time this comes into play is the very first build.  The restore is incremental so as soon as it succeeds one no further projects will run it.  This effectively means that worst case N-1 (where n is the number of MSBuild nodes) projects will be waiting once / serialized while this restore happens, after that this code will not be hit.

I discussed the issue with the nuget/dnx team and they said this is effectively what they plan to do there to fix the bug.

ericstj:
It will be fixed, we just need a workaround because this is failing right now and my previous workaround had side effects.

davidfowl:
Yap!

PR #150, opened 2015-04-23
Add coverage report that shows test and the methods it visits

AlfredoMS:
This commit adds a new coverage report that can be selected. Basically it parses all of the opencover output xml and generates a new file, VisitedMethodsReport.xml. This will be used to match changes in code to tests.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
@@ -14,7 +14,7 @@ namespace Microsoft.DotNet.Build.Tasks
 {
     public class ParseTestCoverageInfo : Task
     {
-        // Path to the directory that contais *.coverage.xml coverage info.
+        // Path to the directory that contains *.coverage.xml coverage info.
         [Required]
         public string CoverageInfoDirectory { get; set; }
 
@@ -23,18 +23,15 @@ namespace Microsoft.DotNet.Build.Tasks
 
         public override bool Execute()
         {
-            if (Directory.Exists(CoverageInfoDirectory))
-            {
-                Log.LogMessage(MessageImportance.Normal, "Parsing coverage files...");
-                if (!ProcessDirectory(CoverageInfoDirectory))
-                {
-                    Log.LogError("There was no coverage file found.");
-                    return false;
-                }
-            }
-            else
+            if (!Directory.Exists(CoverageInfoDirectory))
             {
                 Log.LogError("{0} is not a valid path", CoverageInfoDirectory);
+                return false;
+            }
+
+            if (!ProcessDirectory(CoverageInfoDirectory))
+            {
+                Log.LogError("There was no coverage file found.");
                 return false;
             }
 
@@ -51,8 +48,7 @@ namespace Microsoft.DotNet.Build.Tasks
         /// <returns>True if at least one file that matched the pattern coverage_[name of test assembly].xml is found</returns>
         private bool ProcessDirectory(string directory)
         {
-
-            string[] results = Directory.GetFiles(directory, "*.coverage.xml");
+            string[] results = Directory.GetFiles(directory, "*.coverage.xml", SearchOption.AllDirectories);
 
             bool found = (results.Length > 0);
 
@@ -64,11 +60,6 @@ namespace Microsoft.DotNet.Build.Tasks
                 }
 
                 ParseCoverageFile(file);
-            }
-
-            foreach (string child in Directory.GetDirectories(directory))
-            {
-                found |= ProcessDirectory(child);
             }
 
             return found;
@@ -98,7 +89,7 @@ namespace Microsoft.DotNet.Build.Tasks
                 string moduleName = ModuleNode.SelectSingleNode("ModuleName").InnerText;
 
                 ModuleInfo module = null;
-                if (!visitedModules.ContainsKey(moduleName))
+                if (!visitedModules.TryGetValue(moduleName, out module))
                 {
                     // Initialize module
                     module = new ModuleInfo(moduleName);
@@ -113,10 +104,6 @@ namespace Microsoft.DotNet.Build.Tasks
 
                     visitedModules.Add(moduleName, module);
                 }
-                else
-                {
-                    module = visitedModules[moduleName];
-                }
 
                 string methodSignature = methodNode["Name"].InnerText;
                 // Signature format is "ReturnType Class::method(params)"
@@ -128,14 +115,10 @@ namespace Microsoft.DotNet.Build.Tasks
                 // string methodParams = chunks[3];
 
                 List<string> methodsCovered = null;
-                if (!module.CoveredMethods.ContainsKey(className))
+                if (!module.CoveredMethods.TryGetValue(className, out methodsCovered))
                 {
                     methodsCovered = new List<string>();
                     module.CoveredMethods.Add(className, methodsCovered);
-                }
-                else
-                {
-                    methodsCovered = module.CoveredMethods[className];
                 }
 
                 methodsCovered.Add(methodSignature);
--- a/src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
@@ -16,53 +16,35 @@ namespace Microsoft.DotNet.Build.Tasks
     {
         // Path to the directory that contains *.coverage.xml coverage info.
         [Required]
-        public string CoverageInfoDirectory { get; set; }
+        public ITaskItem[] CoverageReports { get; set; }
+
+        // Name and path of the output XML
+        [Required]
+        public string OutputReport { get; set; }
 
         // One per xml parsed, key = test name, value list of modules visited
-        private Dictionary<string, Dictionary<string, ModuleInfo>> testCoverageInfo = null;
+        private Dictionary<string, Dictionary<string, ModuleInfo>> testCoverageInfo = new Dictionary<string,Dictionary<string,ModuleInfo>>();
 
         public override bool Execute()
         {
-            if (!Directory.Exists(CoverageInfoDirectory))
+            if (CoverageReports.Length == 0)
             {
-                Log.LogError("{0} is not a valid path", CoverageInfoDirectory);
+                Log.LogError("CoverageReports cannot be empty");
                 return false;
             }
 
-            if (!ProcessDirectory(CoverageInfoDirectory))
+            Log.LogMessage(MessageImportance.Normal, "{0} reports found.", CoverageReports.Length);
+
+            foreach (var item in CoverageReports)
             {
-                Log.LogError("There was no coverage file found.");
-                return false;
+                Log.LogMessage(MessageImportance.Normal, "Processing file: {0}", item.ItemSpec);
+                
+                ParseCoverageFile(item.ItemSpec);
             }
 
             SaveToFile();
 
             return true;
-        }
-
-        /// <summary>
-        /// Explore the input path recursively.
-        /// Parse opencoverage output.
-        /// </summary>
-        /// <param name="directory"></param>
-        /// <returns>True if at least one file that matched the pattern coverage_[name of test assembly].xml is found</returns>
-        private bool ProcessDirectory(string directory)
-        {
-            string[] results = Directory.GetFiles(directory, "*.coverage.xml", SearchOption.AllDirectories);
-
-            bool found = (results.Length > 0);
-
-            foreach (string file in results)
-            {
-                if (testCoverageInfo == null)
-                {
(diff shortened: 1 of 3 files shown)


AlfredoMS:
@weshaggard can you please take a look?

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class ParseTestCoverageInfo : Task
+    {
+        // Path to the directory that contais *.coverage.xml coverage info.
```
Petermarcu: contais -> contains

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+                Log.LogMessage(MessageImportance.Normal, "Parsing coverage files...");
+                if (!ProcessDirectory(CoverageInfoDirectory))
+                {
+                    Log.LogError("There was no coverage file found.");
+                    return false;
+                }
+            }
+            else
+            {
+                Log.LogError("{0} is not a valid path", CoverageInfoDirectory);
```
Petermarcu: One suggestion would be to move your error cases to checks at the beginning. The same would be true for the missing coverage file case. You'd have less nested if's and the success path would be succinct.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+        public string CoverageInfoDirectory { get; set; }
+
+        // One per xml parsed, key = test name, value list of modules visited
+        private Dictionary<string, Dictionary<string, ModuleInfo>> testCoverageInfo = null;
+
+        public override bool Execute()
+        {
+            if (Directory.Exists(CoverageInfoDirectory))
+            {
+                Log.LogMessage(MessageImportance.Normal, "Parsing coverage files...");
```
Petermarcu: Should this be inside ProcessDirectory?

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        /// <summary>
+        /// Explore the input path recursively.
+        /// Parse opencoverage output.
+        /// </summary>
+        /// <param name="directory"></param>
+        /// <returns>True if at least one file that matched the pattern coverage_[name of test assembly].xml is found</returns>
+        private bool ProcessDirectory(string directory)
+        {
+
```
Petermarcu: Suggest remove whitespace

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                if (testCoverageInfo == null)
+                {
+                    testCoverageInfo = new Dictionary<string, Dictionary<string, ModuleInfo>>();
+                }
+
+                ParseCoverageFile(file);
+            }
+
+            foreach (string child in Directory.GetDirectories(directory))
```
Petermarcu: Have you considered using the recursive option on Directory.GetFiles so you don't have to do the recursion yourself? That way I think you'd get the full list of coverage files under the root folder in one call.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
Petermarcu: If you change the recursion below, you could move this message into the ProcessDirectory method.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+        private Dictionary<string, Dictionary<string, ModuleInfo>> testCoverageInfo = null;
+
+        public override bool Execute()
+        {
+            if (Directory.Exists(CoverageInfoDirectory))
+            {
+                Log.LogMessage(MessageImportance.Normal, "Parsing coverage files...");
+                if (!ProcessDirectory(CoverageInfoDirectory))
+                {
+                    Log.LogError("There was no coverage file found.");
```
Petermarcu: Is it always an error if no coverage files are found or there a successful case where no coverage was run? I just don't know enough about this space to know.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+            Dictionary<string, ModuleInfo> visitedModules = new Dictionary<string, ModuleInfo>();
+
+            foreach (XmlNode methodNode in methods)
+            {
+                // method < methods < class < classes < module
+                XmlNode ModuleNode = methodNode.ParentNode.ParentNode.ParentNode.ParentNode;
+                string moduleName = ModuleNode.SelectSingleNode("ModuleName").InnerText;
+
+                ModuleInfo module = null;
+                if (!visitedModules.ContainsKey(moduleName))
```
Petermarcu: Consider using TryGetValue pattern here.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,219 @@
...
+            return found;
+        }
+
+        /// <summary>
+        /// Parse all coverage .xml files
+        /// </summary>
+        private void ParseCoverageFile(string file)
+        {
+            XmlDocument coverageXml = new XmlDocument();
+            coverageXml.Load(file);
```
Petermarcu: I don't know enough about the structure of the file but did you consider and rule out a forward only XmlReader as opposed to XmlDocument?

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
AlfredoMS: I prefer this to be the first thing we do. After all that is the only parameter of the Task and if this is invalid nothing will work.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
AlfredoMS: Yes, if no coverage files are found it is an error. Since this is a report generator based on  parsed information from the coverage output and we have to call it explicitly as a parameter of the build.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
AlfredoMS: will do

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
Petermarcu: I meant the Log.LogMessage, not the check for existence.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
AlfredoMS: Yes. Two reasons to use Xdocument 1) XPath query  which allows 2) that I don't have to iterate through every level of nesting.
This is why I have the bottom up parsing comment.

Let me know if I am missing some advantage of XmlReader (maybe faster but I prefer nicer code)

Petermarcu: One reason would be memory pressure. If these are large files, you are loading the whole doc into memory rather than just reading through it for the data you need. If in memory is acceptable, you should decide whether XmlDocument and XDocument. XDocument is the newer and more Linq friendly version and is the default recommendation.

src/Microsoft.DotNet.Build.Tasks/Targets/CodeCoverage.targets
```diff
@@ -22,7 +22,8 @@
     <CoverageEnabledForProject Condition="'$(CoverageEnabledForProject)'=='' and '$(IsTestProject)'=='true'">$(CodeCoverageEnabled)</CoverageEnabledForProject>
 
     <GenerateIndividualCoverageReport Condition="'$(BuildAllProjects)'!='true' and '$(CoverageEnabledForProject)'=='true'">true</GenerateIndividualCoverageReport>
-    <GenerateFullCoverageReport Condition="'$(GenerateCodeCoverageReportForAll)'=='true' and '$(CodeCoverageEnabled)'=='true'">true</GenerateFullCoverageReport>
+    <GenerateFullCoverageReport Condition="'$(GenerateCodeCoverageReportForAll)'=='true' and '$(CodeCoverageEnabled)'=='true' and '$(SkipFullCoverageReport)' != 'true'">true</GenerateFullCoverageReport>
```
weshaggard: Where do you plan to set SkipFullCoverageReport and GenerateVisitedReport? I was expecting the same property to be used one in the positive and one in the negative case. If you expect these to be mutually exclusive then I would use the same property but if you want potentially do both in some case then having the two properties make sense.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        /// <summary>
+        /// Explore the input path recursively.
+        /// Parse opencoverage output.
+        /// </summary>
+        /// <param name="directory"></param>
+        /// <returns>True if at least one file that matched the pattern coverage_[name of test assembly].xml is found</returns>
+        private bool ProcessDirectory(string directory)
+        {
+            string[] results = Directory.GetFiles(directory, "*.coverage.xml", SearchOption.AllDirectories);
```
weshaggard: You should avoid hardcoding file locations/patterns/names in the tasks and instead accept that as a parameter passed in via the target that is calling the task.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        /// <summary>
+        /// Save parsed information into a new file
+        /// </summary>
+        public void SaveToFile()
+        {
+            XmlWriterSettings settings = new XmlWriterSettings();
+            settings.Indent = true;
+            string outputFile = Path.Combine(CoverageInfoDirectory,"VisitedMethodsReport.xml");
```
weshaggard: This should also be an parameter passed from the task to the target. We don't want to force people to recompile task libraries to change this if there is a problem.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
```diff
@@ -0,0 +1,202 @@
...
+                ParseCoverageFile(file);
+            }
+
+            return found;
+        }
+
+        /// <summary>
+        /// Parse all coverage .xml files
+        /// </summary>
+        private void ParseCoverageFile(string file)
```
weshaggard: Have you had a look at the OpenCover libraries/framework to see if they don't already have a support for parsing these files? That would be the preferred mechanism if they already have that.

src/Microsoft.DotNet.Build.Tasks/Targets/CodeCoverage.targets
AlfredoMS: I think having the possibility to do both or only one is good. One reason is doing the runs locally is better to have the options. In practice (CI) coverage run and this runs should be done together.

Both properties are to be set manually.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
AlfredoMS: OpenCover doesn't include that. Even the tool that generates the .html reports implemented their custom parser. Although they use more Linq (https://github.com/danielpalme/ReportGenerator/blob/1deabff63aff52c419bbc811407cf7a0738c2262/ReportGenerator/Parser/OpenCoverParser.cs)

This report simplifies the OpenCover report so I think a custom one is appropiate.

weshaggard:
I didn't dig deeply into the parser but the other pieces look reasonable to me.

src/Microsoft.DotNet.Build.Tasks/ParseTestCoverageInfo.cs
AlfredoMS: I'll considering the future if we get to a point were this task becomes an issue.

PR #151, opened 2015-04-23
Add support for building assemblies with ECMA key

ericstj:
This change adds a property UseECMAKey that can be used to indicate that an assembly
should be signed with the ECMA key b77a5c561934e089 instead of the framework key
b03f5f7f11d50a3a.


stephentoub:
In support of fixing https://github.com/dotnet/corefx/issues/1485

stephentoub:
LGTM

weshaggard:
LGTM

PR #154, opened 2015-05-04
Fix build break when spaces in path

ellismg:
If there are spaces in the path to the repository that we are building
in, the partialfacade targets fail, because the exec command and
argument parsing logic gets confused.

Wrap invocations with quotes, so this doesn't happen.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
@@ -134,11 +134,11 @@
     <Error Text="No single matching contract found." Condition="'@(ResolvedMatchingContract->Count())' != '1'" />
 
     <PropertyGroup>
-      <GenFacadesArgs>-partialFacadeAssemblyPath:$(GenFacadesInputAssembly)</GenFacadesArgs>
-      <GenFacadesArgs>$(GenFacadesArgs) -contracts:%(ResolvedMatchingContract.Identity)</GenFacadesArgs>
-      <GenFacadesArgs>$(GenFacadesArgs) -seeds:@(GenFacadesSeeds, ';')</GenFacadesArgs>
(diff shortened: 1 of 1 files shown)


ellismg:
Noticed this while trying to build at home.

/cc @weshaggard @JeremyKuhne @mellinoe

JeremyKuhne:
LGTM, thanks!

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -134,11 +134,11 @@
     <Error Text="No single matching contract found." Condition="'@(ResolvedMatchingContract->Count())' != '1'" />
 
     <PropertyGroup>
-      <GenFacadesArgs>-partialFacadeAssemblyPath:$(GenFacadesInputAssembly)</GenFacadesArgs>
-      <GenFacadesArgs>$(GenFacadesArgs) -contracts:%(ResolvedMatchingContract.Identity)</GenFacadesArgs>
-      <GenFacadesArgs>$(GenFacadesArgs) -seeds:@(GenFacadesSeeds, ';')</GenFacadesArgs>
-      <GenFacadesArgs>$(GenFacadesArgs) -facadePath:$(GenFacadesOutputPath)</GenFacadesArgs>
-      <GenFacadesArgs Condition="'@(SeedTypePreference)' != ''">$(GenFacadesArgs) -preferSeedType:@(SeedTypePreference->'%(Identity)=%(Assembly)', ',')</GenFacadesArgs>
+      <GenFacadesArgs>-partialFacadeAssemblyPath:&quot;$(GenFacadesInputAssembly)&quot;</GenFacadesArgs>
+      <GenFacadesArgs>$(GenFacadesArgs) -contracts:&quot;%(ResolvedMatchingContract.Identity)&quot;</GenFacadesArgs>
```
weshaggard: You don't actually need to use `&quot;` normal quotes `"` work in property groups.

ellismg:
@weshaggard Updated to use `"` instead of `&quot;` when possible.

PR #160, opened 2015-05-09
InnerLoop support is added to xunit.

Priya91:
This enables user to pass innerloop as a valid value to xunit trait
category. Warnings are no longer reported if there are no tests found in
the test project. The OuterLoop attribute takes OuterLoopCategory as
argument to specify special OuterLoop categories for additional filtering.
To exclude running these special outerloop categories from regular
outerloop runs, the runByDefault value is set to false. With this change,
any new category that is not innerloop should be added by creating a new
OuterLoopCategory object and exposed as a readonly field.

cc @weshaggard @krwq @roncain


Priya91:
closing this PR, as this implementation is are not valid. Seems like attribute parameters should be primitive types or arrays of primitive types.

PR #163, opened 2015-05-13
Check for token type before casting

JeremyKuhne:
On Linux I'm seeing

        "compile": {
          "ref/any/System.Collections.dll": {}
        },

instead of

        "compile": [
          "ref/any/System.Collections.dll"
        ],

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/ResolveNuGetPackageAssets.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/ResolveNuGetPackageAssets.cs
@@ -192,30 +192,42 @@ namespace Microsoft.NuGet.Build.Tasks
 
         private IEnumerable<ITaskItem> CreateItems(string packageName, string packageVersion, JToken packageObject, string key)
         {
-            var values = (JArray)((JObject)packageObject)[key];
+            List<string> values = new List<string>();
+
+            JToken tokenValue = packageObject[key];
+            if (tokenValue != null)
+            {
+                switch (tokenValue.Type)
+                {
+                    case JTokenType.Array:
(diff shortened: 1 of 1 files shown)


JeremyKuhne:
@weshaggard, @akoeplinger, @ericstj

akoeplinger:
Hmm interesting, I'm curious why there'd be a difference, can you point me to a sample where I can see this myself? LGTM otherwise.

JeremyKuhne:
@akoeplinger You can see it in the lock files that get created in the source directories. Well I can, anyway. :) It might possibly be the newer drop of DNU I've got causing this...

akoeplinger:
@JeremyKuhne I'm still seeing the old format with the current DNU. Might causes by https://github.com/aspnet/Announcements/issues/24 on the newer ones?

akoeplinger:
Yeah, the lock file format seems to have changed, with the latest DNX I'm seeing this on Windows as well in an unrelated DNX project:

```
        "compile": {
          "lib/dnx451/EntityFramework.Core.dll": {}
        },
```

JeremyKuhne:
@jasonmalinowski

ericstj:
seems reasonable to me, thanks for making it backwards compatible

jasonmalinowski:
Yep, seems reasonable. Not yet sure if I'm going to pull these back into the internal copy yet since it might not need it there. @ericstj if that becomes a headache then I'll pull it in.

ericstj:
@jasonmalinowski I think if it's showing up in latest DNX it's just a matter of time until that code makes it to Nuget.

jasonmalinowski:
Oh, hmm, I already thought I had consumed that change internally but I guess not.

PR #165, opened 2015-05-19
Fix path combine on Unix.

JeremyKuhne:
If a path starts with '\' on Unix it isn't considered rooted. This causes Path.Combine to not take just the second argument because we've converted it from rooted to unrooted.

@ericstj; @weshaggard


weshaggard:
LGTM

bartonjs:
:shipit:

Issue #125, opened 2015-03-25
BuildTools tests


eatdrinksleepcode:
There are no tests in the BuildTools repo. The Tasks project could use tests both for the custom tasks themselves, as well as the targets files (e.g. importing all of the targets into a master build file and at least ensuring that they parse correctly). I haven't looked enough at the other projects in the repo to know what kinds of tests would be appropriate, but they do contain non-trivial logic that could be tested.

weshaggard:
Yes this is something that I think needs to be done as well it just hasn't come to the top of the list yet. Thanks for filing the issue however. My basic thoughts here are that we should have a template repo layout that consumes the build tools and does some basic validation.

eatdrinksleepcode:
I agree, a template repo would be useful both for testing and for demonstrating the expected usage of the build tools to consuming projects. I would be interested in helping out with that.

weshaggard:
@eatdrinksleepcode That would be awesome if you wanted to put that together. We will put you down as working on that.

ViktorHofer:
Closing as the buildtools repository has been archived and replace by https://github.com/dotnet/arcade.

PR #185, opened 2015-06-24
Allow override of test host and executable

### LINKED ISSUE #132: Support testing against full framework
BuildTools currently only supports testing against Core. However, Microsoft.DotNet.Build.Tasks is built against the full framework, and must continue to be at least until MSBuild is available for Core. Therefore the corresponding unit test project must also be built and run against the full framework. In order for BuildTools to continue dogfooding itself, it must support testing against the full framework.

There are two potential ways to go about this:

1. Allow the TestHost and XunitExecutable properties to be overridden. The project would then be required to set these properties correctly, as well as provide the full framework version of xunit.
2. Expose a UseCoreForTest property (or similar) that makes the required changes automatically, and also pulls in the correct version of xunit. This would make this functionality more reusable, however I am leaning away from it for two reasons:
  * Core vs full framework is already not transparent to the test project; it has to have the correct xunit reference to start with, and the build tools do not provide the compile-time references (although perhaps they could??).
  * It is my impression that few if any projects intend to use these build tools while targeting the full framework. If we put this functionality into the build tools, and then MSBuild becomes available on Core and the BuildTools are converted to Core, then it will be a change to the BuildTools to remove this functionality. If instead we allow these properties to be overridden, and put the full framework references into the unit test project, then it is just a change to the unit test project to take them out. On the other hand, if my assumption is wrong and there are more projects that target the full framework that could benefit from using BuildTools, then that is a stronger argument for 1 over 2.

eatdrinksleepcode:
Fixes #132

Allows the test host and the test executable to be overridden. Also paves the way for #125.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
@@ -19,20 +19,51 @@
     <UnsupportedPlatformsItems Include="$(UnsupportedPlatforms)"/>
   </ItemGroup>
 
+  <PropertyGroup>
+    <TestWithCore Condition="'$(TestWithCore)' == ''">true</TestWithCore>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TestWithCore)' == 'true'">
+    <TestRuntimeProjectJson Condition="'$(TestRuntimeProjectJson)' == ''">$(MSBuildThisFileDirectory)test-runtime\project.json</TestRuntimeProjectJson>
+    <TestRuntimeProjectLockJson Condition="'$(TestRuntimeProjectLockJson)' == ''">$(MSBuildThisFileDirectory)test-runtime\project.lock.json</TestRuntimeProjectLockJson>
+    <TestHost>corerun.exe</TestHost>
+    <XunitExecutable Condition="'$(XunitExecutable)' == ''">xunit.console.netcore.exe</XunitExecutable>
+    <DebugTestFrameworkFolder>dnxcore50</DebugTestFrameworkFolder>
+    <DebugEngines>{2E36F1D4-B23C-435D-AB41-18E608940038}</DebugEngines>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TestWithCore)' != 'true'">
+    <XunitExecutable Condition="'$(XunitExecutable)' == ''">xunit.console.exe</XunitExecutable>
+    <DebugTestFrameworkFolder>dnx45</DebugTestFrameworkFolder>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TestWithCore)' == 'true'">
+    <TestTargetFramework Include="DNXCore,Version=v5.0">
+      <Folder>dnxcore50</Folder>
+    </TestTargetFramework>
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TestWithCore)' != 'true'">
+    <TestTargetFramework Include=".NETFramework,Version=v4.5">
+      <Folder>net45</Folder>
+    </TestTargetFramework>
+  </ItemGroup>
+
   <!-- General xunit options -->
   <PropertyGroup>
-    <XunitHost>corerun.exe</XunitHost>
-    <TestHost>$(XunitHost)</TestHost>
     <XunitResultsFileName>testResults.xml</XunitResultsFileName>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -21,10 +21,12 @@
 
   <!-- General xunit options -->
   <PropertyGroup>
-    <XunitHost>corerun.exe</XunitHost>
-    <TestHost>$(XunitHost)</TestHost>
+    <UseTestHost Condition="'$(UseTestHost)' == ''">true</UseTestHost>
```
weshaggard: So are you assuming that this will be set to false when the test itself it an exe that will run on the full desktop framework and doesn't need a specific host?

weshaggard:
Other than figuring out how to plumb the launch exe through the other extensibility points look good.

eatdrinksleepcode:
Good point about ensuring that the change goes through VS also. I have amended the commit to include changing the VS StartProgram and StartArguments appropriately. However, testing this locally, it seems to work for Start Without Debugging, but attempting to debug results in an error:

> Error while trying to run project: Unable to start program 'C:\Users\danelson\Projects\Personal\dotnet\buildtools\bin\tests\Windows_NT.AnyCPU.Debug\Microsoft.DotNet.Build.Tasks.Tests\dnxcore50\xunit.console.exe'.

> The debugger is not properly installed.  Cannot debug the requested type of code.  Run setup to install or repair the debugger.

I'm not sure what would cause that, or how to troubleshoot it. The VS activity log doesn't show anything useful.

eatdrinksleepcode:
The problem with debugging from VS is demonstrated in my fork at [eatdrinksleepcode/buildtools@dotnet#125-unit-tests](https://github.com/eatdrinksleepcode/buildtools/tree/dotnet%23125-unit-tests), if anyone wants to take a look.

nguerrera:
Which version of VS are you using? The debug engine we specify is only available in Visual Studio 2015 and in some builds, Web Tools were required to be installed.

eatdrinksleepcode:
Thanks @nguerrera for the pointer to the debug engine; that was the problem.

I wanted to allow debugging tests on the full framework and in VS 2013, so I wanted to allow for overriding the DebugEngines. That started to feel like too many things to override individually when they were actually all related. So I decided to move towards option 2 from #132, creating collections of properties that could be selected en masse. At first I tried to do that in the metadata of the TestTargetFramework item group. The problem is that metadata can only be used in targets; but we need to access these properties to set the VS Start properties, which (as far as I can tell) can't be done in a target. And far as I know, the only thing that using metadata gets us is the ability to run the related tasks for multiple different target frameworks, which at least in the short term is not the goal.

So I reverted to setting the properties with a condition based on a new "TestWithCore" property. I don't exactly love the solution, but it is the best I can come up with right now. Happy to hear other opinions.

nguerrera:
I'm OK with the TestWithCore approach as a starting point. We can evolve this as our needs evolve. This change looks good to me.

src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets
```diff
@@ -19,20 +19,51 @@
...
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TestWithCore)' == 'true'">
+    <TestTargetFramework Include="DNXCore,Version=v5.0">
+      <Folder>dnxcore50</Folder>
+    </TestTargetFramework>
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TestWithCore)' != 'true'">
+    <TestTargetFramework Include="DNX,Version=v4.5">
```
nguerrera: Should this be .NETFramework,Version=v4.5 and net45? We aren't set up to actually use DNX. Even the DNXCore part above will change I believe.

/cc @ericstj

ericstj: Yes, don't use DNX here.  The only reason we are using DNXCore above is for asset selection (we want the libraries that are coincidentally set up for the same OS/arch pivots as DNX).  Eventually we should probably switch to netcore, or potentially "dotnet" for the coreclr based test run if we are able to make the runtime/facade packages work with that.

eatdrinksleepcode:
Changed the TestTargetFramework name and folder.

nguerrera:
LGTM

SajayAntony:
What is the action in VS that we need to perform so that the desktop version is picked up?
I would like to make sure we get this other repos as well.

PR #186, opened 2015-06-25
Exclude injected SR.cs/vb file from code coverage

stephentoub:
This avoids any additional custom attributes in corefx and avoids churning the SR.cs/vb files.  We'll just need to remember to update build tools if we ever move those files to a different location or otherwise rename them, or if we put additional code into those files that for some reason shouldn't be excluded.


stephentoub:
cc: @weshaggard, @tarekgh, @nguerrera

tarekgh:
even better :-) LGTM

weshaggard:
LGTM

stephentoub:
Thanks, guys.

Issue #182, opened 2015-06-18
Copying of test assets should resolve to highest package version.


nguerrera:
Currently, in the CopyTestsToTestDirectories target, we sometimes end up with the same dll from different package versions (I believe this happens when product dependencies outpace the coreclr/buildtool package dependencies but I did not confirm). Right now,  we just arbitrarily copy the last one in the list over the previous ones with the same file name. In addition to not being very well defined, it breaks SkipCopyUnchnagedFiles.

nguerrera:
I'm making a tactical fix for the SkipCopyUnchangedFiles issues, but the root issue will still need to be addressed.

MichaelSimons:
Triage: no longer applicable.

PR #189, opened 2015-07-02
Fix VS debugging iteration experience

### LINKED ISSUE #181: VS debugging experience broken
CopyTestToTestDirectory now only happens when the RunTestsForProject target is invoked. This broke the VS flow where build got everything ready and F5 could debug. Now I can only debug in VS if I run the tests on the command line first to copy the right files.

nguerrera:
Move test deployment to build phase and speed it up

The experience around building and running tests in VS got broken by a
change to only copy tests when they're run from msbuild. Move them back
to PrepareForRun to bring back the VS workflow.

In doing so, it became apparent that the test deployment step was
copying files on every invocation regardless of whether there were any
changes. That turned out to be because there are different versions of
files with the same simple name destind for the same output folder. As
such, every iteration would swap them back and forth. The fix here is
too trim out the duplicates before invoking the Copy task. Some of the
duplication is by design (e.g. to override nuget assets with local
assets), but we are also getting multiple nuget package versions out
of the nuget asset resolution, which seems like a bug, which is now filed
as #182.

To further speed things up, using hard-links is now always on for the
test deployment step, which currently copies tons of the same files
to each test folder.

Fix #181

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/RemoveDuplicatesWithLastOneWinsPolicy.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/RemoveDuplicatesWithLastOneWinsPolicy.cs
@@ -1,0 +1,62 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    /// <summary>
+    /// Given a list of items, remove duplicate items. Metadata is not considered. Case insensitive.
+    /// Unlike <see cref="RemoveDuplicates"/>, the last item in the list wins rather than the first.
+    /// That distinction is important when the items have different metadata.
+    /// </summary>
+    public class RemoveDuplicatesWithLastOneWinsPolicy : Task
+    {
+        /// <summary>
+        /// The list of items from which to remove duplicates.
+        /// </summary>
+        public ITaskItem[] Inputs
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// The list of items with duplicates removed.
+        /// </summary>
+        [Output]
+        public ITaskItem[] Filtered
+        {
+            get;
+            set;
+        }
+
+        public override bool Execute()
+        {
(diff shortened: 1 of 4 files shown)


nguerrera:
cc @JeremyKuhne @FiveTimesTheFun @stephentoub

stephentoub:
> to bring back the VS workflow

yay

src/Microsoft.DotNet.Build.Tasks/RemoveDuplicatesWithLastOneWinsPolicy.cs
```diff
@@ -0,0 +1,62 @@
...
+                }
+                else
+                {
+                    filteredList.Add(item);
+                    existingIndexMap.Add(item.ItemSpec, filteredList.Count - 1);
+                }
+            }
+
+            Filtered = filteredList.ToArray();
+            return true;
```
stephentoub: As a complete aside, the proposal https://github.com/dotnet/corefx/issues/2146 would help here.  With a DistinctBy method, this whole thing could become:
```C#
filtered = Inputs.DistinctBy(item => item.ItemSpec, StringComparer.OrdinalIgnoreCase).ToArray();
return true;
```
It could also be simplified if we had the equivalent of ```Comparer<T>.CreateComparer``` but for ```IEqualityComparer<T>```, e.g. ```public static IEqualityComparer<T> Create(Func<T, int> getHashCode, Func<T, T, bool> equals);```.  With something like that, you could do the same thing with Distinct.

stephentoub:
LGTM.  Thanks for doing this.

src/Microsoft.DotNet.Build.Tasks/RemoveDuplicatesWithLastOneWinsPolicy.cs
nguerrera: Interesting. It looks like me like Distinct picks the first match and not the last one. I assume DistinctBy would do the same. 

https://github.com/dotnet/corefx/blob/872bb70953bcfe0a47dcc6e9789fbd330e295bc3/src/System.Linq/src/System/Linq/Enumerable.cs#L924

Mind you, I suppose one could reverse the input beforehand.

stephentoub: Ah, yeah, I completely missed the comment about explicitly desiring to pick the last.

JeremyKuhne:
LGTM.

Issue #192, opened 2015-07-12
Make buildtools defer to built-in nuget package resolution when available


nguerrera:
The buildtools resolution of nuget assets was a bootstrapping mechanism as we got things to build with the .NET Core packages pending official support. Once the official support is broadly available, we should change the targets to detect this and use the built-in targets wherever possible. 

I'm currently preparing a temporary hack to rename some things such that buildtools does not conflict with VS and always uses its own targets. This is just to unblock things for internal devs quickly. 

Once possible, we will need to decide if we can take a required dependency on a version of VS with the built-in targets or if we want to allow building with older versions of VS by falling back to the code in buildtools.

cc @ericstj @ljw1004

MichaelSimons:
Triage: nothing we can do for the moment.

PR #193, opened 2015-07-12
Avoid conflict with built-in nuget targets

nguerrera:
This is a quick change to avoid naming collision and always disable
the built-in nuget resolution targets in favor of what we've used
in buildtools to bootstrap building against the .NET Core packages.

#192 tracks getting things fixed so that we actually consume the
built-in targets when they're available.

cc @ericstj @ljw1004


weshaggard:
Is "Downlevel" the best prefix? I assume you are referring to downlevel VS but it is kind of overloaded with OS down level. What about Prerelease or CoreFx? Otherwise the change looks good to me.

nguerrera:
That is what I meant, but I'll change it to Prerelease because I like it better. :)

JeremyKuhne:
LGTM- thanks!

PR #194, opened 2015-07-16
Skip restoring packages in VS2015 background build

### LINKED ISSUE #191: BuildTools can cause VS 2015 to deadlock in the design-time build when loading a project that uses them
Roslyn uses the buildtools for core-clr projects and we've found that VS 2015 deadlocks while running the dnx.exe binary continuously in the background when loading a project that imports the buildtools targets. Try opening https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/CscCore/CscCore.csproj in VS 2015 for a repro of the issue.

JeremyKuhne:
This avoids a deadlock with VS endlessly resolving references. Fixes #191. @weshaggard, @nguerrera

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003"  InitialTargets="CheckDesignTime">
 
   <PropertyGroup>
     <!-- A number of the imports below depend on these default properties -->
@@ -105,6 +105,19 @@
         DnuRestoreCommand - Used to restore the project packages from project.json
         PackagesDir - Packages are restored to and resolved from this location    
    -->
-  <Import Project="$(MSBuildThisFileDirectory)tests.targets" Condition="'$(IsTestProject)'=='true' and '$(ExcldueTestsImport)'!='true'"/> 
+  <Import Project="$(MSBuildThisFileDirectory)tests.targets" Condition="'$(IsTestProject)'=='true' and '$(ExcldueTestsImport)'!='true'"/>
 
+  <Target Name="CheckDesignTime">
+    <!--
+      Visual Studio does a number of background builds to do a variety of tasks such as resolving references and preparing for intellisense.
+      These are called "design time" builds. You can only determine this state within a target as the properties VS sets are added at build time.
+
+      To see design time logs set TRACEDESIGNTIME=true before launching Visual Studio. Logs will go to %TEMP%.
+
+      Note that the existing $(DesignTimeBuild) is not set for all background builds.
+    -->
+    <PropertyGroup>
+      <VSDesignTimeBuild Condition="'$(BuildingInsideVisualStudio)'=='true' and '$(BuildingOutOfProcess)'=='false'">true</VSDesignTimeBuild>
+    </PropertyGroup>
+  </Target>
 </Project>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
```diff
@@ -105,6 +105,17 @@
         DnuRestoreCommand - Used to restore the project packages from project.json
         PackagesDir - Packages are restored to and resolved from this location    
    -->
-  <Import Project="$(MSBuildThisFileDirectory)tests.targets" Condition="'$(IsTestProject)'=='true' and '$(ExcldueTestsImport)'!='true'"/> 
+  <Import Project="$(MSBuildThisFileDirectory)tests.targets" Condition="'$(IsTestProject)'=='true' and '$(ExcldueTestsImport)'!='true'"/>
 
+  <Target Name="CheckDesignTime">
+    <!--
+      Visual Studio does a number of background builds to do a variety of tasks such as resolving references and preparing for intellisense.
+      These are called "design time" builds. You can only determine this state within a target as the properties VS sets are added at build time.
```
nguerrera: I see '$(DesignTimeBuild)'=='true' checks in the common targets. Why is that not enough to use here? If it truly isn't, calling that out here in the comment would be helpful.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -23,9 +22,10 @@
     <ResolveNugetPackages>false</ResolveNugetPackages>
   </PropertyGroup>
 
-  <Target Name="RestorePackages" 
+  <!-- Restoring packages during a background (designtime) build will cause VS 2015 (v14) to get into an endless loop of resolving references. -->
+  <Target Name="RestorePackages"
           BeforeTargets="ResolveNuGetPackages"
-          Condition="'$(RestorePackages)'=='true'">
+          Condition="'$(RestorePackages)'=='true' and !('$(VSDesignTimeBuild)'=='true' and '$(VisualStudioVersion)' >= '14.0')">
```
nguerrera: I would not qualify this on VS version. Shouldn't we simply stop doing this everywhere?

nguerrera:
In general, LGTM, but I left a couple of comments/questions.

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
JeremyKuhne: It isn't actually set for all of the background builds, resolve references in particular. I waffled on terminology here but settled on DesignTime because of the logging variable. I'll add a remark here that $(DesignTimeBuild) isn't always set.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
JeremyKuhne: Probably should, wanted to keep it tight for now to unblock the hang and make a different change to skip initial target stuff entirely.

nguerrera: OK.

nguerrera:
LGTM.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
weshaggard: I would prefer that we actually use those extra conditions to set RestorePackages property instead of adding it here. We already do that for the DesignTimeBuild property already.

JeremyKuhne: Unfortunately we can only do that if we do it in target as it isn't available until build time. I didn't want to introduce another dependent target for RestorePackages. I also wanted to keep the initial target I added central and clean as I expect we'll need to consume the property elsewhere.

weshaggard: I missed the fact it was being set in the target. Now I have a different question however which is where is the CheckDesignTime actually being called?

JeremyKuhne: InitialTargets in our Build.Common.targets above

PR #196, opened 2015-07-17
Fix and extend TestCopyLocal

JeremyKuhne:
Allow overriding where the local libraries come from and search in .CoreCLR directories as well.

@weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
@@ -44,15 +44,21 @@
     <ItemGroup>
       <TestCopyLocal Include="@(RunTestsForProjectInputs)" Exclude="@(PackagesConfigs)" />
     </ItemGroup>
-    
+
     <!-- Test using locally built libraries if requested. Again, later entries override earlier
          entries. -->
+    <PropertyGroup Condition="'$(TestWithLocalLibraries)'=='true'">
+      <BaseLocalLibrariesPath Condition="'$(BaseLocalLibrariesPath)'==''">$(BaseOutputPath)</BaseLocalLibrariesPath>
+      <LocalLibrariesPath Condition="'$(LocalLibrariesPath)'==''">$(BaseLocalLibrariesPath)$(OSPlatformConfig)</LocalLibrariesPath>
+    </PropertyGroup>
     <ItemGroup Condition="'$(TestWithLocalLibraries)'=='true'">
       <!-- Replace some of the resolved libraries that came from nuget by exploring the list of files that we are going to copy
            and replacing them with local copies that were just built -->
-      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(BaseOutputPath)$(OSPlatformConfig)\%(filename)\%(filename).dll')" />
-      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(BaseOutputPath)$(OSPlatformConfig)\%(filename)\%(filename).pdb')" />
-      <_ExistingReplacementCandidate Include="@(_ReplacementCandidates)" Condition="Exists('%(_ReplacementCandidates.FullPath)')" />      
+      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(LocalLibrariesPath)\%(filename)\%(filename).dll')" />
+      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(LocalLibrariesPath)\%(filename)\%(filename).pdb')" />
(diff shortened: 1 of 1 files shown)


JeremyKuhne:
@AlfredoMS, @nguerrera

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -44,15 +44,21 @@
     <ItemGroup>
       <TestCopyLocal Include="@(RunTestsForProjectInputs)" Exclude="@(PackagesConfigs)" />
     </ItemGroup>
-    
+
     <!-- Test using locally built libraries if requested. Again, later entries override earlier
          entries. -->
+    <PropertyGroup Condition="'$(TestWithLocalLibraries)'=='true'">
+      <BaseLocalLibrariesPath Condition="'$(BaseLocalLibrariesPath)'==''">$(BaseOutputPath)</BaseLocalLibrariesPath>
```
AlfredoMS: in which case is 'BaseLocalLibrariesPath' going to be set beforehand?
I just want to understand if the new properties are for simplifying the transformations in theItemGroups below or if there is any other reason.

JeremyKuhne: Using it for a CoreFx based console app that consumes the build tools. Allows me to consume the CoreFx libraries I'm building when I'm not in the same location.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -44,15 +44,21 @@
...
     <ItemGroup Condition="'$(TestWithLocalLibraries)'=='true'">
       <!-- Replace some of the resolved libraries that came from nuget by exploring the list of files that we are going to copy
            and replacing them with local copies that were just built -->
-      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(BaseOutputPath)$(OSPlatformConfig)\%(filename)\%(filename).dll')" />
-      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(BaseOutputPath)$(OSPlatformConfig)\%(filename)\%(filename).pdb')" />
-      <_ExistingReplacementCandidate Include="@(_ReplacementCandidates)" Condition="Exists('%(_ReplacementCandidates.FullPath)')" />      
+      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(LocalLibrariesPath)\%(filename)\%(filename).dll')" />
+      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(LocalLibrariesPath)\%(filename)\%(filename).pdb')" />
+      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(LocalLibrariesPath)\%(filename).CoreCLR\%(filename).dll')" />
+      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(LocalLibrariesPath)\%(filename).CoreCLR\%(filename).pdb')" />
```
weshaggard: @mellinoe Do you know why we have a .CoreCLR postfix on the partial facades? Can we eliminate that?

mellinoe: We just have that to distinguish them from our ".NETNative" ones internally. Out in the open it makes less sense because they are the only ones there for the time being.

weshaggard: OK we will need to cross this bring in the future but @JeremyKuhne I would prefer we not have the output folder resolve logic hardcoded here but instead update the partial façade targets to not put these into a folder with the ".CoreCLR" postfix.

JeremyKuhne: @weshaggard  Can we open a separate issue on removing .CoreCLR and remove this when we do? This is blocking me. Are you also suggesting that we remove \%(filename)\ and search recursively, or some other sort of finding assemblies?

weshaggard: Actually I was hoping that this was just a convention we could tweak in the targets but I don't think that is the case now as we just get that directory from the project name. 
`<OutputPath Condition="'$(OutputPath)'==''">$(BaseOutputPath)$(OSPlatformConfig)\$(MSBuildProjectName)\</OutputPath>`

And I would like to keep the project names with such conventions so we will need to figure out some other logic for this at some point. As "CoreCLR" will not work once we also have "Net Native" and "Desktop" libraries as well. We will likely need to have some configurable convention to search for base on the target you want to run on. That can happen at some later point however please go ahead and move forward with this for now.

weshaggard:
LGTM

PR #198, opened 2015-07-17
Move to modern buildtools package and targets

mellinoe:
This updates the build targets and NuGet package version that is being used to build this repo. My basic process for updating this was as follows:

1. Copy targets files over from corefx
2. Remove corefx-specific things from the targets, like specific workarounds for corefx projects, etc.
3. Update packages.config files to project.json files for our .NET Core projects here (the two Xunit projects).
4. Re-lock the dependencies
5. Verify repo builds from clean

NOTE: I have left some of the x-plat specific targets files, like mono.targets and roslyn.xplat.targets, but I haven't built this on Linux. The stuff in those files is relatively harmless (disabling PDB's, etc.) and will probably be needed to build this on Linux anyways.

This is in preparation for getting GenFacades building in this repository.


mellinoe:
@weshaggard, @FiveTimesTheFun , @JeremyKuhne

JeremyKuhne:
LGTM

mellinoe:
Thanks, Jeremy. The previous CI build failed because I accidentally had a local NuGet package source in the props file that I was using for testing.

mellinoe:
#199 was merged in, so re-committed on top of that, and included the new Xunit version in the project.json.

mellinoe:
@weshaggard could you take a quick look?

src/xunit.console.netcore/project.json
```diff
@@ -0,0 +1,22 @@
...
+    "System.Reflection": "4.0.10-beta-*",
+    "System.Resources.ResourceManager": "4.0.0-beta-*",
+    "System.Runtime.Extensions": "4.0.10-beta-*",
+    "System.Xml.XDocument": "4.0.10-beta-*",
+    "xunit.abstractions": "2.0.0.0",
+    "xunit.extensibility.execution": "2.1.0-beta2-build2981",
+    "xunit.runner.utility": "2.1.0-beta2-build2981",
+  },
+  "frameworks": {
+    "dnxcore50": {}
```
weshaggard: doesn't this need netcore50 as well?

weshaggard:
Man there is a lot of stuff that is needed for a repo to consume buildtools... we really need to work on making it less work.

Changes look good to me.

src/xunit.console.netcore/project.json
mellinoe: I'm not sure it makes a difference actually, especially for the set of things we are referencing (contracts mostly)

weshaggard:
One other verification that is worth looking at is using a build tools package build after your changes inside of corefx to ensure we didn't break any of the dependencies while switching how the libraries build. I'm fine with that happening as part of your other work however but we need to ensure we don't block the next build tools package release.

mellinoe:
> Man there is a lot of stuff that is needed for a repo to consume buildtools... we really need to work on making it less work.

Yeah, I have set up a few independent repos over the past couple months and it has grown more and more complex over time. Some of the files we have added should eventually go away (like all of the x-plat, mono, roslyn workaround files), but that still leaves quite a few targets and properties.

weshaggard:
@mellinoe would you mind filing a work item about documenting and minimizing the set of things needed to for a repo to consume build tools?

src/xunit.console.netcore/project.json
weshaggard: Yes I do believe we want all these to eventually be only netcore50 but I'm fine with this as is for now as long as it works.

mellinoe:
Filed #200

JeremyKuhne:
@mellinoe, @weshaggard For some reason this change hangs the official build. It isn't clear by the limited logs that are produced as to why. Normal builds finish in 5-6 minutes, let one go for 45 and another for around 12. Built fine after reverting.

Issue #195, opened 2015-07-16
Buildtools has an outdated build process


mellinoe:
Currently, we're using a really old version of buildtools, and the .NET Core Xunit assemblies we build in this repo are building against very old versions of our NuGet packages (very different from the latest packages), and are still using packages.config files. We should update these to be in line with corefx.

ericstj:
I'd like to fix this with https://github.com/dotnet/corefx/issues/2427 instead

mellinoe:
Did you mean to close this? I'm fine with making the same improvement to both repos if that's what we intend to do.

weshaggard:
Given @JeremyKuhne reverted your update @mellinoe we should reopen this issue. I think @ericstj is referring to a more general update away from dnx and less of you intended with this issue which is to update the buildtools repo itself to use the latest buildtools.

mellinoe:
Sounds good to me. I think it makes sense to get this repo on the latest buildtools, because then when we make improvements to corefx's build process, we can just apply them here as well, instead of having to jump straight from this old build process.

mellinoe:
#205 merged and the official build went through fine. This particular issue (being on an old version of buildtools) should be resolved now.

weshaggard:
@mellinoe Have you tried the new build against CoreFX to ensure everything still works as expected?

mellinoe:
CoreFx is already on the new tools and hasn't hit any issues so far. There aren't really any functional changes here, other than possibly the fact that we are using new packages as opposed to the months-old ones.

PR #205, opened 2015-07-21
Move to modern buildtools package and targets

mellinoe:
This reverts commit c27806b8604141172dca595f757ee6438ee7f0f0.

This was reverted sometime over the weekend, but I would still like to get this through to move my other work forward. I'm not sure why our internal build had failed, but we can address that if it comes up. I copied that internal build definition and pointed it to my fork of this repo and saw that it built fine in a few minutes, so there may have just been some point-in-time issue, perhaps with NuGet and package restore(?)

Regarding #195 and dotnet/corefx#2427, I think this change will help, because it will bring the build process here in sync with corefx. If we make changes to move corefx to a newer system, we can then apply the same changes here, instead of moving this ancient build process to a brand new one.


mellinoe:
Please take a look, @weshaggard , @JeremyKuhne

mellinoe:
I kicked off another internal build pointed at this branch and it succeeded again. The only differences are my build didn't publish the NuGet package automatically, and didn't update the build numbers, which shouldn't have been affected by any of my changes anyways.

PR #207, opened 2015-07-21
Add a build for GenFacades.exe and dependencies

mellinoe:
This adds the source code for GenFacades and its dependencies, with builds for both Desktop (.NETFramework 4.6) and CoreCLR (using CoreConsole.exe as a host). There is an extra targets file (publishexe.targets) which handles some of the extra configuration needed to create a CoreCLR-based Console application, including the runtime, host, framework libraries, etc. I've included nuspec files to create nupkg's for both versions of GenFacades (CoreCLR and Desktop) which contain fully-standalone tools. The Desktop verison currently requires .NET 4.6 since it compiles against the contracts and therefore needs some of the in-box facades.

The bulk of the commit is the Microsoft.Cci.Extensions library. We will probably eventually trim this dependency out, at least for our basic tools like GenFacades.

When we start shipping our cross-plat NuGet packages for our framework libraries, and for the runtime/hosts, we should be able to add another build configuration and easily produce cross-plat versions of GenFacades.


mellinoe:
@weshaggard , @nguerrera

mellinoe:
I will need to upload the Microsoft.Cci NuGet package from our internal build for this to pass the CI build.

mellinoe:
@dotnet-bot  Test this, please

mellinoe:
@dotnet test this, please

mellinoe:
The CI build passed but I believe we are going to need to install the .NET 4.6 SDK onto our internal build machines in order to build the GenFacades.Desktop project.

src/GenFacades.Desktop/GenFacades.Desktop.csproj
```diff
@@ -0,0 +1,55 @@
...
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <OutputType>Exe</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>GenFacades</RootNamespace>
+    <AssemblyName>GenFacades</AssemblyName>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
```
weshaggard: Is there a reason this needs to be 4.6 as opposed to 4.5?

src/GenFacades.Desktop/project.json
```diff
@@ -0,0 +1,40 @@
+﻿{
+    "dependencies": {
+        "System.Collections": "4.0.10-beta-*",
```
weshaggard: I assume these are all dependencies that Microsoft.CCI.Extensions uses? At first glance it definitely looks like more than is necessary. I'm pretty sure that CCI.Extensions doesn't use Immutable or Concurrent or AppContext for example. I also thought you removed the console dependency.

src/Microsoft.Cci.Extensions/project.json
```diff
@@ -0,0 +1,36 @@
+﻿{
+    "dependencies": {
+        "System.AppContext": "4.0.0-beta-*",
```
weshaggard: Would you please strip these down to just what is necessary? I want to understand the real dependencies this guy has.

src/common/CommandLine.cs
```diff
@@ -0,0 +1,1820 @@
...
+
+            if (_mustParseHelpStrings)
+                AddHelp(new CommandLineParameter(name, null, helpText, typeof(bool), true, _noDashOnParameterSets, true));
+            if (_skipParameterSets)
+                return false;
+
+            // Have we just finish with the parameter set that was actually on the command line?
+            if (_parameterSetName != null)
+            {
+                _skipDefinition
```
weshaggard: This could be changed to typeof(<Type this is n)).GetTypeInfo().Assembly

src/GenFacades/Program.cs
```diff
@@ -0,0 +1,714 @@
...
+            ErrorTreatment seedLoadErrorTreatment = ErrorTreatment.Default;
+            ErrorTreatment contractLoadErrorTreatment = ErrorTreatment.Default;
+            string[] seedTypePreferencesUnsplit = null;
+            bool forceZeroVersionSeeds = false;
+            string partialFacadeAssemblyPath = null;
+
+#if COREFX
+            string commandLine = "GenFacades.exe " + string.Join(" ", args); // CommandLine parser assumes first arg is exe name.
+#else
+            string commandLine = Environment.CommandLine;
```
weshaggard: Isn't there an overload that just takes the args that we can switch to?

src/Microsoft.Cci.Extensions/Differs/ElementDifferenceFactory.cs
```diff
@@ -0,0 +1,76 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using Microsoft.Cci.Mappings;
+
+#if COREFX
```
weshaggard: MEF is an interesting dependency. We may be able to switch to System.Composition all the time for our usages in this library.

src/Microsoft.Cci.Extensions/Extensions/AssemblyIdentityHelpers.cs
```diff
@@ -0,0 +1,45 @@
...
+#endif
+            return name.ToString();
+        }
+
+        public static AssemblyIdentity Parse(INameTable nameTable, string formattedName)
+        {
+            var name = new System.Reflection.AssemblyName(formattedName);
+            return new AssemblyIdentity(nameTable.GetNameFor(name.Name),
+#if COREFX
+                                        name.CultureName,
```
weshaggard: Should we switch to this version all the time?

src/Microsoft.Cci.Extensions/Extensions/AssemblyIdentityHelpers.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Microsoft.Cci.Extensions
+{
+    public static class AssemblyIdentityHelpers
+    {
+        public static string Format(this AssemblyIdentity assemblyIdentity)
+        {
+            var name = new System.Reflection.AssemblyName();
+            name.Name = assemblyIdentity.Name.Value;
+#if !COREFX
+            name.CultureInfo = new CultureInfo(assemblyIdentity.Culture);
```
weshaggard: Some of these are going to potentially cause differences and maybe even bugs it is unclear looking at this in isolation.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
```diff
@@ -0,0 +1,419 @@
...
+                WriteFakeAttribute(typeName, writeInline: true, parameters: enumValue);
+            }
+        }
+
+        private System.Runtime.CompilerServices.MethodImplOptions CreateMethodImplOptions(IMethodDefinition method)
+        {
+            System.Runtime.CompilerServices.MethodImplOptions options = default(System.Runtime.CompilerServices.MethodImplOptions);
+            if (method.IsUnmanaged)
+#if COREFX
+                options |= System.Runtime.CompilerServices.MethodImplOptionsEx.Unmanaged;
```
weshaggard: Did you define MethodImplOptionsEx yourself somewhere?

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
```diff
@@ -0,0 +1,419 @@
...
+                case '\v': return @"\v";
+                case '\0': return @"\0";
+                case '\a': return @"\a";
+                case '\b': return @"\b";
+                case '\\': return @"\\";
+                case '\'': return inString ? "'" : @"\'";
+                case '"': return inString ? "\\\"" : "\"";
+            }
+#if COREFX
+            var cat = CharUnicodeInfo.GetUnicodeCategory(c);
```
weshaggard: We should consider switching to this all the time.

src/common/CommandLine.cs
```diff
@@ -0,0 +1,1820 @@
...
+
+        // These routines are typically are not needed because ParseArgsForConsoleApp does the work.  
+#if !COREFX
+        public CommandLineParser() : this(Environment.CommandLine) { }
+#endif
+        public CommandLineParser(string commandLine)
+        {
+            ParseWords(commandLine);
+        }
+        public CommandLineParser(string[] args)
```
weshaggard: We should just use this overload in our programs.

src/common/CommandLine.cs
```diff
@@ -0,0 +1,1820 @@
...
+            if (parameterSetBody == 0 && parameterSetName != String.Empty)
+                return String.Empty;
+
+            // At this point parameterSetBody and globalParametersEnd are properly set. Start generating strings
+            StringBuilder sb = new StringBuilder();
+
+            // Create the 'Usage' line;
+            // TODO: Find a better way to obtain this information.
+#if COREFX
+            string appName = "<AppName>";
```
weshaggard: It is probably worth adding a overload to this method that passes in the AppName.

weshaggard:
I did a highlevel overview and searched for all usages of the COREFX define and reviewed those changes. It seems reasonable overall but I provided some feedback we should address.

src/common/CommandLine.cs
mellinoe: Did you put this comment on the wrong line?

src/Microsoft.Cci.Extensions/project.json
mellinoe: Yeah I will trim this down, it has way too much unnecessary stuff in it right now.

src/GenFacades.Desktop/GenFacades.Desktop.csproj
mellinoe: Correct

src/GenFacades/Program.cs
mellinoe: I should probably add one. Previously there wasn't any way to pass in the arguments, and Environment.CommandLine was read automatically. There's also some logic baked into the parsing code to ignore the first argument (it's assumed that it is the executable name) so that would have to be updated as well.

src/Microsoft.Cci.Extensions/Extensions/AssemblyIdentityHelpers.cs
mellinoe: It's definitely possible. I looked through the code using it  and I believe the only things using this particular helper are just using it for string matching / hashing in ApiReviewer and ApiPort. They may just continue to work if all the strings generated by this are still consistent with each other. If we move those tools to CoreCLR, we will have to make sure that any behavior differences caused by this are okay. In the meantime, the original behavior is still available if you don't define COREFX.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
mellinoe: Yes, it's in PortingHelpers.cs

src/common/CommandLine.cs
weshaggard: Looks like something got miss aligned there :) I was trying to add that comment for the place where we were calling GetExecutingAssembly()

mellinoe: Ah yes it actually shows up correctly if I open the "Full View". And you're right, since this is compiled into the application now instead of being a separate library, that would actually work fine and return the correct name.

weshaggard: Even as a separate library that change is correct. GetExecutingAssembly() is different from GetEntryAssembly() (perhaps this code should have been using GetEntryAssembly I didn't dig to closely).

mellinoe: Yeah I had to do a double-take here because I was a bit confused... It uses GetExecutingAssembly which would normally return Microsoft.Fx.CommandLine (or whatever it's called), but we have been ILMerge'ing all of the assemblies together, which would basically cause it to return the "Entry" assembly, or whatever it has been ILMerge'd into. In any case, GetExecutingAssembly() will probably give us what we want as long as this is included as a source in compilation.

mellinoe:
@weshaggard I added some fixes addressing most of what you suggested. This includes the changes I submitted internally as well.

src/Microsoft.Cci.Extensions/Differs/Rules/TokenListDiffer.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.Cci.Filters;
+
+namespace Microsoft.Cci.Differs.Rules
+{
+    [ExportDifferenceRule]
+    public class TokenListDiffer : DifferenceRule
+    {
+        [Import(AllowDefault = true)]
+#if COREFX
+        private IDiffingService _diffingService { get; set; }
```
weshaggard: We should just switch this to a properly named property for both cases. Both versions of MEF support properties.

src/common/CommandLine.cs
weshaggard: Is there a reason we cannot switch to using the arg overload here instead of doing the forking in GenFacades?

src/GenFacades/Program.cs
weshaggard: It looks like there is already one, I called it out with another comment. I would switch both cases to using that here.

weshaggard:
Just a few more pieces of feedback that I think we should handle before this gets merges.

src/common/CommandLine.cs
mellinoe: Switched. I had thought that overload did something else originally, my mistake.

mellinoe:
@weshaggard Fixed the rest of the suggestions

weshaggard:
LGTM

PR #208, opened 2015-07-22
Only hard-link by default on Windows

### LINKED ISSUE #197: Recent buildtools update broke building corefx on Linux
In https://github.com/dotnet/buildtools/commit/2343d6f36d96970b97cea6f210db90d54d2f7183 the Copy task in publishtest.targets was set to `UseHardlinksIfPossible="true"`. However, the xplat MSBuild doesn't implement hardlinks in a cross-platform way currently (it just pinvokes into CreateHardlink Windows API), so this crashes on !Windows.

We should probably check for the OS in publishtest.targets and try hardlinks only on Windows (since tests aren't run via msbuild on other platforms yet anyway).

/cc @JeremyKuhne @nguerrera

nguerrera:
CreateHardLinksIfPossible does not yet work in x-plat msbuild.

Fix #197 

cc @chcosta @JeremyKuhne @akoeplinger

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
@@ -74,8 +74,13 @@
       <Output TaskParameter="Filtered" ItemName="_TestCopyLocalByFileNameWithoutDuplicates" />
     </RemoveDuplicatesWithLastOneWinsPolicy>
 
-    <!-- NOTE: UseHardLinksIfPossible is hard-coded to true because we copy tons of the 
-         same files for every single test project here. -->
+    <!-- Ideally, we'd have UseHardLinksIfPossible on by default because we copy tons of the
+         same files for every single test project here, but it doesn't currently work x-plat.
+         so we only turn it on by default for Windows builds for now. -->
+    <PropertyGroup>
+      <CreateHardLinksForCopyTestToTestDirectoryIfPossible Condition="'$(CreateHardLinksForCopyTestToTestDirectoryIfPossible)'=='' and '$(OS)' == 'Windows_NT'">true</CreateHardLinksForCopyTestToTestDirectoryIfPossible>
+      <CreateHardLinksForCopyTestToTestDirectoryIfPossible Condition="'$(CreateHardLinksForCopyTestToTestDirectoryIfPossible)'==''">$(CreateHardLinksForCopyFilesToOutputDirectoryIfPossible)</CreateHardLinksForCopyTestToTestDirectoryIfPossible>
+   </PropertyGroup>
     <Copy
       SourceFiles="@(_TestCopyLocalByFileNameWithoutDuplicates->'%(SourcePath)')"
       DestinationFolder="$(TestPath)%(TestTargetFramework.Folder)"
@@ -83,7 +88,7 @@
       OverwriteReadOnlyFiles="$(OverwriteReadOnlyFiles)"
       Retries="$(CopyRetryCount)"
       RetryDelayMilliseconds="$(CopyRetryDelayMilliseconds)"
-      UseHardlinksIfPossible="true">
+      UseHardlinksIfPossible="$(CreateHardLinksForCopyTestToTestDirectoryIfPossible)">
       
       <Output TaskParameter="DestinationFiles" ItemName="FileWrites" />
     </Copy>


src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -83,7 +83,7 @@
       OverwriteReadOnlyFiles="$(OverwriteReadOnlyFiles)"
       Retries="$(CopyRetryCount)"
       RetryDelayMilliseconds="$(CopyRetryDelayMilliseconds)"
-      UseHardlinksIfPossible="true">
+      UseHardlinksIfPossible="$(CreateHardLinksForCopyFilesToOutputDirectoryIfPossible)">
```
bartonjs: If x-plat is the only concern, could we set this to true by default when building on Windows?

bartonjs: That is, a default value for CreateHardLinksForCopyFilesToOutputDirectoryIfPossible of true when $(OS)=='Windows_NT'

nguerrera: Ah, was about to ask how to detect host OS in msbuild, but that would be good enough for the common case of iterating on Windows.

JeremyKuhne: That should be fine until we get an updated MSBuild package.

nguerrera: Ugh, I'm having trouble testing that condition. When I build test projects with /p:OSGroup=Linux, it sets OS to Windows_NT (!?), and building with /p:OS=Linux fails outright (!?).

nguerrera: I'm still confused about OS vs OSGroup, but OSGroup is the one that we seem to use now so updated PR using that.

akoeplinger: "OS" is automatically set by msbuild.

nguerrera: Is it meant to indicate the OS of the build system or the OS of the artifacts being compiled. If the former, then it is the right one to use and I'm just confused. :)

akoeplinger: It is indeed the former and yeah, it

nguerrera: OK. Fixed.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -74,8 +74,15 @@
...
-         same files for every single test project here. -->
+    <!-- Ideally, we'd have UseHardLinksIfPossible on by default because we copy tons of the
+         same files for every single test project here, but it doesn't currently work x-plat.
+         so we only turn it on by default for windows builds for now. -->
+    <PropertyGroup>
+      <CreateHardLinksForCopyTestToTestDirectoryIfPossible Condition="'$(CreateHardLinksForCopyTestToTestDirectoryIfPossible)'=='' and '$(OS)' == 'Windows_NT'">true</CreateHardLinksForCopyTestToTestDirectoryIfPossible>
+      <CreateHardLinksForCopyTestToTestDirectoryIfPossible Condition="'$(CreateHardLinksForCopyTestToTestDirectoryIfPossible)'==''">$(CreateHardLinksForCopyFilesToOutputDirectoryIfPossible)</CreateHardLinksForCopyTestToTestDirectoryIfPossible>
+   </PropertyGroup>
+    <!-- Ideally, we'd have UseHardLinksIfPossible on by default because we copy tons of the 
+         same files for every single test project here, but it doesn't currently work x-plat. -->
```
bartonjs: Probably don't need the comment twice, and once you're editing it, you could capitalize "Windows" on line 79.

nguerrera: Oops. Fixed.

akoeplinger:
LGTM, I also verified it works on Ubuntu :+1:

bartonjs:
LGTM

PR #211, opened 2015-07-27
Update xunit versions to latest

mellinoe:
This moves us to the latest version of all of the xunit assemblies, including in our test-runtime project.json. We were previously using a "xunit.runner.dependencies.netcore" NuGet package, which we aren't building from anywhere public (it seems to be a holdover from our very early build process). This switches to just referencing the official xunit NuGet packages directly.

This also includes changes in our custom xunit assemblies (the console runner and our extensions assembly) to build directly against the official xunit packages.

NOTE: We also recently un-published some of our old NuGet packages, including one we were actually still using (PerfTools version 0.0.1-prerelease-00022), so this moves some things to floating versions to avoid that in the future.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
@@ -4,7 +4,7 @@
        "Microsoft.NETCore.Windows.ApiSets-x86": "1.0.0-beta-*",
        "Microsoft.NETCore.TestHost-x86": "1.0.0-beta-*",
        "Microsoft.NETCore.Runtime.CoreCLR-x86": "1.0.0-beta-*",
-       "Microsoft.DotNet.PerfTools": "0.0.1-prerelease-00022",
+       "Microsoft.DotNet.PerfTools": "0.0.1-prerelease-*",
        "OpenCover": "4.6.166",
        "ReportGenerator": "2.1.6.0",
        "System.Collections": "4.0.10-beta-*",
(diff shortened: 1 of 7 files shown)


mellinoe:
@jhendrixMSFT

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
```diff
@@ -32,10 +32,12 @@
        "System.Threading": "4.0.10-beta-*",
        "System.Threading.Tasks": "4.0.10-beta-*",
        "System.Threading.ThreadPool": "4.0.10-beta-*",
+       "System.Threading.Thread": "4.0.0-beta-*",
        "System.Xml.ReaderWriter": "4.0.10-beta-*",
        "System.Xml.XDocument": "4.0.10-beta-*",
-       "xunit.console.netcore": "1.0.2-prerelease-00036",
-       "xunit.runner.dependencies.netcore": "1.0.1-prerelease"
+       "xunit": "2.1.0-beta3-*",
```
weshaggard: I'm not convinced we want to use floating versions for these. 

@ericstj @nguerrera thoughts?

mellinoe: We have them locked, i agree we don't want to just grab the latest all the time as this is used for every test project.

mellinoe:
@dotnet-bot test this, please

mellinoe:
@weshaggard Any more thoughts on this? I can move this to explicit versions if you'd prefer that (although we have a lock file)

weshaggard:
@mellinoe I'm fine with this for the time being we are likely going to make some larger changes around the * versioning but I agree we need to update the versions of those packages. Go ahead and merge but please make sure these changes don't break corefx repo.

mellinoe:
> please make sure these changes don't break corefx repo.

I will also need to update all of the test project.json files in corefx, but I'm working on that right now.

PR #212, opened 2015-07-28
Update package dependencies.

jhendrixMSFT:
Update package dependencies to the latest bits from dotnet-core.
Update perf tools package to latest build (the previously referenced
package no longer exists).

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
-    <BuildToolsVersion>1.0.25-prerelease-00063</BuildToolsVersion>
+    <BuildToolsVersion>1.0.25-prerelease-00065</BuildToolsVersion>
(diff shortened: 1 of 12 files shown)


mellinoe:
Where did the version of buildtools you are referencing here come from (1.0.25-prerelease-00065) ? The repo is still at build number 64.

mellinoe:
For some reason the desktop build of GenFacades seems to still be pulling in an old version of a few packages. Strange.

jhendrixMSFT:
This is a private build I produced in order to break out of a "chicken-and-egg problem", i.e. build tools depending on an older build of build tools that has missing nuget packages.

jhendrixMSFT:
This build error is unique to building with VS 2013, I don't see it with VS 2015.

jhendrixMSFT:
CC @ericstj

PR #213, opened 2015-07-29
Make buildtools compatible with xbuild

agocke:
The current buildtools break on XBuild because of a syntax error (using a property without quotes) and using metadata references inside a metadata definition of an item. This should fix both problems.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
@@ -70,7 +70,7 @@
       <NugetTargetFrameworkMoniker Condition="'$(NugetTargetFrameworkMoniker)' == '.NETPortable,Version=v4.5,Profile=Profile7'">DNXCore,Version=v5.0</NugetTargetFrameworkMoniker>
     </PropertyGroup>
 
-    <PrereleaseResolveNuGetPackageAssets Condition="Exists($(ProjectLockJson))"
+    <PrereleaseResolveNuGetPackageAssets Condition="Exists('$(ProjectLockJson)')"
                                AllowFallbackOnTargetSelection="true"
                                IncludeFrameworkReferences="false"
                                NuGetPackagesDirectory="$(PackagesDir)"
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -85,22 +85,9 @@
...
-      </_ReferencesFileNames>
-
-      <_NoneFileNames Include="@(None->'%(FileName)%(Extension)')">
-        <OriginalIdentity>%(Identity)</OriginalIdentity>
-      </_NoneFileNames>
-      
       <!-- Intersect project-refs with package-refs -->
-      <_ReferenceFileNamesToRemove Include="@(_ReferencesFileNames->'%(OriginalIdentity)')" Condition="'@(_ProjectReferenceFilenames)' == '@(_ReferencesFileNames)' and '%(Identity)' != ''"/>
-      <_NoneFileNamesToRemove Include="@(_NoneFileNames->'%(OriginalIdentity)')" Condition="'@(_ProjectReferenceFilenames)' == '@(_NoneFileNames)' and '%(Identity)' != ''"/>
+      <_ReferenceFileNamesToRemove Include="@(Reference)" Condition="'@(_ResolvedProjectReferencePaths->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)' and '%(Identity)' != ''" />
```
weshaggard: @ericstj as best I can tell this set logic looks equivalent but it is always a little difficult to know for with these crazy conditions. 

@agocke @akoeplinger do you guys know why your this change works with xbuild but the old one doesn't?

agocke: @weshaggard Yes, I mentioned in the PR -- xbuild doesn't support metadata references inside the metadata declarations on items, so all those %(OriginalIdentity) references cause NREs in xbuild.

agocke:
Could I get a pointer to a new build when this PR is accepted? This is blocking the Roslyn CoreCLR tests on OSX/Linux since we don't use MSBuild-mono and I'd like to get it in ASAP.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
weshaggard: OK. Please verify that this change works. From some local hacking with an msbuild project file I think you need to remove `and '%(Identity)' != ''` from the condition with this method.

agocke:
I've added another change which removes the %(Identity) references. After making this change and running the installed buildtools inside the CoreFX repo I get exactly the same references to csc2 as I did before making this PR.

weshaggard:
Thanks for testing. The change looks good to me but you need to resolve your conflicts before we can merge. Please also squash the commits.

agocke:
Done.

agocke:
Thanks for merging! When can I expect to see a new build  with these changes?

PR #217, opened 2015-07-30
Show _AssemblyInfo.cs in Solution Explorer

naamunds:
The Multilingual App Toolkit (MAT) requires the AssemblyInfo file to be
included in EnvDTE so that it can read the assembly's neutral language.
This change sets Visible to true for the AssemblyInfo file and gives it
an appropriate Link.  It also moves it outside of the
GenerateAssemblyInfo target, because it only is shown in Solution
Explorer if it's not in a target.


src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -92,13 +94,16 @@
...
       Overwrite="true" />
-
-    <ItemGroup>
-      <Compile Include="$(AssemblyInfoFile)" />
-      <FileWrites Include="$(AssemblyInfoFile)" />
-    </ItemGroup>
   </Target>
 
+  <ItemGroup Condition="'$(GenerateAssemblyInfo)'=='true'">
+    <Compile Include="$(AssemblyInfoFile)">
```
weshaggard: What happens if you open a project before the _AssemblyInfo.cs file is generated? Does it show up as missing? Does it refresh correctly after the first build?

mellinoe:
Is this file generated as readable? If so, it may be a good idea to try to make it readonly, as I wouldn't want anyone thinking they could make modifications in VS and actually have an effect.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
naamunds: For some solutions, such as Microsoft.CSharp, _AssemblyInfo.cs is generated when the solution is opened, if it hadn't been generated previously.  For others, such as System.IO.sln, the link is broken until the project is built.

weshaggard: Any ideas why they are different? I would expect them all to be consistent.

weshaggard:
@mellinoe yes the file is writable but I don't think we should try to make it readonly that may simply cause other issues. Instead it will be worth adding a comment to the top of the file saying it is generated and not to edit it. In that way it will be very similar to the other designer generated files. @naamunds can you add a line to have the generated file have a comment to that effect?

naamunds:
@mellinoe That's a valid point, but I don't see a way to make the file read-only.  The Compile section on https://msdn.microsoft.com/en-us/library/bb629388.aspx shows the schema for the Compile elements, and I don't see a way to make it read-only; I tried using AutoGen, and that didn't do it.  So, I've decided to follow the pattern used by the generated file Resources.Designer.cs, which is to include a comment at the top that explains that the file was generated, and so changes made to it will be lost.  I'll submit another commit with that change.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
naamunds: After further testing, it looks like if the "bin" directory is empty, _AssemblyInfo.cs is not generated for any of the projects until they are built.  When I was testing it before, I was just deleting the project's directory in "bin\obj", which is where _AssemblyInfo.cs was used, but perhaps the system was getting the file from somewhere else in some cases.

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -41,6 +43,8 @@
     <Error Condition="!Exists('$(IntermediateOutputPath)')" Text="GenerateAssemblyInfo failed because IntermediateOutputPath isn't set to a valid directory" />
 
     <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
+      <AssemblyInfoComments Include="// This code was generated by versioning.targets in the Microsoft.DotNet.BuildTools NuGet package." />
```
weshaggard: I would avoid being specific about what generated this because this target gets uses in other places as well and so it will be wrong and is likely to get outdated and maintained for little benefit. I woulds simply say that it is generated by a tool an any manual changes will be overwritten.

naamunds: I've changed this comment as you suggested.

weshaggard:
LGTM.

PR #219, opened 2015-08-03
Update buildtools to use latest xunit beta

mellinoe:
After filing this issue here: https://github.com/xunit/xunit/issues/515, I found there is a small modification we can make to our tests to accommodate for the difference in behavior with MemberDataAttribute. I have a set of changes made in corefx that will be able to consume these changes:

https://github.com/mellinoe/corefx/commits/latest-xunit
PR coming shortly there as well.

### DIFF
--- a/src/GenFacades.Desktop/project.lock.json
+++ b/src/GenFacades.Desktop/project.lock.json
@@ -3,24 +3,24 @@
   "version": -9996,
   "targets": {
     ".NETFramework,Version=v4.6": {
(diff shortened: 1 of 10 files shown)


mellinoe:
Please take a look: @weshaggard.

I'll also need to make another commit on top of this in order to consume the new xunit console runner package in the test-runtime project.json, once that NuGet package is published from an official build.

mellinoe:
I'm going to merge this in so I can move forward with the rest of the changes.

PR #220, opened 2015-08-03
Re-lock test-runtime dependencies

mellinoe:
Second half of the buildtools changes needed to move to the latest xUnit beta.

This change just re-locks the dependencies of the test-runtime package, which brings us to the version of the packages built from the last official build of this repo. The next official build should give us a package with version prerelease-00069 which is what my corefx PR is set up to consume.


FiveTimesTheFun:
Hi Eric, in your last pull request you updated the versions of packages you depended on in your project.jsons to 23127, but then updated xunit.netcore.extensions.nuspec dependencies to 23121.  Would you like to bump your nuspec dependencies to 23127?

mellinoe:
Good point, I'll update those.

mellinoe:
Merging this in to unblock the corefx PR

PR #223, opened 2015-08-04
Two improvements to xunit.console.netcore

stephentoub:
1. Console.CancelKeyPress is now exposed from the Console contract, so we no longer have to exclude ctrl-C behavior.
2. Brad added the ability to disable discovery enumeration of theories, which is only necessary when there's UI associated with showing test cases (https://github.com/xunit/xunit/commit/07178782a6ea455ae42f1fa6f4b3bfec935cd587).  Just as in xunit.console, I'm disabling such preenumeration of theories as it's unnecessary overhead.

Note that I've not yet tested this against corefx; I'll do so once @mellinoe's xunit changes in corefx go in.

### DIFF
--- a/src/xunit.console.netcore/Program.cs
+++ b/src/xunit.console.netcore/Program.cs
@@ -39,6 +39,7 @@ namespace Xunit.ConsoleClient
 
 #if !NETCORE
                 AppDomain.CurrentDomain.UnhandledException += OnUnhandledException;
+#endif
                 Console.CancelKeyPress += (sender, e) =>
                 {
                     if (!cancel)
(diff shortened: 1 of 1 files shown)


stephentoub:
Separately, our xunit.console.netcore seems to have drifted out-of-date with regards to the real xunit.console.  Do we have plans for making a pull request to xunit to push the required changes for .NET Core back to xunit proper?

mellinoe:
>  Do we have plans for making a pull request to xunit to push the required changes for .NET Core back to xunit proper?

I would hope this is our long-term plan, but I'm not sure how easy it is to build portable executables outside of our build system. It seems that there are issues building a "Portable, v5.0" project from VS 2015 with OutputType set to "exe". Right now we are essentially building a portable library project with OutputType set to exe, but with all of our special stuff on top of it to resolve NuGet packages. Hopefully it will be easy enough to build portable executables soon enough.

The changes here LGTM.

stephentoub:
Thanks, Eric.

weshaggard:
@stephentoub Yes it is our goal to contribute these changes back to xunit. I think it is a task on @FiveTimesTheFun's list to tackle at some point.

PR #224, opened 2015-08-06
Add support for reference assemblies

ericstj:
This change adds support to the targets for building reference assemblies.
Reference asseblies shall be located in ref\<major>.<minor>\AssemblyName.csproj
as a convention.
Partial facade targets will now prefer reference assemblies over packages, if they exist.
Reference assemblies will be built to a ref\<major>.<minor> folder under the
assembly's output directory.

/cc @weshaggard @mellinoe


src/Microsoft.DotNet.Build.Tasks/Targets/coreAssembly.rsp
```diff
@@ -0,0 +1,2 @@
+﻿# CSC requires the runtimemetadataversion flag when building a core assembly (containing System.Object)
+/runtimemetadataversion:v4.0.30319
```
weshaggard: Note this will break for anyone not using roslyn compiler. For example anyone building in VS2013. We should think about what we can do here.

ericstj: how do we build a core assembly on pre-roslyn?

ericstj: I see how it's done: https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/GenerateCompilerResponseFile.targets

weshaggard: pre-roslyn the core assembly doesn't need anything special.

src/Microsoft.DotNet.Build.Tasks/Targets/ReferenceAssemblies.targets
```diff
@@ -0,0 +1,19 @@
...
+    <!-- The first two parts of the assembly version represent API version, 
+         the third part is for bugfixes, the fourth is for hotfixes/securtity-updates -->
+    <APIVersion>$([System.Version]::Parse('$(AssemblyVersion)').ToString(2))</APIVersion>
+    
+    <CompilerResponseFile Condition="'$(IsCoreAssembly)' == 'true'">$(MSBuildThisFileDirectory)coreAssembly.rsp</CompilerResponseFile>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IsReferenceAssemblyProject)'=='true'">
+    <OutputPath>$([System.IO.Path]::Combine('$(OutputPath)', 'ref', '$(APIVersion)'))</OutputPath>
+    <IntermediateOutputPath>$([System.IO.Path]::Combine('$(IntermediateOutputPath)', 'ref', '$(APIVersion)'))</IntermediateOutputPath>
```
weshaggard: Why are you calling Path.Combine instead of just doing the concat?

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -44,6 +44,7 @@
       $(ResolveAssemblyReferencesDependsOn);
       ResolveNuGetPackages;
     </ResolveAssemblyReferencesDependsOn>
+    <NugetTargetFrameworkMoniker Condition="'$(NugetTargetFrameworkMoniker)' == '' and '$(IsReferenceAssemblyProject)' == 'true'">.NETPlatform,Version=v5.0</NugetTargetFrameworkMoniker>
```
weshaggard: What is ".NETPlatform"? I don't recall seeing that anywhere before.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -52,13 +52,27 @@
...
   </PropertyGroup>
+  
+  <!-- If a reference assembly project exists for the partial facade, prefer that over the package  -->
+  <Target Name="FindPartialFacadeProjectReference" Condition="'$(IsPartialFacadeAssembly)' == 'true'" BeforeTargets="AssignProjectConfiguration">
+    <PropertyGroup>
+      <_referenceAssemblyProjectRelativeToRoot>$([System.IO.Path]::Combine('ref', '$(APIVersion)', '$(AssemblyName).csproj'))</_referenceAssemblyProjectRelativeToRoot>
+      <_referenceAssemblyProject>$([System.IO.Path]::Combine($([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), $(_referenceAssemblyProjectRelativeToRoot))), $(_referenceAssemblyProjectRelativeToRoot)))</_referenceAssemblyProject>
+    </PropertyGroup>
+    <ItemGroup>
+      <ProjectReference Include="$(_referenceAssemblyProject)" Condition="'$(_referenceAssemblyProject)' != ''">
```
weshaggard: Should the condition be "Exists"

src/Microsoft.DotNet.Build.Tasks/Targets/coreAssembly.rsp
ericstj: I see, any good detection condition for roslyn? $(VisualStudioVersion) >= 14.0?  /cc @Pilchie

src/Microsoft.DotNet.Build.Tasks/Targets/ReferenceAssemblies.targets
ericstj: I want this to work with cross-plat

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
ericstj: This is what nuget translates "dotnet" to.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
ericstj: GetDirectoryNameOfFileAbove does an exists check but I guess I do need that since the path combine will give me a bogus path in the case GetDirectoryNameOfFileAbove fails.

src/Microsoft.DotNet.Build.Tasks/Targets/coreAssembly.rsp
weshaggard: Last time I checked there was no good detection. I think we were considering using the msbuild version number.

Pilchie: I'm not aware of detection.  +@jaredpar

jaredpar: Neither am I.  @jasonmalinowski may have a good idea of what to do here.

src/Microsoft.DotNet.Build.Tasks/Targets/ReferenceAssemblies.targets
akoeplinger: The xplat MSBuild turns `\` -> `/` if that's what you're concerned about.

src/Microsoft.DotNet.Build.Tasks/Targets/coreAssembly.rsp
ericstj: So far we have either MSBuildToolsVersion or VisualStudioVersion.  I've found that both seem to be specified wether building in a vs env or not, so they seem like reasonable approximations.  I'll go with MSBuildToolsVersion for now unless @jasonmalinowski has a better recommendation.

ericstj: I diffed a diag log from 12 and 14 and these are really the only properties that seem usefully different.

src/Microsoft.DotNet.Build.Tasks/Targets/ReferenceAssemblies.targets
ericstj: In that case, why is our build littered with both `\` and `/`?  I'll simplify to just use `\` if that is all that is required.

src/Microsoft.DotNet.Build.Tasks/Targets/coreAssembly.rsp
jasonmalinowski: For the record, I have no better recommendation. :smile:

ericstj:
Last call for feedback.  I need to merge this today.  Please let me know in the next hour.

weshaggard:
LGTM

PR #226, opened 2015-08-07
Generate _AssemblyInfo.cs in neutral output path

naamunds:
_AssemblyInfo.cs was previously generated in a configuration-specific
path, such as bin\obj\FreeBSD.AnyCPU.Debug\System.Console.  For projects
which build with an architecture other than Windows_NT.AnyCPU.Debug, the
_AssemblyInfo.cs link in Solution Explorer was broken because it was
pointing to Windows_NT.AnyCPU.Debug.

Because _AssemblyInfo.cs is not specific to the OS or configuraiton,
this change fixes the link by moving _AssemblyInfo.cs to a neutral
output path.


src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -28,8 +28,9 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(GenerateAssemblyInfo)'=='true'">
-    <AssemblyInfoFile Condition="'$(MSBuildProjectExtension)' == '.csproj'">$(IntermediateOutputPath)_AssemblyInfo.cs</AssemblyInfoFile>
-    <AssemblyInfoFile Condition="'$(MSBuildProjectExtension)' == '.vbproj'">$(IntermediateOutputPath)_AssemblyInfo.vb</AssemblyInfoFile>
+    <AssemblyInfoFileDir>$(BaseIntermediateOutputPath)$(MSBuildProjectName)\</AssemblyInfoFileDir>
```
weshaggard: What about the other version files generated like versions.h for the native stuff. Does that also need to be in a neutral location?

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
naamunds: It doesn't need to be in a neutral location for the work I'm doing now.

ericstj:
This breaks when multiple projects have the same name but different metadata.  I also disagree with this statement:
>Because _AssemblyInfo.cs is not specific to the OS or configuraiton,
this change fixes the link by moving _AssemblyInfo.cs to a neutral
output path.

The metadata is driven by properties in the project file which could be conditioned on platform.  The right fix here is to just hide this file from solution explorer.  It has no business being there since it is not a user-editable file.

ericstj:
I just hit this for the reference assembly work I am doing, so I will fix it.

ericstj:
I see from your other PR that you need this to show up in Solution explorer due to some tool dependency so I won't remove it.  I think the right fix here would have been to split only the neutralresourcelanguage attribute into a stable file that appears in the solution explorer IDE.   Potentially even in a pre-canned CS that ships in the build tools, rather than generating one.

naamunds:
@ericstj Right, we considered splitting the assembly info files, but decided that it would be simpler to just make the existing one show up in Solution Explorer.  I didn't realize that there are projects with the same name but different metadata.  I'll fix it for https://github.com/dotnet/buildtools/issues/231.

PR #227, opened 2015-08-07
Use 64 bit runtime for tests

ellismg:
c398e63a was an opportunistic change.  It was my understanding that we no longer use the feeds?

I validated that with a private build tools with these changes CoreFX uses the expected 64 bit runtime.

/cc @weshaggard


stephentoub:
@weshaggard should review, but LGTM.

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
```diff
@@ -43,6 +43,6 @@
        "dnxcore50": { }
     },
     "runtimes": {
-       "win7-x86": { }
+       "win7-x64": { }
```
weshaggard: You should include both there.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -30,7 +30,7 @@
                                AllowFallbackOnTargetSelection="true"
                                IncludeFrameworkReferences="false"
                                NuGetPackagesDirectory="$(PackagesDir)"
-                               RuntimeIdentifier="win7-x86"
```
weshaggard: Can you make this an overrideable property? I'm fine setting it to win7-x64 but it would allow people to also set it to win7-x86.

ellismg: Can do.  How does $(TestRuntimeIdentifier) sound?

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
ellismg: Why?

If we do add this, do we need to do something so we depend on the 32 bit versions of Microsoft.NETCore packages?

weshaggard: cc @ericstj 
You would do that so that someone could pass in the other runtime identifier. My understanding here is that all this does is tell the restore command what set of items to prepopulate in the lock file so that they can be resolved by our build system. It is saying I want to know about both of those runtimes.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
weshaggard: Seems reasonable to me.

ellismg:
I tired to make the changes that @weshaggard outlined, but ran into a problem where the generated project.lock.json file contained both the 32 and 64 bit versions of mscorlib.ni.dll for all the DNXCore,Version=v5.0[/*] targets (see the generated [project.lock.json](https://github.com/ellismg/buildtools/blob/3e546cb917561ef2365be0e31666708e5c9c412e/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.lock.json) file)

I would have expected they would not be in the DNXCore,Version=v5.0 target and only the x86 version would be in the DNXCore,Version=v5.0/win7-x86 target (with a similar layout for the x64 target).  With them both included, we end up trying to use a 32bit mscorlib.ni.dll with a 64 bit runtime leading to problems.

After talking to Wes about it, I think we are going to push the change to move to 64 bit in (so we can shake out any issues that appear in the CI system) and then once @ericstj explains to us what we are doing wrong and support for running on 32 bit as well.

PR #230, opened 2015-08-11
Generate EncodingTable.Data as a task

Clockwork-Muse:
This task generates the data-cache file for System.Text.Encoding.CodePages

@tarekgh


tarekgh:
@Clockwork-Muse usually when updating the build tools we have to build and publish it. I think I can help with that after merging this change.
Your other change in corefx repro has to wait this change to get populated and get the new build tool package number and update it the corefx to use this new version to be able to use the new changes.

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
```diff
@@ -0,0 +1,395 @@
...
+        public string IANAMappings { get; set; }
+
+        [Required]
+        public string PreferedIANANames { get; set; }
+
+        [Required]
+        public string OutputDataTable { get; set; }
+
+        public string Namespace { get; set; }
+        public string ClassName { get; set; }
```
tarekgh: should we mark the Namespace ad ClassName with optional attribute?

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
```diff
@@ -0,0 +1,395 @@
+﻿using Microsoft.Build.Framework;
```
tarekgh: do you think it is good idea to regenerate the encoding table every time we build System.Text.Encoding.CodePages? I don't expect the encoding tables will not change that frequently and it doesn't make sense to have it build every time. 

I prefer to keep the current mechanism which the encoding data checked in corefx and have the tool to generate this file in some place in corefx (or other repo) and generate the data on demand. 

What you think?

tarekgh:
@weshaggard could you have a look too?

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
Clockwork-Muse: Do you mean every time somebody submits a new PR (during test builds)?  As a task, it only rebuilds the output file when the input file(s) changes.  This is what happens with all the resource generation, too, which probably don't change that much more.

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
Clockwork-Muse: ... I don't think there is one?  There isn't one in the same namespace (which is what I would expect).  Unless you mean from somewhere else?  
At least one existing task doesn't mark the optional attributes at all (well, I assumed it was optional).

I guess I could comment it.

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
tarekgh: I mean anyone doing clean build (like msbuild /t:rebuild) will regenerate the file. resources are changing frequently so this is why we are regenerating it on the fly but this encoding data tables is not expected to change that frequently.

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
tarekgh: please ignore this one. you are right there is no optional attribute. I confused this with other issue :-)

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
tarekgh: by the way, your changes LGTM. just wait for @weshaggard to have a quick look too and we should be good to go.

Clockwork-Muse: Well, okay, I guess.
(The purist in me decries storing derived files... heck, I was debating trying to find a way to generate directly from the .nlp and .resx files).

I already know how to drop it in the source directory, do you want it to appear in the listing in VS too (not too sure about how to do that, yet)?

tarekgh: if you'll auto generate the file during the build time then it is not a good idea to generate the file in the source directory because this will have github show there is a new change in the sources. it'll be better to generate the file in the intermediate folder and reference it from there. 
to include this file in the project you have to do it in a target element. you may look at the file resources.targets in the build tools in the line 
      <Compile Include="$(IntermediateResOutputFileFullPath)" />

Clockwork-Muse: Ah, that's actually what I'm doing right now (minus showing in VS).  So you're saying not to do this as a build-time task, but as a run-on-command target?

Clockwork-Muse:
@tarekgh , @weshaggard 

Also - even though this is "working", I have concerns about the supplied data (as I noted in the related corefx PR).  We need to get that straightened out a bit before accepting this, I would imagine, in case this needs any changes.

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
weshaggard: Yes I think we should avoid this work as part of everyone's build if we can and only run the update on demand.

src/Microsoft.DotNet.Build.Tasks/Targets/encoding.targets
```diff
@@ -0,0 +1,46 @@
...
+         <PreferredIANANamesPath>$(MSBuildThisFileDirectory)\nameOfPreferredNameFile</PreferredIANANamesPath>
+       </PropertyGroup>
+       <Import Project="$(ToolsDir)encoding.targets" Condition="Exists('$(ToolsDir)encoding.targets')" />
+  -->
+
+  <PropertyGroup>
+    <IntermediateEncOutputFileFullPath>$(IntermediateOutputPath)EncodingData.Table.cs</IntermediateEncOutputFileFullPath>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IANAMappingPath)'!='' And '$(PreferredIANANamesPath)'!=''">
```
weshaggard: As I mentioned above lets not hook this into the default build, so I would remove this CompileDependsOn property update.

src/Microsoft.DotNet.Build.Tasks/GenerateEncodingTable.cs
Clockwork-Muse: Done.  I also loosened the validation back up, after reading a bit more about (at least one of) the code pages/encodings.  (We probably need to add/change a couple of entries, but that's independent of the program itself).

tarekgh: Excellent! 
I assume there will be no change in the corefx side except updating the new generated cs encoding data file. right?

Clockwork-Muse: Had to:
Update .proj file (add output location)
Add "invalid" preferred entries back in (unpick)
Add file back to src (unpick) and regenerate (mostly white-space only diffs).

So yeah, pretty much.

PR #232, opened 2015-08-12
Fix clash on AssemblyInfo.cs

ericstj:
versioning.targets was setting a path to AssemblyInfo.cs that would clash if two projects
have the same name.  I'm working around that for now.  I've filed https://github.com/dotnet/buildtools/pull/217
to track removing this.

/cc @weshaggard @naamunds


naamunds:
LGTM

PR #233, opened 2015-08-12
Add ReferenceAssemblyAttribute + others

ericstj:
These attributes are added to all reference assemblies in our internal build.

/cc @weshaggard @nguerrera


mellinoe:
LGTM

nguerrera:
LGTM. 

I think we can consider stripping APTCA and all security transparency attributes from future reference assemblies as we're moving away from them as the .NET Core hosts we support disable enforcement. 

The comment about GenProfile won't make much sense to external developers, so perhaps add a clarification e.g. "(internal tool used to generate reference assembly source)."

ericstj:
Right now there are cases where our reference assemblies are used for OOB libraries.  So I'll leave APTCA in for now.  My main goal with these changes is parity with internal.  If you think it can be removed and you'd like to do some investigation feel free to file an issue.
I updated the comment.

PR #234, opened 2015-08-13
Add common localization code

naamunds:
This localization code is common to all C# projects, and is the code
added by the Multilingual App Toolkit (MAT) for maintaining localization
in a project.

When building a project on a machine with MAT installed, MAT
synchronizes the .xlf files in the project with the Strings.resx, and
synchronizes the localized Strings.resx files with the .xlf files.  If
MAT is not installed on the machine, the loc file synchronization is
skipped and a message is written to the log file.

For the files to show up in Solution Explorer, it is necessary to set
Visible to True.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
@@ -8,6 +8,12 @@
     <AssemblyVersion Condition="'$(AssemblyVersion)'==''">999.999.999.999</AssemblyVersion>
     <CLSCompliant Condition="'$(CLSCompliant)'=='' and '$(IsTestProject)'=='true'">false</CLSCompliant>
     <CLSCompliant Condition="'$(CLSCompliant)'==''">true</CLSCompliant>
+
+    <!--
+      Check if the project has been localized by looking for the existence of the .de.xlf file.  By convention, we assume that if 
+      a project is localized in German, then it is localized in all languages.
+    -->
+    <ExcludeLocalizationImport Condition="'$(ExcludeLocalizationImport)'=='' And !Exists('$(MSBuildProjectDirectory)\MultilingualResources\$(MSBuildProjectName).de.xlf')">true</ExcludeLocalizationImport>
   </PropertyGroup>
   
   <!--
@@ -83,6 +89,11 @@
       BuildNumberTarget - If this property is set it will try to import the file which allows for it to override the properties above.
   -->
   <Import Project="$(MSBuildThisFileDirectory)versioning.targets" Condition="'$(ExcludeVersioningImport)'!='true'" />
+
+  <!--
+    Import the localization target
+  -->
+  <Import Project="$(MSBuildThisFileDirectory)localization.targets" Condition="'$(ExcludeLocalizationImport)'!='true'" />
 
   <!-- 
     Import the partial facade generation targets
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/localization.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/localization.targets
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="MultilingualAppToolkit">
+    <MultilingualAppToolkitVersion>4.0</MultilingualAppToolkitVersion>
+    <MultilingualFallbackLanguage>en-US</MultilingualFallbackLanguage>
+  </PropertyGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.de.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.es.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.fr.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.it.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.ja.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.ko.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.ru.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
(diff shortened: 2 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/localization.targets
```diff
@@ -0,0 +1,69 @@
...
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.ru.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.zh-Hans.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+    <EmbeddedResource Include="$(MSBuildProjectDirectory)\Resources\Strings.zh-Hant.resx">
+      <Visible>True</Visible>
+    </EmbeddedResource>
+  </ItemGroup>
```
weshaggard: Given that we are importing these in the common targets do they still correctly show up in VS? What about for projects that don't have any or only partial of these files? Should we add some Exists conditions?

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
```diff
@@ -84,6 +84,11 @@
   -->
   <Import Project="$(MSBuildThisFileDirectory)versioning.targets" Condition="'$(ExcludeVersioningImport)'!='true'" />
 
+  <!--
+    Import the localization target
+  -->
+  <Import Project="$(MSBuildThisFileDirectory)localization.targets" Condition="'$(IncludeLocalizationImport)'=='true'" />
+
```
weshaggard: Right now it looks like this imports is opt-in, but should we use some other trigger to determine whether or not to import? Perhaps the existents of some xlf file?

src/Microsoft.DotNet.Build.Tasks/Targets/localization.targets
naamunds: They do show up correctly in VS, but only with "Visible" explicitly set to True.

I agree with conditioning the importing of the localization target on the existence of an .xlf file.  I don't think we have a case where a project would only have some of the .xlf or localized .resx files; if we see later that we do actually have cases like that, we can make a change at that point.

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
naamunds: I agree that checking for the existence of an xlf file is a better approach.

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
```diff
@@ -84,6 +84,11 @@
   -->
   <Import Project="$(MSBuildThisFileDirectory)versioning.targets" Condition="'$(ExcludeVersioningImport)'!='true'" />
 
+  <!--
+    Import the localization target
+  -->
+  <Import Project="$(MSBuildThisFileDirectory)localization.targets" Condition="'$(ExcludeLocalizationImport)'!='true' And Exists('$(MSBuildProjectDirectory)\MultilingualResources\$(MSBuildProjectName).de.xlf')" />
```
weshaggard: To enable better readability of this please pull the condition out and set ExcludeLocalizationImport=true if that condition is true.

src/Microsoft.DotNet.Build.Tasks/Targets/localization.targets
weshaggard: Agreed we can do this later if necessary.

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
naamunds: Do you mean put `<ExcludeLocalizationImport Condition="'$(ExcludeLocalizationImport)'=='' And !Exists('$(MSBuildProjectDirectory)\MultilingualResources\$(MSBuildProjectName).de.xlf')">true</ExcludeLocalizationImport>` in the PropertyGroup at the top of the file?  I'm not sure if I would consider that more readable, but I'll defer to your opinion on it, as I don't have a lot of experience with MSBuild.

weshaggard: Yes I was thinking something along those lines. I agree that the readability isn't way better overall but it does make the import condition a simple Boolean value and we can control it in various ways. For example we could explicitly set it to false and override the Exists check on xlf if we needed that flexibility, or change the check to look for a different file in some cases. I think it enables us more flexibility so we don't have to wait on a build tools update to unblock ourselves if we need to.

Also is "de.xlf" the right file to key on? It might be OK for now but we may want to consider *.xlf at some point, although I don't think that works with Exists we could use an item group if we needed that support.

naamunds: Ah, that makes sense.  Giving projects the way to override the Exists check makes sense.

I think looking for de.xlf is fine.  Really, all 9 .xlf files and all 9 localized .resx files are necessary; otherwise the build will failed.  But as we discussed earlier in the other thread here, I think it's fine to just check for one of them.

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
```diff
@@ -8,6 +8,8 @@
     <AssemblyVersion Condition="'$(AssemblyVersion)'==''">999.999.999.999</AssemblyVersion>
     <CLSCompliant Condition="'$(CLSCompliant)'=='' and '$(IsTestProject)'=='true'">false</CLSCompliant>
     <CLSCompliant Condition="'$(CLSCompliant)'==''">true</CLSCompliant>
+
+    <ExcludeLocalizationImport Condition="'$(ExcludeLocalizationImport)'=='' And !Exists('$(MSBuildProjectDirectory)\MultilingualResources\$(MSBuildProjectName).de.xlf')">true</ExcludeLocalizationImport>
```
jhendrixMSFT: Why is this hard-coded to de?

naamunds: This condition checks if the project has been localized.  If it has the .de.xlf file, then I think we can assume that it has the .xlf files for all 9 languages.  The alternative would be to check for the existence of all 9, perhaps using an ItemGroup, but I don't really see any value in doing that way instead.  I don't think there's a use case yet for a project having xlf files in some languages but not others.

jhendrixMSFT: For shipping I agree, however could there be some transition state for loc where some languages are done but not others?  Or do we agree on some convention that when initially adding loc one must add it for all languages?

naamunds: My thinking is that we can use a convention that when loc is added, it must be done for all languages.  It doesn't have to be translated in all languages initially, of course; the loc files just have to be put in place.

jhendrixMSFT: Seems reasonable, be sure to call this out in the loc workflow doc.  You might also want to add a comment here explaining why we only check for the existence of de.  Other than that LGTM.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/Targets/Build.Common.targets
naamunds: Good idea; I've added a comment.  I'll also call it out in the loc OneNote page.

PR #235, opened 2015-08-13
Allow override of reference assembly output path

ericstj:
This enables us to have the reference assemblies built to the contracts folder for internal builds.


weshaggard:
LGTM

PR #237, opened 2015-08-14
Allow OuterLoop on a class

mellinoe:
Technically, a trait can be applied to a class in addition to a method. This would let us mark a whole class as OuterLoop, something I might be interested in doing if I were to re-organize some of the Tasks tests.


bartonjs:
Assuming you've validated that it behaves correctly, LGTM.

stephentoub:
Could you do the same for PlatformSpecific while you're at it? And do the discoverers need to be updated as well, or is this already handled?

mellinoe:
The discoverers don't need to be updated, as far as I know. I only did a very basic test with ActiveIssue and everything seemed to work as expected without touching the discoverer.

PR #238, opened 2015-08-14
List correct dependencies for xunit.netcore.extensions

### LINKED ISSUE #236: xunit.netcore.extensions should declare its dependency on RuntimeInformation
A recent change switched us to using this contract in the xunit.netcore.extensions assembly, but we didn't add the dependency to the NuGet package, so users of the package will have to add it themselves after they see the error.

mellinoe:
This moves to stable versions of dependencies for xunit.netcore.extensions, and also lists the new dependency on System.Runtime.InteropServices.RuntimeInformation in its nuspec. Fixes #236 .


mellinoe:
By the way, I moved away from the * dependencies here for the two non-stable packages so that the nuspec dependencies don't drift apart just from re-locking the dependencies.

mellinoe:
Looks like @ericstj covered this with #240  and #242

PR #240, opened 2015-08-14
Update dependencies in buildtools to stable versions where available

ericstj:
Now that we've shipped stable versions of most of these packages we need to update the dependencies to stable to get the latest bits.  A stable dependency still permits updating to latest pre-release, however a pre-release dependency (-*) will not.  Due to the way -* works this will only ever match a pre-release version of the package.

This change is needed in order to get our corefx repo entirely onto the latest packages.


src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
```diff
@@ -1,43 +1,43 @@
...
+       "System.Runtime": "4.0.20",
+       "System.Runtime.Extensions": "4.0.10",
+       "System.Runtime.Handles": "4.0.0",
+       "System.Runtime.InteropServices": "4.0.20",
+       "System.Text.Encoding": "4.0.10",
+       "System.Text.Encoding.Extensions": "4.0.10",
+       "System.Text.RegularExpressions": "4.0.10",
+       "System.Threading": "4.0.10",
+       "System.Threading.Tasks": "4.0.10",
        "System.Threading.ThreadPool": "4.0.10-beta-*",
```
weshaggard: Should we also be removing any of the beta-* and replace them with just *?

ericstj: One thing at a time, please.

weshaggard:
LGTM

Issue #231, opened 2015-08-12
AssemblyInfo.cs path clash


ericstj:
https://github.com/dotnet/buildtools/pull/226 moved AssemblyInfo.cs to a path that can be common if projects have the same name.  I hit this in the reference assembly projects and worked around it.  I think the right fix here is to split out just the attributes needed by https://github.com/dotnet/buildtools/pull/217 and make those in a stable path that is shown in solution explorer.  Ideally you can just ship a static file in build tools so you don't have to worry about building before it shows up.  I didn't do this myself since I'm not familar with the tool you mention in the PR.  Can you please have a look and clean this up?  When you do feel free to delete my workaround.

naamunds:
Fixed by https://github.com/dotnet/buildtools/commit/5b053072daed9abc674508acbe72f0bf5e069034 and https://github.com/dotnet/buildtools/commit/d4509423b630bacd446237241dadfe60710af9ec and https://github.com/dotnet/corefx/commit/1b8334fc397eef0ab0d811caff22f2eb731bfdbd.

I've verified the fix by successfully building "corefx\src\Microsoft.CSharp\ref\Microsoft.CSharp.csproj" and by verifying that the Multilingual App Toolkit functionality still works for the non-reference assemblies.

PR #241, opened 2015-08-14
Remove superfluous MakeDir call

naamunds:
With https://github.com/dotnet/buildtools/commit/a85967b2aab1bbed12c2dcb523cba9f0d4e51c8b, the Error 
in GenerateAssemblyInfo was changed to occur when AssemblyInfoFileDir
does not exist.  However, in the case where AssemblyInfoFileDir is set
to "$(BaseIntermediateOutputPath)$(MSBuildProjectName)\", it is expected
to not exist, and is created by a MakeDir task later in the target.

This is causing the build to fail when using the BuildTools in corefx (see 
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_mac_release_prtest/2428/console)

With the fix to https://github.com/dotnet/buildtools/issues/231, this file will probably be reverted
back to how it was before the _AssemblyInfo.cs changes made for loc,
meaning that the Error will look for the existence of
$(IntermediateOutputPath).

Until that fix, though, I think the right workaround is to remove the
Error.

@ericstj @jhendrixMSFT


jhendrixMSFT:
I think this is no longer required.  In change 82e19ef a MakeDir task was added to unconditionally create the directory if it doesn't exist.

naamunds:
> I think this is no longer required. In change  82e19ef  a MakeDir task was added to unconditionally create the directory if it doesn't exist.

@jhendrixMSFT I'm not sure I understand what you mean by this.  Are you agreeing with the code change in this pull request, but saying that the Error will not need to be added back later with the fix for #231?  I'm thinking that that the fix for #231 will entail removing the MakeDir task, along with the rest of the changes I made to versioning.targets.

naamunds:
I'll squash this to one commit before submitting.

jhendrixMSFT:
LGTM

ericstj:
LGTM

PR #242, opened 2015-08-14
Various fixes to enable migrating to latest build tools in corefx

ericstj:
AssemblyInfoFileDir will not exist on a clean open build.  Ensure it is created.
xunit.netcore.extensions took a dependency on RuntimeInformation but didn't add it to the nuspec, add it.
The test architecture change broke some tests.  Add an opt out.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
@@ -1,9 +1,9 @@
 {
     "dependencies": {
        "coveralls.io": "1.4",
-       "Microsoft.NETCore.Windows.ApiSets-x64": "1.0.0",
        "Microsoft.NETCore.TestHost-x64": "1.0.0-beta-*",
-       "Microsoft.NETCore.Runtime.CoreCLR-x64": "1.0.0",
+       "Microsoft.NETCore.TestHost-x86": "1.0.0-beta-*",
+       "Microsoft.NETCore.Runtime": "1.0.0",
        "Microsoft.DotNet.PerfTools": "0.0.1-prerelease-*",
        "OpenCover": "4.6.166",
        "ReportGenerator": "2.1.6.0",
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
```diff
@@ -1,9 +1,9 @@
 {
     "dependencies": {
        "coveralls.io": "1.4",
-       "Microsoft.NETCore.Windows.ApiSets-x64": "1.0.0",
        "Microsoft.NETCore.TestHost-x64": "1.0.0-beta-*",
-       "Microsoft.NETCore.Runtime.CoreCLR-x64": "1.0.0",
+       "Microsoft.NETCore.TestHost-x86": "1.0.0-beta-*",
```
weshaggard: Should this be * instead of beta-*

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -41,6 +41,7 @@
       Outputs="$(AssemblyInfoFile)"
       Condition="'$(GenerateAssemblyInfo)'=='true'">
 
+    <MakeDir Condition="!Exists('$(AssemblyInfoFileDir)')" Directories="$(AssemblyInfoFileDir)" ContinueOnError="true" />
```
weshaggard: I believe MakeDir already handles if the dir already exists, so I don't believe the condition is necessary.

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
ericstj: NO, as I have mentioned before we aren't switching syntax yet.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
ericstj: It isn't necessary, but it saves the expensive task invocation at the cost of a cheap existence check.

PR #245, opened 2015-08-16
Disable "warning" for no tests in assembly

stephentoub:
The xunit.console.netcore runner is printing out "Warning: " when there are no tests to run in the specified assembly.  This causes msbuild to treat that as a warning, which gets flagged and accumulated in the warnings statistics when doing a full build.  That results in 17 warnings when building corefx today, which is very noisy and ends up making it difficult to spot real warnings that should be fixed.

This commit simply changes the string to be "Info" instead of "Warning", and also uses a slightly different shade of yellow in order to differentiate it from more important warnings.  This brings the number of warnings generated during a full corefx build back down to 0.

### DIFF
--- a/src/xunit.console.netcore/Program.cs
+++ b/src/xunit.console.netcore/Program.cs
@@ -299,8 +299,8 @@ namespace Xunit.ConsoleClient
                     {
                         lock (consoleLock)
                         {
-                            Console.ForegroundColor = ConsoleColor.Yellow;
-                            Console.WriteLine("Warning:       {0} has no tests to run", Path.GetFileNameWithoutExtension(assembly.AssemblyFilename));
+                            Console.ForegroundColor = ConsoleColor.DarkYellow;
+                            Console.WriteLine("Info:        {0} has no tests to run", Path.GetFileNameWithoutExtension(assembly.AssemblyFilename));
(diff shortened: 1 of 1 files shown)


bartonjs:
Personally, I'm not sure why it's even news.  I'd just have it print out 
```
  === TEST EXECUTION SUMMARY ===
     The.Assembly.Name  Total: 0, Errors: 0, Failed: 0, Skipped: 0, Time: 0.000s
```

But, this change is a good step.

weshaggard:
I agree with @bartonjs that we should just print it out, presumably it would have a non-zero number of skipped tests.

stephentoub:
> presumably it would have a non-zero number of skipped tests.

That'd be nice, but right now at least most (all?) of them show 0 skipped: they only show as skipped out they're marked as Skip in the Fact, not if our discoverers ignore them due to being outer loop or for the wrong platform or am active issue.

> we should just print it out

I have no opinion on that. The thing that bothered me was it manifesting as a warning, which this fixed. If we want to remove the remaining text, that's fine. I will say that seeing the yellow message has saved me a few times when I accidentally run the tests against System.Whatever.dll instead of against System.Whatever.Tests.dll.

PR #248, opened 2015-08-26
Add GenAPI to build tools

ericstj:
This change adds GenAPI to the buildtools repo and a package for shipping the desktop version.

Referencing this package brings in a targets file that emits a CS with file to the output
directory (overridable to any path chosen by the project).  In practice I expect projects
to override this to point to a location under source control in another project that builds
the reference assembly.

This can be used to automate the production/update of reference assemblies.
I've chosen to emit source since any API diffs will clearly show up in source control.

This change uncovered a problem in the Microsoft.Cci package which I am fixing internally.  I'll update this PR once a new version of that package is published.


src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
```diff
@@ -0,0 +1,33 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <GenAPITargetPath Condition="'$(GenAPITargetPath)' == ''">$(TargetDir)$(TargetName).cs</GenAPITargetPath>
```
weshaggard: Do you think TargetDir is the right default here? What about Obj/Intermediate?

src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
```diff
@@ -0,0 +1,33 @@
...
+    <!-- Sequence ourselves last in PrepareForRun 
+         This ensures the output has been copied to TargetPath, but still permits
+         us to participate in FileWrites. -->
+    <PrepareForRunDependsOn>
+      $(PrepareForRunDependsOn);
+      GenerateReferenceAssemblySource;
+    </PrepareForRunDependsOn>
+  </PropertyGroup>
+
+  <Target Name="GenerateReferenceAssemblySource"
```
weshaggard: What do you think about actually running csc on the generated CS file defaulting most of the properties like name and version from the implementation project?

src/GenFacades.Desktop/project.json
```diff
@@ -15,8 +15,5 @@
     },
     "frameworks": {
        "net46": { }
-    },
-    "runtimes": {
-       "win7-x64": { }
```
weshaggard: Just to help with my understanding why don't we still need win7-x64 here?

src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
ericstj: I didn't see a need for an extra intermediate file in this case.  Indeed, the goal was for folks to commit the file.  If someone wants to build a "direct to dll" package that imports this they can, overriding the path to the file and calling CSC in that target.

src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
ericstj: I specifically did not want to do that in this package.  Folks have lots of different preferences about how their projects get that metadata: properties, generated source files, checked in source files, etc.  I'd like for this package to just be responsible for the hard part: generating the source.  Folks can choose to do things like we do and check in the source (indeed a follow up change to this will be to use these targets in our own build), or they can choose to define other targets or another package to go direct to dll.  Longer term I think the direct to dll solution will be addressed better by https://github.com/dotnet/roslyn/issues/2184.

weshaggard: Makes sense.

For our builds do you envision that we generate the CS file every time as part of the build and overwrite the committed file in place? If so, in general I'm not a fan of overwriting committed files in the source tree as part of the build. That will of course fail in TFS where the files are read-only. On top of that you would need to get the project build ordering correct which might be tricky in some cases.

src/GenFacades.Desktop/project.json
ericstj: This is an AnyCPU exe and all of it's dependencies are AnyCPU.  At the moment none of the dependency packages require a runtime.  For the most part this is the expectation for the desktop implementations of our packages: it is already implicitly windows only and the assets are only facades which aren't specific to an architecture.  If the shape of the dependency packages changed in the future we'd need to change this project to use a runtime and change how we packaged it (since we'd need to have runtime specific folders that redisted those runtime specific dependencies).

src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
ericstj: Yeah, I wasn't planning on continuing the build when an update occurred.  Updates are only expected on dev machines.  If an update occurs in the official build it'll mean that someone forgot to check in an API change.

Once I implement that support it's possible that I'll put some of the logic in this package.  For now I'm just getting the basics working.

ericstj: And your comment on sequencing is very accurate: we already inject a P2P from impl to ref in the partial facade projects.  If we wanted to sequence ref after impl (as is implied by impl generating ref source) that'd be a circular dependency.  That's why I'm just planning to fail the build.

weshaggard:
Changes LGTM

ericstj:
Ok, should be feature complete now.  Please have a look.

ericstj:
Wes reviewed these latest changes over the shoulder and signed off.  Merging.

PR #251, opened 2015-09-01
Move NeutralResourcesLanguage attribute to separate file

naamunds:
This commit fixes issue https://github.com/dotnet/buildtools/issues/231 by moving the NeutralResourcesLanguage attribute
out of the generated _AssemblyInfo file and into a separate
AssemblyInfoPartial file.  This file is the same for every project, so
it is shipped as a static file in buildtools.

This file is shown in Solution Explorer so that it is recognized by the
Multilingual App Toolkit, as described in https://github.com/dotnet/buildtools/pull/217.

This is an alternative approach to that used in https://github.com/dotnet/buildtools/pull/217, so other than
the changes to include AssemblyInfoPartial, versioning.targets has been
reverted to its state before https://github.com/dotnet/buildtools/pull/217.


naamunds:
cc: @jhendrixMSFT @ericstj @weshaggard

jhendrixMSFT:
LGTM

weshaggard:
The change looks reasonable, although I'm a little reluctant to have this static AssemblyInfo file. However I understand the reasoning so I'm fine with it.

PR #252, opened 2015-09-02
Add basic integration with the xUnit performance tools

mellinoe:
This adds some basic integration with the xunit performance tools into the build.

I took a slightly different approach from the last PR. Instead of having a separate set of targets for perf tests, I sort of modeled after what the code coverage targets do. If /p:Performance=true is specified in your build arguments, the PerfTesting.targets file will basically override some test properties, causing the xUnit performance console runner to wrap the test runner.

The Code Coverage support does basically the same thing (it wraps the test runner in OpenCover.exe), so I have disallowed both targets from running. Technically, we could try to make coverage.exe wrap performance.exe, which would wrap corerun.exe, which would wrap xunit.console.exe, which would actually call your test... but that seemed a bit too complicated and not very valuable. If you want coverage numbers for the perf tests, you can just run the coverage tool by itself.

Consuming this is fairly straightforward:
* Create or open an existing test project
* Add a reference to the Microsoft.DotNet.xunit.performance NuGet package
* Create a benchmark test, following the perf test guidelines (to be published?)
* Decorate it with the [Benchmark] attribute (from the above NuGet package)
* Run the tests by doing the following ````msbuild /t:(re)buildandtest /p:Performance=true````
* An analysis report will automatically be generated and the path to it will be printed out.
    * Analysis report will only be generated if the test project actually had any performance tests in it. If you try to run this command on a project that doesn't have any performance tests in it, then it will basically just tell you that no tests were discovered and won't even try to create a report.

For now, I've enabled this only on Windows operating systems. We do have plans to make this run on, at the very least, Linux in the near future. I believe we are going to need to create a separate NuGet package (Microsoft.DotNet.xunit.performance.runner.Linux ?), and when that is available, we should just be able to do a switch on the OS.

cc: @ericeil , @ianhays , @Sridhar-MS , @pharring


ericeil:
I like the approach; others are probably better qualified to comment on the particulars of the various project file changes.  One thing to think about is that the eventual goal for a local perf run (on a dev's machine) is to get a comparison to a baseline, and we'll likely need to *run the baseline* as well as the "candidate" build.  I'm not sure what that should look like, but it's something to start thinking about.  I wouldn't let it block this change. :)

mellinoe:
Yes, I am definitely eager to get the baseline stuff working, that should be a really nice feature for us. I am hoping we can provide a mechanism where all the developer needs to do is pass something like a git commit hash to msbuild that represents the "baseline" build configuration. Something like:
````msbuild /t:buildandtest /p:performance=true /p:baseline=asdf1234jkl````
Then we could add a bunch of logic into the targets here to handle preparing two builds, one with the current git commit, and one from the commit pointed to by that hash, and pointing the performance tool at those two builds. I'm not sure how feasible something like that would be, but it would be a pretty cool developer experience.

**Comment by ianhays** [comment]
+1 for the baseline Property. I do not know what work that would entail, but it would be very useful.

File: src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets [hunk 1]
```diff
@@ -1,36 +1,59 @@
...
+
+  <Target Name="AnalyzePerfResults"
+          AfterTargets="RunTestsForProject"
+          Condition="'$(AnalyzePerfResults)' == 'true' and Exists('$(PerfRunOutputFile)')">
+    <Exec Command="$(PerfAnalysisCommandLine)"
+          WorkingDirectory="$(TestPath)%(TestTargetFramework.Folder)"
+          ContinueOnError="false">
+      <Output PropertyName="PerfTestRunExitCode" TaskParameter="ExitCode" />
+    </Exec>
+    <Message Text="Performance test analysis report is available at $(TestPath)$(DebugTestFrameworkFolder)/$(AnalysisReportFileName)" />
```
Sridhar-MS: Is the perf report always going to be created in the Debug test folder? From the naming it seems like it.

mellinoe: Ah, I should double check this. I may have to use a different property.

Sridhar-MS:
@mellinoe Can you make changes to print out a warning if the perf test is going to be run on debug builds? Something like "Perf run is triggered on a debug build. The results may not be reliable. Use release builds instead".

Sridhar-MS:
@mellinoe I am curious about the 'build.cmd /p:performance=true' scenario. This is probably gonna be used in the CI system in the future. Does it produce a single report for all the perf tests in the repo?

mellinoe:
It does not produce a single report currently, but I should be able to add that. Good point about the warning in debug builds, will add as well.

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.lock.json
```diff
@@ -4,10 +4,78 @@
...
     "DNXCore,Version=v5.0": {
       "coveralls.io/1.4": {},
-      "Microsoft.DotNet.PerfTools/0.0.1-prerelease-00077": {},
-      "Microsoft.NETCore.Runtime/1.0.1-beta-23213": {},
-      "Microsoft.NETCore.TestHost-x64/1.0.0-beta-23213": {},
-      "Microsoft.NETCore.TestHost-x86/1.0.0-beta-23213": {},
+      "Microsoft.DotNet.xunit.performance.analysis/1.0.0-alpha-build0012": {},
+      "Microsoft.DotNet.xunit.performance.runner.Windows/1.0.0-alpha-build0012": {
+        "dependencies": {
+          "xunit.runner.console": "2.1.0-beta4-build3109"
```
pharring: Is this the right version? I thought we were sticking with Beta3 for now?

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.lock.json
```diff
@@ -494,6 +612,7 @@
           "lib/dnxcore50/xunit.execution.dnx.dll": {}
         }
       },
+      "xunit.runner.console/2.1.0-beta4-build3109": {},
```
pharring: Should this be beta3?

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.lock.json
mellinoe: I'm actually not sure where this dependency is being pulled in from, as I did update the nuspec for the runner to point at beta3. Perhaps there was some issue with my NuGet cache or something. I'll check this out tomorrow.

mellinoe: For some reason this keeps getting locked to this version, even though the dependency listed in Microsoft.DotNet.xunit.performance.analysis/1.0.0-alpha-build0012 is actually beta3. I have no idea why that's happening. Regardless, we aren't actually using this dependency anywhere (it's the desktop console runner), so it is fairly harmless. We may just want to remove this dependency, as you don't need to use the desktop console runner with the performance runner.

src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
mellinoe: This property is a bit weirdly named, but it's the right one. "TestPath" actually includes most of the path, i.e. ````bin\tests\Windows_NT.AnyCPU.Debug\<projectname> ```` whereas "DebugTestFrameworkFolder" is just ````dnxcore50```` or ````net45````

mellinoe:
@Sridhar-MS I added a warning that triggers if you are in any configuration other than "Release". I realize that we technically allow any configuration name at all, we don't just limit it to "Debug" and "Release" for whatever reason. So, if you aren't in a "Release" config, I emit a warning.

Sridhar-MS:
LGTM

mellinoe:
I'm going to merge this in a minute after I do a couple more tests so we can get unblocked in corefx. Currently this will produce separate perf reports for each project, but I'm planning on adding some logic like we have for CodeCoverage, which generates a single report per-project, but has another property set in the full repo build ("GenerateCodeCoverageReportForAll"), which I will probably emulate for the perf tests. I'll do that separately. At that time I'll also probably move the perf report HTML files out into a separate folder like they are for code coverage so it's a bit cleaner (they go next to the tests right now).

PR #255, opened 2015-09-04
Fix PrereleaseResolveNuGetPackageAssets on Mono

ellismg:
When using the build tools on MSBuild running on Mono (which is what
corefx does in build.sh) we were running into a few exceptions due to
some differences in the two environments.

 - We end up using some MSBuild assemblies implemented by Mono which
   seem to have slightly different semantics in when the Log property on
   a task is actually set.  Unlike Microsoft's implementation, which
   sets the property during construction of the Task object, Mono
   appears to defer this work until the task is assigned a BuildEngine.
   The fix is to just move our setup of the resource manager for the log
   to before we execute the task.

 - The task in question wants to use some exports on mscoree to parse
   CLR version information from metadata. This doesn't work on non
   windows platforms. MSBuild already faced this problem internally and
   wrote a light weight PE Reader to pull out the info, so we pick that
   up and fall back to it when we can't find the entry point we want on
   mscoree.


ellismg:
/cc @weshaggard @FiveTimesTheFun

This was starting to get frustrating, so I cleaned up the debt here.  My understanding is that @pgavlin was also feeling pain and has basically locked him self back to earlier build tools so he didn't hit this issue.

With the changes, most projects build via build.sh in corefx again, three fail, two of which are due to file casing issues in a csproj and the other is something I don't understand yet but doesn't seem to be related to build tools at all.

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
```diff
@@ -669,22 +670,239 @@ private static string TryGetRuntimeVersion(string path)
...
+                return ManagedRuntimeVersionReader.GetRuntimeVersion(path);
             }
-            else
+        }
+    }
+
+    /// <summary>
+    /// Managed implementation of a reader for getting the runtime version of an assembly
+    /// </summary>
+    static class ManagedRuntimeVersionReader
```
weshaggard: /cc @nguerrera 

Can the MetadataReader be used to get this information instead of implementing it ourselves?

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
```diff
@@ -669,22 +670,239 @@ private static string TryGetRuntimeVersion(string path)
...
-            do
+            try
             {
-                runtimeVersion = new StringBuilder(bufferLength);
-                hresult = NativeMethods.GetFileVersion(path, runtimeVersion, bufferLength, out actualBufferSize);
-                bufferLength = bufferLength * 2;
+                do
+                {
+                    runtimeVersion = new StringBuilder(bufferLength);
+                    hresult = NativeMethods.GetFileVersion(path, runtimeVersion, bufferLength, out actualBufferSize);
```
weshaggard: Is there a reason we shouldn't always use the PEReader to avoid the duplication?

ellismg: Was just trying to follow what MSBuild did more or less (they actually switch on the platform using the legacy Environment.OSVersion stuff).  Can change to just use the PE Reader.

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
ellismg: I will take a look.

akoeplinger: @ellismg back when this code was added to MSBuild you said there's some stuff missing from PE Reader that needs to be exposed: https://github.com/Microsoft/msbuild/pull/43#issuecomment-88251329

ellismg:
Rebased on top of master, moved all paths to use the managed version reader, opened dotnet/corefx#3125 to track exposing an API to do this from the Metadata Reader.

Also build CoreFX locally with these changes.

@weshaggard Can you take a look again?

nguerrera:
I think you can already get what you want using System.Relection.Metadata: https://github.com/dotnet/corefx/issues/3125#issuecomment-138665105

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
```diff
@@ -664,27 +667,228 @@ private void TryParseRuntimeVersion(string imageRuntimeVersion, out bool isWinMD
...
+                return string.Empty;
+            }
+        }
+    }
 
-            do
+    /// <summary>
+    /// Managed implementation of a reader for getting the runtime version of an assembly
+    /// </summary>
+    static class ManagedRuntimeVersionReader
```
weshaggard: I assume this is no longer needed after you switched to using the metadata reader.

weshaggard:
Other than cleaning out the code that isn't needed any longer it looks good.

ellismg:
Thanks @weshaggard, you are correct about the code being able to be removed.

PR #257, opened 2015-09-10
Support partial facade generation on X-Plat

ellismg:
This change does the minimial amount of work to get partial facade
generation into a state where it can be run as part of the build when
running on Linux or OSX.

The major problem was that since we don't produce PDBs when building
with Roslyn on Linux or OSX, GenFacades needs to be able to handle the
case where PDBs are missing.  The version of GenFacades that we had in
build tools can handle this (it issues a warning when doing so) but we
actually don't consume this version from BuildTools today. This change
addresses that issue.

 - Update partialfacades.targets to consume the GenFacades.Desktop
   pacakge.  In doing so, I have introduced a new targets file that
   ships as part of build tools that has version number information
   about the revision of build tools in question so we don't have to
   hardcode the ever changing revision into the targets.  This does
   require that folks that restore the build tools and want to use
   partial facades need to now start restoring the GenFacades.Desktop
   pacakge as well.

 - Fix up some logic in the targets to not try to move PDBs if
   DebugSymbols is false, as it is on Linux and OSX.

 - Add a way to supress the warning to GenFacades for the case where the
   PDB is known not to exist

With these changes, GenFacades can run and fill partial facades on Linux
when running on a custom version of Mono (Mono is missing
Array.Empty<T>() and some StringBuilder overloads which
Microsoft.Cci.dll depends on, these will be added when Mono consumes an
updated reference assembly drop, for now it is easy enough to add them
and then build locally if you really want this to work, which I do.
Longer term we should just move to running GenFacades.exe on top of
CoreCLR, once we have cross platform coreclr runtime packages we can
consume.)

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
@@ -1,10 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <Import Project="$(MSBuildThisFileDirectory)buildtoolsversion.targets" />
+
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
-    <GenFacadesNuGetName>Microsoft.DotNet.BuildTools.ApiTools.1.0.0-prerelease</GenFacadesNuGetName>
+    <GenFacadesNuGetName>GenFacades.Desktop.1.0.0-experimental-$(BuildToolsRevisionNumber)</GenFacadesNuGetName>
     <GenFacadesPath>$(PackagesDir)$(GenFacadesNuGetName)\tools\GenFacades.exe</GenFacadesPath>
   </PropertyGroup>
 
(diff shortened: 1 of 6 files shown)


ellismg:
/cc @mellinoe @weshaggard 

This is to help @pgavlin and myself in order to be able to build enough of CoreFX on Linux such that we can use run-tests.sh without having to point it at and windows builds.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -1,10 +1,12 @@
...
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <Import Project="$(MSBuildThisFileDirectory)buildtoolsversion.targets" />
+
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
-    <GenFacadesNuGetName>Microsoft.DotNet.BuildTools.ApiTools.1.0.0-prerelease</GenFacadesNuGetName>
+    <GenFacadesNuGetName>GenFacades.Desktop.1.0.0-experimental-$(BuildToolsRevisionNumber)</GenFacadesNuGetName>
     <GenFacadesPath>$(PackagesDir)$(GenFacadesNuGetName)\tools\GenFacades.exe</GenFacadesPath>
```
weshaggard: All these build numbers inside the build tools targets are getting a little out of hand... I'm going to think about a way to address them.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -1,10 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <Import Project="$(MSBuildThisFileDirectory)buildtoolsversion.targets" />
+
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
-    <GenFacadesNuGetName>Microsoft.DotNet.BuildTools.ApiTools.1.0.0-prerelease</GenFacadesNuGetName>
+    <GenFacadesNuGetName>GenFacades.Desktop.1.0.0-experimental-$(BuildToolsRevisionNumber)</GenFacadesNuGetName>
```
weshaggard: Where is this package restored?

weshaggard:
How pressing is this for you guys? I'm currently working on trying to get a version of CoreCLR setup to run tools like bclrewriter and I plan to use genfacades as a test case for that work. If you can wait a little while perhaps this will not matter.

src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
ellismg: Presently it needs to be restored by whatever restores the build tools (just like it's done today).  Would you like the target to restore it instead?

pgavlin:
This is quite pressing for me, as it affects my ability to build and test the cross-platform code for Sockets.

ellismg:
@weshaggard At a minimum, you will want to pick up the actual GenFacades changes as well as the targets changes to handle the PDB not needing to exist.  Getting on CoreCLR based GenFacades would be super great as it would reduce the need for a private copy of mono with some of the v4.6 surface area that Microsoft.Cci.dll depends on.

Do you have an idea on how long you think your work is going to take?

weshaggard:
I've carved out a hour or so this afternoon to try and hack this together. I will know more by EOD on how long this might take. 

I agree we will still need the GenFacades change. Perhaps we can split this up into that change and get it in independent. If I don't think we are very close on getting GenFacades working on CoreCLR by say tomorrow we can go ahead with the change similar to what you are proposing here.

ellismg:
@weshaggard Sure.  Let me know tomorrow what you think the status is and I'll either merge this or break out the GenFacades changes and put them in independently.  Pat and I can remain unblocked with a private build and a custom package source.

ellismg:
Closing in favor of #258, which is the long term direction for this work.  @pgavlin and I can work off my private hacks for now.

PR #258, opened 2015-09-11
Tool runtime support

weshaggard:
Add support for running tools on top of .NET Core, things like GenFacades.

This change will restore a version of the Console .NET Core meta-package and any other dependencies into a shared directory, $(ToolRuntimePath), also including a host set via $(ToolHost).

Right now this only works on windows because we don't have all the necessary packages for x-plat yet. I'm actively working to get those enabled.

### DIFF
--- a/src/nuget/Microsoft.DotNet.BuildTools.nuspec
+++ b/src/nuget/Microsoft.DotNet.BuildTools.nuspec
@@ -30,6 +30,8 @@
     <file src="Microsoft.DotNet.Build.Tasks\coreAssembly.rsp" target="lib" />
     <file src="Microsoft.DotNet.Build.Tasks\test-runtime\project.json" target="lib\test-runtime\project.json" />
     <file src="Microsoft.DotNet.Build.Tasks\test-runtime\project.lock.json" target="lib\test-runtime\project.lock.json" />
+    <file src="Microsoft.DotNet.Build.Tasks\tool-runtime\project.json" target="lib\tool-runtime\project.json" />
+    <file src="Microsoft.DotNet.Build.Tasks\tool-runtime\project.lock.json" target="lib\tool-runtime\project.lock.json" />
     <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DotNet.Build.Tasks.dll" target="lib" />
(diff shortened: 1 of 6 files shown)


weshaggard:
@ellismg @joperezr I address your feedback but given I rebased and you made comments on the commits the comments got pushed aside. See https://github.com/weshaggard/buildtools/commit/4a5729faf95ae4eb925e152b003a9247dfc614a2#commitcomment-13217758 for my follow-up comments.

cc @mellinoe as FYI

I'm doing some more testing from the corefx repo with these changes and then I will merge and update the build tools version.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -41,6 +43,8 @@
     <file src="Microsoft.DotNet.Build.Tasks\NativeBinaries\amd64\git2-69db893.dll" target="lib\NativeBinaries\amd64" />
     <file src="Microsoft.DotNet.Build.Tasks\NativeBinaries\x86\git2-69db893.dll" target="lib\NativeBinaries\x86" />
     <file src="Microsoft.DotNet.Build.Tasks\AssemblyInfoPartial.*" target="lib" />
+    <file src="GenFacades\GenFacades.dll" target="lib" />
```
mellinoe: If we plan to use this with CoreRun, we can probably stop building it with CoreConsole and having it rename the extension from exe -> dll. There's a post-step in the publishexe.targets file that renames the file if it detects CoreConsole.exe next to the project output.

mellinoe:
LGTM

src/nuget/Microsoft.DotNet.BuildTools.nuspec
weshaggard: I do plan to clean-up the GenFacades projects and also delete their packages. However I'm going to do that in a separate PR.

joperezr: I can do that once I checkin my stuff with BclRewriter since I'll have to do it as well as I'm also using CoreConsole.

weshaggard: I just cleaned it up, see https://github.com/dotnet/buildtools/pull/263

PR #260, opened 2015-09-12
Stop GenFacade errirs when DebugSymbols is false

ellismg:
When DebugSymbols is set to false (which is the case when building on
OSX or Linux, the FillPartialFacade task fails since it wants to the the
PDB (which was not produced) into the folder GenFacades uses as
input. With this fixed, GenFacades still issues a warning saying it will
not produce a PDB.

 - Update the MSBuild authoring to not try to move a PDB when
   DebugSymbols is false.
 - Add a flag to GenFacades telling it to not try to produce a PDB and
   use that flag to supress the warning when DebugSymbols is false.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
@@ -163,12 +163,19 @@
       <GenFacadesArgs>$(GenFacadesArgs) -contracts:"%(ResolvedMatchingContract.Identity)"</GenFacadesArgs>
       <GenFacadesArgs>$(GenFacadesArgs) -seeds:"@(GenFacadesSeeds, ';')"</GenFacadesArgs>
       <GenFacadesArgs>$(GenFacadesArgs) -facadePath:"$(GenFacadesOutputPath.TrimEnd('\'))"</GenFacadesArgs>
+      <GenFacadesArgs Condition="'$(DebugSymbols)' == 'false'">$(GenFacadesArgs) -producePdb:false</GenFacadesArgs>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/partialfacades.targets
```diff
@@ -163,14 +163,21 @@
       <GenFacadesArgs>$(GenFacadesArgs) -contracts:"%(ResolvedMatchingContract.Identity)"</GenFacadesArgs>
       <GenFacadesArgs>$(GenFacadesArgs) -seeds:"@(GenFacadesSeeds, ';')"</GenFacadesArgs>
       <GenFacadesArgs>$(GenFacadesArgs) -facadePath:"$(GenFacadesOutputPath.TrimEnd('\'))"</GenFacadesArgs>
+      <GenFacadesArgs Condition="'$(DebugSymbols)' == 'false'">$(GenFacadesArgs) -producePdb:false</GenFacadesArgs>
       <GenFacadesArgs Condition="'@(SeedTypePreference)' != ''">$(GenFacadesArgs) -preferSeedType:"@(SeedTypePreference->'%(Identity)=%(Assembly)', ',')"</GenFacadesArgs>
```
weshaggard: Where do you expect DebugSymbols=false to be set?

ellismg: CoreFX sets it as part of the xplat.overides.targets which it conditionally imports.  This is a well known MSBuild property (the CSC task looks at it for example)

weshaggard:
LGTM

PR #263, opened 2015-09-15
Clean-up GenFacade individual packages

weshaggard:
We currently don't need the individual GenFacades packages now
that we are adding this to the BuildTools package so deleting them.

Also updating GenFacades.csproj to be a Library project instead of exe.


weshaggard:
cc @mellinoe

PR #264, opened 2015-09-15
Revert "Clean-up GenFacade individual packages"

weshaggard:
Reverts dotnet/buildtools#263

This actually break running GenFacades on .NETCore because it no longer has an entry point. I will revisit this after I push an updated buildtools to unblock other issues.


joperezr:
I believe that in order to fix it you just need to keep the Output as an Exe, and then have a target that renames it back to .dll

weshaggard:
@joperezr I think your probably right. Given my quick attempt didn't work I will let you go ahead and handle this clean-up work.

mellinoe:
Right, GenFacades still needs to be built as an executable, because CoreRun needs the target to have an entrypoint method defined. We don't really need a target that renames the extension, either. It should be "GenFacades.exe" because it is an executable assembly.

joperezr:
I'll do that then. Although I believe we do want to add a target to rename to .dll because I think Wes wants to do this so people don't use GenFacades (and BclRewriter) directly and call them through CoreRun instead.

weshaggard:
I'm open to entertaining the idea of keeping these as .exe files as well.

PR #265, opened 2015-09-15
Updated build tools to use xunit-rc1

ianhays:
For performance tests to function correctly with corefx, we need to upgrade from xunit-beta3. Since xunit-rc1 is already out, I skipped beta4 and updated buildtools straight to rc1 instead.

I've excluded the project.lock.jsons with the intention of uploading them when they have an updated buildtools nuget package to reference.

I tested this with CoreFX and unit tests passed with the exception of a System.Console import error that is known and is being fixed by @weshaggard. With that said, I would greatly appreciate another tester :)

cc: @mellinoe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
@@ -37,9 +37,9 @@
        "System.Threading.Thread": "4.0.0-beta-*",
        "System.Xml.ReaderWriter": "4.0.10",
        "System.Xml.XDocument": "4.0.10",
-       "xunit": "2.1.0-beta3-*",
+       "xunit": "2.1.0-rc1-*",
        "xunit.console.netcore": "1.0.2-prerelease-*",
-       "xunit.runner.utility": "2.1.0-beta3-*"
+       "xunit.runner.utility": "2.1.0-rc1-*"
     },
     "frameworks": {
        "dnxcore50": { }
--- a/src/xunit.console.netcore/CommandLine.cs
+++ b/src/xunit.console.netcore/CommandLine.cs
@@ -44,7 +44,6 @@ namespace Xunit.ConsoleClient
                 {
                     AssemblyFilename = Path.GetFullPath(assembly.Item1),
                     ConfigFilename = assembly.Item2 != null ? Path.GetFullPath(assembly.Item2) : null,
-                    ShadowCopy = true
                 });
 
             return result;
@@ -167,7 +166,7 @@ namespace Xunit.ConsoleClient
                 {
                     GuardNoOptionValue(option);
                     foreach (var assembly in project.Assemblies)
-                        assembly.ShadowCopy = false;
+                        assembly.Configuration.ShadowCopy = false;
                 }
                 else if (optionName == "-trait")
                 {
--- a/src/xunit.console.netcore/Program.cs
+++ b/src/xunit.console.netcore/Program.cs
@@ -284,7 +284,7 @@ namespace Xunit.ConsoleClient
                         Console.WriteLine("Discovering: {0}", Path.GetFileNameWithoutExtension(assembly.AssemblyFilename));
                 }
 
-                using (var controller = new XunitFrontController(assembly.AssemblyFilename, assembly.ConfigFilename, assembly.ShadowCopy))
+                using (var controller = new XunitFrontController(AppDomainSupport.Denied, assembly.AssemblyFilename, assembly.ConfigFilename, assembly.Configuration.ShadowCopyOrDefault))
                 using (var discoveryVisitor = new TestDiscoveryVisitor())
                 {
                     controller.Find(includeSourceInformation: false, messageSink: discoveryVisitor, discoveryOptions: discoveryOptions);
--- a/src/xunit.console.netcore/project.json
+++ b/src/xunit.console.netcore/project.json
@@ -12,8 +12,8 @@
     "System.Resources.ResourceManager": "4.0.0",
     "System.Runtime.Extensions": "4.0.10",
     "System.Xml.XDocument": "4.0.10",
-    "xunit": "2.1.0-beta3-*",
-    "xunit.runner.utility": "2.1.0-beta3-*"
+    "xunit": "2.1.0-rc1-*",
+    "xunit.runner.utility": "2.1.0-rc1-*"
   },
   "frameworks": {
     "dnxcore50": {}
--- a/src/xunit.netcore.extensions/project.json
+++ b/src/xunit.netcore.extensions/project.json
@@ -10,7 +10,7 @@
     "System.Resources.ResourceManager": "4.0.0",
     "System.Runtime.Extensions": "4.0.10",
     "System.Runtime.InteropServices.RuntimeInformation":"4.0.0-beta-*",
-    "xunit": "2.1.0-beta3-*"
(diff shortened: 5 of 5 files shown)


src/xunit.netcore.extensions/xunit.netcore.extensions.csproj
```diff
@@ -10,6 +10,8 @@
     <RootNamespace>Xunit.NetCore.Extensions</RootNamespace>
     <CLSCompliant>false</CLSCompliant>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetFrameworkVersion>v5.0</TargetFrameworkVersion>
```
weshaggard: Why is this needed?

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
```diff
@@ -36,9 +37,9 @@
        "System.Threading.Thread": "4.0.0-beta-*",
        "System.Xml.ReaderWriter": "4.0.10",
        "System.Xml.XDocument": "4.0.10",
-       "xunit": "2.1.0-beta3-*",
```
weshaggard: Do you know if the rc version still keeps wanting to add a .props file whenever you build/restore it from VS in one of our projects?

src/xunit.netcore.extensions/xunit.netcore.extensions.csproj
ianhays: Since xunit-rc1 is built as a v5.0, I concluded that these extensions will have to specific that they are v5.0 as well. I could try removing it and rebuilding and seeing it still works with CoreFX, though.

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
ianhays: I do not; I haven't been using VS for package restoration/building, though that issue does sound familiar. Has that behavior arisen elsewhere that I can look at?

src/xunit.netcore.extensions/xunit.netcore.extensions.csproj
ianhays: Just tried this. Buildtools compiled without the <TargetFrameWork*> tags causes the following error once it is imported into CoreFX and then CoreFX tests are ran:
```
CoreRun.exe xunit.console.netcore.exe System.Collections.Tests.dll  -xml testResults.xml -notrait category=nonwindowstests  -notrait ca
  tegory=OuterLoop -notrait category=failing

  Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'xunit.runner.utility.dnx, Version=2.1.0.3029, Cu
  lture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.
     at Xunit.ConsoleClient.Program.Main(String[] args)
```

src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
weshaggard: I don't believe it has been reported anywhere but I started hitting it locally after we switched to 2.1 version of xunit and it has been polluting my local projects.

src/xunit.netcore.extensions/xunit.netcore.extensions.csproj
weshaggard: I'm not sure that is related. Can you follow-up to figure out what that changes in the package or output?

ianhays: You are correct; i was importing the packages incorrectly. I tried it with both versions a few times to ensure it works either way and will be updating the PR with them removed

mellinoe: Yeah, this shouldn't be necessary anymore. It looks like the latest builds are built against a portable profile, as of this change:

https://github.com/xunit/xunit/commit/4089905f0e7ba13160f6048b32b3a3988eb203dc#diff-7ffcd805a457da27f45456f40843a2d9R16
(Warning: Might crash your browser)

ianhays:
Any other feedback before I squash commits and merge?

weshaggard:
LGTM. Please test it against CoreFx if you haven't already. Use https://github.com/dotnet/buildtools/wiki/Local-testing-of-BuildTools-in-another-repo as a way of local testing.

I also think it would be good for you to follow-up to make sure this version of xunit works correctly in VS 2015 and doesn't keep causing files to churn.

ianhays:
I was testing in a slightly different (and far more manual) method than the document listed, so thanks for the reference. I tested CoreFX and there were no more failures than without the changes.

What I did:
- Create localpublish.props
```
  <ItemGroup>
    <LocalPackages Include="$(PackagesOutDir)">
      <PackageName>Microsoft.DotNet.BuildTools</PackageName>
      <PackageVersion>1.0.25-prerelease-00087</PackageVersion>
      <InstallLocation>d:\git\corefx\packages</InstallLocation>
    </LocalPackages>
  </ItemGroup>
</Project>
```
- from ```d:git\buildtoools\``` run ```msbuild /p:LocalPublishPropsFile="D:\git\buildtools\localpublish.props"```
- from ```d:\git\corefx\``` run ```./build.cmd```

This isn't consuming the new packages for xunit.console.netcore and xunit.netcore.extensions though, is it?

ianhays:
>I also think it would be good for you to follow-up to make sure this version of xunit works correctly in VS 2015 and doesn't keep causing files to churn.

I've toyed around with this a bit and these are my results:

Without any changes to CoreFX (other than using the new BuildTools package from the above comment), building System.IO.Compression from VS creates a System.IO.Compression.Tests.nuget.props file within the tests directory that makes a reference to xunit.core\beta3.

When the project.json of the tests folder is changed from referencing ```    "xunit": "2.1.0-beta3-*",
```to instead reference ```    "xunit": "2.1.0-rc1-*",```, subsequent building of the solution from within VS does <b>not</b> produce a props file.

This would suggest that rc1 fixed the annoyance that you mentioned, but we will need to modify the project.jsons of all test libraries to benefit from the change.

**Comment by weshaggard** [comment]
Thanks for the follow-up on xunit we will probably want to clean this up at some point.

As for testing your local changes it will not be consuming those unless you add entries for them in the props file. The entry you have is just for the BuildTools package but you can add similar ones for the others.

**Comment by ianhays** [comment]
Modified my localpublish.props to:
```
  <ItemGroup>
    <LocalPackages Include="$(PackagesOutDir)">
      <PackageName>Microsoft.DotNet.BuildTools</PackageName>
      <PackageVersion>1.0.25-prerelease-00087</PackageVersion>
      <InstallLocation>d:\git\corefx\packages</InstallLocation>
    </LocalPackages>
	<LocalPackages Include="$(PackagesOutDir)">
      <PackageName>xunit.console.netcore</PackageName>
      <PackageVersion>1.0.2-prerelease-00087</PackageVersion>
      <InstallLocation>d:\git\corefx\packages</InstallLocation>
    </LocalPackages>
	<LocalPackages Include="$(PackagesOutDir)">
      <PackageName>xunit.netcore.extensions</PackageName>
      <PackageVersion>1.0.0-prerelease-00087</PackageVersion>
      <InstallLocation>d:\git\corefx\packages</InstallLocation>
    </LocalPackages>
  </ItemGroup>
</Project>
```
And the CoreFX tests still pass.

mellinoe:
LGTM

PR #266, opened 2015-09-15
IL.Targets: Support Multiple Compile Items and Additional Arguments

### LINKED ISSUE #262: IL.targets fails for IL Projects with multiple Compile Items
In the exec task, all of the compile items get listed with semicolons between them.

It should be as simple as turning this:
@(Compile)

Into this:
@(Compile, ' ')

brthor:
Add property $(IlasmFlags) to the exec command to support additional arguments if necessary.

Change the separator between compile items to a space, so it doesn't break.

Fixes #262


weshaggard:
LGTM

PR #267, opened 2015-09-17
Update the lock files for xunit-rc1

ianhays:


ianhays:
@weshaggard @mellinoe 

I'm running into an odd issue with CoreFX where System.Linq.Expressions isn't being transferred to the dnxcore50 test folder when building the tests. It seems unrelated, but I don't know enough to say for sure. I ran into this problem with System.Console before, but recent changes fixed it. Any idea @weshaggard?

xunit-rc1 depends on System.Linq.Expressions.

mellinoe:
Other than the issue with System.Linq.Expressions, this should fix the build break in corefx, I think. This just moves everything forward to RC1 rather than move back to beta3. If we can figure out the issue with Linq.Expressions then we should go with this, I think.

mellinoe:
Have you tried the very latest version of System.Linq.Expressions (4.0.11-*) ?

weshaggard:
System.Linq.Expressions is a new dependency coming in from the xunit update but the 4.0.0.0 doesn't contain any runtime assets. To fix this you need to explicitly add a higher version. 

By the way I'm in the process of updating all of buildtools to dnu beta7 and buildtools 87 so I will take care of this update.

weshaggard:
PR https://github.com/dotnet/buildtools/pull/268 contains this update as well as the dnu beta7 update which is also needed to make sure the lock files aren't invalidated every build.

Issue #253, opened 2015-09-03
Delete Microsoft.DotNet.PerfTools


pharring:
I was talking to @jhendrixMSFT today and we both agreed that Microsoft.DotNet.PerfTools should be deleted from this repo (now that we know it will be replaced with xunit.performance).

PR #269, opened 2015-09-17
Remove perf tools prototype.

jhendrixMSFT:
The perf tools prototype is no longer required as we're moving to an
xunit based solution.

This will fix https://github.com/dotnet/buildtools/issues/253


mellinoe:
LGTM

ianhays:
Uhh, this also deleted the new xunit perf targets that enable us to run msbuild /p:performance=true within CoreFX. Was that intended?

jhendrixMSFT:
Ah I didn't realize it had been repurposed.  I will restore it shortly.

ianhays:
Could you also include the changes made in https://github.com/dotnet/buildtools/pull/270 to that restoration?

jhendrixMSFT:
Yes I'll add them to the change.

PR #270, opened 2015-09-17
Updated perf targets for xunit-rc1

ianhays:
Note that for this version of xunit I had to copy the runner to the run directory and run it from there

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
@@ -11,17 +11,32 @@
   <!-- Perf Runner NuGet package paths -->
   <PropertyGroup>
     <XunitPerfRunnerPackageId>Microsoft.DotNet.xunit.performance.runner.Windows</XunitPerfRunnerPackageId>
-    <XunitPerfRunnerPackageVersion>1.0.0-alpha-build0012</XunitPerfRunnerPackageVersion>
+    <XunitPerfRunnerPackageVersion>1.0.0-alpha-build0014</XunitPerfRunnerPackageVersion>
     <XunitPerfRunnerPackageDir>$(PackagesDir)$(XunitPerfRunnerPackageId)/$(XunitPerfRunnerPackageVersion)</XunitPerfRunnerPackageDir>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
```diff
@@ -11,17 +11,32 @@
...
+    <ItemGroup>
+	  <PerfRunners Include="$(XunitPerfRunnerPackageDir)/tools/xunit.performance.run.exe" />
+	  <PerfRunners Include="$(XunitPerfRunnerPackageDir)/tools/xunit.performance.metrics.dll" />
+	  <PerfRunners Include="$(XunitPerfRunnerPackageDir)/tools/xunit.performance.logger.exe" />
+	  <PerfRunners Include="$(XunitPerfRunnerPackageDir)/tools/ProcDomain.dll" />
+	  <PerfRunners Include="$(XunitPerfRunnerPackageDir)/tools/Microsoft.Diagnostics.Tracing.TraceEvent.dll" />
+	  <PerfRunners Include="$(XunitPerfRunnerPackageDir)/tools/*/*.*" />
+	</ItemGroup>
+    <Copy SourceFiles="@(PerfRunners)"
+          DestinationFiles="@(PerfRunners->'$(TestPath)$(DebugTestFrameworkFolder)\%(RecursiveDir)%(Filename)%(Extension)')"
```
mellinoe: Any reason we can't just use "DestinationFolder" like before?

ianhays: It reacted oddly with the wildcard folder inclusions for PerfRunners e.g. not copying subdirectories properly. I didn't want to copy xunit.core from the xunit.performance,runners directory so I couldnt just include all of it with "tools/**/*.*"s, I had to pick and choose.

I've never made an msbuild target before though, so I expect this could be optimized.

mellinoe: Ah, that sounds fine to me. I realize now that there are subfolders in the NuGet package. There may be a way to just copy the whole folder into another folder, but I've never done it. This way works fine

mellinoe:
LGTM

jhendrixMSFT:
This was fixed in https://github.com/dotnet/buildtools/pull/271 so I think we can close this PR.

PR #271, opened 2015-09-17
Add PerfTesting.targets back

jhendrixMSFT:
PerfTesting.targets was inadvertently deleted in commit ca22795; this
change adds it back along with some fixes provided by ianhays.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'false' and '$(OS)' == 'Windows_NT'">true</RunPerfTestsForProject>
+    <AnalyzePerfResults Condition="'$(RunPerfTestsForProject)' == 'true' And '$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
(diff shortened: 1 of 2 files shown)


ianhays:
Do you also need to add this line:
```
<!-- import settings for perf testing --> 
<Import Project="$(MSBuildThisFileDirectory)PerfTesting.targets" />  
```
back to 
src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets 

?

jhendrixMSFT:
Yes good catch will add it.

ianhays:
LGTM

ianhays:
Just imported this into a fresh CoreFX installation and it's failing with the following error for every test directory
```
"D:\git\corefx\src\System.IO.Compression\tests\System.IO.Compression.Tests.csproj" (BuildAndTest target) (1) ->
(RunTestsForProject target) ->
  D:\git\corefx\packages\Microsoft.DotNet.BuildTools.1.0.25-prerelease-00090\lib\tests.targets(126,5): warning MSB3073: The command "
D:\git\corefx\bin/tests/Windows_NT.AnyCPU.Debug\System.IO.Compression.Tests\dnxcore50/tools/xunit.performance.run.exe System.IO.Compr
ession.Tests.dll -runnerhost CoreRun.exe -runner xunit.console.netcore.exe -runid latest-perf-build -verbose " exited with code 3. [D
:\git\corefx\src\System.IO.Compression\tests\System.IO.Compression.Tests.csproj]
```
when running ```msbuild /t:TestAndBuild /p:performance=true``` or ```./build.cmd /p:performance=true```

running without the performance property enabled runs executes without issue.

Using my version from the PR #270 I put out (that doesn't include changes from #369) also works fine. So I'm guessing something from #269 or #271 must have been incompatible with my changes.

ianhays:
To be more specific, I was using myget build 00090

ianhays:
Alright, looks like I found the problem. The perf runner path is being created incorrectly. I'll post a fix in a sec.

ianhays:
#272

PR #273, opened 2015-09-18
Remove '*' versions from test and tool runtime project.json

weshaggard:
Having the '*' versioning wih these keeps getting builds into
unpredictable states because they are pulling the latest packages
more often then we would like. This commit fixes the versions
to a specific version so we can have more deterministic results.


joperezr:
Great! Should we also use this PR to add ubuntu as a runtime to the tool runtime file since we are already touching it?...other than that LGTM

weshaggard:
@joperezr we should do the as a separate commit from this one.

weshaggard:
FYI @mellinoe @nguerrera

nguerrera:
Thanks.

Issue #274, opened 2015-09-18
GenFacades should not emit nested exported type entries if the nested type is not visible outside the assembly


nguerrera:
See [Program.cs#L488](https://github.com/dotnet/buildtools/blob/4607a6f4036458006ed5ca232236bcf345c9b64c/src/GenFacades/Program.cs#L488)

We replicated the behavior of csc.exe at the time. However, some tools get upset when they find an exported type entry that can't resolve even if there is no actual use of the entry encountered. Those tools should be hardened appropriately, but we might as well also help them out by not emitting what isn't necessary.

Things to consider:
* InternalsVisibleTo: I am not certain how type forwards behave relative to it, but I believe it would depend on whether the caller is a friend of the destination. We may want to keep anything that is visible to friends and only cull things invisible outside assembly even in the face of InternalsVisibleTo.

* Check the behavior of Roslyn. Does it still emit these from a [TypeForwardedTo(TopLevelTypeWithPrivateNestedTypes)]?) If so, we should file a bug against the compiler as well.

joperezr:
@nguerrera Thanks for filing the issue, I'll fix this.

nguerrera:
Thanks. I can't assign it to you for some reason, so please assign yourself.

nguerrera:
cc @weshaggard

joperezr:
Closed with #275

PR #275, opened 2015-09-18
Only add a type forward if the nested type is visible outside of the assembly

joperezr:
Only add a type forward if the nested type is visible outside of the assembly. This is the fix for issue #274

### DIFF
--- a/src/GenFacades/Program.cs
+++ b/src/GenFacades/Program.cs
@@ -446,7 +446,7 @@ namespace GenFacades
                         continue;
                     }
 
-                    AddTypeForward(assembly, seedType, contractAssembly);
+                    AddTypeForward(assembly, seedType);
                 }
 
                 if (error)
@@ -488,7 +488,7 @@ namespace GenFacades
                 string preferredSeedAssembly;
                 if (_seedTypePreferences.TryGetValue(docId, out preferredSeedAssembly))
                 {
-                    return seedTypes.SingleOrDefault(t => String.Equals(GetContainingUnitNamespaceFromType(t), preferredSeedAssembly, StringComparison.OrdinalIgnoreCase));
+                    return seedTypes.SingleOrDefault(t => String.Equals(t.GetAssembly().Name.Value, preferredSeedAssembly, StringComparison.OrdinalIgnoreCase));
                 }
 
                 return null;
@@ -500,19 +500,11 @@ namespace GenFacades
 
                 foreach (INamedTypeDefinition type in seedTypes)
                 {
-                    Trace.TraceError("  /preferSeedType:{0}={1}", docId.Substring("T:".Length), GetContainingUnitNamespaceFromType(type));
+                    Trace.TraceError("  /preferSeedType:{0}={1}", docId.Substring("T:".Length), type.GetAssembly().Name.Value);
                 }
             }
 
-            private static string GetContainingUnitNamespaceFromType(INamedTypeDefinition type)
-            {
-                INamespaceTypeDefinition namespaceTypeDefinition = type as INamespaceTypeDefinition;
-                if (namespaceTypeDefinition != null)
-                    return namespaceTypeDefinition.ContainingUnitNamespace.Unit.Name.Value;
-                return type.GetAssembly().Name.Value;
-            }
-
-            private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType, IAssembly contractAssembly)
+            private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType)
             {
                 var alias = new NamespaceAliasForType();
                 alias.AliasedType = ConvertDefinitionToReferenceIfTypeIsNested(seedType, _seedHost);
--- a/src/GenFacades/Program.cs
+++ b/src/GenFacades/Program.cs
@@ -417,7 +417,7 @@ namespace GenFacades
                         continue;
                     }
 
-                    AddTypeForward(assembly, seedType);
+                    AddTypeForward(assembly, seedType, contractAssembly);
                 }
 
                 if (error)
@@ -475,7 +475,7 @@ namespace GenFacades
                 }
             }
 
-            private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType)
+            private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType, IAssembly contractAssembly)
             {
                 var alias = new NamespaceAliasForType();
                 alias.AliasedType = ConvertDefinitionToReferenceIfTypeIsNested(seedType, _seedHost);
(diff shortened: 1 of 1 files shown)


joperezr:
cc @nguerrera @weshaggard

src/GenFacades/Program.cs
```diff
@@ -495,7 +495,8 @@ private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType)
                 //       but the runtime currently throws a TypeLoadException without explicit forwarders for the nested 
                 //       types.
                 foreach (var nestedType in seedType.NestedTypes.OrderBy(t => t.Name.Value))
```
akoeplinger: The comment above still mentions this is done for all nested types regardless of their accessibility, etc. Would be good to update it.

joperezr: right, forgot to change that, will fix that.

nguerrera: +1

nguerrera:
What did you think about my question about InternalsVisibleTo in the issue. I believe it's OK for GenFacades to ignore this case, but I would:

(1) test whether the following works:

**Assembly A**
``` C#
[assembly: InternalsVisibleTo(C)]
public class X { internal class Y {} }
```

** Assembly B, generated by GenFacades -> A**
``` C#
[assembly: TypeForwardedTo(typeof(X))]
```
Assembly C, uses A via type forward to B:
class Z : X.Y { }
```

(2) decide if GenFacades need to support. I think probably not, but we should deliberate doc that here. If it does need to support it, then I think we have to keep anything with internal visibility because the InternalsVisibleTo of target could change over time.

Also, did you check what Roslyn compiler now emits in these cases? I am curious and think we should follow up with them if they still emit entries for private nested classes.

Finally, please do an IL diff of Facades in corefx generated before and after this change and make sure that nothing unexpected is getting stripped.

**Comment by joperezr** [comment]
I also think that is a non common case, but if you think we want to support it I can do it. If not, where do you want me to document it? I haven't check yet what Roslyn compiler does nor have I done the IL diff. I'll do both tasks today and report back on the tracking issue, just wanted to start the PR first.

**Comment by nguerrera** [comment]
Actually, I just realized a better fix that would make the whole internal vs. private thing moot. Instead of enumerating the nested types of the destination, we should enumerate the nested types of the contract and forward those. We won't generally have any private nested classes in contracts.

**Comment by weshaggard** [comment]
+1 to Nick's last suggestion that we should use the contract as the filter.

**Comment by joperezr** [comment]
@weshaggard @nguerrera Let me know if the last commits are what you envisioned. I already tested it in the corefx build and it is passing, and also I've checked the ILs of the assemblies produced and seems like it's doing the right thing.

**Comment by joperezr** [comment]
Ping! :)

File: src/GenFacades/Program.cs [hunk 2]
```diff
@@ -485,17 +485,24 @@ private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType)
...
+                    foreach (var nestedType in contractType.NestedTypes.OrderBy(t => t.Name.Value))
+                    {
+                        var seedNestedType = seedType.NestedTypes.Where(nt => nt.Name.Value == nestedType.Name.Value).FirstOrDefault();
+                        if (seedNestedType == null)
+                            throw new Exception("Nested type found in the contract, but not on the seedType.");
+                        AddTypeForward(assembly, seedNestedType, contractAssembly);
+                    }
+                }
+                else // Add all the nested types for ExportedTypes.
+                    foreach (var nestedType in seedType.NestedTypes.OrderBy(t => t.Name.Value))
```
weshaggard: Please put braces around this foreach statement which is part of the else statement.

src/GenFacades/Program.cs
```diff
@@ -485,17 +485,24 @@ private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType)
...
-                // that were built from source vs. those that were produced by this tool.
-                //
+                // Recursively add forwarders for nested types.
                 // NOTE: Some design-time tools can resolve forwarded nested types with only the top-level forwarder,
                 //       but the runtime currently throws a TypeLoadException without explicit forwarders for the nested 
                 //       types.
-                foreach (var nestedType in seedType.NestedTypes.OrderBy(t => t.Name.Value))
-                    AddTypeForward(assembly, nestedType);
+                INamedTypeDefinition contractType = contractAssembly.GetAllTypes().Where(t => t.FullName() == seedType.FullName()).FirstOrDefault();
+                if (contractType != null) // Only add forwarders for the nested types on the contract
```
weshaggard: I wonder if we should start building a table of the contract types from docId to type as well so we don't have to keep repeating this type search for each type. At first glance it might be a perf hit. It is also not finding the types exactly the same way. I would prefer we also use the docid to align the types here so things are all correct.

src/GenFacades/Program.cs
```diff
@@ -485,17 +485,24 @@ private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType)
...
+                {
+                    foreach (var nestedType in contractType.NestedTypes.OrderBy(t => t.Name.Value))
```
weshaggard: In fact we can actually create a table for the contract assembly that contains the nested type docids as well and we can align them, the same way we are aligning the other docids. That way we can eliminate this special casing and we can do the nested types the same as we do the non-nested types. All we have to do is ensure we traverse the nested types when building the docid lists above.

src/GenFacades/Program.cs
joperezr: Done.

src/GenFacades/Program.cs
joperezr: I changed it now.

src/GenFacades/Program.cs
joperezr: Done.

joperezr:
I already tested my last changes with corefx repo, and everything seems to work correctly except that I'm running into the known issue dotnet/corefx#3726 The rest of the tests pass just fine.

joperezr:
@dotnet-bot test this please

src/GenFacades/Program.cs
```diff
@@ -459,23 +488,31 @@ private INamespaceTypeDefinition GetSeedType(string docId, IReadOnlyList<INamesp
...
-            private static void TraceDuplicateSeedTypeError(string docId, IReadOnlyList<INamespaceTypeDefinition> seedTypes)
+            private static void TraceDuplicateSeedTypeError(string docId, IReadOnlyList<INamedTypeDefinition> seedTypes)
             {
                 Trace.TraceError("The type '{0}' is defined in multiple seed assemblies. If this is intentional, specify one of the following arguments to choose the preferred seed type:", docId);
 
-                foreach (INamespaceTypeDefinition type in seedTypes)
+                foreach (INamedTypeDefinition type in seedTypes)
                 {
-                    Trace.TraceError("  /preferSeedType:{0}={1}", docId.Substring("T:".Length), type.ContainingUnitNamespace.Unit.Name.Value);
+                    Trace.TraceError("  /preferSeedType:{0}={1}", docId.Substring("T:".Length), GetContainingUnitNamespaceFromType(type));
```
weshaggard: Just always call type.GetAssembly().Name.Value

src/GenFacades/Program.cs
```diff
@@ -459,23 +488,31 @@ private INamespaceTypeDefinition GetSeedType(string docId, IReadOnlyList<INamesp
...
-            private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType)
+            private static string GetContainingUnitNamespaceFromType(INamedTypeDefinition type)
+            {
+                INamespaceTypeDefinition namespaceTypeDefinition = type as INamespaceTypeDefinition;
+                if (namespaceTypeDefinition != null)
+                    return namespaceTypeDefinition.ContainingUnitNamespace.Unit.Name.Value;
+                return type.GetAssembly().Name.Value;
+            }
+
+            private void AddTypeForward(Assembly assembly, INamedTypeDefinition seedType, IAssembly contractAssembly)
```
weshaggard: You don't need contractAssembly any more.

src/GenFacades/Program.cs
```diff
@@ -459,23 +488,31 @@ private INamespaceTypeDefinition GetSeedType(string docId, IReadOnlyList<INamesp
...
             {
                 var alias = new NamespaceAliasForType();
```
weshaggard: Does this need to be something other than NamespaceAliasForType in the nested case?

src/GenFacades/Program.cs
joperezr: Done.

joperezr:
Forgot to squash before merging. Sorry for the trouble!

PR #276, opened 2015-09-19
Deploy xunit.console.netcore.exe via NuGet

dsplaisted:
This will allow the test runner to be deployed by NuGet when xunit.console.netcore package is referenced from project.json in test projects.  For example, see [here](https://github.com/dsplaisted/msbuild/blob/00d48e80ff5c46c86e08c3bf0f1ede98863ba7bc/src/Framework/UnitTests/project.json#L26)

### DIFF
--- a/src/nuget/xunit.console.netcore.nuspec
+++ b/src/nuget/xunit.console.netcore.nuspec
@@ -33,11 +33,16 @@
       <dependency id="System.Threading" version="4.0.10" />
       <dependency id="System.Threading.Tasks" version="4.0.10" />
       <dependency id="System.Xml.XDocument" version="4.0.10" />
+      <dependency id="xunit.runner.utility" version="2.1.0-rc1-build3168" />
     </dependencies>
   </metadata>
   <files>
     <!-- xunit runner for .NET Core -->
     <file src="xunit.console.netcore\xunit.console.netcore.exe" target="lib\aspnetcore50" />
+    <file src="xunit.console.netcore\_._" target="lib\dnxcore50" />
+    <file src="xunit.console.netcore\_._" target="lib\dotnet" />
+  
+    <file src="xunit.console.netcore\xunit.console.netcore.exe" target="runtimes\any\native" />
   </files>
   
 </package>
--- a/src/xunit.console.netcore/project.json
+++ b/src/xunit.console.netcore/project.json
@@ -13,7 +13,7 @@
     "System.Runtime.Extensions": "4.0.10",
     "System.Xml.XDocument": "4.0.10",
     "xunit": "2.1.0-rc1-*",
-    "xunit.runner.utility": "2.1.0-rc1-*"
+    "xunit.runner.utility": "2.1.0-rc1-build3168"
   },
   "frameworks": {
     "dnxcore50": {}
--- a/src/xunit.console.netcore/project.lock.json
+++ b/src/xunit.console.netcore/project.lock.json
@@ -35,7 +35,7 @@
           "lib/dotnet/System.Collections.Concurrent.dll": {}
         }
       },
-      "System.Console/4.0.0-beta-23316": {
+      "System.Console/4.0.0-beta-23318": {
         "type": "package",
         "dependencies": {
           "System.IO": "4.0.0",
@@ -618,10 +618,10 @@
         "System.Collections.Concurrent.nuspec"
       ]
     },
-    "System.Console/4.0.0-beta-23316": {
+    "System.Console/4.0.0-beta-23318": {
       "type": "package",
       "serviceable": true,
-      "sha512": "pYDflUHKD2oIllR+HsbgfkbwNsjivEgYsFjM3eZWS6KVqa5vYVMl93OvC+WfB0Va3Q9gAZYzxXQJUv6gWoJGLw==",
+      "sha512": "7O8pbiNDUzI4arqyeMUFOFBiUAY1eHh0z4HclT8iPi+P+i5TZcBP65JoqdS784HQGdkdKUAidJlpCIi7knHvzA==",
       "files": [
         "lib/MonoAndroid10/_._",
         "lib/MonoTouch10/_._",
@@ -635,8 +635,8 @@
         "ref/xamarinios10/_._",
         "ref/xamarinmac20/_._",
         "runtime.json",
-        "System.Console.4.0.0-beta-23316.nupkg",
-        "System.Console.4.0.0-beta-23316.nupkg.sha512",
+        "System.Console.4.0.0-beta-23318.nupkg",
+        "System.Console.4.0.0-beta-23318.nupkg.sha512",
         "System.Console.nuspec"
       ]
     },
@@ -1731,7 +1731,7 @@
       "System.Runtime.Extensions >= 4.0.10",
       "System.Xml.XDocument >= 4.0.10",
       "xunit >= 2.1.0-rc1-*",
-      "xunit.runner.utility >= 2.1.0-rc1-*"
+      "xunit.runner.utility >= 2.1.0-rc1-build3168"
     ],
     "DNXCore,Version=v5.0": []
   }
--- a/src/xunit.console.netcore/xunit.console.netcore.csproj
+++ b/src/xunit.console.netcore/xunit.console.netcore.csproj
@@ -62,6 +62,9 @@
   <ItemGroup>
     <None Include="App.config" />
     <None Include="project.json" />
+    <None Include="_._">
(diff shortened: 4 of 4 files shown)


weshaggard:
@dsplaisted the changes look reasonable to me. Even if you get this deployed how are you planning to deploy corerun.exe (or whichever host you are using)?

mellinoe:
Is this what the NuGet team recommends we do? We are packaging a bunch of other exe's in NuGet packages, but those are all native executables (CoreRun, CoreConsole, etc), and so we have architecture-specific packages for them. For an AnyCPU executable like the xunit runner, I feel like there should be a way to just package it up in a normal NuGet package (not even architecture-specific if we don't need it to be).

dsplaisted:
@weshaggard Currently we're referencing `Microsoft.NETCore.TestHost-x86` for CoreRun.exe.  Checking the feed, it looks like we can probably reference `Microsoft.NETCore.TestHost` instead and automatically get the right host depending on what runtime we're targeting.

> I feel like there should be a way to just package it up in a normal NuGet package (not even architecture-specific if we don't need it to be).

@mellinoe I agree. The currently recommended way to do this is to use a .props file which adds the items to be copied to the output directory.  This is what [xUnit does](https://github.com/xunit/xunit/blob/47fdc2669ae6aa28f6d642e202840193dfc7dbd7/src/xunit.core/build/xunit.core.props), for example.  However, that doesn't work for us, because we have multiple frameworks declared in project.json.  I tried various ways to get the .exe deployed without being specific to a given runtime (such as putting it in a native/ folder at the root of the package), but couldn't find a way that worked.

dsplaisted:
It looks like we're going to run into the same issue with with deploying exe's with the Roslyn compilers.  I don't want to try the same hack with them, so I'm going to see if I can get this to work with a .props file in the NuGet package.

dsplaisted:
OK, I came up with a [solution](https://github.com/Microsoft/msbuild/pull/223) in the MSBuild repo which doesn't require renaming an EXE to a DLL.  So I've updated this pull request to use a .props file to deploy the EXE.  How does this look?

src/nuget/xunit.console.netcore.nuspec
```diff
@@ -33,11 +33,21 @@
       <dependency id="System.Threading" version="4.0.10" />
       <dependency id="System.Threading.Tasks" version="4.0.10" />
       <dependency id="System.Xml.XDocument" version="4.0.10" />
+      <dependency id="xunit.runner.utility" version="2.1.0-rc1-build3168" />
     </dependencies>
   </metadata>
   <files>
     <!-- xunit runner for .NET Core -->
     <file src="xunit.console.netcore\xunit.console.netcore.exe" target="lib\aspnetcore50" />
```
ericstj: This can be removed.

src/nuget/xunit.console.netcore.nuspec
```diff
@@ -33,11 +33,21 @@
...
+    <file src="xunit.console.netcore\_._" target="lib\dnxcore50" />
+    <file src="xunit.console.netcore\_._" target="lib\dotnet" />
+    
+    <!--
+      NuGet ignores .exe files in lib folders, so use .props files to copy the exe
+      to the output folder.
+    -->
+    <file src="xunit.console.netcore\xunit.console.netcore.exe" target="build\dotnet" />
+    <file src="xunit.console.netcore\xunit.console.netcore.props" target="build\dotnet" />
   </files>
```
ericstj: You don't need the props file if you just put the exe in a native folder.

ericstj: Specifically:
```
native\xunit.console.netcore.exe
```
Native is a bit of a misnomer.  Nuget treats native assets as copy-local assets for runtime and not reference.  It seems to be exactly what you want.

dsplaisted: > native\xunit.console.netcore.exe

@ericstj I tried this, but it didn't work.  It goes in project.lock.json as expected, but the ResolveNuGetPackageAssets task doesn't deploy anything from a native folder that's not specific to a runtime.  The logic is in [GetCopyLocalItems](https://github.com/dotnet/roslyn-internal/blob/275671ee97283bbff71b59e173ed9319ac415b83/Closed/MSBuildNuGet/Tasks/ResolveNuGetPackageAssets.cs#L234).

I'm not sure a fix for this would go entirely in the task, either.  You'd need to define what happens when you have a native folder that's specific to a runtime a runtime-agnostic one.  Should you take all files from both (preferring the runtime-specific ones when there is the same file in both places), or should the existence of a runtime-specific native folder mean the runtime-agnostic folder isn't used at all?  That type of decision seems like maybe it should be part of NuGet instead of the ResolveNuGetPackageAssets task.

src/nuget/xunit.console.netcore.nuspec
dsplaisted: > This can be removed.

@ericstj What makes you say this?  When I remove this reference from project.json, I get build errors.

src/nuget/xunit.console.netcore.nuspec
ericstj: >when you have a native folder that's specific to a runtime a runtime-agnostic one...
That's already well defined.  Runtimes/{rid} wins, same as lib.

What version of ResolveNuGetPackageAssets are you looking at?  Seems like a bug if it is in the lock file but not returned by the task.  /cc @jasonmalinowski

src/nuget/xunit.console.netcore.nuspec
ericstj: No-one should be using the "aspnetcore50" moniker any more.

src/nuget/xunit.console.netcore.nuspec
dsplaisted: After some investigation, it looks like the bug is with the NuGet restore command, not the ResolveNuGetPackageAssets task.  I've filed a bug (NuGet/Home#1454) and sent a PR (NuGet/NuGet3#132) to fix it.

In the meantime we need a solution to deploy the xUnit .NET Core Console runner for MSBuild.  Can we go ahead and merge this PR to unblock us?  I think this is the cleanest solution until the NuGet bug is fixed.  NuGet v3 has improved support for .props and .targets files and no longer needs to modify your project files to include them.

weshaggard: You don't have to modify your project files but you do end up with a *.nuget.props/targets file in every directory for projects that consume these. I would prefer if we can avoid those and instead include this copy logic in a common test.targets from build tools, assuming you are using build tools package.

dsplaisted: @weshaggard We've added *.nuget.props and *.nuget.targets to our .gitignore.  Does that address your concerns?

weshaggard: That doesn't help all it does is hide them from git status and require all repos consuming this package to do something similar or commit them. It also manages to get the projects into inconsistent states on different machines based on whether or not they did a restore in VS that updates this or not.

weshaggard: Are you guys using the testing targets we have in buildtools? If so it would be good to just get those updated to copy, if they don't already I cannot remember, instead.

ericstj: Without the workaround you could use runtimes/any/native/xunit.console.netcore.exe, assuming you bring in our runtime graph package (ms.netcore.platforms, or one of the packages that refs it: ms.netcore, ms.netcore.targets).

dsplaisted:
@ericstj's suggestion of using /runtimes/any/native works.  I think this makes the discussion of the .props files moot, since we don't need them anymore.  I've rebased this PR down to one commit.

How does this look now?

src/nuget/xunit.console.netcore.nuspec
dsplaisted: @ericstj The aspnetcore50 moniker is currently still in use in [tests.targets](https://github.com/dotnet/buildtools/blob/65d93460c83bdcabfc1c0e53ff74f1f344cafa47/src/Microsoft.DotNet.Build.Tasks/Targets/tests.targets#L47)

ericstj: We should file an issue to get rid of that.

ericstj:
LGTM

src/nuget/xunit.console.netcore.nuspec
dsplaisted: I've filed #287 to get rid of the aspnetcore50 TFM

PR #277, opened 2015-09-20
Fix code coverage report generation

stephentoub:
Code coverage is failing because the path to the report generator tool is
being created incorrectly.  It appears this is due to the version number
of the report generator changing from 2.3.0.0 to 2.3.0.  As long as I was
fixing that, I updated it to the most recent version, 2.3.1.  I verified
code coverage works now with the change.

cc: @weshaggard, @mmitche

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/CodeCoverage.targets
@@ -6,7 +6,7 @@
   -->
   <PropertyGroup>
     <OpenCoverVersion>4.6.166</OpenCoverVersion>
-    <ReportGeneratorVersion>2.3.0.0</ReportGeneratorVersion>
+    <ReportGeneratorVersion>2.3.1</ReportGeneratorVersion>
     <CoverallsUploaderVersion>1.4</CoverallsUploaderVersion>
   </PropertyGroup>
   
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/test-runtime/project.json
@@ -6,7 +6,7 @@
 
       "coveralls.io": "1.4",
       "OpenCover": "4.6.166",
-      "ReportGenerator": "2.3.0.0",
+      "ReportGenerator": "2.3.1",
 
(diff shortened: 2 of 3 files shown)


weshaggard:
Yes after our switch to DNU beta7 it is much stricter about using normalizing to 3 part version numbers which is why this started popping up.

Other than my comment about the other projects this looks good. If you want you can simply undo those and I can handle the fixing of those project.json files.

src/Microsoft.DotNet.Build.Tasks/Targets/CodeCoverage.targets
```diff
@@ -6,7 +6,7 @@
   -->
   <PropertyGroup>
     <OpenCoverVersion>4.6.166</OpenCoverVersion>
-    <ReportGeneratorVersion>2.3.0.0</ReportGeneratorVersion>
+    <ReportGeneratorVersion>2.3.1</ReportGeneratorVersion>
     <CoverallsUploaderVersion>1.4</CoverallsUploaderVersion>
```
weshaggard: Actually i suspect we have the same problem with the CoverallsUploader version. You may need to change this to 1.4.0.

stephentoub: Hmm... I don't think we're doing anything with coveralls anymore.  @mmitche?  Maybe a separate PR should just delete that stuff entirely?

stephentoub:
Thanks, Wes.

> Other than my comment about the other projects this looks good

Just to confirm, you're referring to the coveralls comment?  That's the only other comment I see, but I'm verifying because it comes after this comment in the GitHub UI, so I'm concerned I've somehow missed some other comment from you.

src/xunit.console.netcore/project.lock.json
```diff
@@ -35,7 +35,7 @@
           "lib/dotnet/System.Collections.Concurrent.dll": {}
         }
       },
-      "System.Console/4.0.0-beta-23316": {
```
weshaggard: Can you please fix the project.json version of this to 23316 for this project and the other? I want to avoid updating these versions and I want to start removing the * versions from them.

weshaggard:
@stephentoub I just noticed I didn't post my other comment about the other projects. I posted it now.

stephentoub:
> I just noticed I didn't post my other comment about the other projects. I posted it now.

Ok, thanks.  I removed the other automatic changes.

PR #278, opened 2015-09-20
Update Test-Runtime to use Console meta-package

weshaggard:
By using the Console meta-package we don't have to reference
all a ton of individual packages and keep them updated. This allows
us to update all the dependent packages just by updating the one
meta-package.


weshaggard:
cc @iamjasonp this update should get you the later versions of the dependent packages you need to test wcf.

iamjasonp:
Thanks @weshaggard! We'll try it out.

PR #281, opened 2015-09-22
Update BuildTools to support x86 and ubuntu

weshaggard:
tool-runtime\project.lock.json file can now resolve assets for win7-x86 and ubuntu-x64. 

Also update GenFacades to be build as AnyCPU so we can run it on win7-x86, which we need to do for our internal builds.


weshaggard:
@joperezr FYI I did the GenFacades exe update because I needed it for another fix I'm working on.

joperezr:
@weshaggard yes, thanks for letting me know.

PR #285, opened 2015-09-22
Add the ability to force test runs.

JeremyKuhne:
When diagnosing test runs it is useful to run multiple times without
changes.

@weshaggard


weshaggard:
I was actually thinking about this the other day and the approach I was thinking about is detecting the passed target and if it was "Test" or "BuildAndTest" that it would clean out the semaphore file so it would force them to run again. However I couldn't figure out a good way to do that, baring that I think your approach seems reasonable.

JeremyKuhne:
@weshaggard We can always iterate on the design here. I'll put this one in for now. I'll just wait until the next build tools update as well.

PR #289, opened 2015-09-29
Support a project type to deploy assets from nuget packages

ericstj:
This adds depProj.targets which minimally extend Microsoft.Common.Targets
to merely copy assets out of packages referenced by project.json.  The targets
can be configured to copy a specific asset type: (default) ReferenceCopyLocalPaths
for runtime assets, Reference for compile assets, or Analyzer for analyzers.

They do very little to change the nuget resolution process so that remains
driven as it is today by the normal project properties like TargetFramework,
TargetPlatform, and PlatformTarget.

I also made a change to plug these in to build tools for .depproj extension,
and a change to our reference assembly targets to set the property to indicate
that compile time assets should be used.

I plan to follow this up with changes in corefx that use .depproj's to deploy
older versions of our previously shipped reference assemblies.  These will
be consumed by future packages using dotnet "generations" to have
packages include their entire history of surface area and thus support
older platforms rather than drop them.

/cc @weshaggard @chcosta @FiveTimesTheFun

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/depProj.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/depProj.targets
@@ -53,8 +53,11 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
       <!-- Don't set IntermediateAssembly since this is not produced -->
       <IntermediateAssembly Remove="@(IntermediateAssembly)" />
       
-      <!-- filter to items that came from packages -->
-      <NuGetDeploy Include="@($(NuGetDeploySourceItem))" Condition="'%($(NuGetDeploySourceItem).NuGetPackageId)' != ''" />
+      <NuGetDeploy Include="@($(NuGetDeploySourceItem))" />
+      <!-- filter to only items that came from packages -->
+      <!-- the following condition must be applied after the include because msbuild doesn't seem
+           to support property-defined-item-names in a metadata statement -->
+      <NuGetDeploy Remove="@(NuGetDeploy)" Condition="'%(NugetDeploy.NuGetPackageId)' == ''" />
       
       <!-- remove all existing items from NuGet packages we'll be defining these in our own item -->
       <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' != ''"/>
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/depProj.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/depProj.targets
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+***********************************************************************************************
+depProj.targets
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks/Targets/depProj.targets
```diff
@@ -0,0 +1,74 @@
...
+                      '$(NuGetDeploySourceItem)' != 'Reference' AND
+                      '$(NuGetDeploySourceItem)' != 'Analyzer'"
+           Text="Unexpected value for NuGetDeploySourceItem:'$(NuGetDeploySourceItem)'.  Expected ReferenceCopyLocalPaths, Reference, or Analyzer." />
+
+    <ItemGroup>
+      <!-- Don't set IntermediateAssembly since this is not produced -->
+      <IntermediateAssembly Remove="@(IntermediateAssembly)" />
+      
+      <!-- filter to items that came from packages -->
+      <NuGetDeploy Include="@($(NuGetDeploySourceItem))" Condition="'%($(NuGetDeploySourceItem).NuGetPackageId)' != ''" />
```
weshaggard: Interesting this is the first time I've ever seen an msbuild property being used as the item name.

src/Microsoft.DotNet.Build.Tasks/Targets/depProj.targets
```diff
@@ -0,0 +1,74 @@
...
+      
+      <!-- remove all existing items from NuGet packages we'll be defining these in our own item -->
+      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' != ''"/>
+      <Reference Remove="@(Reference)" Condition="'%(Reference.NuGetPackageId)' != ''"/>
+      <Analyzer Remove="@(Analyzer)" Condition="'%(Analyzer.NuGetPackageId)' != ''"/>
+
+      <!-- add items defined by NuGetDeployItem property to ReferenceCopyLocalPaths -->
+      <ReferenceCopyLocalPaths Include="@(NuGetDeploy)" />
```
weshaggard: Why reuse the ReferenceCopyLocalPaths? and not just use the NugetDeploy item group? That would also not force you to clear the other item groups and they could be used in some potential cases.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/Targets/depProj.targets
ericstj: ReferenceCopyLocalPaths is already populated by the resolution target and common targets will copy it to the output.  If we didn't clear ReferenceCopyLocalPaths we'd binclash with the resolved copy-local runtime assets.  
I could leave Reference and Analyzers defined, but those have no use for a 'depProj' (and references causes RAR to do extra work) so I've cleared them.

ericstj:
This is actually going to start copying a lot of things to the output directory that previously were not copied (due to arguably a bug in our implementation of package resolution targets).  I need to fix that up so that I don't start bloating people's output directories.

ericstj:
I've fixed the extra copies and tried to get us more in-sync with the shipping targets.  Validated this on corefx to ensure that the bin folder is the same size before/after this change.

PR #290, opened 2015-09-29
Initial support for Helix test runner.

jhendrixMSFT:
Update packageresolve.targets to generate AssemblyList.txt for test
projects.
Update publishtest.targets to conditionally compress test build output.
Added supporting msbuild tasks for identifying test assemblies containing
performance tests, writing test list json and archiving directories.


akoeplinger:
What is the "Helix test runner" ?

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -103,4 +103,20 @@
     <Message Text="Excluding @(_ReferenceFileNamesToRemove);@(_NoneFileNamesToRemove) from package references since the same file is provided by a project refrence."
              Condition="'@(_ReferenceFileNamesToRemove)' != '' or '@(_NoneFileNamesToRemove)' != ''"/>
   </Target>
+
+  <!-- the helix test infrastructure uses the generated assembly list to determine the runtime binaries to restore -->
```
weshaggard: Lets avoid using code names like helix here.

weshaggard:
cc @pallavit as she is working on this task library to make it .NET Core compliant.

src/Microsoft.DotNet.Build.Tasks/GetPerfTestAssemblies.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public override bool Execute()
+        {
+            Log.LogMessage(MessageImportance.High, "About to inspect {0} test assemblies.", TestBinaries.Length);
+            var perfTests = new List<ITaskItem>();
+
+            foreach (var testBinary in TestBinaries)
+            {
+                Log.LogMessage(MessageImportance.Low, "Inspecting assembly {0}.", testBinary.ItemSpec);
+                Assembly assembly = Assembly.ReflectionOnlyLoadFrom(testBinary.ItemSpec);
```
weshaggard: We should not be using ReflectionOnlyLoad. We should use the MetadataReader for this activity.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -103,4 +103,20 @@
...
+  <Target Name="GenerateAssemblyListTxt" AfterTargets="ResolveNuGetPackages" Condition="'$(PrereleaseResolveNuGetPackages)'=='true' and '$(IsTestProject)'=='true'">
+    <Message Text="Generating $(OutDir)AssemblyList.txt" />
+
+    <ItemGroup Condition="Exists('$(ProjectLockJson)')">
+      <AssemblyPaths Include="$([MSBuild]::MakeRelative($(PackagesDir), %(CopyLocalItems.FullPath)))" />
+    </ItemGroup>
+
+    <WriteLinesToFile Condition="Exists('$(ProjectLockJson)')"
```
weshaggard: Can you help me understand why it needs a separate list as opposed to resolving these and copying them to a directory and zipping that up?

Also is this even the complete list? I would expect it would need more things like the test-runtime file set as well.

src/Microsoft.DotNet.Build.Tasks/Targets/packageresolve.targets
```diff
@@ -87,7 +87,7 @@
                                OmitTransitiveCompileReferences="$(OmitTransitiveCompileReferences)">
       <Output TaskParameter="ResolvedAnalyzers" ItemName="Analyzer" />
       <Output TaskParameter="ResolvedReferences" ItemName="Reference" />
-      <Output TaskParameter="ResolvedCopyLocalItems" ItemName="None" />
+      <Output TaskParameter="ResolvedCopyLocalItems" ItemName="CopyLocalItems" />
```
weshaggard: This conflicts with @ericstj PR https://github.com/dotnet/buildtools/pull/289

ericstj: This should be ReferenceCopyLocalPaths to be consistent with the inbox targets.

ericstj: Please merge in my changes.  If you need to get just nuget copy local assets the right way to do that is 
`<Foo Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NugetPackageId)' != ''
" />`

As was mentioned in the other comment you might want to consider more than just the packages.  Consider either all of ReferenceCopyLocalPaths (defined after ResolveReferences target closure runs), or more.  To get an idea of everything that might be relevant have a look at the CopyFilesToOutputDirectory target in ms.common.targets and all the targets it depends on.

ellismg:
To answer @akoeplinger's question, Helix is an internal test execution system.  We want to start running our tests on it in addition to the runs we do today in Jenkins.

src/Microsoft.DotNet.Build.Tasks/CreateAzureContainer.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(string.Format("DefaultEndpointsProtocol=https;AccountName={0};AccountKey={1}", AccountName, AccountKey));
+            CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
+            CloudBlobContainer newContainer = blobClient.GetContainerReference(ContainerName);
+            newContainer.CreateIfNotExists();
+            StorageUri = newContainer.Uri.ToString();
+
+            // if ReadOnlyToken or WriteOnlyToken != null they should contain the number of days before the token expires
+
+            if (ReadOnlyToken != null)
```
ChadNedzlek: Isn't ReadOnlyToken an output parameter? As such, it shouldn't ever be set at this point, so this code will never run (and then it strangely overwrites it at the end of the if block anyway). Or am I missing something?

src/Microsoft.DotNet.Build.Tasks/CreateAzureContainer.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public sealed class CreateAzureContainer : Task
+    {
+        [Required]
```
ChadNedzlek: There need to be comments about what these properties are and what format they are in.

src/Microsoft.DotNet.Build.Tasks/CreateAzureFileShare.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public sealed class CreateAzureFileShare : Task
+    {
+        [Required]
+        public string AccountName { get; set; }
+
+        [Required]
+        public string AccountKey { get; set; }
+
+        [Output]
```
ChadNedzlek: It looks a little strange to have an output parameter mixed in with the input ones, though I understand that it's because that's alphabetically correct...

src/Microsoft.DotNet.Build.Tasks/CreateAzureContainer.cs
jhendrixMSFT: I was trying to reduce the number of properties.  How this works is when calling CreateAzureContainer if a RO token is required you specify the number of days for which it should last in ReadOnlyToken.  The token is created and then placed in ReadOnlyToken.  I suppose the complexity is the same either way (i.e. lots of properties or this behavior) so if people find it confusing I will change it to explicit properties.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -122,6 +124,29 @@
     </ItemGroup>
 
     <Copy Condition="'$(BUILDTOOLS_OVERRIDE_RUNTIME)' != ''" SourceFiles="@(TestRuntimeSource)" DestinationFolder="$(TestPath)%(TestTargetFramework.Folder)" />
+
+    <!-- the helix test infrastructure uses the generated assembly list to determine the runtime binaries to restore -->
+    <Message Text="Generating $(OutDir)AssemblyList.txt" />
```
ChadNedzlek: This is a little odd in this location, if I'm reading this correctly, because it's in the "CopyTestToTestDirectory" target... which is something we wouldn't want to do for tests that are about to be zipped/uploaded, since that will contain all the product binaries inside the test package.

src/Microsoft.DotNet.Build.Tasks/SendToEventHub.cs
```diff
@@ -0,0 +1,35 @@
...
+    public sealed class SendToEventHub : Task
+    {
+        [Required]
+        public string ConnectionString { get; set; }
+
+        [Required]
+        public string EventHubPath { get; set; }
+
+        [Required]
+        public string EventJson { get; set; }
```
ChadNedzlek: We don't actually care that this is Json, so maybe rename it to just "EventData"?

src/Microsoft.DotNet.Build.Tasks/WriteItemsToJson.cs
```diff
@@ -0,0 +1,77 @@
...
+                            var mdValue = customMd[key].ToString();
+
+                            jsonWriter.WritePropertyName(mdString);
+
+                            if (mdValue.Length > 0 && mdValue[0] == '[' && mdValue[mdValue.Length - 1] == ']')
+                            {
+                                mdValue = mdValue.Substring(1, mdValue.Length - 2);
+                                jsonWriter.WriteStartArray();
+
+                                var parts = mdValue.Split(new char[] { ',' });
```
ChadNedzlek: This is going to cause a problem if the data contains commas, which seems like something it should handle.  It seems to be a sort of JSON encoded thing, since it's checking for [...], but doesn't handle quoting.

I'm a little confused as to what the format of these values are supposed to be.

src/Microsoft.DotNet.Build.Tasks/CreateAzureContainer.cs
ChadNedzlek: Is it actually possible to have a parameter be both input and output?  I don't ever recall seeing something like that in other MSBuild tasks.  Even if it is possible, it might confuse people, since it changes format.  As input it's "string of an integer that is number of days", and output it's "query parameter without the ?".

ChadNedzlek:
LGTM

PR #291, opened 2015-09-29
Removing GenFacades nuget packages and minor changes on the runtime targets so that it runs in Ubuntu

joperezr:
Changes:	
       1.-Removing GenFacades Nuget packages since it will now be
	includded in buildtools.
	2.-Modifications to the targets to make it run on Ubuntu.


joperezr:
cc @weshaggard

src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
```diff
@@ -11,11 +11,13 @@
          This is probably because the code coverage target invocation does not canonicalize the
          process name. -->
     <ToolHost Condition="'$(OS)' == 'Windows_NT'">CoreRun.exe</ToolHost>
-    <ToolHost Condition="'$(OS)' != 'Windows_NT'">corerun</ToolHost>
```
joperezr: I am actually confirming that this is right, since I was seeing that the right name was with caps on my Linux machine but it doesn't look that way on the TestHost nuget package, so I'll make sure that this is correct, or revert this line if it's no longer correct.

weshaggard: It should be all lower case. This was fixed recently in our packages to ensure it is all lowercase 'corerun'.

src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
```diff
@@ -54,7 +56,7 @@
 
     <Exec
       Condition="'$(OS)' != 'Windows_NT'"
-      Command="chmod a+x &quot;$(ToolRuntimePath)/corerun&quot;" />
+      Command="chmod a+x &quot;$(ToolRuntimePath)/CoreRun&quot;" />
```
weshaggard: This should also be all lowercase. In fact we should probably just use the ToolHost property.

joperezr: Ok, I've fixed that and pushed the changes to the PR.

src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
```diff
@@ -14,8 +14,10 @@
     <ToolHost Condition="'$(OS)' != 'Windows_NT'">corerun</ToolHost>
 
     <ToolTargetFramework Condition="'$(ToolTargetFramework)' == ''">DNXCore,Version=v5.0</ToolTargetFramework>
+    <ToolOS Condition="'$(OS)' == 'Windows_NT'">win7</ToolOS>
+    <ToolOS Condition="'$(OS)' != 'Windows_NT'">ubuntu.14.04</ToolOS>
```
weshaggard: @ellismg is this a reasonable default or would you prefer a different detection mechanism?

ellismg: I am not sure.  I decided against doing something like this in #293 since not only is it wrong for different distros, it's also wrong for different non windows oses (e.g OSX).  I think no matter the default, the build consuming build tools is going to need to setup ToolNugetRuntimeId, there really isn't a good default.  I sort of wish we would just always restore the windows version so folks that need to set it discover it quickly rather than it working if they are on ubuntu and blowing up everywhere else?

joperezr: Ok, I will remove this then and from coreclr repo I'll do something similar to your PR where you check if corerun exists, and set the NugetRuntimeId over there. Thanks for the feedback! @weshaggard with this in mind, should we add more distros and osx to the tool-runtime\project.json runtimes?

weshaggard: We probably should go ahead and add osx. We should figure out how we make this extensible so that people can do this without requiring a build tools update, at least temporarily. I wonder if we should generate this project.json file on the fly and include the runtime id's based on what is setup and then generate the lock file at that point so we don't have to keep updating the lock file for all possible runtimes.

joperezr: Ok, I'm adding osx and I'll start looking into generating the project.json on the fly. I'll push my changes shortly.

weshaggard:
LGTM

weshaggard:
Actually it looks like you are missing the updated lock file after adding osx to the runtimes please add that update.

joperezr:
Good catch, I just pushed the change fixing it. Let me know if it's ok to merge.

PR #293, opened 2015-09-29
Fix build breaks on Linux with 00097

### LINKED ISSUE #292: Microsoft.DotNet.BuildTools.1.0.25-prerelease-00097 is broken on Linux
...at least on non-Ubuntu distros: an incorrect tool runtime is being deployed, which causes `chmod a+x <ToolRuntimePath>/corerun` to fail.

ellismg:
By default, build tools would restore the windows version of the runtime
packages (even when '$(OS)' was not 'Windows_NT'). While a project
consuming the build tools could override this, it meant if they did not
the chmod would fail because it couldn't find corerun (since we instead
restored CoreRun.exe).

Guard the chmod with a test to ensure corerun actually exists.

Fixes #292

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
@@ -53,7 +53,7 @@
       RetryDelayMilliseconds="$(CopyRetryDelayMilliseconds)" />
 
     <Exec
-      Condition="'$(OS)' != 'Windows_NT'"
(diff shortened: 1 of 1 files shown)


ellismg:
I considered trying to use a better RID when OS != Windows_NT, but there's no good choice since in not only do we not understand what distro we are on, we also don't even understand if we are on OSX or Linux.

pgavlin:
LGTM. Have you filed an issue to follow up on this workaround, or is https://github.com/dotnet/corefx/issues/3538 sufficient?

ellismg:
I think the dotnet/corefx issue is sufficient.  Understanding the requirements here will inform the final design of the system.

weshaggard:
LGTM

PR #294, opened 2015-09-30
Disable RAR warning about mismatched portable references

weshaggard:
While we label ourselves as .NETPortable v4.5 we aren't a traditional
portable project and we can correctly reference a .NETPortable v5.0 project.
So this disabled the RAR warning for projects that target the default.

We will eventually switch our defaults to .NETPortable v5.0 as that is
closer to what we truely are be we aren't ready to force everyone to use
VS 2015 yet and require the new portable targeting pack when we don't even
really use it.

cc @ericstj @nguerrera


ericstj:
I'm OK with this for our projects.

ericstj:
@Pilchie it comes to mind that we might want to stop writing a target framework attribute on libraries that use package-based portable since it doesn't really have any meaning.

Pilchie:
Tagging @rchande too.  Can you file an issue in our VSO repo for PCL so we don't forget about this?

PR #296, opened 2015-10-01
Updated lock files and netcore nuspec

ianhays:
This should make build tools 101 what we want for CoreFX. Will update https://github.com/dotnet/corefx/pull/3599/files accordingly.

@pgavlin @ellismg @mellinoe


ellismg:
Should we not update https://github.com/dotnet/buildtools/blob/master/src/nuget/xunit.console.netcore.nuspec as well?

ellismg:
LGTM

ianhays:
Need to first update the nuspecs for version 101, then I will pull consume that new package and make version 102 the one CoreFX uses

PR #299, opened 2015-10-02
Updated perf targets for new xunit

ianhays:


ianhays:
@mellinoe @weshaggard

src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
```diff
@@ -40,7 +39,9 @@
 
   <!-- Executable properties -->
   <PropertyGroup>
-    <XunitRunnerArgs>$(TargetFileName) -runnerhost $(TestHost) -runner $(XunitExecutable) -runid $(PerfRunId) -verbose</XunitRunnerArgs>
+    <PerfRunId Condition="'$(PerfRunId)' == ''">$(TargetFileName)-WindowsCore</PerfRunId>
```
mellinoe: Minor note: should we just use $(OS), or do we think running on Linux will probably be different enough that this won't be re-used there?

mellinoe:
LGTM

src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
ianhays: I can't imagine it would be reusable, especially since the unix runner is dnx based now

ianhays:
Thanks @mellinoe!

src/Microsoft.DotNet.Build.Tasks/Targets/PerfTesting.targets
```diff
@@ -4,8 +4,6 @@
   <PropertyGroup>
     <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'false' and '$(OS)' == 'Windows_NT' and Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" >true</RunPerfTestsForProject>
```
weshaggard: orthogonal but can you please remove this check "Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" after my change https://github.com/dotnet/buildtools/pull/294 I don't expect this to be required any longer.

PR #301, opened 2015-10-05
Adding support for tool-runtime in different linux distros

joperezr:
Adding support to set an Env Variable that points to a directory with the tool runtime, so that we support building mscorlib in different linux distros like Fedora

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
@@ -23,7 +23,7 @@
     <!-- If COMPLUS_InstallRoot is set clear it before calling the ToolHost otherwise some root activations like PDB COM activation will fail -->
     <ToolHostCmd Condition="'$(COMPLUS_InstallRoot)' != ''">(set COMPLUS_InstallRoot=) &amp; $(ToolHostCmd)</ToolHostCmd>
   </PropertyGroup>
-
+  
   <Target Name="EnsureBuildToolsRuntime"
       Inputs="$(BuildToolsSemaphore)"
       Outputs="$(ToolRuntimeSempahore)">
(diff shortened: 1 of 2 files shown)


joperezr:
@ellismg Here is the PR that we talked about for not breaking building coreclr on different distros

ellismg:
Can we change the structure of this such that we always restore from NuGet and then if the environment variable is set we just copy what's in there over the tool runtime?  That way I can just but native components in that folder instead of having to have a complete tool runtime.

src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
```diff
@@ -24,9 +24,22 @@
     <ToolHostCmd Condition="'$(COMPLUS_InstallRoot)' != ''">(set COMPLUS_InstallRoot=) &amp; $(ToolHostCmd)</ToolHostCmd>
   </PropertyGroup>
 
+  <!-- If BUILDTOOLS_TOOLS_RUNTIME is set, then we will use that as the tool and test runtime, else we will restore it from NuGet. -->
+  <Target Name="UseBuildToolsRuntimeFromEnvVariable"
```
weshaggard: I don't think any of this is necessary. If folks want to use a different runtime they can simply set ToolRuntimePath to point at the directory they have setup. They would also need to copy the ToolRuntime.semaphore file there as well.

ellismg: I would like us to have a way to make it easy for someone to bring just the native components instead of having to build out an entire tool runtime.  Otherwise folks are going to have to fight with NuGet to pull down all the managed stuff + our tools.

weshaggard: Currently this change will not enable that either, but we should be careful about mixing and matching these components. I agree we need to have a way to override but I'm not sure we want people to do it in a piece-wise fashion.

ellismg: > Currently this change will not enable that either

Yes, I provided feedback to that point.

> but we should be careful about mixing and matching these components. I agree we need to have a way to override but I'm not sure we want people to do it in a piece-wise fashion.

You will have a catch-22 for folks like Pat if you require that they bring an entire tool-runtime for their system but provide no way to build it.  Given that you will need a tool runtime to build managed code it seems like the best strategy is going to be restore the managed code using nuget and then fix up the native bits that are wrong for your system.

If you would rather provide a script that handles this (and just have folks override ToolRuntimePath), I'm fine with that approach, but it needs to be dead simple for me to produce a tool runtime on a machine where I don't have a functional version of CoreCLR to start with.

weshaggard:
This looks like a reasonable way to enable an extension point for folks to override some libraries in both the test and tool runtime if they need to.

PR #302, opened 2015-10-05
Add a hook for dropping marker files

ericstj:
Our internal build needs to distinguish newly built files from
redistributed binaries.  Provide a hook in depProj.targets to
do this for assets that are copied from nuget packages.

/cc @weshaggard


weshaggard:
LGTM

PR #303, opened 2015-10-06
Show starting/finishing test progress in xunit console

stephentoub:
When debugging problems in hanging tests, it's very useful to be able to know which tests have started but haven't finished.  This commit adds a "-showprogress" option to the xunit.console.netcore.exe command line which, when set, will result in xunit tracing out to the console the starting and finishing of each individual test, e.g.

```
   System.Threading.Tasks.Dataflow.Tests.ActionBlockTests.TestPost [STARTING]
   System.Threading.Tasks.Dataflow.Tests.ActionBlockTests.TestPost [FINISHED] Time: 0.0316922s
```

cc: @FiveTimesTheFun, @mellinoe, @weshaggard, @ellismg

### DIFF
--- a/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
+++ b/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
@@ -12,17 +12,20 @@ namespace Xunit.ConsoleClient
         readonly object consoleLock;
         readonly ConcurrentDictionary<string, ExecutionSummary> completionMessages;
         readonly string defaultDirectory;
+        readonly bool showProgress;
 
         public StandardOutputVisitor(object consoleLock,
                                      string defaultDirectory,
                                      XElement assemblyElement,
                                      Func<bool> cancelThunk,
-                                     ConcurrentDictionary<string, ExecutionSummary> completionMessages = null)
+                                     ConcurrentDictionary<string, ExecutionSummary> completionMessages = null,
+                                     bool showProgress = false)
             : base(assemblyElement, cancelThunk)
         {
             this.consoleLock = consoleLock;
             this.defaultDirectory = defaultDirectory;
             this.completionMessages = completionMessages;
+            this.showProgress = showProgress;
         }
 
         protected override bool Visit(ITestAssemblyStarting assemblyStarting)
@@ -93,7 +96,26 @@ namespace Xunit.ConsoleClient
 
         protected override bool Visit(ITestStarting testStarting)
         {
+            if (showProgress)
+            {
+                lock (consoleLock)
+                {
+                    Console.WriteLine("   {0} [STARTING]", Escape(testStarting.Test.DisplayName));
+                }
+            }
             return base.Visit(testStarting);
+        }
+
+        protected override bool Visit(ITestFinished testFinished)
+        {
+            if (showProgress)
+            {
+                lock (consoleLock)
+                {
+                    Console.WriteLine("   {0} [FINISHED] Time: {1}s", Escape(testFinished.Test.DisplayName), testFinished.ExecutionTime);
+                }
+            }
+            return base.Visit(testFinished);
         }
 
         protected override bool Visit(IErrorMessage error)
--- a/src/xunit.console.netcore/CommandLine.cs
+++ b/src/xunit.console.netcore/CommandLine.cs
@@ -31,6 +31,8 @@ namespace Xunit.ConsoleClient
 
         public bool? ParallelizeTestCollections { get; set; }
 
+        public bool ShowProgress { get; protected set; }
+
         public bool TeamCity { get; protected set; }
 
         public bool Wait { get; protected set; }
@@ -161,6 +163,11 @@ namespace Xunit.ConsoleClient
                 {
                     GuardNoOptionValue(option);
                     AppVeyor = true;
+                }
+                else if (optionName == "-showprogress")
+                {
+                    GuardNoOptionValue(option);
+                    ShowProgress = true;
                 }
                 else if (optionName == "-noshadow")
                 {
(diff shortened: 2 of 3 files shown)


src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -93,9 +96,22 @@ protected override bool Visit(ITestSkipped testSkipped)
 
         protected override bool Visit(ITestStarting testStarting)
         {
+            if (showProgress)
```
FiveTimesTheFun: I believe you need to lock on consoleLock to prevent this output from being intertwined with other messages or when the console colors are changed for other messages.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -93,9 +96,22 @@ protected override bool Visit(ITestSkipped testSkipped)
...
+                Console.Error.WriteLine("   {0} [STARTING]", Escape(testStarting.Test.DisplayName));
+            }
             return base.Visit(testStarting);
         }
 
+        protected override bool Visit(ITestFinished testFinished)
+        {
+            if (showProgress)
+            {
+                Console.Error.WriteLine("   {0} [FINISHED] Time: {1}s", Escape(testFinished.Test.DisplayName), testFinished.ExecutionTime);
```
FiveTimesTheFun: consoleLock

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
stephentoub: Fixed

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
stephentoub: Fixed

FiveTimesTheFun:
Hi Stephen - can you please add a line to the help output here
https://github.com/dotnet/buildtools/blob/master/src/xunit.console.netcore/Program.cs#L129
documenting -showprogress?

stephentoub:
> can you please add a line to the help output here

I already did:
https://github.com/dotnet/buildtools/pull/303/files#diff-f69859fd9066ece5bff9066a6a2cbe53R129
Is that not sufficient?

FiveTimesTheFun:
Thanks, looks good to me :)

stephentoub:
:)

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -93,9 +96,28 @@ protected override bool Visit(ITestSkipped testSkipped)
 
         protected override bool Visit(ITestStarting testStarting)
         {
+            if (showProgress)
+            {
+                lock (consoleLock)
+                {
+                    Console.Error.WriteLine("   {0} [STARTING]", Escape(testStarting.Test.DisplayName));
+                }
```
weshaggard: I didn't realize these things were getting written to stderr that seems wrong to me. Any ideas why they should go there as opposed to stdout?

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -12,17 +12,20 @@ public class StandardOutputVisitor : XmlTestExecutionVisitor
...
         readonly string defaultDirectory;
+        readonly bool showProgress;
 
         public StandardOutputVisitor(object consoleLock,
                                      string defaultDirectory,
                                      XElement assemblyElement,
                                      Func<bool> cancelThunk,
-                                     ConcurrentDictionary<string, ExecutionSummary> completionMessages = null)
+                                     ConcurrentDictionary<string, ExecutionSummary> completionMessages = null,
+                                     bool showProgress = false)
```
weshaggard: Do we know if constructor is used outside this library? If so adding this new default parameter would be a binary breaking change.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
stephentoub: Don't know; I was just following the precedent used for some of the other ones, like SKIP.  I can certainly change to just use Console.WriteLine if you'd prefer.

weshaggard: Yes I figured you were following the precedent which is why I asked if you knew why that is. I personally think stdout would make more since for this given that it isn't actually an error.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
stephentoub: This is part of the .exe, in our own fork.  I'm not quite understanding the concern.  Can you elaborate?  Who else are you concerned is relying on this file?

weshaggard: I don't have a specific concern, I just wanted to ask the question as I always do when I see public APIs changing like this. I'm OK with this if we don't know of any external consumers of this API, I just wanted to pose the question to ensure it was thought about.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
stephentoub: Fixed.

stephentoub:
Thanks for the review, guys.

ViktorHofer:
@stephentoub any reason why you haven't suggested this to be added to the official xunit repo? I'm currently walking through the commits of our forked runner and this seems to be one of the only changes that haven't yet landed in the parent project. If there wasn't any particular reason, are you willing to create an issue?

ViktorHofer:
I created https://github.com/xunit/xunit/issues/1720 to track this.

stephentoub:
No reason other than I saw they already had the start logging, which represented the 90% case for why I added this in the first place (hanging tests).  But the timing information is useful as well, so that would be good to add.

ViktorHofer:
okay cool. I will submit the PR for the finish event + time tomorrow.

Issue #286, opened 2015-09-23
Should ToolRuntime be deployed to obj, not bin?


ellismg:
Today the runtime to run our tools under gets deployed to a ToolRuntime folder in `$(BaseOutputPath)\$(OSPlatformConfig)`, I think that `$(IntermediateOutputRootPath)` (e.g. `bin\obj\Linux.AnyCPU.Debug`) is a better place for it.  What do you think @weshaggard?

weshaggard:
@ellismg I put it under bin because that is where we deploy the test-runtime as well, however I don't have a strong opinion about this. What benefits do you feel intermediate has over bin?

ellismg:
My preference is that things produced by the build are in bin and things used to produce the build are in obj.  I think both the tools and test runtime should be in obj (and for now when we build out the tests folders for each test, we continue to copy the test runtime into there).

weshaggard:
I'm cool with this. I will try to get this this at some point but feel free to submit a PR.

mellinoe:
Just wanted to add my support for this. I'd like if everything under bin/* were actual product binaries. It would even be nice to split test binaries and framework product binaries into separate directories.

joperezr:
I'll do this now since I need it as part of consuming the new version of buildtools from the coreclr repo. Will send PR soon.

PR #304, opened 2015-10-08
Changing ToolRuntimePath to be in the Intermediate folder instead of bin output

joperezr:
Changing ToolRuntimePath to be in the Intermediate folder instead of bin output

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
@@ -17,7 +17,7 @@
     <ToolArchitecture Condition="'$(ToolArchitecture)' == ''">x64</ToolArchitecture>
     <ToolNugetRuntimeId Condition="'$(ToolNugetRuntimeId)' == ''">win7-$(ToolArchitecture)</ToolNugetRuntimeId>
 
-    <ToolRuntimePath Condition="'$(ToolRuntimePath)' == ''">$(BaseOutputPath)$(OSPlatformConfig)\ToolRuntime\</ToolRuntimePath>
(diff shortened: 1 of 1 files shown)


joperezr:
cc @ellismg @weshaggard

joperezr:
This is to address issue #286 
We need this change in order to be able to upgrade buildtools used in coreclr repo

ellismg:
LGTM, but Wes should take a look as well.

src/Microsoft.DotNet.Build.Tasks/Targets/toolruntime.targets
```diff
@@ -17,7 +17,7 @@
     <ToolArchitecture Condition="'$(ToolArchitecture)' == ''">x64</ToolArchitecture>
     <ToolNugetRuntimeId Condition="'$(ToolNugetRuntimeId)' == ''">win7-$(ToolArchitecture)</ToolNugetRuntimeId>
 
-    <ToolRuntimePath Condition="'$(ToolRuntimePath)' == ''">$(BaseOutputPath)$(OSPlatformConfig)\ToolRuntime\</ToolRuntimePath>
+    <ToolRuntimePath Condition="'$(ToolRuntimePath)' == ''">$(IntermediateOutputPath)ToolRuntime\</ToolRuntimePath>
```
weshaggard: This should be $(IntermediateOutputRootPath) because we want it to be shared with all the projects and $(IntermediateOutputPath) includes the project name in the path. See https://github.com/dotnet/corefx/blob/master/dir.props#L226

joperezr: Ok I'll change it. The reason why I chose that variable is because that si the one defiined in coreclr repo which is the one I used to test my change. I'll change this and then define the RootPath in coreclr repo if it's not defined.

weshaggard:
LGTM

PR #306, opened 2015-10-09
Put quotes around InstallLocation to handle paths with spaces

weshaggard:


joperezr:
In line 105 of the file changed, we are passing the same variable to the task of WriteLinesToFile, do we also need to add quotes there?

weshaggard:
No in fact it fails if you do. The difference here is that goes into a Task string which can handle strings where as the ones I needed to fix were command line strings going to Exec which need to be properly quoted.

joperezr:
I see. LGTM.

PR #307, opened 2015-10-09
Support automatic nuget package creation for test packages

FiveTimesTheFun:
Support automatically creating nuget packages for test packages,
with the GenerateTestPackage msbuild property.  This brings in and
uses NuProj (https://github.com/nuproj/nuproj/).  The static nuspec
fields are stored in testpackage.nuproj, and the dynamic fields are
calculated and set by publishtest.targets.  Currently, the content of
the test package is the same as the test runtime directory without
dependencies.


FiveTimesTheFun:
Goes with https://github.com/dotnet/corefx/pull/3762

FiveTimesTheFun:
cc @weshaggard @terrajobst @Chrisboh

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -20,6 +20,7 @@
       <dependency id="Microsoft.DotNet.TestHost" version="1.0.3-prerelease" />
       <dependency id="Microsoft.DotNet.CoreCLR" version="1.0.0-prerelease" />
       <dependency id="Microsoft.DotNet.BuildTools.ApiTools" version="1.0.0-prerelease" />
+      <dependency id="NuProj" version="0.10.4-beta-gf7fc34e7d8" />
```
weshaggard: I just removed all the BuildTools package dependencies and similarly this shouldn't be a reference package dependency.

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
```diff
@@ -15,11 +15,33 @@
...
+      <TestPackageContent>@(_TestCopyLocalByFileNameWithoutDuplicates->'%(SourcePath)')</TestPackageContent>
+      <TestPackageBuild Condition="'$(TestPackageBuild)'==''">$(BuildNumberMajor)</TestPackageBuild>
+      <TestPackageBuild Condition="'$(TestPackageBuild)'==''">000000</TestPackageBuild>
+      <TestPackageVersion Condition="'$(TestPackageVersion)'==''">1.0.0-prerelease-$(TestPackageBuild)</TestPackageVersion>
+    </PropertyGroup>
+    <MSBuild Projects="$(MSBuildThisFileDirectory)testpackage.nuproj"
+      Properties="
+      IntermediateOutputPath=$(IntermediateOutputPath);
+      OutputPath=$(OutputPath);
+      NuProjPropsPath=$(NuProjDir)NuProj.props;
```
weshaggard: Where is NuProjDir getting set?

weshaggard:
@ericstj what do you think about the nuproj dependency?

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
FiveTimesTheFun: Hi Wes, I saw you found where this is set in https://github.com/dotnet/corefx/pull/3762#discussion-diff-41768313R45.

I put the definition of NuProjDir there following the pattern of PackagesDir being set in the corefx repo and consumed in the buildtools repo.  Since NuProjDir depends on PackagesDir and NuProjVersion both of which we know in buildtools, I could move the definition of NuProjDir here, and agree that would be better.  I'll update both pull requests with that change.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
FiveTimesTheFun: Hi Wes, do you mean the NuProj binary and targets should be embedded in this package, or the dependency should be implicit and the consumers responsible for adding it?  What's your thinking here?

src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets
FiveTimesTheFun: Updated.  I added condition guards on NuProjDir and NuProjVersion so if a consuming repo takes a different version, they can override these values.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
weshaggard: If it is necessary it should be included in the BuildTools package itself. Please resolve your conflicts and you will see that I deleted all other dependencies in the nuspec.

FiveTimesTheFun: Thanks - I wasn't seeing the change when I tried to view the history of the nuspec file through this PR, but see it now.  I'll update to include the NuProj binary and targets in this package.

src/BuildValues.props
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
jasonwilliams200OK: Seems like notepad added BOM at the begining of the file. It can be removed using [HexEdit](http://hexedit.com/) on Windows. While at it, this file is also missing end of file linefeed. :)

src/BuildValues.props
```diff
@@ -8,6 +8,6 @@
       prerelease version number and without the leading zeroes foo-20 is
       smaller than foo-4.
     -->
-    <RevisionNumber>00105</RevisionNumber>
+    <RevisionNumber>00106</RevisionNumber>
```
weshaggard: You shouldn't edit this revision directly it happens automatically as part of our official build.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -20,6 +20,8 @@
       <dependency id="Microsoft.DotNet.TestHost" version="1.0.3-prerelease" />
       <dependency id="Microsoft.DotNet.CoreCLR" version="1.0.0-prerelease" />
       <dependency id="Microsoft.DotNet.BuildTools.ApiTools" version="1.0.0-prerelease" />
+      <!-- This version should be kept up to date with the version in publishtest.targets -->
+      <dependency id="NuProj" version="0.10.4-beta-gf7fc34e7d8" />
```
weshaggard: As we chatted offline I'm expecting you will be removing this dependency.

FiveTimesTheFun: Yes, thank you, updated.

src/BuildValues.props
FiveTimesTheFun: Okay, delegating to dotnet-bot.

src/BuildValues.props
FiveTimesTheFun: Thanks Jason - this file is automatically generated and I reverted my changes in the most recent commit.  @dotnet-bot will update it as part of our official build.

FiveTimesTheFun:
Hi @weshaggard  - how does this look after addressing your feedback?

src/.nuget/packages.Windows_NT.config
```diff
@@ -1,5 +1,7 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Microsoft.DotNet.BuildTools" version="1.0.25-prerelease-00087" />
   <package id="dnx-coreclr-win-x86" version="1.0.0-beta7" />
+  <!-- This version should be kept up to date with dir.props -->
+  <package id="NuProj" version="0.10.4-beta-gf7fc34e7d8" />
```
weshaggard: Why here as opposed to in the project.json for the task library? Given that the task library is the one collecting and publishing these I would put it there.

dir.props
```diff
@@ -16,6 +16,8 @@
     <DnxPackageName Condition="'$(DnxPackageName)' == '' and '$(OsEnvironment)'=='Unix'">dnx-mono.$(DnxVersion)</DnxPackageName>
     <RoslynVersion>1.0.0-rc3-20150510-01</RoslynVersion>
     <RoslynPackageName>Microsoft.Net.ToolsetCompilers</RoslynPackageName>
+    <!-- This version should be kept up to date with packages.Windows_NT.config -->
+    <NuProjVersion>0.10.4-beta-gf7fc34e7d8</NuProjVersion>
```
weshaggard: I don't think there is much value in having this in dir.props. Again I would just keep this in the task project.

src/.nuget/packages.Windows_NT.config
FiveTimesTheFun: Thanks, addressed.

dir.props
FiveTimesTheFun: Thanks, addressed.

weshaggard:
LGTM

FiveTimesTheFun:
Thanks Wes, merging.

PR #308, opened 2015-10-10
Updated perf testing targets

ianhays:
Update to the latest xunit-performance package. Also added a target that will add the new OptimizeForBenchmarks assembly tag to any perf test build. This tag was added in https://github.com/Microsoft/xunit-performance/pull/94

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
@@ -9,7 +9,7 @@
   <!-- Perf Runner NuGet package paths -->
   <PropertyGroup>
     <XunitPerfRunnerPackageId>Microsoft.DotNet.xunit.performance.runner.Windows</XunitPerfRunnerPackageId>
-    <XunitPerfRunnerPackageVersion>1.0.0-alpha-build0022</XunitPerfRunnerPackageVersion>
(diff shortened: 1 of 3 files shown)


ianhays:
@mellinoe

src/Microsoft.DotNet.Build.Tasks/Targets/versioning.targets
```diff
@@ -61,6 +63,11 @@
       <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'" 
         Include="[assembly:System.Reflection.AssemblyMetadata(&quot;%(AssemblyMetadata.Identity)&quot;, &quot;%(AssemblyMetadata.Value)&quot;)]" />
     </ItemGroup>
+    
+    <ItemGroup Condition="'$(RunPerfTestsForProject)' == 'true' and '$(DisableTests)' != 'true'">
```
weshaggard: Let not add this in our general versioning.targets file. Please include this in the perftesting.targets, that is why we have this extensibility point so people can add to it outside this file.

ianhays: Sure thing, I can certainly move it to perftesting.Targets instead.

weshaggard:
I've renamed Targets=>PackageFiles yesterday so you will need to sync master to resolve your conflicts.

ianhays:
@weshaggard done and done

weshaggard:
LGTM. Note you will also need to update test-runtime\project.json in CoreFX repo as well.

PR #314, opened 2015-10-22
Update test-runtime with latest DNX

ericstj:


weshaggard:
Just as an FYI in case you didn't figure it out yet we have a local copy of this project.json and lock file in CoreFX directly.

ericstj:
But this one was being hit when we did the batch restore and it was triggering this issue: https://github.com/aspnet/dnx/issues/2771

PR #318, opened 2015-10-23
Updated perf targets

ianhays:
My old targets were buggy and didn't allow perf tests to be included when building the tests but not running them. I've modified them and renamed them to better accomodate this. 

I also removed the v5.0 TargetFramework declarations that are no longer required.

@mellinoe @weshaggard @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
@@ -2,8 +2,7 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'false' and '$(OS)' == 'Windows_NT' and Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" >true</RunPerfTestsForProject>
-    <AnalyzePerfResults Condition="'$(RunPerfTestsForProject)' == 'true' And '$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
+    <AnalyzePerfResults Condition="'$(IncludePerformanceTests)' == 'true' And '$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
   </PropertyGroup>
 
   <!-- Perf Runner NuGet package paths -->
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -2,8 +2,7 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'false' and '$(OS)' == 'Windows_NT' and Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" >true</RunPerfTestsForProject>
```
mellinoe: There doesn't seem to be a replacement for this line. Is that intended? As far as I can tell now, $(IncludePerformanceTests) won't ever get set to true unless you pass it into msbuild. Before, you would pass /p:Performance=true to trigger all of this stuff.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -22,13 +21,13 @@
...
   <!-- Perf testing assembly attribute -->
-  <ItemGroup Condition="'$(RunPerfTestsForProject)' == 'true' and '$(TestDisabled)' != 'true'">
+  <ItemGroup Condition="'$(IncludePerformanceTests)' == 'true'">
     <AssemblyInfoUsings Include="using Microsoft.Xunit.Performance%3B" />
     <AssemblyInfoLines Include="[assembly:OptimizeForBenchmarks]" />
   </ItemGroup>
 
   <!-- Copy over the performance runners to the test directory-->
-  <Target Name="CopyXunitExecutionDesktop" BeforeTargets="RunTestsForProject" Condition="'$(RunPerfTestsForProject)' == 'true'">
+  <Target Name="CopyXunitExecutionDesktop" AfterTargets="RestorePackages" Condition="'$(IncludePerformanceTests)' == 'true'">
```
mellinoe: This seems like a weird target to hook, but I guess it will get the job done. Isn't there some target like "CopyTestToTestDirectory" or something like that we could hook into? Also, does this target get executed if you have already restored the packages?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
ianhays: Yes it is intentional. It is now being set in CoreFX within the csprojs of the test libraries that contain performance tests. 

I forgot to link the PR that accompanies this one: Check out https://github.com/dotnet/corefx/pull/4115

mellinoe: Okay, I understand now. I'll echo what I said in the corefx PR, if you have to have a special project configuration for the perf tests, they could probably just go into a separate project file, IMO. I'm not really sure we need to be doing that at all, though, can't we just always include all of the source files in the project? Or am I missing some obvious issue with that?

ianhays:
closed per comments in https://github.com/dotnet/corefx/pull/4115

PR #319, opened 2015-10-23
Updated perf targets

ianhays:
My old targets were buggy and didn't allow perf tests to be included when building the tests but not running them. I've modified them and renamed them to better accomodate this.

I also removed the v5.0 TargetFramework declarations that are no longer required.

@mellinoe @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
@@ -48,6 +48,12 @@
     <AnalyzePerfResults Condition="'$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
     <TestCommandLine>$(PerfTestCommandLine)</TestCommandLine>
   </PropertyGroup>
+  
+    <!-- Optimizations to configure Xunit for performance -->
+  <ItemGroup Condition="'$(IncludePerformanceTests)' == 'true'">
+    <AssemblyInfoUsings Include="using Microsoft.Xunit.Performance%3B" />
+    <AssemblyInfoLines Include="[assembly:OptimizeForBenchmarks]" />
+  </ItemGroup>
 
   <Target Name="AnalyzePerfResults"
           AfterTargets="RunTestsForProject"
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
@@ -1,11 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <PropertyGroup>
-    <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'false' and '$(OS)' == 'Windows_NT' and Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" >true</RunPerfTestsForProject>
-    <AnalyzePerfResults Condition="'$(RunPerfTestsForProject)' == 'true' And '$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
-  </PropertyGroup>
-
   <!-- Perf Runner NuGet package paths -->
   <PropertyGroup>
     <XunitPerfRunnerPackageId>Microsoft.DotNet.xunit.performance.runner.Windows</XunitPerfRunnerPackageId>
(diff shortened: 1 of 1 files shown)


ianhays:
Goes with https://github.com/dotnet/corefx/pull/4118

mellinoe:
GitHub is having a hard time displaying the diff here, not exactly sure why. Could you try keeping the tabs at 2 spaces instead of 4 and seeing if it's easier to diff?

ianhays:
Huh, sorry about that. Not sure why it replaced my spaces. Fixed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -55,16 +44,16 @@
...
+    <AnalyzePerfResults Condition="'$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
     <TestCommandLine>$(PerfTestCommandLine)</TestCommandLine>
   </PropertyGroup>
   
-  <!-- Perf tests require TargetFrameworkVersion=v5.0. If it isn't installed, throw a helpful warning and run unit tests instead -->
-  <Target Name="ValidateFrameworkVersion" BeforeTargets="RestorePackages">
-    <Error Text="To run performance tests, .NET Portable v5.0 must be installed with VS 2015. Installation instructions available at: https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/performance-tests.md"
-      Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'true'" />
-  </Target>
+    <!-- Optimizations to configure Xunit for performance -->
```
mellinoe: This used to be further up in the file, could we keep it up there to minimize diff?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -1,11 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <PropertyGroup>
-    <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'false' and '$(OS)' == 'Windows_NT' and Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" >true</RunPerfTestsForProject>
-    <AnalyzePerfResults Condition="'$(RunPerfTestsForProject)' == 'true' And '$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
-  </PropertyGroup>
```
mellinoe: Why did you remove these from here? You redefine AnalyzePerfResults again down below on line 48, but I think it's effectively the same as it was up here (it gets set to true if it wasn't previously set to false, and if Performance is set to true). It's slightly different because the assignment of RunPerfTestsForProject checked a few other things than just Performance.

mellinoe:
I'm trying to understand the overall diff here, meaning the overall difference in behavior given different commands. You replaced $(RunPerfTestsForProject) with just $(Performance) everywhere, but the two were mostly equivalent before, modulo a few extra checks. I think the only real diff is that we copy the xunit assemblies over after "CopyTestToTestDirectory" rather than before "RunTestsForProject", and we condition the inclusion of the assembly attribute on something different. I think we could accomplish those two things with a lot less change. Perhaps we can leave everything else the same but just condition those two things on the new attribute added to the test projects themselves?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -55,16 +44,16 @@
...
-  <PropertyGroup Condition="'$(RunPerfTestsForProject)' == 'true'">
+  <PropertyGroup Condition="'$(Performance)' == 'true'">
+    <AnalyzePerfResults Condition="'$(AnalyzePerfResults)' != 'false'">true</AnalyzePerfResults>
     <TestCommandLine>$(PerfTestCommandLine)</TestCommandLine>
   </PropertyGroup>
   
-  <!-- Perf tests require TargetFrameworkVersion=v5.0. If it isn't installed, throw a helpful warning and run unit tests instead -->
-  <Target Name="ValidateFrameworkVersion" BeforeTargets="RestorePackages">
-    <Error Text="To run performance tests, .NET Portable v5.0 must be installed with VS 2015. Installation instructions available at: https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/performance-tests.md"
-      Condition="'$(Performance)' == 'true' and '$(RunPerfTestsForProject)' != 'true'" />
```
weshaggard: Thanks for cleaning this up. I assume you are also planning to remove the conditions in the project file where you set their TargetFramework to v5.0 when compiled with Performance?

ianhays: Yup! I removed all of those checks in https://github.com/dotnet/corefx/pull/3764/

ianhays:
Sorry @mellinoe , I didn't prioritize diff minimization <b>at all</b>.

I want to accomplish a few things:

- I wanted to get rid of the RunPerfTestsForProject variable; the only thing that RunPerfTestsForProject added on top of the Performance property was a check of the current OS and an outdated call to Exists that is no longer necessary. 
- Minimize repeated checks. AnalyzePerfResults was  moved closer to the target where it is used (in the middle of the file) where it could be combined with the TestCommandLine declaration that also depends on $(Performance). I'm completely fine with reverting this  to minimize diff as it's a purely a stylistic change.
- Modify the perf copy to occur after CopyTestToTestDirectory
- Modify the Assembly inclusions to depend on the IncludePerformanceTests property. This didn't need to be moved in the file; I can revert that.

mellinoe:
Sorry, I was just trying to understand the overall difference in behavior here the easiest, because I was thinking that the first two bullet points you listed were sort of a no-op in terms of behavior differences and were more for cleaning up the target (which is definitely a good thing).

> I wanted to get rid of the RunPerfTestsForProject variable; the only thing that RunPerfTestsForProject added on top of the Performance property was a check of the current OS and an outdated call to Exists that is no longer necessary.

That seems reasonable to me. There was also the difference in that you could explicitly set RunPerfTestsForProject to 'false' in order to avoid actually running the performance tests while still building them, so it's a bit less clear to me if this is the same as before. But it seems like this property wasn't really helpful or clear, or perhaps didn't even work correctly. I think it makes more sense now in that you set Performance=true if you want to build everything for performance tests, and then you build the "Test" target if you actually want to run them. Does that sound like an accurate description of how things will work?

>Minimize repeated checks. AnalyzePerfResults was moved closer to the target where it is used (in the middle of the file) where it could be combined with the TestCommandLine declaration that also depends on $(Performance). I'm completely fine with reverting this to minimize diff as it's a purely a stylistic change.

No, we should leave it, it makes more sense where it is now, especially if RunPerfTestsForProject is gone, because then it would be the only thing up there :smile:. I think when I made that note above I didn't realize that RunPerfTestsForProject was removed (which is fine, I don't think it was that useful).

Overall this LGTM.

ianhays:
Thanks, @mellinoe!

>Does that sound like an accurate description of how things will work?

Yup, exactly.

PR #320, opened 2015-10-26
Enable testing when running cross plat

ellismg:
Some minor changes needed to the build tools to enable the testing
targets to work cross platform.

 - We need to ensure corerun is executable before we run it.
 - corerun is not on the path and neither with the CWD, so we pass a
   full path to corerun when invoking it.
 - Use OSGroup to compute the platform specific attribute to pass to
   xunit to disable tests that are not valid for the OS in question.


ellismg:
/cc @weshaggard @naamunds 

I need these changes in buildtools so I can enable the testing targets when building CoreFX on Linux or OSX.

weshaggard:
LGTM

PR #321, opened 2015-10-27
Add post-build targets and support for cloud testing.

jhendrixMSFT:
Add Build.Post.targets to be wired up after product and test binaries have
been built (see https://github.com/dotnet/corefx/pull/4141 for context).

Add CloudTest.targets to be conditionally imported post-build.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
@@ -1,0 +1,235 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <UsingTask TaskName="CreateAzureContainer" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="CreateAzureFileShare" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="GetPerfTestAssemblies" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="SendToEventHub" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="UploadToAzure" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="WriteItemsToJson" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+
+  <PropertyGroup>
+    <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)</ContainerName>
+    <ContainerName>$(ContainerName.ToLower())</ContainerName>
+    <FuncTestListFilename>FuncTests.$(OSGroup).$(Platform)$(ConfigurationGroup).json</FuncTestListFilename>
+    <PerfTestListFilename>PerfTests.$(OSGroup).$(Platform)$(ConfigurationGroup).json</PerfTestListFilename>
+    <ArchivesRoot>$(TestWorkingDir)$(OSPlatformConfig)\archive\</ArchivesRoot>
+    <TestArchivesRoot>$(ArchivesRoot)Tests\</TestArchivesRoot>
+    <PackagesArchiveFilename>Packages.zip</PackagesArchiveFilename>
+    <PackagesArchiveFile>$(ArchivesRoot)$(PackagesArchiveFilename)</PackagesArchiveFile>
+    <FuncTestListFile>$(TestWorkingDir)$(OSPlatformConfig)\$(FuncTestListFilename)</FuncTestListFile>
+    <PerfTestListFile>$(TestWorkingDir)$(OSPlatformConfig)\$(PerfTestListFilename)</PerfTestListFile>
+  </PropertyGroup>
+
+  <!-- main entrypoint -->
+  <Target Name="CloudBuild"
+          AfterTargets="Build"
+          DependsOnTargets="VerifyInputs;CreateTestListJson;UploadContent" />
+
+  <!-- gather the test archives for upload -->
+  <ItemGroup>
+    <ForUpload Include="$(TestArchivesRoot)*.zip" />
+  </ItemGroup>
+
+  <Target Name="VerifyInputs">
+    <!-- verify all required properties have been specified -->
+    <Error Condition="'$(Creator)' == ''" Text="Missing required property Creator." />
+    <Error Condition="'$(TargetQueue)' == ''" Text="Missing required property TargetQueue." />
+    <Error Condition="'$(TestProduct)' == ''" Text="Missing required property TestProduct." />
+    <Error Condition="'$(BuildMoniker)' == ''" Text="Missing required property BuildMoniker." />
+    <Error Condition="'$(Branch)' == ''" Text="Missing required property Branch." />
+    <Error Condition="'$(CloudProductDropAccountName)' == ''" Text="Missing required property CloudProductDropAccountName." />
+    <Error Condition="'$(CloudOutputAccountName)' == ''" Text="Missing required property CloudOutputAccountName." />
+    <Error Condition="'$(DropAccessToken)' == ''" Text="Missing required property DropAccessToken." />
+    <Error Condition="'$(FileShareAccessToken)' == ''" Text="Missing required property FileShareAccessToken." />
+    <Error Condition="'$(BuildCompleteConnection)' == ''" Text="Missing required property BuildCompleteConnection." />
+    <Error Condition="'$(BuildIsOfficial)' == 'true' and '$(BuildIsOfficialConnection)' == ''" Text="Missing required property BuildIsOfficialConnection." />
+    <!-- verify the test archives were created -->
+    <Error Condition="'@(ForUpload->Count())' == '0'" Text="Didn't find any test archives in '$(ArchivesRoot)'." />
+    <!-- add relative blob path metadata -->
+    <ItemGroup>
+      <ForUpload>
+        <RelativeBlobPath>$(Platform)$(ConfigurationGroup)\Tests\%(Filename)%(Extension)</RelativeBlobPath>
+      </ForUpload>
+    </ItemGroup>
+  </Target>
+
+  <!-- create Azure containers and file shares -->
+  <Target Name="CreateAzureStorage">
+    <CreateAzureContainer
+      AccountKey="$(DropAccessToken)"
+      AccountName="$(CloudProductDropAccountName)"
+      ContainerName="$(ContainerName)"
+      ReadOnlyTokenDaysValid="30">
+        <Output TaskParameter="StorageUri" PropertyName="DropUriRoot" />
+        <Output TaskParameter="ReadOnlyToken" PropertyName="DropUriReadOnlyToken" />
+    </CreateAzureContainer>
+    <!-- append the build arch and type to the root URI -->
+    <CreateProperty Value="$(DropUriRoot)/$(BuildArch)$(BuildType)">
+      <Output TaskParameter="Value" PropertyName="DropUri" />
+    </CreateProperty>
+    <CreateAzureFileShare
+      AccountKey="$(FileShareAccessToken)"
+      AccountName="$(CloudOutputAccountName)"
+      ShareName="results"
+      ReadOnlyTokenDaysValid="30"
+      WriteOnlyTokenDaysValid="1">
+        <Output TaskParameter="StorageUri" PropertyName="ResultsUri" />
+        <Output TaskParameter="ReadOnlyToken" PropertyName="ResultsReadOnlyToken" />
+        <Output TaskParameter="WriteOnlyToken" PropertyName="ResultsWriteOnlyToken" />
+    </CreateAzureFileShare>
+  </Target>
+
+  <Target Name="CreateTestListJson"
+          DependsOnTargets="CreateFuncTestListJson;CreatePerfTestListJson" />
+
+  <Target Name="CreateFuncTestListJson" DependsOnTargets="CreateAzureStorage">
+    <!-- create item group of functional tests -->
+    <ItemGroup>
+      <FunctionalTest Include="$(TestArchivesRoot)*.zip" />
+    </ItemGroup>
+    <ItemGroup>
+      <FunctionalTest>
+        <Command>%HELIX_PYTHONPATH% %HELIX_SCRIPT_ROOT%\xunitrunner.py --dll %(Filename).dll</Command>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+    <GetPerfTestAssemblies TestBinaries="@(TestBinary)">
+      <Output TaskParameter="PerfTestAssemblies" ItemName="PerfTestAssembly" />
+    </GetPerfTestAssemblies>
+    <!-- don't add any items to the group if no perf tests were found -->
+    <ItemGroup Condition="'@(PerfTestAssembly->Count())' != '0'">
+      <PerfTest Include="$(TestArchivesRoot)%(PerfTestAssembly.Identity).zip" />
+    </ItemGroup>
+    <ItemGroup Condition="'@(PerfTestAssembly->Count())' != '0'">
+      <PerfTest>
+        <Command>%HELIX_PYTHONPATH% %HELIX_SCRIPT_ROOT%\xunitrunner.py --dll %(Filename).dll --perf-runner Microsoft.DotNet.xunit.performance.runner.Windows</Command>
```
weshaggard: What about these paths should they be named CLOUD? It appears there a new dependency on python. We should figure out how to add a check for that.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+  <UsingTask TaskName="UploadToAzure" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="WriteItemsToJson" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+
+  <PropertyGroup>
+    <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)</ContainerName>
+    <ContainerName>$(ContainerName.ToLower())</ContainerName>
+    <FuncTestListFilename>FuncTests.$(OSGroup).$(Platform)$(ConfigurationGroup).json</FuncTestListFilename>
+    <PerfTestListFilename>PerfTests.$(OSGroup).$(Platform)$(ConfigurationGroup).json</PerfTestListFilename>
+    <ArchivesRoot>$(TestWorkingDir)$(OSPlatformConfig)\archive\</ArchivesRoot>
```
weshaggard: Who is putting things into the archive folder?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+    <TestArchivesRoot>$(ArchivesRoot)Tests\</TestArchivesRoot>
+    <PackagesArchiveFilename>Packages.zip</PackagesArchiveFilename>
+    <PackagesArchiveFile>$(ArchivesRoot)$(PackagesArchiveFilename)</PackagesArchiveFile>
+    <FuncTestListFile>$(TestWorkingDir)$(OSPlatformConfig)\$(FuncTestListFilename)</FuncTestListFile>
+    <PerfTestListFile>$(TestWorkingDir)$(OSPlatformConfig)\$(PerfTestListFilename)</PerfTestListFile>
+  </PropertyGroup>
+
+  <!-- main entrypoint -->
+  <Target Name="HelixBuild"
```
weshaggard: CloudBuild?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: The test list JSON file is consumed by the controller service and these commands are sent verbatim to the test machines.  Changing the variable name would also require changes to the test clients.
The python dependency is only on the test client.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Will fix.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: This happens when tests are built if the ArchiveTests property is set to true.

weshaggard: OK yeah I see it at https://github.com/dotnet/buildtools/blob/a58b33a0cc964a96bf99cea408f3769eb42c5ea8/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets#L146

jhendrixMSFT:
CC @ChadNedzlek, @MattGal

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
ChadNedzlek: This feels like this same path is being calculated in multiple places... so anyone changing one of them is likely to break something unexpectedly.  Is there any way these can be calculated in a single place and referenced by both pieces?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+    <Error Condition="'$(TargetQueue)' == ''" Text="Missing required property TargetQueue." />
+    <Error Condition="'$(TestProduct)' == ''" Text="Missing required property TestProduct." />
+    <Error Condition="'$(BuildMoniker)' == ''" Text="Missing required property BuildMoniker." />
+    <Error Condition="'$(Branch)' == ''" Text="Missing required property Branch." />
+    <Error Condition="'$(AzureContainerAccountName)' == ''" Text="Missing required property AzureContainerAccountName." />
+    <Error Condition="'$(AzureFileShareAccountName)' == ''" Text="Missing required property AzureFileShareAccountName." />
+    <Error Condition="'$(DropAccessToken)' == ''" Text="Missing required property DropAccessToken." />
+    <Error Condition="'$(FileShareAccessToken)' == ''" Text="Missing required property FileShareAccessToken." />
+    <Error Condition="'$(BuildCompleteConnection)' == ''" Text="Missing required property BuildCompleteConnection." />
+    <Error Condition="'$(BuildIsOfficialConnection)' == ''" Text="Missing required property BuildIsOfficialConnection." />
```
ChadNedzlek: If the build is "not official", is the connection string really necessary?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+    <ForUpload Include="$(TestArchivesRoot)*.zip" />
+  </ItemGroup>
+
+  <Target Name="VerifyInputs">
+    <!-- verify all required properties have been specified -->
+    <Error Condition="'$(TargetQueue)' == ''" Text="Missing required property TargetQueue." />
+    <Error Condition="'$(TestProduct)' == ''" Text="Missing required property TestProduct." />
+    <Error Condition="'$(BuildMoniker)' == ''" Text="Missing required property BuildMoniker." />
+    <Error Condition="'$(Branch)' == ''" Text="Missing required property Branch." />
+    <Error Condition="'$(AzureContainerAccountName)' == ''" Text="Missing required property AzureContainerAccountName." />
```
ChadNedzlek: The relevant part of this value isn't necessary that is a container, it's really the "Product drop account name".  I feel like we'll want the flexibility to choose either blobs or files (or something else later), and it would be nice not to have to change this variable name when that happens.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Good point will update the condition to include '$(BuildIsOfficial)' == 'true'.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+    <Error Condition="'$(AzureFileShareAccountName)' == ''" Text="Missing required property AzureFileShareAccountName." />
```
ChadNedzlek: Same deal here.  This bit is the "CloudOutpuAccountName" really.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+  </Target>
+
+  <Target Name="CreateTestListJson"
+          DependsOnTargets="CreateFuncTestListJson;CreatePerfTestListJson" />
+
+  <Target Name="CreateFuncTestListJson" DependsOnTargets="CreateAzureStorage">
+    <!-- create item group of functional tests -->
+    <ItemGroup>
+      <FunctionalTest Include="$(TestArchivesRoot)*.zip" />
+    </ItemGroup>
```
ChadNedzlek: Can't these two itemGroups be one? Where you calculate the include and add the metadata at the same time?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+    <ItemGroup>
+      <FunctionalTest>
+        <Command>%HELIX_PYTHONPATH% %HELIX_SCRIPT_ROOT%\xunitrunner.py --dll %(Filename).dll</Command>
+        <CorrelationPayloadUris>[$(DropUri)$(Platform)$(ConfigurationGroup)/Packages.zip$(DropUriReadOnlyToken);;http://dotnetbuildscripts.blob.core.windows.net/scripts/xunit_210.zip;;http://dotnetbuildscripts.blob.core.windows.net/scripts/amd64ret/corerun.zip]</CorrelationPayloadUris>
+        <PayloadUri>$(DropUri)$(Platform)$(ConfigurationGroup)/Tests/%(Filename)%(Extension)$(DropUriReadOnlyToken)</PayloadUri>
+        <WorkItemId>FunctionalTest.%(Filename)</WorkItemId>
+      </FunctionalTest>
+    </ItemGroup> 
+    <WriteItemsToJson JsonFileName="$(FuncTestListFile)" Items="@(FunctionalTest)" />
```
ChadNedzlek: This isn't how the ToF builds do it, as it become a real nightmare to maintain this way. (For example, this doesn't work for linux or mac, because of the use of % and \.  We had a template system in ToF that tried to reduce all the duplication of data here... can we not port that?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+      </TestListFile>
+    </ItemGroup>
+  </Target>
+
+  <!-- compress the packages dir in preparation for uploading -->
+  <Target Name="CompressPackagesDir" Condition="'$(SkipArchive)' != 'true'">
+    <ItemGroup>
+      <ExcludeFromArchive Include="nupkg$" />
+      <ExcludeFromArchive Include="Microsoft.DotNet.BuildTools" />
+      <ExcludeFromArchive Include="TestData" />
```
ChadNedzlek: Isn't the TestData actually necessary for the tests that use it? I know it contains a bunch of certs for the crypto tests, for example.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
```diff
@@ -0,0 +1,234 @@
...
+    <ItemGroup>
+      <BuildComplete Include="%(TestListFile.BuildCompleteJson)">
+        <CorrelationId>%(TestListFile.CorrelationId)</CorrelationId>
+        <DropContainerSAS>$(DropUriReadOnlyToken)</DropContainerSAS>
+        <ListUri>$(DropUri)%(TestListFile.Filename)%(TestListFile.Extension)$(DropUriReadOnlyToken)</ListUri>
+        <QueueId>$(TargetQueue)</QueueId>
+        <ResultsUri>$(ResultsUri)/%(TestListFile.CorrelationId)</ResultsUri>
+        <ResultsUriRSAS>$(ResultsReadOnlyToken)</ResultsUriRSAS>
+        <ResultsUriWSAS>$(ResultsWriteOnlyToken)</ResultsUriWSAS>
+        <Creator>$(USERNAME)</Creator>
```
ChadNedzlek: This data, in the open, might be... misleading, or personal information that someone might not want to divulge to our service (since it's going to give us the username they use to log into whatever system they are logged into when this task runs)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Yes, unfortunately including this in the archive was causing the archive step to crash.  It will need to be fixed, however I don't want to block our daily builds on a small subset of tests.

ChadNedzlek: zipping up the test data folder caused it to crash??? That's... alarming. I wonder if there are other files someone might check in that will cause a similar crash
We should probably open an issue to investigate and address that if we aren't going to fix this right now.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: It would need to go into a common props file (e.g. CloudTest.props) and get included by both targets.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Will rename accordingly.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Will rename accordingly.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Unfortunately it doesn't work, at least I haven't found a way to make it work.  The metadata for the included items (e.g. %(Filename)) are empty.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: I have opened a task in VSO to track fixing this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
ChadNedzlek: Comment withdrawn.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Good point, I'll make this a required property.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CloudTest.targets
jhendrixMSFT: Agreed this is not optimal.  I will look at porting the template-based system during the Linux/Mac bring-up.

ChadNedzlek:
LGTM

PR #323, opened 2015-10-29
Extracting Cloud Test changes into a new task library

joperezr:
Extracting CLoud Test tasks into a new task library called Microsoft.DotNet.Build.CloudTestTasks.dll (I'm open to name changes if anybody has any suggestions.)

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -3,7 +3,7 @@
 
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="RemoveDuplicatesWithLastOneWinsPolicy" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
-  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileCreateFromDirectory" Condition="'$(ArchiveTests)' == 'true'" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   
   <PropertyGroup>
     <SerializeProjects Condition="'$(TestWithLocalLibraries)'=='true'">true</SerializeProjects>
--- a/src/nuget/Microsoft.DotNet.BuildTools.nuspec
+++ b/src/nuget/Microsoft.DotNet.BuildTools.nuspec
@@ -19,13 +19,13 @@
     </dependencies>
   </metadata>
   <files>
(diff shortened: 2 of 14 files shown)


joperezr:
cc @weshaggard @jhendrixMSFT

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -3,7 +3,7 @@
 
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="RemoveDuplicatesWithLastOneWinsPolicy" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
-  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.CloudTestTasks.dll"/>
```
jhendrixMSFT: Is this going to cause problems when doing non-cloud test builds on .NET Core?  This task is only used by target ArchiveTestBuild when the ArchiveTests property is set to true, should this UsingTask also be predicated on the same condition?

joperezr: That is a good point, I'll add the condition to the UsingTask. Also, I think it makes sense to move the CloudTest.targets into this new project, since it is not relevant anymore on the other package.

weshaggard: This also seems like a more general task. If it is .NET Core compliant please just keep it in the other Tasks library. If it isn't compliant then add some conditions around this and the Archive task in some way to make sure we don't accidently try loading it.

jhendrixMSFT: Good point, I believe this task is .NET Core compliant so it should remain in the tasks dll.

joperezr: Sounds good, I'm working on it now and will update the PR once I'm done.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -19,13 +19,15 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="Microsoft.DotNet.Build.Tasks\package\**\*.*" target="lib" />
+    <file src="Microsoft.DotNet.Build.CloudTestTasks\Microsoft.ServiceBus.dll" target="lib" />
```
weshaggard: Should we consider putting these assemblies in a subfolder under lib given they are netfx specific and might conflict with the other .NET Core versions in the root?

weshaggard:
Other than the one consideration this LGTM.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
joperezr: done, I moved everything that has to do with CloudTestTasks.dll into lib\net45

PR #325, opened 2015-10-29
Fix for perf commandline

ianhays:
Fix for the change made in https://github.com/dotnet/buildtools/pull/320

@mellinoe @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/test-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/test-runtime/project.json
@@ -8,8 +8,8 @@
       "OpenCover": "4.6.166",
       "ReportGenerator": "2.3.1",
 
-      "Microsoft.DotNet.xunit.performance.analysis": "1.0.0-alpha-build0023",
-      "Microsoft.DotNet.xunit.performance.runner.Windows": "1.0.0-alpha-build0023",
+      "Microsoft.DotNet.xunit.performance.analysis": "1.0.0-alpha-build0025",
+      "Microsoft.DotNet.xunit.performance.runner.Windows": "1.0.0-alpha-build0025",
 
       "xunit": "2.1.0",
       "xunit.console.netcore": "1.0.2-prerelease-00101",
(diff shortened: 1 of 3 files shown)


ellismg:
@ianhays Does my change in #326 mean you don't have to do e8ef648dc48e8c297de77770f4e66a8e059c3489 anymore?

ianhays:
Yes, I can leave that out now. Thanks.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/test-runtime/project.json
```diff
@@ -8,8 +8,8 @@
       "OpenCover": "4.6.166",
       "ReportGenerator": "2.3.1",
 
-      "Microsoft.DotNet.xunit.performance.analysis": "1.0.0-alpha-build0023",
-      "Microsoft.DotNet.xunit.performance.runner.Windows": "1.0.0-alpha-build0023",
+      "Microsoft.DotNet.xunit.performance.analysis": "1.0.0-alpha-build0025",
```
weshaggard: I assume you will also update this in the test-runtime project in corefx as well? This coupling between the target files and the test-runtime\project.json seems like it is asking for some trouble. At a minimum I think it is worth allowing  folks to override your PackageVersion properties in the targets file. That way folks that have their own test-runtime have a choice of either overriding the versions used or updating their test-runtime project.json.

ianhays: Yes, the CoreFX PR with the buildtools update will update those also. 

>allowing folks to override your PackageVersion properties in the targets file

What would this look like? Would it be as simple as adding a property for the version (e.g. 1.0.0-alpha-build0023) instead of hard coding it? The relationship between the project.json files for buildtools/test-runtime, corefx/common/test-runtime, and the project itself is still a bit blurry to me.

ianhays: >What would this look like? Would it be as simple as adding a property for the version (e.g. 1.0.0-alpha-build0023) instead of hard coding it? The relationship between the project.json files for buildtools/test-runtime, corefx/common/test-runtime, and the project itself is still a bit blurry to me.

I wasn't thinking this through all the way. I see what you mean now. I've added a check around the version property to allow overriding.

ianhays:
Any other feedback before I squash and merge?

PR #327, opened 2015-10-30
Filter out non-Windows tests for cloud test runs.

jhendrixMSFT:
Current test runs are running all tests on Windows which is creating false
failures, i.e. the non-Windows tests are failing.  This change filteres
them out of test runs.


jhendrixMSFT:
CC @ChadNedzlek

PR #328, opened 2015-11-04
Update SendToEventHub with PartitionKey property any use it.

alexperovich:
This change adds an optional PartitionKey property to the SendToEventHub build task and then consumes it to send the official build event to the appropriate partition.

This fixes a caching issue where test runs could be duplicated.


alexperovich:
@jhendrixMSFT

jhendrixMSFT:
LGTM

PR #329, opened 2015-11-05
Fix running test projects in VS

ellismg:
$(TestProgram) is now a full path, so we don't need to prepend it with
the working directory.


ellismg:
/cc @nguerrera @stephentoub 

Sorry for breaking this @nguerrera.

nguerrera:
:+1: Sorry for the emoji.

PR #330, opened 2015-11-05
Fix build races with BUILDTOOLS_OVERRIDE_RUNTIME

ellismg:
When BUILDTOOLS_OVERRIDE_RUNTIME is set, if we are hard linking the shared
runtime in the test folder, then trying to copy an updated coreclr (or
other files used by the runtime) will cause races due to a file being in
use.

Update the logic to fold the over-ridden files into the general ItemGroup
we filter and then copy as needed.


stephentoub:
LGTM

PR #331, opened 2015-11-06
Upgrade to ReportGenerator 2.3.4

stephentoub:
cc: @ellismg


ellismg:
Looks good.  Folks that are consuming this and have a custom test-runtime project.json file (like corefx and wcf) will need to make the corresponding updates in their local copies.

PR #332, opened 2015-11-07
disable perf as unit tests temporarily

ianhays:
https://github.com/dotnet/corefx/pull/4118 made it so that perf tests ran as unit tests by default. Although this is the desired behavior, the InnerIterations structure that they use is causing an unacceptable amount of slowdown to the other CoreFX unit tests.

I need to get a more permanent solution but in the meantime this commit is a quick and easy reversion that disables the perf tests by default. Running with Performance=true still works as expected.

@stephentoub @mellinoe


stephentoub:
LGTM. Thanks.

ianhays:
Thanks Steve!

PR #334, opened 2015-11-13
Port a fix from xunit.console.exe to xunit.console.netcore.exe

stephentoub:
From:
https://github.com/xunit/xunit/commit/9749ad1bd871342ba3f4ca936a0904c01673bd42

Will need this to fix https://github.com/dotnet/corefx/issues/2370

(Editorial: We should really try to get off our own custom runner.)

cc: @mellinoe, @weshaggard


weshaggard:
Port looks good to me. As for moving to our own runner I'm happy to do so we just need someone to drive the changes back into xunit itself, so if anyone is interested in doing that it would be most welcome.

PR #339, opened 2015-11-19
Added modifications for AssemblyList generated during Cloud Tests

karajas:
Added modifications for AssemblyList generated during Cloud Tests to add pattern support. This is to help differentiate Runtime dependencies and xunit dependencies during test execution.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -5,6 +5,8 @@
   <UsingTask TaskName="RemoveDuplicatesWithLastOneWinsPolicy" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="ZipFileCreateFromDirectory" Condition="'$(ArchiveTests)' == 'true'" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
   
+  <UsingTask TaskName="GenerateAssemblyList" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  
   <PropertyGroup>
     <SerializeProjects Condition="'$(TestWithLocalLibraries)'=='true'">true</SerializeProjects>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
```diff
@@ -25,6 +25,7 @@
   <ItemGroup>
     <Compile Include="CreateAzureContainer.cs" />
     <Compile Include="CreateAzureFileShare.cs" />
+    <Compile Include="GenerateAssemblyList.cs" />
```
joperezr: Can this task run in .Net Core? if yes, then perhaps we want to move it back to Microsoft.DotNet.Build.Tasks.csproj. This project is more intended to hold only the tasks that are not able to be converted yet due to external dependencies that are not in .NetCore yet.

src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
joperezr: I'm not sure why it thinks that you modified the line, but just make sure you didn't change the encoding on the file and undo this line's change

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -134,10 +134,14 @@
       <AssemblyPaths Include="$([MSBuild]::MakeRelative($(PackagesDir), %(_TestCopyLocalByFileNameWithoutDuplicates.SourcePath)))" Condition="'%(_TestCopyLocalByFileNameWithoutDuplicates.NugetPackageId)' != ''" />
     </ItemGroup>
     <WriteLinesToFile
-      File="$(OutDir)AssemblyList.txt"
+      File="$(OutDir)AssemblyList.flat.txt"
       Lines="@(AssemblyPaths)"
       Overwrite="true"
       Encoding="Ascii" />
+    <GenerateHelixAssemblyList
```
joperezr: Aren't you missing a <UsingTask> to import it from the dll?

joperezr:
If you want to test your changes in the corefx repo, you can go to [follow this Wiki](https://github.com/dotnet/buildtools/wiki/Local-testing-of-BuildTools-in-another-repo) to restore your version of buildtools into a different repo.

joperezr:
Once the encoding issue on the CloudTest task library is fixed, this LGTM

ericstj:
I'm OK with this for now since it is need to unblock the tests, but we really need to switch the test deployment to consume `project.lock.json` we really shouldn't be using our own format.

I understand the TPA requirement and think that should be handled with the deployment code on the client with the parameters it passes to corerun and/or doing the directory split there: just for mscorlib.  We shouldn't be putting all test dependencies in the TPA list, I think that's just an artifact of the "the way it used to be done" and not an appropriate simulation of the real world app environment.

ericstj:
LGTM

PR #340, opened 2015-11-23
Remove the workarounds in partialfacades.targets

mellinoe:
These workarounds were in place because, in order to reference mscorlib in our partial facades, we were referencing a Microsoft.DotNet.CoreCLR NuGet package, which had a variety of extra assemblies that got automatically pulled in as references. Also, the mscorlib we were referencing was architecture-specific, which brought in a compiler warning that we suppressed. None of this is necessary any longer now that we have published proper reference assembly packages for mscorlib and the rest.

@weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
@@ -3,41 +3,6 @@
 
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
-  <!-- Default properties for partial facade projects -->
-  <PropertyGroup Condition="'$(IsPartialFacadeAssembly)' == 'true'">
-    <PostFilterNugetReferences>true</PostFilterNugetReferences>
-    <IgnoreArchitectureMismatches>true</IgnoreArchitectureMismatches>
-  </PropertyGroup>
-
-  <!-- It seems this needs to be set here, and cannot be set in individual project files -->
-  <PropertyGroup Condition="'$(IgnoreArchitectureMismatches)' == 'true'">
-    <!-- Needed because we are using mscorlib.dll from the CoreCLR package, rather than a true reference assembly -->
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
```diff
@@ -9,35 +9,6 @@
     <IgnoreArchitectureMismatches>true</IgnoreArchitectureMismatches>
```
weshaggard: Looks like these properties can be cleaned out as well.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
```diff
@@ -9,35 +9,6 @@
...
-    <!-- Needed because we are using mscorlib.dll from the CoreCLR package, rather than a true reference assembly -->
-    <ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
-  </PropertyGroup>
-
-  <!-- PostFilterNuGetReferences
-       Filters the set of references generated from the ResolveNugetPackages target. This is a default target for partial facade
-         assemblies, and can be optionally used in other projects as well, by setting PostFilterNugetReferences=true. The target
-         contains some logic for removing well-known assemblies from the "Microsoft.DotNet.CoreCLR" package, except for "mscorlib.dll".
-  -->
-  <Target Name="PostFilterNugetReferences"
```
weshaggard: While this isn't an issue for CoreCLR when we enabling this for .NET Native and Desktop I do think I want to figure out a good way to filter to only the references that are explicitly listed in the partial facade project. I don't want people to blindly have access to all the types in the full framework for example.

mellinoe: Ah, I see what you are saying. Microsoft.TargetingPack.Private.NETNative has ~20 private assemblies in it. We'll have to think of something to do about that if we don't want all of them exposed in a project. Splitting it into separate NuGet packages is the obvious option, but that's a lot of NuGet packages then.

weshaggard: I don't think we should split these packages but instead do some sort of filtering on them, but we can do that as a separate work item.

weshaggard:
LGTM

Issue #337, opened 2015-11-17
PrereleaseResolveNuGetPackageAssets throws a null pointer exception on OSX when Log.LogErrorFromResources is called


cdmihai:
Under OSX the following line: 

https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs#L169

Throws this exception:

```
Target "ResolveNuGetPackageAssets: (TargetId:49)" in file "/Users/cdmihai/Dropbox/projects/msbuild/override.targets" from project "/Users/cdmihai/Dropbox/projects/msbuild/src/Framework/Microsoft.Build.Framework.csproj" (target "ResolveAssemblyReferences" depends on it):
Set Property: _ExplicitReference=
Set Property: _OmitTransitiveReferences=false
Using "PrereleaseResolveNuGetPackageAssets" task from assembly "/Users/cdmihai/Dropbox/projects/msbuild/packages/Microsoft.DotNet.BuildTools.1.0.25-prerelease-00120/lib/Microsoft.DotNet.Build.Tasks.dll".
Task "PrereleaseResolveNuGetPackageAssets" (TaskId:28)
  Task Parameter:AllowFallbackOnTargetSelection=False (TaskId:28)
  Task Parameter:IncludeFrameworkReferences=True (TaskId:28)
  Task Parameter:ProjectLanguage=C# (TaskId:28)
  Task Parameter:ProjectLockFile=project.lock.json (TaskId:28)
  Task Parameter:TargetMonikers=DNXCore,Version=v5.0 (TaskId:28)
  Task Parameter:RuntimeIdentifier=osx.10.10-x64 (TaskId:28)
  Task Parameter:OmitTransitiveCompileReferences=False (TaskId:28)
  Resolved references from Microsoft.CSharp: (TaskId:28)
  Resolved references from Microsoft.NETCore: (TaskId:28)
  Resolved references from Microsoft.NETCore.Platforms: (TaskId:28)
  Resolved references from Microsoft.NETCore.Portable.Compatibility: (TaskId:28)
  Resolved references from Microsoft.NETCore.Runtime: (TaskId:28)
  Resolved references from Microsoft.NETCore.Runtime.CoreCLR: (TaskId:28)
  Resolved references from Microsoft.NETCore.Runtime.Native: (TaskId:28)
  Resolved references from Microsoft.NETCore.Targets: (TaskId:28)
[...]

PR #341, opened 2015-11-25
Throw exception if ProjectLanguage is null

cdmihai:
After I bumped the build tools version up I started getting NPEs from the resolve assets tasks. Apparently the ProjectLanguage task parameter is now mandatory.

In order to save investigation time I added a null check and an error message.

An alternative implementation would be to mark the task parameter as required.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
@@ -401,6 +401,12 @@ namespace Microsoft.DotNet.Build.Tasks
 
         private bool IsFileValid(string file, string expectedLanguage, string unExpectedLanguage)
         {
+
+            if(ProjectLanguage == null)
+            {
+                throw new ExceptionFromResource("NoProgrammingLanguageSpecified");
+            }
+
             var expectedProjectLanguage = expectedLanguage;
             expectedLanguage = expectedLanguage == "C#" ? "cs" : expectedLanguage;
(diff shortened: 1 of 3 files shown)


ericstj:
Maybe rather than this we should just return either true or false of ProjectLanguage is null.  @jasonmalinowski what's the right behavior here?

jasonmalinowski:
I treat throwing exceptions to ultimately return false back from the build task. I have an ExceptionFromResource type that if you throw that will just print the error message and will skip the stack trace. So if you added the string and then threw one of those, that'd be the "correct fix".

This is fine though. :-)

jasonmalinowski:
Also, if a bug hasn't been filed on NuGet/Home, we probably should do that too.

ericstj:
So you want an unspecified ProjectLanguage to fail the build rather than just not returning language specific analyzers?

jasonmalinowski:
Probably. Language-specific analyzers are the common case for analyzers, really, since lots of analyzers want/need to do something with syntax. Just silently dropping them would probably be too "surprising".

ProjectLanguage is also used now for the immutable content stuff too, so again not having it set and silently dropping assets would be bad.

ericstj:
Ok, @cdmihai can you move the string into Strings.resx and use ExceptionFromResource like all the other exceptions in this file?

cdmihai:
Sure, no problem!

cdmihai:
Made the update. I remember now why I did not first use ExceptionFromResource. On OSX catching and re-throwing that exception caused a NPE: #337.

ericstj:
LGTM

PR #342, opened 2015-11-26
Bringing back Full Framework version of the task library to be able to build in razzle and linux

joperezr:
Bringing back Full Framework version of the task library to be able to build in razzle and linux.

When we moved this Task library to .Net Core we verified that it would be able to be loaded also when using full framework. What we didn't validate, is that it requires full framework 4.6 which means that it won't work in linux(mono msbuild doesn't support 4.6) nor Razzle builds(razzle runs with full framework 4.5). Because of this, after upgrading the buildtools version broke corefx repo in both environments. This change will add back the 4.5 version of the task library and store it under the 'net45' folder inside the package, so that razzle and linux builds can load that instead of the .Net Core version.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
@@ -6,7 +6,7 @@
   <UsingTask TaskName="SendToEventHub" AssemblyFile="$(ToolsDir)net45\Microsoft.DotNet.Build.CloudTestTasks.dll"/>
   <UsingTask TaskName="UploadToAzure" AssemblyFile="$(ToolsDir)net45\Microsoft.DotNet.Build.CloudTestTasks.dll"/>
   <UsingTask TaskName="WriteItemsToJson" AssemblyFile="$(ToolsDir)net45\Microsoft.DotNet.Build.CloudTestTasks.dll"/>
-  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
     <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)</ContainerName>
--- a/src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
@@ -1,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProductVersion>8.0.30703</ProductVersion>
+    <SchemaVersion>2.0</SchemaVersion>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Microsoft.DotNet.Build.Tasks</RootNamespace>
+    <AssemblyName>Microsoft.DotNet.Build.Tasks</AssemblyName>
+    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
+    <CopyNuGetImplementations>true</CopyNuGetImplementations>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <SkipSigning>true</SkipSigning>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="..\Microsoft.DotNet.Build.Tasks\Delegates.cs" />
+    <Compile Include="..\Microsoft.DotNet.Build.Tasks\ExecWithMutex.cs" />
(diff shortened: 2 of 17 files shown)


joperezr:
cc @weshaggard

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -6,7 +6,7 @@
   <UsingTask TaskName="SendToEventHub" AssemblyFile="$(ToolsDir)net45\Microsoft.DotNet.Build.CloudTestTasks.dll"/>
   <UsingTask TaskName="UploadToAzure" AssemblyFile="$(ToolsDir)net45\Microsoft.DotNet.Build.CloudTestTasks.dll"/>
   <UsingTask TaskName="WriteItemsToJson" AssemblyFile="$(ToolsDir)net45\Microsoft.DotNet.Build.CloudTestTasks.dll"/>
-  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
```
weshaggard: In one of the targets files in BuildTools you should default BuildToolsTaskDir to ToolsDir so that not everyone using build tools is forced to set this independently in their repo.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
```diff
@@ -0,0 +1,118 @@
...
+    <SchemaVersion>2.0</SchemaVersion>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Microsoft.DotNet.Build.Tasks</RootNamespace>
+    <AssemblyName>Microsoft.DotNet.Build.Tasks</AssemblyName>
+    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <ResolveNuGetPackages>false</ResolveNuGetPackages>
```
weshaggard: If possible I would still use the project.json even for the desktop version as opposed to the old packages.config.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
```diff
@@ -0,0 +1,118 @@
...
+    <SkipSigning>true</SkipSigning>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="..\Microsoft.DotNet.Build.Tasks\Delegates.cs">
+	
+	</Compile>
```
weshaggard: Any ideas why there is an end tag and why it is formatted like this? You should be able to simply close it off with "/>".

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -19,7 +19,16 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="Microsoft.DotNet.Build.CloudTestTasks\*.dll" target="lib\net45" />
+    <file src="Microsoft.DotNet.Build.Tasks.net45\*.dll" target="lib\net45" />
+    <file src="Microsoft.DotNet.Build.CloudTestTasks\Microsoft.Azure.KeyVault.Core.dll" target="lib\net45" />
```
weshaggard: Why did you need to expand this instead of using *.dll now? Is it because there are file conflicts?

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
joperezr: No idea why it got imported like this, good catch, I'll fix it

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
joperezr: Fixed.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
joperezr: There are three dependencies that exist in both the cloudTestTasks and the net45 task library, they are Newtonsoft.Json, System.Reflection.Metadata and System.Collections.Immutable. This causes no issues for now as they both projects have the same version of all three assemblies. When this stops being true, we can start dividing each project into a different folder within net45.

ericstj:
Can you make a more verbose commit message to explain the problem?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
```diff
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <PropertyGroup>
```
weshaggard: I think this needs to be in a more core location like the Build.Common.targets itself.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
ericstj: Lets avoid this duplication.  Instead can we import the other project so that we don't duplicate?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
joperezr: Will move.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
ericstj: If you want you can check in with the duplication and I will come back and clean it up.

joperezr: I merged this PR to unblock stakeholders and I will send out a new PR shortly with this refactoring. Thanks Eric!

PR #344, opened 2015-12-01
Add a msbuild property to disable functional tests for cloud testing.

jhendrixMSFT:
There is currently no way to submit just the performance tests to the
cloud test infrastructure.  This new property, FuncTestsDisabled, will
disable the functional tests when perf tests are enabled.
Stop excluding test data from packages.zip as it causes test failures.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
@@ -83,7 +83,10 @@
   <Target Name="CreateTestListJson"
           DependsOnTargets="CreateFuncTestListJson;CreatePerfTestListJson" />
 
-  <Target Name="CreateFuncTestListJson" DependsOnTargets="CreateAzureStorage">
+  <!-- allow skipping this target if perf testing is enabled -->
+  <Target Name="CreateFuncTestListJson"
+          DependsOnTargets="CreateAzureStorage"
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -83,7 +83,8 @@
   <Target Name="CreateTestListJson"
           DependsOnTargets="CreateFuncTestListJson;CreatePerfTestListJson" />
 
-  <Target Name="CreateFuncTestListJson" DependsOnTargets="CreateAzureStorage">
+  <!-- allow skipping this target if perf testing is enabled -->
+  <Target Name="CreateFuncTestListJson" DependsOnTargets="CreateAzureStorage" Condition="'$(Performance)' != 'true' or '$(FuncTestsDisabled)' != 'true'">
```
ChadNedzlek: This line is a bit long.

I'd also prefer it wasn't a negative property. Instead of "disablefoo" that defaults to false, I'd rather have it be "enablefoo" that defaults to true... "DisableFoo=false" to not not do a thing always hurts my brain, and I frequently get it backward.

jhendrixMSFT: I will move the properties to new lines to shorten the overall length.

Regarding negative property, as the behavior is to be opt-out it feels overly verbose to conditionally set it to true, plus it is also the pattern used in other parts of build tools (e.g. disabling tests post-build follows the same pattern).

PR #345, opened 2015-12-01
Add project.json tooling for an alternative to checking in lockfiles

dagood:
This adds a new target `ValidatePackageVersions` and two tasks `UpdatePrereleaseDependencies` and `UpdatePackageDependencyVersion`. This diff shows what I needed to do to use them in corefx: https://github.com/dotnet/corefx/compare/master...dagood:dagood/no-lockfile-tooling

(Note: this adds a NuGet source of the nugetbuild MyGet repository to `dir.props` for `NuGet.Versioning`, but it was already in `src/NuGet.config` anyway.)

---

`ValidatePackageVersions` runs after `ResolveNuGetPackages` and can check all project.json files for `*` (floating) dependencies and/or can make sure all packages match a given prerelease version. What it does depends on the properties/item collections when the target is run. By default, the task runs but no validation happens when no properties are passed.

This is what I added to corefx to configure validation: https://github.com/dotnet/corefx/compare/master...dagood:dagood/no-lockfile-tooling#diff-0b192804a6349e8c26d2b027afbd89a2R60

This adds some confidence that `project.json` dependencies don't get out of sync and enforces the new update workflow.

---

`UpdatePrereleaseDependencies` is a task that rewrites given project.json files with specified changes. After adding a task to `build.proj` in corefx, example usage:

    msbuild /t:UpdatePrereleaseDependencies /p:OldPrerelease=beta-45678;NewPrerelease=beta-45679

Updates *all packages* of version `*-beta-45678` to `*-beta-45679`.

---

`UpdatePackageDependencyVersion` rewrites given project.json files to replace exact matches of PackageId of OldVersion to NewVersion. Using a task added to `build.proj` in corefx:

    msbuild /t:UpdatePackageDependencyVersion /p:PackageId=xunit;OldVersion=2.0.0;NewVersion=2.1.0

Updates all dependencies on `xunit 2.0.0` to `xunit 2.1.0` in all project.json files.

---

The purpose of the update tasks is to have a way to update all package versions across the repository that's an easy alternative to `*` dependencies, even when upgrades need to happen often. (Where previously we would use `*` dependencies and relock.) Without `*` dependencies we no longer need the lockfile in source control to have a repeatable build.

/cc @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.net45/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/project.json
@@ -6,6 +6,7 @@
 	"NuGet.Core": "2.8.3",
 	"NuGet.NuManifest": "1.0.1-alpha",
 	"NuGet.NuManifest.DotNet": "1.0.1-alpha",
+    "NuGet.Versioning": "3.3.0",
 	"System.Collections.Immutable": "1.1.32-beta",
 	"System.Reflection.Metadata": "1.0.17-beta",
 	"xunit.runners": "2.0.0-beta5-build2785"
--- a/src/Microsoft.DotNet.Build.Tasks.net45/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/project.json
@@ -1,15 +1,15 @@
 {
   "dependencies": {
     "LibGit2Sharp": "0.19.0.0",
-	"Microsoft.Web.Xdt": "2.1.1",
-	"Newtonsoft.Json": "6.0.8",
-	"NuGet.Core": "2.8.3",
-	"NuGet.NuManifest": "1.0.1-alpha",
-	"NuGet.NuManifest.DotNet": "1.0.1-alpha",
+    "Microsoft.Web.Xdt": "2.1.1",
+    "Newtonsoft.Json": "6.0.8",
+    "NuGet.Core": "2.8.3",
+    "NuGet.NuManifest": "1.0.1-alpha",
+    "NuGet.NuManifest.DotNet": "1.0.1-alpha",
     "NuGet.Versioning": "3.3.0",
-	"System.Collections.Immutable": "1.1.32-beta",
-	"System.Reflection.Metadata": "1.0.17-beta",
-	"xunit.runners": "2.0.0-beta5-build2785"
+    "System.Collections.Immutable": "1.1.32-beta",
+    "System.Reflection.Metadata": "1.0.17-beta",
+    "xunit.runners": "2.0.0-beta5-build2785"
   },
   "frameworks": {
     "net45": { }
--- a/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
@@ -32,6 +32,8 @@
     <Compile Include="ParseTestCoverageInfo.cs" />
     <Compile Include="RemoveDuplicatesWithLastOneWinsPolicy.cs" />
     <Compile Include="PrereleaseResolveNuGetPackageAssets.cs" />
+    <Compile Include="UpdateProjectDependencies.cs" />
+    <Compile Include="ValidateProjectDependencyVersions.cs" />
     <Compile Include="ZipFileCreateFromDirectory.cs" />
     <Compile Include="Strings.Designer.cs">
       <AutoGen>True</AutoGen>
--- a/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
@@ -32,6 +32,8 @@
     <Compile Include="ParseTestCoverageInfo.cs" />
     <Compile Include="RemoveDuplicatesWithLastOneWinsPolicy.cs" />
     <Compile Include="PrereleaseResolveNuGetPackageAssets.cs" />
+    <Compile Include="UpdatePackageDependencyVersion.cs" />
+    <Compile Include="UpdatePrereleaseDependencies.cs" />
     <Compile Include="UpdateProjectDependencies.cs" />
     <Compile Include="ValidateProjectDependencyVersions.cs" />
     <Compile Include="ZipFileCreateFromDirectory.cs" />
--- a/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
@@ -34,7 +34,7 @@
     <Compile Include="PrereleaseResolveNuGetPackageAssets.cs" />
     <Compile Include="UpdatePackageDependencyVersion.cs" />
     <Compile Include="UpdatePrereleaseDependencies.cs" />
-    <Compile Include="UpdateProjectDependencies.cs" />
+    <Compile Include="VisitProjectDependencies.cs" />
     <Compile Include="ValidateProjectDependencyVersions.cs" />
     <Compile Include="ZipFileCreateFromDirectory.cs" />
     <Compile Include="Strings.Designer.cs">
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
@@ -101,7 +101,7 @@
 
   <Target Name="ValidatePackageVersions">
     <ValidateProjectDependencyVersions Condition="Exists('$(ProjectJson)')"
-                                       ProjectJson="$(ProjectJson)"
+                                       ProjectJsons="$(ProjectJson)"
                                        ProhibitFloatingDependencies="$(ProhibitFloatingDependencies)"
                                        ValidationPatterns="@(ValidationPattern)" />
   </Target>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
@@ -103,9 +103,8 @@
   <Target Name="ValidatePackageVersions">
     <ValidateProjectDependencyVersions Condition="Exists('$(ProjectJson)')"
                                        ProjectJson="$(ProjectJson)"
-                                       ValidPrereleaseVersion="$(CoreFXDependencyPrereleaseVersion)"
-                                       FloatingDependenciesInvalid="$(FloatingDependenciesInvalid)"
-                                       OurPackageIdentities="$(OurPackageIdentityRegex)" />
+                                       ProhibitFloatingDependencies="$(ProhibitFloatingDependencies)"
+                                       ValidationPatterns="@(ValidationPattern)" />
   </Target>
 
   <Target Name="UpdateDependencies">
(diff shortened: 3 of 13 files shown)


src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
```diff
@@ -32,6 +32,8 @@
     <Compile Include="ParseTestCoverageInfo.cs" />
     <Compile Include="RemoveDuplicatesWithLastOneWinsPolicy.cs" />
     <Compile Include="PrereleaseResolveNuGetPackageAssets.cs" />
+    <Compile Include="UpdateProjectDependencies.cs" />
+    <Compile Include="ValidateProjectDependencyVersions.cs" />
```
ericstj: /cc @joperezr this will need to be merged with https://github.com/dotnet/buildtools/pull/342

src/Microsoft.DotNet.Build.Tasks/UpdateProjectDependencies.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        public string PackageId { get; set; }
+
+        public string OldVersion { get; set; }
+
+        public string NewVersion { get; set; }
+
+        public override bool Execute()
+        {
+            if (new[] { OldPrerelease, NewPrerelease }.All(s => !string.IsNullOrEmpty(s)))
```
ericstj: Lets do some basic parameter validation.  If NewPrerelease is set and OldPrerelease is not, log an error.
```
if (!String.IsNullOrWhiteSpace(NewPrerelease))
{
  if (!String.IsNullOrWhiteSpace(OldPrerelease))
  {
    ReplacePrerelease();
  }
  else
  {
    // log error
  }
}
```

Do something similar for version mode.

If neither are specified, error.  If both are specified, error, since it isn't clear if PackageId should apply to the pre-release.

src/Microsoft.DotNet.Build.Tasks/UpdateProjectDependencies.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        public JObject ReadProject()
+        {
+            using (TextReader projectFileReader = File.OpenText(ProjectJson.ItemSpec))
+            {
+                var projectJsonReader = new JsonTextReader(projectFileReader);
+
+                var serializer = new JsonSerializer();
+                return serializer.Deserialize<JObject>(projectJsonReader);
```
ericstj: You can replace with JObject.Load(projectJsonReader);

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class ValidateProjectDependencyVersions : Task
+    {
+        [Required]
+        public ITaskItem ProjectJson { get; set; }
+
+        public string ValidPrereleaseVersion { get; set; }
+
+        public bool FloatingDependenciesInvalid { get; set; }
```
ericstj: Can you rename to ProhibitFloatingDependencies?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -2,6 +2,8 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ValidateProjectDependencyVersions" AssemblyFile="$(ToolsDir)Microsoft.DotNet.Build.Tasks.dll"/>
```
weshaggard: You will need to update this to BuildToolsTaskDir after @joperezr's https://github.com/dotnet/buildtools/pull/342 change.

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public string OurPackageIdentities { get; set; }
```
dagood: Good idea, I'll change it to that.

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
dagood: Yep! I was struggling to find a good name for it.

src/Microsoft.DotNet.Build.Tasks/UpdateProjectDependencies.cs
```diff
@@ -0,0 +1,140 @@
...
+                    }
+                }
+            }
+            else
+            {
+                Log.LogError(
+                    "Expected properties from one of (OldPrerelease, NewPrerelease) or (PackageId, OldVersion, NewVersion) but found {0}.",
+                    (anyPrereleaseUpdateParams && anyVersionUpdateParams) ? "properties from both" : "none");
+            }
+            return !Log.HasLoggedErrors;
```
dagood: @ericstj Is the above parameter checking at all reasonable? Straightforward `if`s works but gets kinda big with 3 parameters (vs. reasonable for 2).

ericstj: Sure.  This is readable though a little excessive on allocations (arrays & enumerators).  That shouldn't matter for this usage so it's OK.

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
```diff
@@ -0,0 +1,127 @@
...
+                }
+                else if (!string.IsNullOrWhiteSpace(_expectedPrerelease))
+                {
+                    throw new ArgumentException(string.Format(
+                        "Both ExpectedVersion and ExpectedPrerelease metadata found on item {0}, but only one permitted",
+                        item.ItemSpec));
+                }
+            }
+
+            public string FindValidationError(string packageId, string version)
```
ericstj: Consider calling this validate and just passing in the Logger.

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
```diff
@@ -0,0 +1,127 @@
...
+
+                    if (ProhibitFloatingDependencies && version.Contains('*'))
+                    {
+                        Log.LogError("Floating dependency detected: {0}", versionMessage);
+                    }
+                    else
+                    {
+                        string validationError = patterns
+                            .Select(pattern => pattern.FindValidationError(id, version))
+                            .FirstOrDefault(error => error != null);
```
ericstj: You should iterate through all of these and log all of them.  Only logging the first one is a recipe for "wack-a-mole" workflow.

ericstj:
LGTM

src/Microsoft.DotNet.Build.Tasks/UpdateProjectDependencies.cs
dagood: I've centralized the array, but still more enumerations than necessary. Would be simple to use two foreaches and four bools for better perf but I think it's more readable like this.

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
dagood: Did this, keeps it simple, thanks.

src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
dagood: Added these to the net45 project, added the NuGet.Versioning dependency to its project.json, and fixed up other conflicts. It still builds and works for me (although I haven't tested the net45 version).

src/Microsoft.DotNet.Build.Tasks.net45/project.json
```diff
@@ -6,6 +6,7 @@
 	"NuGet.Core": "2.8.3",
 	"NuGet.NuManifest": "1.0.1-alpha",
 	"NuGet.NuManifest.DotNet": "1.0.1-alpha",
+    "NuGet.Versioning": "3.3.0",
```
weshaggard: nit: indention

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -97,8 +100,25 @@
...
   </Target>
 
+  <Target Name="ValidatePackageVersions">
+    <ValidateProjectDependencyVersions Condition="Exists('$(ProjectJson)')"
+                                       ProjectJson="$(ProjectJson)"
+                                       ProhibitFloatingDependencies="$(ProhibitFloatingDependencies)"
+                                       ValidationPatterns="@(ValidationPattern)" />
+  </Target>
+
+  <Target Name="UpdateDependencies">
```
weshaggard: I would prefer we simply have one update all target in corefx that calls this target with a list of ProjectJson files. Similar to https://github.com/dotnet/corefx/blob/master/build.proj#L26.

src/Microsoft.DotNet.Build.Tasks/UpdateProjectDependencies.cs
```diff
@@ -0,0 +1,146 @@
...
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class UpdateProjectDependencies : Task
+    {
+        [Required]
+        public ITaskItem ProjectJson { get; set; }
```
weshaggard: Make this accept a list.

src/Microsoft.DotNet.Build.Tasks/UpdateProjectDependencies.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        public string OldPrerelease { get; set; }
```
weshaggard: these other properties should probably be Required as well.

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            using (TextReader projectFileReader = File.OpenText(ProjectJson.ItemSpec))
+            {
+                var projectJsonReader = new JsonTextReader(projectFileReader);
+
+                var serializer = new JsonSerializer();
+                JObject projectRoot = serializer.Deserialize<JObject>(projectJsonReader);
+
+                foreach (var package in projectRoot["dependencies"].OfType<JProperty>())
```
weshaggard: What about dependencies that are under a particular framework?

src/Microsoft.DotNet.Build.Tasks.net45/project.json
dagood: Some of the file was using tabs, changed to consistent spaces.

src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
dagood: Fixed to look at all descendants to find properties with the name "dependencies" and look through those.

src/Microsoft.DotNet.Build.Tasks/UpdateProjectDependencies.cs
dagood: I'm going to split update into two tasks/targets--it really is two separate commands anyway. That'll let me get rid of the boolean xor, too.

dagood:
@ericstj @weshaggard I made a bunch of changes and edited the PR description to match.

One thing I'm wondering is if `UpdatePrereleaseDependencies` needs some way to filter package names. If our prerelease label happens to overlap with one of our dependencies' someone running this could accidentally upgrade both. A simple global find/replace would be a fix if that happens though. Worth looking at? Could reuse the validation patterns, maybe even make the update target change the expected version automatically... I have more thoughts on this but don't know if it's worth going after.

src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
```
joperezr: Is this an intentional change? I think what happened here is that the encoding on the csproj was changed, lets just make sure that it is the right one

dagood: Unintentional, thanks for catching this. Looking at the change in ASCII shows this:

    -???<?xml version="1.0" encoding="utf-8"?>

After reverting it, it comes back after adding a file in Visual Studio and saving. (But not just when building or loading.) I'm not too familiar with what tools expect which formats... do you have a suggestion?

ericstj: If VS changes the encoding let's keep the encoding that VS wants.  Otherwise we'll always be reverting this line when we change the file.

dagood: Change is now intentional. :+1:

weshaggard:
LGTM

PR #346, opened 2015-12-01
Add support for filtering resolved packages from targeting packages

weshaggard:
This add support for filtering out resolved assets from the
targeting pack packages. It does this based on the following property
TargetingPackNugetPackageId which is set to the id of the package
to filter. It will remove any packages references from that package
except for the ones explicitly listed in the TargetingPackageReference
item group.

cc @mellinoe @ericstj


mellinoe:
Looks good to me. By the way, I believe we are going to need another change in order to build .NET Native projects with buildtools, because of an issue with some common AssemblyInfo include file. See this commit:

https://github.com/mellinoe/buildtools/commit/1be7623bc658d0502454c5a21ff71c8875840183

If the project has resource strings, then we automatically include a source file (AssemblyInfoPartial.cs) which references a type in System.Resources.ResourceManager (which is not in any core assembly in .NET Native).

weshaggard:
Thanks @mellinoe please go ahead and submit a PR for that. I'm curious why I didn't hit that when doing a partial facade for .NET Native. I do think we also need a way to disable the auto-reference of the mscorlib design-time facade as well.

PR #347, opened 2015-12-02
Add initial support for reporting build statistic.

jhendrixMSFT:
These changes are currently scoped to the cloud-based test infrastructure.

Generate a JSON file containing test build statistics and upload it to
DocumentDB for later processing.  At present, the only statistic captured
is the number of successfully built tests.

Stop adding xunit and corerun zip packages to the list of downloads as
they are no longer required (they are pulled from packages.zip).

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/SendJsonToDocumentDb.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/SendJsonToDocumentDb.cs
@@ -1,0 +1,160 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using Microsoft.Azure.Documents;
+using Microsoft.Azure.Documents.Client;
+using Microsoft.Azure.Documents.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.DotNet.Build.Common.Desktop;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    /// <summary>
+    /// Used to send JSON blobs to DocumentDb.
+    /// </summary>
+    public class SendJsonToDocumentDb : Task
+    {
+        /// <summary>
+        /// The account key used to connect to DocumentDB.
+        /// </summary>
+        [Required]
+        public string AccountKey { get; set; }
+
+        /// <summary>
+        /// The name of the document collection.
+        /// </summary>
+        [Required]
+        public string Collection { get; set; }
+
+        /// <summary>
+        /// The name of the document database.
+        /// </summary>
+        [Required]
+        public string Database { get; set; }
+
+        /// <summary>
+        /// The ID of the document to be created.
+        /// </summary>
+        [Required]
+        public string DocumentId { get; set; }
+
+        /// <summary>
+        /// The DocumentDB endpont URI.
+        /// </summary>
+        [Required]
+        public string EndpointUri { get; set; }
+
+        /// <summary>
+        /// The JSON file to be uploaded.
+        /// </summary>
+        [Required]
+        public string JsonFile { get; set; }
+
+        static SendJsonToDocumentDb()
+        {
+            AssemblyResolver.Enable();
+        }
+
+        public override bool Execute()
+        {
+            // load the JSON file
+            JObject json;
+
+            using (StreamReader streamReader = new StreamReader(JsonFile))
+            using (JsonTextReader jsonReader = new JsonTextReader(streamReader))
+            {
+                var jsonSerializer = new JsonSerializer();
+                json = (JObject)jsonSerializer.Deserialize(jsonReader);
+
+                // add the document ID to the JSON blob
+                json.Add("id", DocumentId);
+            }
+
+            using (var con = CreateDocumentDbConnection())
+            {
+                // for specified document ID check if there is a document for it, if there
+                // isn't then create a new one with the JSON blob that was loaded earlier.
+
+                var result = con.Client.CreateDocumentQuery(string.Format("dbs/{0}/colls/{1}", con.Database.Id, con.Collection.Id))
+                    .Where(doc => doc.Id == DocumentId)
+                    .AsEnumerable()
+                    .FirstOrDefault();
+
+                if (result == null)
+                {
+                    Log.LogMessage(MessageImportance.Normal, "Creating new document with ID '{0}'", DocumentId);
+                    con.Client.CreateDocumentAsync(con.Collection.SelfLink, json).Wait();
+                }
+                else
+                {
+                    Log.LogWarning("A document with ID '{0}' already exists, no content was uploaded.", DocumentId);
+                }
+            }
+
+            return true;
+        }
+
+        private class DocumentConnection : IDisposable
+        {
+            public DocumentConnection(Database db, DocumentClient client, DocumentCollection collection)
+            {
+                Database = db;
+                Client = client;
+                Collection = collection;
+            }
+
+            public Database Database { get; }
+
+            public DocumentClient Client { get; }
+
+            public DocumentCollection Collection { get; }
+
+            public void Dispose()
+            {
+                Client.Dispose();
+            }
+        }
+
+        private DocumentConnection CreateDocumentDbConnection()
+        {
+            var client = new DocumentClient(new Uri(EndpointUri), AccountKey);
+
+            // get the database and if it doesn't exist create it
+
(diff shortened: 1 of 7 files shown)


jhendrixMSFT:
CC @ChadNedzlek, @karajas, @alexperovich

src/Microsoft.DotNet.Build.CloudTestTasks/AssemblyResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.IO;
+using System.Reflection;
+using System.Threading;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    /// <summary>
+    /// Used to enable custom assembly resolution.
+    /// </summary>
+    class AssemblyResolver
```
ChadNedzlek: I think we are supposed to explicitly write down default access modifiers.

src/Microsoft.DotNet.Build.CloudTestTasks/SendJsonToDocumentDb.cs
```diff
@@ -0,0 +1,167 @@
...
+
+                var result = con.Client.CreateDocumentQuery(string.Format("dbs/{0}/colls/{1}", con.Database.Id, con.Collection.Id))
+                    .Where(doc => doc.Id == DocumentId)
+                    .AsEnumerable()
+                    .FirstOrDefault();
+
+                if (result == null)
+                {
+                    Log.LogMessage("Creating new document with ID '{0}'", DocumentId);
+                    using (var task = con.Client.CreateDocumentAsync(con.Collection.SelfLink, json))
```
ChadNedzlek: Why are we using the Task? I've never seen that pattern before (usually just task.Wait())

src/Microsoft.DotNet.Build.CloudTestTasks/SendJsonToDocumentDb.cs
```diff
@@ -0,0 +1,167 @@
...
+                .AsEnumerable()
+                .FirstOrDefault();
+
+            if (database == null)
+            {
+                Log.LogMessage(MessageImportance.Low, "The database {0} does not exist, will create it.", Database);
+                using (var task = client.CreateDatabaseAsync(new Database { Id = Database }))
+                {
+                    task.Wait();
+                    database = task.Result;
```
ChadNedzlek: Result is already blocking, so calling Wait() before it doesn't actually do anything, to my knowledge.

src/Microsoft.DotNet.Build.CloudTestTasks/AssemblyResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            // unify up to a local file if it exists.
+            string fileName = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), referenceName.Name + ".dll");
+            if (File.Exists(fileName))
+            {
+                var localName = AssemblyName.GetAssemblyName(fileName);
+
+                // if the local file has a version greater than what was
+                // requested assume it's backward compatible and load it
+                if (localName.Version > referenceName.Version)
```
ChadNedzlek: This is a really broad statement to make... "All assemblies are always backward compatible"... And I don't think that's actually true (if it were, that would presumably just be what .NET does).

What problem is this trying to solve?

src/Microsoft.DotNet.Build.CloudTestTasks/WriteBuildStatsJson.cs
```diff
@@ -0,0 +1,117 @@
...
+                public FailedProject(string name)
+                {
+                    m_name = name;
+                }
+
+                public string ProjectName { get { return m_name; } }
+            }
+
+            private List<string> m_aggregates;
+            private List<FailedProject> m_failedProjects;
```
ChadNedzlek: This violates the dotnet coding style guidelines (which I think apply to this project).

src/Microsoft.DotNet.Build.CloudTestTasks/WriteBuildStatsJson.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                foreach (var buildFailure in buildFailures)
+                    m_failedProjects.Add(new FailedProject(buildFailure.ItemSpec));
+
+                m_logUri = logUri;
+                m_builtCount = builtCount;
+                m_testCount = testCount;
+            }
+
+            public List<string> Aggregates { get { return m_aggregates; } }
```
ChadNedzlek: I'm not sure what compiler this project uses, but you could just have get only properties that you assign in the constructor... should make the file a bit shorter and easier to fit on a screen. :-)

src/Microsoft.DotNet.Build.CloudTestTasks/AssemblyResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+    class AssemblyResolver
+    {
+        static int s_enabled;
+
+        /// <summary>
+        /// Call to enable the assembly resolver for the current AppDomain.
+        /// </summary>
+        public static void Enable()
+        {
+            if (Interlocked.CompareExchange(ref s_enabled, 1, 0) == 0)
```
ericstj: Avoid the flag entirely and just put the code in the static constructor for AssemblyResolver.  Have an Enable method that does nothing.  The act of calling it will trigger the static constructor which is guaranteed to be thread safe and run once.

src/Microsoft.DotNet.Build.CloudTestTasks/AssemblyResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+                AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
+        }
+
+        private static Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEventArgs args)
+        {
+            // apply any existing policy
+            AssemblyName referenceName = new AssemblyName(AppDomain.CurrentDomain.ApplyPolicy(args.Name));
+
+            // unify up to a local file if it exists.
+            string fileName = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), referenceName.Name + ".dll");
```
ericstj: Consider using args.RequestingAssembly.

src/Microsoft.DotNet.Build.CloudTestTasks/AssemblyResolver.cs
ericstj: That is indeed what CoreCLR and the .NET loader in store apps does.  We cannot make .NET desktop do it for compat reasons.  As a result we have bindingRedirects on classic .NET but MSBuild doesn't expose bindingRedirects to task DLLs, so we have this workaround that mimics the CoreCLR behavior.

src/Microsoft.DotNet.Build.CloudTestTasks/SendJsonToDocumentDb.cs
jhendrixMSFT: Since it implements IDisposable I figured it needed to be disposed.  I did a little more digging and found http://blogs.msdn.com/b/pfxteam/archive/2012/03/25/10287435.aspx, apparently it's not needed in most cases.

src/Microsoft.DotNet.Build.CloudTestTasks/WriteBuildStatsJson.cs
jhendrixMSFT: What is in violation here?

jhendrixMSFT: Oh the "m_".  These have been removed.

src/Microsoft.DotNet.Build.CloudTestTasks/AssemblyResolver.cs
ericstj: Never mind, looks like that can be null.

src/Microsoft.DotNet.Build.CloudTestTasks/AssemblyResolver.cs
ChadNedzlek: Is there a specific assembly that's causing this? Mostly curious "why now"?

ericstj: Newtonsoft.Json.  Folks hit this problem a lot.  Usually not for msbuild tasks since they tend to be small and you can control the entire closure.  As such MSBuild has no convention for specifying the bindingRedirects.  In this case Joel has two dependencies he doesn't own which use different versions of Newtonsoft.Json.

ChadNedzlek: Gotcha. Makes perfect sense.

PR #348, opened 2015-12-03
Allow multiple dependency updates to one project.json

dagood:
When I used `.Any(...)` to check if dependencies needed to change after being visited, only the first change happens because `.Any` short-circuits. Changed to `.Select(...).ToArray().Any()` so that all visits are always evaluated.

/cc @ericstj @weshaggard


weshaggard:
LGTM

ericstj:
LGTM

PR #349, opened 2015-12-03
Add AssemblyResolver to load conflicting assemblies without bindingRedirects

ericstj:
MSBuild does not provide a convention for task dll's to specify bindingRedirects.
Sometimes a task may need to use two different libraries that use use different versions
of a common library (EG: Newtonsoft.JSON).  NuGet ensures that the higher version, is
used by the task, but one dependency will still have static reference to the lower version.

Without binding redirects this will result in a failure to load the assembly with the
lower version.  Address this by hooking AssemblyResolve and probing for the
assembly by simple name and choosing the first assembly that satisfies the request,
permitting higher versions.

### DIFF
--- a/src/common/AssemblyResolver.cs
+++ b/src/common/AssemblyResolver.cs
@@ -1,0 +1,108 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.DotNet.Build.Common.Desktop
+{
+    /// <summary>
+    /// Used to enable app-local assembly unification.
+    /// </summary>
+    internal static class AssemblyResolver
+    {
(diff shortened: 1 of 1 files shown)


src/common/AssemblyResolver.cs
```diff
@@ -0,0 +1,108 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.DotNet.Build.Tasks.Packaging
```
jhendrixMSFT: I think the namespace should change, perhaps Microsoft.DotNet.Build.Common.Desktop?  I include "Desktop" in the namespace as presumably this class is only necessary when running on desktop CLR?

ericstj: Good call.  I'll make that change.

jhendrixMSFT:
LGTM

PR #352, opened 2015-12-03
Import BuildTools project rather than duplicating

ericstj:
Add an `UpdateImportedProjectRelativePaths` to remap items that came from an imported project.
Have Microsoft.DotNet.Build.Tasks.net45.csproj import Microsoft.DotNet.Build.Tasks.csproj 
rather than duplicating the project.


weshaggard:
LGTM

PR #358, opened 2015-12-04
Add tool-runtime and test-runtime project.json validation

dagood:
Add two new targets that run automatically during the build to validate `tool-runtime/project.json` and `test-runtime/project.json` automatically.

These new validation steps (and the earlier project.json validation) are individually optional. This should be useful because it makes sense to keep the `test-runtime/project.json` override in the corefx repo in step, but with `tool-runtime/project.json` in the buildtools package it doesn't seem worth it to keep that in step.

/cc @ericstj @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
@@ -99,9 +99,9 @@
              Condition="'@(_ReferenceFileNamesToRemove)' != '' or '@(_ReferenceCopyLocalPathsFileNamesToRemove)' != ''"/>
   </Target>
 
-  <Target Name="ValidatePackageVersions">
-    <ValidateProjectDependencyVersions Condition="Exists('$(ProjectJson)')"
-                                       ProjectJsons="$(ProjectJson)"
+  <Target Name="ValidatePackageVersions"
+          Condition="'$(RestorePackages)'=='true' and '$(ValidatePackageVersions)'=='true' and Exists('$(ProjectJson)')">
+    <ValidateProjectDependencyVersions ProjectJsons="$(ProjectJson)"
                                        ProhibitFloatingDependencies="$(ProhibitFloatingDependencies)"
                                        ValidationPatterns="@(ValidationPattern)" />
   </Target>


src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -16,6 +17,14 @@
           Condition="'$(RestorePackages)'=='true' AND '$(IsTestProject)' == 'true' AND '$(TestRuntimeProjectJson)' != ''">
     <Exec Command="$(DnuRestoreCommand) &quot;$(TestRuntimeProjectJson)&quot;" StandardOutputImportance="Low" CustomErrorRegularExpression="^Unable to locate .*" />
   </Target>
+  
+  <Target Name="ValidateTestRuntimeDependencies"
+          AfterTargets="RestoreTestRuntimePackage"
```
dagood: @ericstj @weshaggard When I enable this and build from the corefx repo root, the same set of validation errors are reported for each project (if there are any errors). Is there a better way to hook this target in, or is that behavior ok?

ericstj: I suppose it could be run up front at the same time we're restoring when building from the root.  Similar to how we disable per-project restore when building from the root.

weshaggard: Unlike the ToolRuntime it doesn't look like we have a semaphore to ensure we don't run the RestoreTestRuntimePackage more than once. We should consider fixing that and when we do we should also do the ValidateTestRuntimeDependencies at the same time. I'm fine with this being here but just want to be sure that it doesn't run for every test project.

dagood: Thanks! It looks like it might be straightforward to add that semaphore. Looking deeper into it, I'm seeing that a run from the root *sometimes* skips `RestoreTestRuntimePackage` because `RestorePackages` is false, but for some test projects it runs anyway:

    System.Globalization\tests\System.Globalization.Tests.csproj
    System.Composition.Convention\tests\System.Composition.Convention.Tests.csproj
    System.Composition\tests\System.Composition.Tests.csproj
    System.Diagnostics.Tools\tests\System.Diagnostics.Tools.Tests.csproj
    System.IO\tests\System.IO.Tests.csproj
    System.Runtime\tests\System.Runtime.Tests.csproj
    System.Text.Encodings.Web\tests\System.Text.Encodings.Web.Tests.csproj
    System.Runtime.InteropServices\tests\System.Runtime.InteropServices.Tests.csproj
    System.Runtime.Extensions\tests\System.Runtime.Extensions.Tests.csproj

I'm trying to find a reason for this. I noticed though that for both toolsruntime and testruntime, `BatchRestorePackages` restores the project.json and then later it's restored again, so there might be other strangeness going on. Should I look further into this or just set up a semaphore for now?

dagood: @ericstj @weshaggard I've been trying out a semaphore and I think it would be a good fix to make to eliminate some unnecessary restores but as far as I can tell it can't help with making validation only happen once. The main problem is because after writing a semaphore the task won't happen again unless the inputs change: if validation fails before touching the semaphore, the other projects don't know that the file's already failed validation and will also try to validate it. If the semaphore is touched before validating, it will report one error, but subsequent (root or project) builds will skip validation entirely and report success. While figuring this out I've added a semaphore for `RestoreTestRuntimePackage`, so I'll leave it in the PR, but I think I'll need to do the same thing as restore with root and project validation being distinct.

weshaggard: Instead of trying to go down the semaphore file route I would prefer we simply call the validate for these two directly as part of the BatchRestorePackages target that only ever gets run once. I think that is the easiest way forward without adding addition logic or runtime to every test project build.

dagood: The way I've got it going now without semaphores (working in my Razzle build and a git corefx repo) is that a BuildAllProjects build will run a `ValidateAllProjectDependencies` target (similar to `BatchRestorePackages`), and the code in buildtools is skipped. The validation targets in buildtools only run when `RestorePackages=true`, which in corefx's case is during an individual project build.

I don't know if leaving it all to a `BatchRestorePackages`-like step is great, since someone working on and building only one project would miss test-runtime and tool-runtime validation until they (or CI) do an all-project build.

On the other hand, we could validate all project.jsons in the entire repository on every build. Validation is pretty fast, so maybe that simplicity vs. speed tradeoff is good. What do you think?

(The `TestRuntimeSemaphore` in this PR was just a fix to avoid restoring more than necessary, and doesn't affect the validation logic. I'll split off that change to keep things pure.)

weshaggard:
LGTM

PR #359, opened 2015-12-07
Make xUnit test args configurable.

jhendrixMSFT:
Remove hard-coded xUnit test args and replace with a property so the value
can be specified at build time.


jhendrixMSFT:
CC @ChadNedzlek

PR #360, opened 2015-12-09
Updating Dev Workflow so that the ToolRuntime is restored using dotnet cli. Creating props file with the Common properties.

joperezr:
As part of the new Dev Workflow, the bootstrapping of buildtools and the ToolRuntime will be handled by dotnet cli. This are the required changes for this in the buildtools side, and there will be a PR comming on the corefx side that is consuming this new changes.


joperezr:
cc @weshaggard

joperezr:
It's worth noting that when this gets merged, a change in corefx to adapt to this new workflow is required in order to upgrade buildtools since the way to restore ToolRuntime is now changed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -0,0 +1,230 @@
...
+  if the code is the same between some OS's (so if you have some project that just calls POSIX APIs, we still have
+  OSX_[Debug|Release] and Linux_[Debug|Release] configurations.  We do this so that we place all the output under
+  a single binary folder and can have a similar experience between the command line and Visual Studio.
+  
+  Since now have multiple *Debug and *Release configurations, ConfigurationGroup is set to Debug for any of the
+  debug configurations, and to Release for any of the release configurations.
+  -->
+
+  <!-- Set default Configuration and Platform -->
+  <PropertyGroup>
```
weshaggard: This file is out of sync with the latest changes in dir.props in corefx. Please sync it up with the current properties in corefx.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -0,0 +1,33 @@
+@echo off
+setlocal
+
+set PROJECT_DIR=%~1
+set TOOLS_DIR=%~2
+set DOTNET_CMD=%~3
+
+echo PROJECT_DIR=[%PROJECT_DIR%]
```
weshaggard: I assume these are just for debugging, do we want to keep them always echoing?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -0,0 +1,33 @@
...
+  echo ERROR: Cannot find tools path at [%TOOLS_DIR%]. Please pass in the tools directory as the 2nd parameter.
+  exit /b 1
+)
+
+if not exist "%DOTNET_CMD%" (
+  echo ERROR: Cannot find dotnet command at [%DOTNET_CMD%]. Please pass in the dotnet command as the 3rd parameter.
+  exit /b 1
+)
+
+set TOOLRUNTIME_DIR=%PROJECT_DIR%\bin\obj\Windows_NT.AnyCPU.Debug\ToolRuntime
```
weshaggard: We are likely going to need to change this location because if someone does a clean we don't want to be try and delete the tools we are running on as that will fail badly.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -0,0 +1,31 @@
...
+   exit 1
+fi
+
+if [ ! -e "$__DOTNET_CMD" ]; then
+   echo "ERROR: Cannot find dotnet.exe at path '$__DOTNET_CMD'. Please pass in the path to dotnet.exe as the 3rd parameter."
+   exit 1
+fi
+
+# Temporary hack to make dnu restore more reliable.
+export MONO_THREADS_PER_CPU=2000
```
weshaggard: Do we need this? Is dotnet running on MONO?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packages.Unix.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="dnx-mono" version="1.0.0-rc2-16128" />
```
weshaggard: I assume we don't need these packages.*.config files any longer?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
```diff
@@ -24,9 +24,12 @@
...
   
-  <Target Name="EnsureBuildToolsRuntime"
-      Inputs="$(BuildToolsSemaphore)"
-      Outputs="$(ToolRuntimeSempahore)">
+  <Target Name="EnsureBuildToolsRuntime">
+    <Error Condition="!Exists('$(ToolRuntimeSempahore)')"
+           Text="The Tool Runtime directory [$(ToolRuntimePath)] does not exist. Please run $(SourceDir)Build.cmd to ensure the tools are installed before attempting to build an individual project." />
+  </Target>
+
+  <Target Name="RestoreToolRuntime">
```
weshaggard: I assume we can delete this target now?

weshaggard:
Can you point to a branch with the corresponding corefx changes?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
joperezr: I thought that I had synced it but I'll go over it again with the latest one in corefx

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: right, forgot to remove them.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: That makes sense, I had not thought of that. so do you want the folder to be in {repo_root}/ToolRuntime ?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: I had this in place for when we used to restore everything using dnx as it made the restore more reliable, dotnet cli is using dnx under the covers, but I'm not sure if it is running on MONO or not, @brthor do you know if we still need this to make restore more reliable?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packages.Unix.config
joperezr: It actually depends, if we want to restore dnx on the corefx side or on the buildtools side. The way I have it setup right now is that we do it in corefx side so yes, this files have no use for now, but I could change it so that corefx only restores buildtools, and buildtools restores dnx.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
joperezr: Yes, It looks like dotnet-publish works as expected, so I will go ahead and remove this unused target.

joperezr:
I'm still making changes on it since the build still doesn't seem to work just yet(hence the name of the branch), but the changes I have as of now are [in my temp branch.](https://github.com/joperezr/corefx/tree/temp)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
brthor: The cli restore uses core or based DNx under the hood for the time being.

We actually set this in the restore script, so setting it here is redundant.
https://github.com/dotnet/cli/blob/master/scripts/dotnet-restore.sh

davidfowl: It's not needed ever. The dotnet CLI is using a CoreCLR based DNX so mono settings don't apply.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packages.Unix.config
brthor: If you are using dotnet cli why do you need to restore dnx? Just curious :smile:

weshaggard: I think as part of our workflow shift we should be able to remove all these *.config files that are used for restoring. We will no longer need to restore dnx only dotnet cli and the fact it uses dnx behind the scenes right now we don't need to care about.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: Lets create a tools folder %PROJECT_DIR%\Tools

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: Ok, I'll remove from here then.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -0,0 +1,28 @@
...
+   exit 1
+fi
+
+if [ ! -e "$__DOTNET_CMD" ]; then
+   echo "ERROR: Cannot find dotnet.exe at path '$__DOTNET_CMD'. Please pass in the path to dotnet.exe as the 3rd parameter."
+   exit 1
+fi
+
+$__DOTNET_CMD restore $__TOOLS_DIR/tool-runtime/project.json
+$__DOTNET_CMD publish $__TOOLS_DIR/tool-runtime/project.json -f dnxcore50 -r ubuntu.14.04-x64 -o $__TOOLRUNTIME_DIR
```
brthor: Sorry if this is a dumb question, I don't have full context here. But would this bash script potentially be called on platforms like Mac and CentOS?

joperezr: For the time being, our scenarios are only officially supported on Linux and Windows, however we have tried adding shell commands that could potentially run in Mac as well.

brthor: I also do not have a mac to test on :smile:

That said:
`-r ubuntu.14.04-x64`
This will definitely cause some issues, especially for native components. 

A basic way to solve this is to do something like this:
```bash
if [[ "$(uname)" == "Darwin" ]]; then
    $RID="osx.10.10-x64"
elif [[ "$(uname)" == "Linux" ]]; then
    $RID="Ubuntu.14.04-x64"
fi
```

This doesn't solve the issue completely though, because there is no concept of version or what flavor of Linux you are on.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -0,0 +1,34 @@
+@echo off
+setlocal
+
+set PROJECT_DIR=%~1
+set DOTNET_CMD=%~2
+set TOOLRUNTIME_DIR=%~3
+set TOOLS_DIR=%~dp0
```
weshaggard: To help eliminate confusion (which I had until I read the rest of this script), can you name this variable something else. Perhaps BuildToolsPackageDir.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -0,0 +1,34 @@
...
+  echo ERROR: Cannot find tools path at [%TOOLS_DIR%]. Please pass in the tools directory as the 2nd parameter.
+  exit /b 1
+)
+
+if not exist "%DOTNET_CMD%" (
+  echo ERROR: Cannot find dotnet command at [%DOTNET_CMD%]. Please pass in the dotnet command as the 3rd parameter.
+  exit /b 1
+)
+
+set tempDirectory=%cd%
```
weshaggard: Why do you need this tempDirectory? Within the scope of setlocal/endlocal in the script the current directory shouldn't be impacted from outside this script.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -0,0 +1,34 @@
...
+cd %TOOLS_DIR%\tool-runtime\
+call %DOTNET_CMD% restore
+call %DOTNET_CMD% publish -f dnxcore50 -r win7-x64 -o %TOOLRUNTIME_DIR%
+cd %tempDirectory%
+
+xcopy /E /Y /Q %TOOLS_DIR%..\lib %TOOLRUNTIME_DIR%
+
+call type NUL > %TOOLRUNTIME_DIR%\ToolRuntime.semaphore
```
weshaggard: Didn't we decide the semaphore was going to have the version and be handled in the individual repo scripts?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -0,0 +1,31 @@
...
+cp -a $__TOOLS_DIR/. $__TOOLRUNTIME_DIR/
+
+chmod a+x $__TOOLRUNTIME_DIR/corerun
+touch $__TOOLRUNTIME_DIR/ToolRuntime.semaphore
```
weshaggard: Same question about this semaphore

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/nuget.config
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
weshaggard: Do we need a specific config for tool-runtime? If so plese remove the coreclr and corefxtestdata feeds as they shouldn't be needed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.lock.json
```diff
@@ -275,6 +275,9 @@
       "Microsoft.NETCore.ConsoleHost/1.0.0-beta-23328": {
         "type": "package"
```
weshaggard: go ahead and delete this lock file as we are trying to avoid having the checked in now.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
```diff
@@ -17,57 +17,15 @@
     <ToolArchitecture Condition="'$(ToolArchitecture)' == ''">x64</ToolArchitecture>
```
weshaggard: ToolRuntimeProjectJson/LockJson property can be deleted.

weshaggard:
@ericstj PTAL

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: I agree, I'll change this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: I had to do this because for some reason "dotnet publish <pathToProject.Json>" wasn't working for me, when project.json was not in the current directory. If you see the next lines, I basically have to change the working directory to be the location of the project.json, and then do the publish from there. The temp Dir helps me navigate back. I believe Daniel Plasteid was also facing this problem.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: We can do that. This semaphore file is the one used by the msbuild target to ensure that buildtools was restored, while the project.json that contains the version is used by init-tools as a semaphore to tell if we need to restore a newer version of buildtools. I'll change the logic so both msbuild and init-tools will use the project.json file as the semaphore file

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/nuget.config
joperezr: Since we are using dotnet restore at the tool-runtime directory to restore the runtime, dnu will grab the nuget.config file in the working directory to get the feeds. If we don't want to add this file, I could also pass in the --source params to the dotnet restore command.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.lock.json
joperezr: Sure, I'll remove it and add it to the .gitignore file so that it doesn't get picked up everytime that people build.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
joperezr: yes I'll clean up this file

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: I'm find changing directories but I'm not sure why you need to do it this way. Why not just change and say there why do you need to go back? At any rate if you do need to do this you should use pushd and popd instead of trying to capture it this way.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: I was trying to do this now and found a problem with it. The project.json file is being located at $(ToolRuntimePath)/<buildToolsversion>/project.json. The only reason why I was still generating the semaphore file is because in our toolruntime targets, we have it in our 'EnsureBuildToolsRuntime' target. The problem is that if I were to  use project.json instead, that would mean that we would need to know the buildToolsVersion in the msbuild side. I'm not sure how to get around that. Would it be fine if I only check if $(ToolRuntimePath) exists in that target instead of checking the semaphore file?

weshaggard: Yes let's start with that.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -1,12 +1,27 @@
...
     "Microsoft.Build.Utilities.Core": "0.1.0-preview-00005",
+    "Microsoft.CSharp": "4.0.1-beta-23328",
+    "System.Linq": "4.0.1-beta-23328",
+    "System.Xml.XDocument": "4.0.11-beta-23328",
+    "System.IO.Compression": "4.0.1-beta-23328",
+    "System.IO.Compression.ZipFile": "4.0.1-beta-23328",
+    "System.Xml.ReaderWriter": "4.0.11-beta-23328",
+    "System.Xml.XPath.XmlDocument": "4.0.0",
+    "System.Reflection": "4.1.0-beta-23328",
+    "System.Reflection.Metadata": "1.1.0",
```
ericstj: Most of these are redundant to `Microsoft.NETCore.Console`.  In some cases they may contradict what is in there (specifically looking at Reflection.Metadata here).  Can you review this to make sure it is the minimal set of references?

joperezr: Done. now the only missing step is to move the packaging dependencies over to tool-rutnime/project.json and to remove the property CopyNuGetImplementations from the packaging csprojs. I'll push that commit later today once its ready.

joperezr:
Done. I'm currently testing on corefx to make sure repo builds fine.

.gitignore
```diff
@@ -2,6 +2,9 @@ syntax: glob
 
 ### VisualStudio ###
 
+# Removing toolruntime lock file so that it doesn't get picked up everytime it is generated.
+src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.lock.json
```
weshaggard: go ahead and add project.lock.json. We haven't deleted them from this repo yet but we will.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -0,0 +1,25 @@
...
+)
+
+if not exist "%DOTNET_CMD%" (
+  echo ERROR: Cannot find dotnet cli at [%DOTNET_CMD%]. Please pass in the path to dotnet.exe as the 2nd parameter.
+  exit /b 1
+)
+
+cd %BUILDTOOLS_PACKAGE_DIR%\tool-runtime\
+call %DOTNET_CMD% restore
+call %DOTNET_CMD% publish -f dnxcore50 -r win7-x64 -o %TOOLRUNTIME_DIR%
```
weshaggard: There is a good chance you will need to add quotes around these env variables because if anyone has a space in the repo path it will fail otherwise.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -0,0 +1,29 @@
...
+fi
+
+if [ ! -e "$__DOTNET_CMD" ]; then
+   echo "ERROR: Cannot find dotnet.exe at path '$__DOTNET_CMD'. Please pass in the path to dotnet.exe as the 3rd parameter."
+   exit 1
+fi
+
+$__DOTNET_CMD restore $__TOOLS_DIR/tool-runtime/project.json
+$__DOTNET_CMD publish $__TOOLS_DIR/tool-runtime/project.json -f dnxcore50 -r ubuntu.14.04-x64 -o $__TOOLRUNTIME_DIR
+cp -a $__TOOLS_DIR/. $__TOOLRUNTIME_DIR/
```
weshaggard: should this include a ../lib like in the cmd file?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/nuget.config
weshaggard: I kind of don't like having his extra nuget.config, so I think we should consider using the --source option.

src/Microsoft.DotNet.Build.Tasks/project.json
```diff
@@ -4,14 +4,14 @@
     "Microsoft.Build.Framework": "0.1.0-preview-00005",
     "Microsoft.Build.Tasks.Core": "0.1.0-preview-00005",
     "Microsoft.Build.Utilities.Core": "0.1.0-preview-00005",
-    "Microsoft.CSharp": "4.0.0",
-    "System.Linq": "4.0.0",
-    "System.Xml.XDocument": "4.0.0",
-    "System.IO.Compression": "4.0.0",
-    "System.IO.Compression.ZipFile": "4.0.0",
-    "System.Xml.ReaderWriter": "4.0.10",
+    "Microsoft.CSharp": "4.0.1-beta-23328",
```
weshaggard: What do we need that requires the beta packages?

weshaggard:
Only a couple of nits but after those I think it is ready. Please only merge this once you are completely ready to update the corefx repo and internal build changes as well.

.gitignore
joperezr: Done.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: right I'll fix that

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: Yes, for some reason I forgot to push that commit.

src/Microsoft.DotNet.Build.Tasks/project.json
joperezr: I forgot to remove this change. Basically when we moved all dependencies up to tool-runtime/project.json I had to take the later version of each dependency in all project.jsons, which was a beta. After that, Eric suggested to clean up tool-runtime/project.json since I was referencing a meta package already, so it wasn't necesary to have dependencies to all the betas so I removed it from there, but didn't remove it from here. I'll fix it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: Not necessary in this case, in cmd we have to do it because in order to copy all contents from a dir then the folder must end with out a slash, which is why we go one folder back, and then go back to lib. basically it looks something like: C:/foo/bar/toolsdir/lib/../lib. In linux this is not required, since just doing '.' notation works.

weshaggard: Wow that hack wasn't at all obvious to me :) Seem like there should be a better way. I assumed that you put the cmd file into another directory beyond the lib folder.

PR #361, opened 2015-12-09
Add dependency validation option to update invalid dependencies to expected versions

dagood:
Added a parameter that allows dependency validation to fix invalid package versions that it finds. The goal is to reuse the validation rules to update packages. The flow with this to update packages is:

1. Update the validation rule items (`dir.props` in corefx) to the new version.
2. Run validation with `UpdateInvalidDependencies` `true`. (In corefx, through a specific target on the root `build.proj`--**not** by setting a property. We don't want it to look like you can update during a build because there may be other issues there.)
3. Done: the project.json files have been updated.

This also adds a `LogMessage` when a project.json is written to make it explicit which files are being rewritten.

cc: @ericstj @weshaggard


src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
```diff
@@ -9,63 +9,107 @@ namespace Microsoft.DotNet.Build.Tasks
 {
     public class ValidateProjectDependencyVersions : VisitProjectDependencies
     {
+        private delegate void LogAction(string format, params object[] args);
+
         private class ValidationPattern
         {
             private Regex _idPattern;
             private string _expectedVersion;
             private string _expectedPrerelease;
```
weshaggard: nit: empty line.

weshaggard:
LGTM. With these changes do need the other tasks (UpdatePackageDependencyVersion/UpdatePrereleaseDependencies ) anymore?

dagood:
I think I can get rid of `UpdatePrereleaseDependencies`, but `UpdatePackageDependencyVersion` is still useful for upgrading packages where the prerelease version bump makes the package invalid. I actually used it a lot when fixing up the corefx dependency versions.

For example if we update the valid prerelease from `beta-100` to `beta-200`, but around `beta-150` we released a stable version of `System.Foo` and prereleases changed from `1.0.0-...` to `1.1.0-...`, we would need to manually change `System.Foo`'s version from `1.0.0-beta-200` to `1.1.0-beta-200`.

PR #363, opened 2015-12-09
Consolidate copying of supplemental test data.

jhendrixMSFT:
Add a new target, CopySupplementalTestData, that projects can use to copy
test data to the test directory.  In addition, the task will also copy the
assets to the output directory so it can be picked up by the cloud test
infrastructure (it will be included in the test archive).
Exclude the test data from packages.zip as it's now being distributed with
the test binary.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -130,6 +130,45 @@
 
   </Target>
 
+  <!--
+    Temporary until we have proper nuget support to deploy content files.
+    Copies supplemental test data to the build output and test directories.
+  -->
+  <Target Name="CopySupplementalTestData"
+          AfterTargets="CopyTestToTestDirectory"
+          Inputs="@(SupplementalTestData)"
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
cc @ericstj @weshaggard

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -130,6 +130,45 @@
 
   </Target>
 
+  <!--
+    Temporary until we have proper nuget support to deploy content files.
+    Copies supplemental test data to the build output and test directories.
+  -->
+  <Target Name="CopySupplementalTestData"
+          AfterTargets="CopyTestToTestDirectory"
+          Inputs="@(SupplementalTestData)"
```
weshaggard: Where is SupplementalTestData being set?

jhendrixMSFT: It would be defined by the projects that require it.  See https://github.com/jhendrixMSFT/corefx/commit/adf0c5927bce1ed7ec97f56577719d63049f24a6 for how this will be consumed.

PR #368, opened 2015-12-16
ReferencedPackages should include TFM specific deps

ericstj:
Previously the ReferencedPackages item list was only
returning the list of packages listed under the root
dependencies section.  This changes it to return that
as well as any TFM specific dependencies.

/cc @weshaggard @jasonmalinowski @dsgouda


dsgouda:
:+1:  Ship it!

PR #369, opened 2015-12-17
Add Pack to packaging targets

ericstj:
.pkgprojs will now produce a nupkg.
This ports some code that was using NuGet.Core
to do the packing and thus it is only in the desktop
build of the tasks.  We'll eventually need a core-CLR
friendly implementation of this.  Once we have that
library from NuGet we can move this to the shared
source.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -53,8 +53,9 @@
   <PropertyGroup>
     <PackageOutputPath Condition="'$(PackageOutputPath)' == ''">$(BaseOutputPath)pkg\</PackageOutputPath>
     <OutputPath>$(PackageOutputPath)</OutputPath>
-    <NuSpecPath>$(OutputPath)$(Id)$(NuspecSuffix).nuspec</NuSpecPath>
-    <RuntimeFilePath Condition="'$(RuntimeFilePath)' == ''">$(OutputPath)$(Id)$(NuspecSuffix)\runtime.json</RuntimeFilePath>
+    <NuSpecOutputPath Condition="'$(NuSpecOutputPath)' == ''">$(PackageOutputPath)specs\</NuSpecOutputPath>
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -92,7 +94,10 @@
     <ShouldGenerateNuSpec Condition="'$(PackagePlatforms)' == '' AND $(PackageTargetRuntime.Contains('-$(PackagePlatform)'))">true</ShouldGenerateNuSpec>
     <!-- build if PackagePlatforms is not specified and arch is x86 or AnyCPU -->
     <ShouldGenerateNuSpec Condition="'$(PackagePlatforms)' == '' AND ('$(PackagePlatform)' == 'x86' OR '$(PackagePlatform)' == 'AnyCPU')">true</ShouldGenerateNuSpec>
+    <!-- if we built a nuspec, also pack unless explicitly set -->
+    <ShouldCreatePackage Condition="'$(ShouldGenerateNuPkg)' == ''">$(ShouldGenerateNuSpec)</ShouldCreatePackage>
     <BuildDependsOn Condition="'$(ShouldGenerateNuSpec)' == 'true'">GenerateNuSpec</BuildDependsOn>
+    <BuildDependsOn Condition="'$(ShouldCreatePackage)' == 'true'">$(BuildDependsOn);CreatePackage</BuildDependsOn>
```
weshaggard: Do we want to do this as part of building the pkgproj? I thought we were thinking to make this optional for individual library builds and do them all at once for the full build?

ericstj: We can support both options with the flags.  I felt that a default of building the nupkgs was what folks want from the open.  Also in the corefx build workflow we have more flexibility for where signing occurs so we need not separate nuspec creation and packing like we do in TFS.

Also the task is reusable to do a batch pack should we need to do so in the open.  It's the same one I'm using in TFS.

weshaggard:
LGTM

PR #371, opened 2015-12-17
Skip ignored references during Generation validation

ericstj:
Previously we weren't skipping ignored references if they
were direct (only indirect).  This fixes that.

I've also added an override in case we hit something similar
again.

Previously I made this fix for some NETNative assemblies
that were claiming to be generational.  We'll need that 
in the future if we make NETNative more generic than
just netcore50.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
@@ -153,7 +153,7 @@
   <Target Name="ValidatePackageTargetFramework"
           Inputs="$(MSBuildAllProjects);@(CustomAdditionalCompileInputs);@(ReferencePath)"
           Outputs="@(IntermediateAssembly)"
-          Condition="'$(PackageTargetFramework)' != ''"
+          Condition="'$(PackageTargetFramework)' != '' AND '$(SkipValidatePackageTargetFramework)' != 'true'"
           AfterTargets="ResolveReferences">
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -153,14 +153,15 @@
...
+          Condition="'$(PackageTargetFramework)' != '' AND '$(SkipValidatePackageTargetFramework)' != 'true'"
           AfterTargets="ResolveReferences">
     <ValidatePackageTargetFramework AssemblyName="$(AssemblyName)"
                                     AssemblyVersion="$(AssemblyVersion)"
                                     GenerationDefinitionsFile="$(GenerationDefinitionFile)"
                                     PackageTargetFramework="$(PackageTargetFramework)"
                                     DirectReferences="@(ReferencePath)"
-                                    CandidateReferences="@(ReferencePath);@(IndirectReference)" />
+                                    CandidateReferences="@(ReferencePath);@(IndirectReference)" 
+                                    IgnoredReferences="@(ValidateIgnoreReference)"/>
```
weshaggard: Can you move your hard coded list into the targets file?

weshaggard:
LGTM other than the one suggestion.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: sure. done.

PR #372, opened 2015-12-18
Workaround nullref in Mono's MSBuild

ericstj:
Mono's implementation of TaskLoggingHelper seems
to capture the BuildEngine property of the task during
construction of the TaskLoggingHelper.  This breaks
when someone constructs a TasksLoggingHelper in
the constuctor of their task, since none of the fields
are initialized.

Work around this by making our TaskLoggingHelper
lazily initialized.


weshaggard:
LGTM

PR #373, opened 2015-12-18
Sync packaging infrastructure with TFS changes

ericstj:
Generations.cs
Make dll probing in version calculation more flexible

We weren't finding a reference assembly for System.Threading.Tasks.Extensions
so were falling back to package store probing.  Package store brobing would
fail if no stable package existed (as is the case here).

To fix this I've made the probing much more flexible.  This is a temporary
fix, the code is dead in the Open-source port of the task which uses
the actual project references to do the generation calculation.

Generations.json
Move System.ServiceModel.Http 5.4 to 4.1.0.0.  API changes supported
by existing desktop framework 4.6.

Packaging.targets
@chcosta - Packaging for native shims.
@weshaggard - Support project-authored stable packages.
            - Project .builds work.

CreateTrimDependencies.cs
@chcosta - NuGetFramework == is not overloaded, use .Equals
         - handle pre-release dependencies w/o assemblyversion.


weshaggard:
LGTM

PR #377, opened 2015-12-23
Packaging fixes for TFS build

ericstj:
Make packages build library references by default.
Allow TFS to set PackagePlatform.
Fix up formatting.

/cc @weshaggard


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -315,6 +318,13 @@
   <Target Name="ExpandProjectReferences"
           DependsOnTargets="SplitProjectReferences">
 
+    <!-- Only rebuild project references when specified -->
```
weshaggard: I'm fine with just Build for now, at some point we should consider if we also want to support Clean/Rebuild from these.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: Sure, I specifically do not want that right now.

PR #378, opened 2015-12-24
UndefineProperties OSGroup and TargetGroup for contract project references

weshaggard:
Now with the builds files we get race conditions for projects like
System.Collections that build multple partial facades in parallel and so
the building of the ref with different OSGroup or TargetGroup combinations
causes this to build multiple times which conflict given that the output
directores aren't parameterized on those properties. This change simply
undefines these global properties so the project references get folded into one
and there is no longer a race.

cc @ericstj @mellinoe


ericstj:
LGTM.

PR #383, opened 2016-01-05
Support SkipCreatePackageOnMissingFiles

ericstj:
Add a property SkipCreatePackageOnMissingFiles that will
skip the creation of the nuget package when missing files
are detected.

/cc @weshaggard @chcosta

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -883,13 +883,26 @@
   <Target Name="CreatePackage"
           Inputs="$(NuSpecPath)"
           Outputs="$(PackageOutputPath)$(Id).$(Version).nupkg">
+
+    <ItemGroup>
+      <_missingFiles Include="@(PackageFile)" Condition="!Exists(%(FullPath))"/>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <_SkipCreatePackage Condition="'$(SkipCreatePackageOnMissingFiles)' == 'true' AND '@(_missingFiles)' != ''">true</_SkipCreatePackage>
+    </PropertyGroup>
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -883,13 +883,26 @@
...
+    </PropertyGroup>
+
+    <Warning Condition="'$(_skipPack)' == 'true'" Text="Skipping package creation for $(NuSpecPath) because the following files do not exist: @(_missingFiles)" />
+
     <NugetPack Nuspecs="$(NuSpecPath)"
                OutputDirectory="$(PackageOutputPath)"
                Properties="$(PackageProperties)"
-               ExcludeEmptyDirectories="true" />
+               ExcludeEmptyDirectories="true"
+               Condition="'$(_skipPack)' != 'true'"/>
```
chcosta: Casing of _skipPack is inconsistent.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -883,13 +883,26 @@
   <Target Name="CreatePackage"
           Inputs="$(NuSpecPath)"
           Outputs="$(PackageOutputPath)$(Id).$(Version).nupkg">
+
+    <ItemGroup>
+      <_missingFiles Include="@(PackageFile)" Condition="!Exists(%(FullPath))"/>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <_SkipPack Condition="'$(SkipCreatePackageOnMissingFiles)' == 'true' AND '@(_missingFiles)' != ''">true</_SkipPack>
```
chcosta: Maybe a name like "_skipCreatePackage" would be more consistent and obviously associated with the "SkipCreatePackageOnMissingFiles" property.

ericstj:
I made the name change suggestions.

weshaggard:
change LGTM. What scenarios are you envisioning using this? All it really does is turn a error into a warning.

ericstj:
Right now it's for the ci which doesn't build native bits before building packages.  We can turn that off after we fix the build workflow.

PR #384, opened 2016-01-06
Fix backslashes in FindPartialFacadeProjectReference

mellinoe:
This target has actually never worked on Ubuntu, or presumably any other platform than Windows. We had a backup target kicking in here ("FindPartialFacadeContractReference"), which would run if FindPartialFacadeProjectReference did not locate the project reference. The backup target did not work for net46 configurations, because it produced a malformed project.json which it tried to restore. I talked with Wes and we agreed to just remove the backup target because we should always be able to find the reference assembly project. It was also using * versioning still which may have been problematic and also made builds less predictable. If we are for some reason not able to locate the reference assembly project still, we will get the same error as before ("No single matching contract found.").

@weshaggard, @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
@@ -27,8 +27,8 @@
   
   <Target Name="FindPartialFacadeProjectReference">
     <PropertyGroup>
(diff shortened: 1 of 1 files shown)


weshaggard:
LGTM - lets just be sure that we test your private BuildTools change out on corefx repo to ensure no projects are using this fallback.

mellinoe:
Just finished a clean build of corefx with the change, and nothing needed the fallback target. Going to merge in now.

PR #386, opened 2016-01-08
More robust UnsupportedPlatform handling

mellinoe:
This was previously hard-coded to only consider things that were not supported on Windows. Instead, I'm just comparing the identities of the unsupported platforms with the current OSGroup. This should fix all of the unsupported tests being run in x-plat builds (outside of run-tests.sh).

@weshaggard , @Priya91


weshaggard:
LGTM. However you cannot update to a newer version of BuildTools yet until @joperezr's changes gets into corefx.

PR #389, opened 2016-01-12
Adding the option to set the Buildtools target runtime instead of always using win7-x64

joperezr:
Adding the option to set the Buildtools target runtime instead of always using win7-x64


weshaggard:
LGTM

PR #392, opened 2016-01-13
Add attribute __BlockReflectionAttribute target

maririos:
Add the attribute __BlockReflectionAttribute to the libraries that have targerGroup as netcore50 or netcore50aot, instead of running FxTransform

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
@@ -185,4 +185,25 @@
       <VSDesignTimeBuild Condition="'$(BuildingInsideVisualStudio)'=='true' and '$(BuildingOutOfProcess)'=='false'">true</VSDesignTimeBuild>
     </PropertyGroup>
   </Target>
+  
+  <!--
+    Providing a definition for __BlockReflectionAttribute in an assembly is a signal to the .NET Native toolchain 
+    to remove the metadata for all non-public APIs. This both reduces size and disables private reflection on those 
+    APIs in libraries that include this. The attribute can also be applied to individual public APIs to similarly block them.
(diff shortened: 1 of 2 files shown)


maririos:
cc @weshaggard

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
```diff
@@ -185,4 +185,20 @@
       <VSDesignTimeBuild Condition="'$(BuildingInsideVisualStudio)'=='true' and '$(BuildingOutOfProcess)'=='false'">true</VSDesignTimeBuild>
     </PropertyGroup>
   </Target>
+
```
weshaggard: Can you please add a comment about what this is? Also add a similar comment in the cs file.

weshaggard:
Other than adding some comments this looks good.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
maririos: Done

weshaggard: No need to mention FxTransform as that is legacy at this point. I would simply add the following comment in both places.

Providing a definition for __BlockReflectionAttribute in an assembly is a signal to the .NET Native toolchain to remove the metadata for all non-public APIs. This both reduces size and disables private reflection on those APIs in libraries that include this. The attribute can also be applied to individual public APIs to similarly block them.

maririos: Thanks Wes. It is more clear now

maririos:
CC: @jhendrixMSFT

PR #394, opened 2016-01-14
Add additional clean targets

naamunds:
Based on the feedback for https://github.com/dotnet/corefx/pull/4643,
I've decided to add clean targets to BuildTools so that they can be
shared between clean.sh and clean.cmd, and used for both CoreFX and
CoreCLR.  Note that these targets are in addition to the Clean target
which is defined in build.proj in CoreFX and removes the bin directory.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/clean.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/clean.targets
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <UserLocalFolder Condition="'$(OsEnvironment)'!='Unix'">$(LocalAppData)/</UserLocalFolder>
+    <UserLocalFolder Condition="'$(OsEnvironment)'=='Unix'">$(HOME)/.local/share/</UserLocalFolder>
+  </PropertyGroup>
+
+  <Target Name="CleanPackages">
+    <RemoveDir Directories="$(PackagesDir)" />
+  </Target>
+
+  <Target Name="CleanPackagesCache">
+    <RemoveDir Directories="$(UserLocalFolder)NuGet/Cache/;$(UserLocalFolder)NuGet/v3-cache/;$(UserLocalFolder)dnu/cache/" />
+  </Target>
+
+</Project>


src/Microsoft.DotNet.Build.Tasks/PackageFiles/clean.targets
```diff
@@ -0,0 +1,21 @@
...
+  <Target Name="CleanPackages">
+    <RemoveDir Directories="$(PackagesDir)" />
+  </Target>
+
+  <Target Name="CleanPackagesCache">
+    <RemoveDir Directories="$(UserLocalFolder)NuGet/Cache/;$(UserLocalFolder)NuGet/v3-cache/;$(UserLocalFolder)dnu/cache/" />
+  </Target>
+
+  <Target Name="CleanTools">
+    <RemoveDir Directories="$(ToolsDir)" />
```
weshaggard: You will not be able to CleanTools when the tool (msbuild) you are using to run this is in the ToolsDir.

naamunds: Good point.  I've pushed a change to move this logic to clean.sh in CoreFX.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/clean.targets
```diff
@@ -0,0 +1,17 @@
...
+    <UserLocalFolder Condition="'$(OsEnvironment)'!='Unix'">$(LocalAppData)/</UserLocalFolder>
+    <UserLocalFolder Condition="'$(OsEnvironment)'=='Unix'">$(HOME)/.local/share/</UserLocalFolder>
+  </PropertyGroup>
+
+  <Target Name="CleanPackages">
+    <RemoveDir Directories="$(PackagesDir)" />
+  </Target>
+
+  <Target Name="CleanPackagesCache">
+    <RemoveDir Directories="$(UserLocalFolder)NuGet/Cache/;$(UserLocalFolder)NuGet/v3-cache/;$(UserLocalFolder)dnu/cache/" />
```
weshaggard: You should test this but I suspect this will fail if one of those directories doesn't exist. You might need to do some exist checks before adding them.

naamunds: I tested it and it simply no-ops if one of the directories doesn't exist.  This implementation is consistent with the definition of the Clean target in build.proj in CoreFX.

weshaggard:
LGTM

PR #395, opened 2016-01-14
Removing unused nuget feed

joperezr:
dotnet-corefx feed was deleted yesterday, hence builds started failing since the calls to get the packages returned a 404. I checked and it looks like all the packages are already on the other feeds, so simply removing it and updating one lock file gave a green build for me.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -75,7 +75,6 @@
     <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-core/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-coreclr/" />
-    <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-corefx/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-corefxtestdata/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-buildtools/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/nugetbuild/" />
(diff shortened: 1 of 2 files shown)


joperezr:
cc @weshaggard

dir.props
```diff
@@ -75,7 +75,6 @@
     <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-core/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-coreclr/" />
```
weshaggard: You can also remove coreclr feed as it is essentially dead as well.

src/Microsoft.Cci.Extensions/project.lock.json
```diff
@@ -3,7 +3,7 @@
   "version": 2,
```
weshaggard: What did the lock file update?

dir.props
joperezr: Ok, I'll remove it

src/Microsoft.Cci.Extensions/project.lock.json
joperezr: There where two indirect dependencies that had very old versions, which existed in dotnet-corefx feed but they were not available in dotnet-core one. The restore was failing because those two dependencies couldn't be found, so I just updated the lock file and that fixed it, by upgrading the versions of both of those to be able to find them in dotnet-core feed.

ericstj:
LGTM, merging to fix CI.  Separately we need to clean up these * dependencies /cc @dagood

weshaggard:
We should also eliminate the lock files from BuildTools repo. It also looks like we didn't remove the coreclr feed which is also dead and should be removed.

dagood:
I created https://github.com/dotnet/buildtools/issues/402 for the * dependency and lockfile work. If the coreclr feed removal hasn't been done by the time I make those changes, I'll add that into the pull request.

PR #396, opened 2016-01-14
Adding scripts that will bootstrap netcore msbuild and that can be called from repos. Also switching forward slashes to backslashes for OS X build.

joperezr:
[DO NOT MERGE YET] Adding scripts that will bootstrap netcore msbuild and that can be called from repos


dir.props
```diff
@@ -74,11 +74,12 @@
   <ItemGroup>
     <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-core/" />
+    <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/roslyn-nightly/" />
```
weshaggard: We should not add this feed dependency. Any packages we need from there should go on our buildtools feed.

dir.props
```diff
@@ -74,11 +74,12 @@
...
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-coreclr/" />
-    <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-corefx/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-corefxtestdata/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-buildtools/" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/nugetbuild/" />
+    <DnuSourceList Include="C:\NuGet\msbuild-netcore\" />
```
weshaggard: I assume this is temporary but just calling it out so you remove it before this work is finished.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -32,7 +32,7 @@
     <!-- Setting a default portable profile, although nothing should resolve from there as we want to use the pacakge refs -->
     <TargetPlatformIdentifier>Portable</TargetPlatformIdentifier>
     <TargetFrameworkIdentifier>.NETPortable</TargetFrameworkIdentifier>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v5.0</TargetFrameworkVersion>
     <TargetFrameworkProfile>Profile7</TargetFrameworkProfile>
```
weshaggard: When we move to v5.0 there are no profiles so you should remove TargetFrameworkProfile. Actually you should probably just set it to empty.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-msbuild.cmd
```diff
@@ -0,0 +1,40 @@
...
+  exit /b 1
+)
+
+if not exist "%TOOLSDIR%" (
+  echo ERROR: Cannot find the buildtools location at [%TOOLSDIR%]. Please pass in the path to buildtools as the 3rd parameter.
+  exit /b 1
+)
+:: Restore most of the runtime
+cd "%BUILDTOOLS_PACKAGE_DIR%\msbuild-runtime\"
+call "%DOTNET_CMD%" restore --source https://www.myget.org/F/roslyn-nightly/ --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/
```
weshaggard: Shouldn't this be restored as a dependency from the BuildTools package itself? It feels like we have a duplication of all the BuildTools restore/publish logic which we should be able to get for free just by adding the necessary package dependencies to BuildTools package.

dir.props
joperezr: Right, this won't go in the real checkin, this is just so that it builds now.

dir.props
joperezr: Correct.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
joperezr: Actually this branch is not up to date with the latest on my local repo, basically since the official portable targets package doesn't work, I moved this back to 4.5 since Daniel's self-built package works with 4.5. Nonetheless, if you want me to remove the property I can certainly do it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-msbuild.cmd
joperezr: right, as I said in a previous comments, this is not up to date. Basically what I did is to add the msbuild runtime into the tool-runtime/project.json so that the Tools directory has everything it needs includding .Net Core MSBuild. Once I have a ready version of buildtools, I'll push all of those updates into this branch.

joperezr:
@dotnet-bot test this please

joperezr:
With my latest change, I've now removed the roslyn and the local feed for the restore since I have succesfully uploaded all required packages into buildtools feed, this means that CI should pass now, and that this PR is ready to be merge once feedback is done.

joperezr:
cc @weshaggard @JohnChen0 @naamunds

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -6,6 +6,7 @@ set DOTNET_CMD=%~2
 set TOOLRUNTIME_DIR=%~3
 IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
+set MSBUILD_CONTENT_JSON={"dependencies": {"Microsoft.Portable.Targets": "0.1.1-dev"},"frameworks": {"dnxcore50": {},"net46": {}}}
```
weshaggard: Can you remind me again whey we cannot include this in the tool-runtime\project.json?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -4,26 +4,36 @@
   },
   "dependencies": {
     "NuProj": "0.10.4-beta-gf7fc34e7d8",
-    "Microsoft.Build.Framework": "0.1.0-preview-00005",
-    "Microsoft.Build.Tasks.Core": "0.1.0-preview-00005",
-    "Microsoft.Build.Utilities.Core": "0.1.0-preview-00005",
+    "Microsoft.Build.Framework": "0.1.0-preview-00015",
```
weshaggard: Please file an issue to remove tool-runtime\project.json and instead have the dependencies directly in BuildTools package.

weshaggard:
LGTM but please be sure to file the extra tracking issues for follow-up work.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: The reason is that the contents of this package needs to be copied (along with the package's structure) into the ToolRuntime/Extensions folder. If we put this in the tool-runtime project.json instead, then the contents of the package will be flattened out and dropped on the root of the ToolRuntime folder which will not work for MSBuild. I'll add a comment inline the scripts to explain this and also to mention that once dotnet cli adds support for content files then we can remove this workaround.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
joperezr: Will do. I already have a branch for that work, but it is still in progress. My branch for that work is https://github.com/joperezr/buildtools/tree/AddBuildToolsPackageDependencies

joperezr:
I created issue #413 to track the package dependencies work, and #414 for removing the portable targets workaround once dotnet cli supports content files.

PR #397, opened 2016-01-15
Remove dotnet-corefx feed since it no longer exists

ericstj:
/cc @weshaggard @markwilkie 
FYI to anyone who breaks with 
```
Error: FindPackagesById: <packageId>
  Response status code does not indicate success: 404 (Not Found).
```
This is the error you'll get whenever any feed in your source list cannot be resolved.  NuGet/DNX fails rather than just skipping that feed.


joperezr:
I already had an open PR for this #395 and for me I also needed to update a lock file in order for the build to pass, so I think we also have to check that in. I can merge my PR as well to have that change if that's ok.

PR #400, opened 2016-01-18
Add shortcuts for common metadata to all ProjectReferences

weshaggard:
The common metadata includes things like OSGroup and TargetGroup,
as well as some common packaging properties.

cc @ericstj


ericstj:
LGTM

PR #403, opened 2016-01-19
Respect BUILDTOOLS_OVERRIDE_RUNTIME in init-tools.

pgavlin:
If the user has specified a value for BUILDTOOLS_OVERRIDE_RUNTIME
(e.g. if they are building on an OS or distro for which a package
is not available), copy the specified runtime into the necessary
locations for dotnet, dnx, etc.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -45,4 +45,9 @@ cd $__TOOLS_DIR/tool-runtime/
 $__DOTNET_CMD restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/
 $__DOTNET_CMD publish -f dnxcore50 -r ${__PUBLISH_RID} -o $__TOOLRUNTIME_DIR
 
+if [ -n $BUILDTOOLS_OVERRIDE_RUNTIME ]; then
+    find $__TOOLRUNTIME_DIR -name *.ni.* | xargs rm 2>/dev/null
+    cp -R $BUILDTOOLS_OVERRIDE_RUNTIME/* $__TOOLRUNTIME_DIR
+fi
+
 exit $__BUILDERRORLEVEL


pgavlin:
@joperezr @ellismg PTAL. This is a companion change of https://github.com/dotnet/corefx/pull/5529.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -45,4 +45,9 @@ cd $__TOOLS_DIR/tool-runtime/
 $__DOTNET_CMD restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/
 $__DOTNET_CMD publish -f dnxcore50 -r ${__PUBLISH_RID} -o $__TOOLRUNTIME_DIR
 
+if [ -n $BUILDTOOLS_OVERRIDE_RUNTIME ]; then
+    find $__TOOLRUNTIME_DIR -name *.ni.* | xargs rm 2>/dev/null
+    cp -R $BUILDTOOLS_OVERRIDE_RUNTIME/* $__TOOLRUNTIME_DIR
```
joperezr: I believe you also need to copy the runtime into the $__TOOLRUNTIME_DIR/dotnetcli/runtime/coreclr/ folder in order for dotnet cli to work

pgavlin: This is handled by the changes in CoreFX.

pgavlin:
FWIW, I have verified that the two changes produce a working build of CoreFX on Fedora when applied together.

brthor:
@pgavlin Have you tried using the CentOS drop of cli?

There is also the native host which I expect would need to be built appropriately.

pgavlin:
> Have you tried using the CentOS drop of cli?

The CentOS drop does not work for me: it is built against different versions of some libraries than I have installed.

> There is also the native host which I expect would need to be built appropriately.

Which native host? If you're referring to `corerun`, it should be placed in the directory pointed to by `$BUILDTOOLS_OVERRIDE_RUNTIME`.

brthor:
> Which native host? If you're referring to  corerun , it should be placed in the directory pointed to by  $BUILDTOOLS_OVERRIDE_RUNTIME .

CLI uses a custom built host which serves as the entrypoint to all commands. 
https://github.com/dotnet/cli/tree/master/src/corehost

pgavlin:
Ah, interesting. AFAICT it only links against libc, so we'll probably be able to get away without it on the majority of systems for now. We'll absolutely need a proper build of this host in the long-term.

weshaggard:
I'm OK with this as a stop-gap until we figure out a better long-term solution.

PR #405, opened 2016-01-21
Downgrade to NuGet 3.4.0-beta

ericstj:
A stable build of NuGet 3.4 was mistakenly
published to myget and we picked it up.

Undo that and depend on the pre-release
version.

/cc @anurse @weshaggard


ericstj:
I've mirrored the 3.4.0-beta-288 build to https://www.myget.org/F/dotnet-buildtools. I will unlist the 3.4.0 version, then delete it after we've upgraded every buildtools consumer.

anurse:
Looks right to me

ericstj:
Looks like a breaking API change, interesting that didn't show up in my local build (incremental fail perhaps?).  Will react to the breaking change...

ericstj:
Ok, I unlisted the 3.4.0 versions.

PR #408, opened 2016-01-23
Update package references

weshaggard:
Delete all the lock files and update package versions including deleting all * versions.


weshaggard:
cc @dagood @joperezr

joperezr:
LGTM. One thing I want to do once I move the tool-runtime to the package dependencies is to go through all project.jsons and make sure the versions match the runtime that is being restored since Im not sure that's the case right now.

PR #412, opened 2016-01-26
Fix ReportGeneratorVersion

stephentoub:
The report generator command is failing because the path gets generated with "-beta2".


mellinoe:
LGTM

PR #415, opened 2016-01-28
Ported ApiCompat

venkat-raman251:
@weshaggard : Please have a look


src/ApiCompat/Mappings.txt
```diff
@@ -0,0 +1,79 @@
+﻿# Mappings file for Silverlight to Jupiter UI stacks.
```
weshaggard: This file isn't needed it was only part of an experiment. Please delete it.

src/ApiCompat/NamespaceRemappingComparer.cs
```diff
@@ -0,0 +1,132 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
weshaggard: We should update all the copyrights to match what @stephentoub is doing right now. https://github.com/dotnet/buildtools/pull/418

weshaggard:
Do you have a way to show the diff of what you had to change to make it .NET Core complaint?

venkat-raman251:
@weshaggard : I have updated the licenses and also created an initial commit (with the code as it previously was) so that you can see the diff.

weshaggard:
The more I think about how to integrate this the more I think it will be best to actually create a wrapper msbuild task for ApiCompat that we can hock into the individual project builds. So after we get the source ported then please play with creating a msbuild task that wraps this.

venkat-raman251:
@weshaggard : Fixed most of the problems that you had pointed out. Please let me know if it looks ok

weshaggard:
What about all my comments/questions directly on commit https://github.com/venkat-raman251/buildtools/commit/b95a2b39467c352d4c4f0e664cc24d2637481a27

venkat-raman251:
I fixed them based off your comments on that commit (CLScompliance,license,assemblyInfo and some renaming).
The questions you had were about a couple of try-catch blocks which didn't make any sense. Must have missed 'em while cleaning up the code. So i removed it.

venkat-raman251:
@weshaggard - Removed AssemblyInfo, Added handling for ExitCode using a static variable.

venkat-raman251:
@weshaggard  Could you please have a look.

weshaggard:
LGTM thanks @venkat-raman251

PR #416, opened 2016-01-28
Upgrading BuildTools to version 157 and new Workflow

joperezr:
Upgrading BuildTools to version 157 and new Workflow


joperezr:
cc: @weshaggard @ericstj @jhendrixMSFT

weshaggard:
The error looks interesting, potentially some race condition. Also it looks like it is still downloading nuget.exe, I'm not sure why that is.

joperezr:
Yeah I'm pushing the fix now for the exception on downloading nuget. The reason we still need it is because we create the nuget packages using nuget, so I had to add it to init-tools.cmd

joperezr:
FYI, even though I didn't change the Windows build on the PR, I did try locally to build the repo using the restored .Net Core MSBuild and it succeeded, so perhaps we can now add the sh scripts and have this repo build x-plat as well with not a lot of work.

weshaggard:
Can we do the package building with dotnet pack?  If so lets avoid nuget at all.

I want to clean this repo to the barebones and only support windows for now. As I plan to use it as a sample and then we can consider adding x-plat support, but it isn't needed right now.

joperezr:
Doing it with dotnet pack is a bit more complicated since currently it only supports project.json and not nuspec files. I can however use [the task](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs) that @ericstj added to create nuget packages based on nuspecs without requiring NuGet.exe so I'll give that a try.

joperezr:
I created #425 to add the extra parameter to the NugetPack task so that we can use that for the packaging without depending on nuget.exe. Once that PR is merged and a new build tools package is generated, then we can use that version here.

dir.props
```diff
@@ -167,7 +167,7 @@
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('Linux'))">Linux</OSGroup>
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('OSX'))">OSX</OSGroup>
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('FreeBSD'))">FreeBSD</OSGroup>
-    <OSGroup Condition="'$(OSGroup)'==''">Windows_NT</OSGroup>
+    <OSGroup Condition="'$(OSGroup)'==''">AnyOS</OSGroup>
```
joperezr: @weshaggard This is the change to the default OS that I had to do to fix the race condition on the build

weshaggard: OK I assume that after this change everything builds into an AnyOS output directory now correct?

joperezr: Correct, AnyOS.AnyCPU.Debug dir

weshaggard:
LGTM although I think it is worth squashing.

PR #417, opened 2016-01-28
Update TargetingPack filtering

weshaggard:
We now always filter both references and copy-local references from
any packages that starts with "Microsoft.TargetingPack". For references
we still keep any reference included in TargetingPackReference item group.

This also add some default entries to the TargetingPackReference to avoid
duplication in all projects that use these.

cc @ericstj @dagood @mellinoe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
@@ -11,7 +11,7 @@
     <RestorePackages Condition="'$(RestorePackages)'!='false' and Exists('$(ProjectJson)') and '$(DesignTimeBuild)' != 'true'">true</RestorePackages>
     <PrereleaseResolveNuGetPackages Condition="'$(PrereleaseResolveNuGetPackages)'!='false' and Exists('$(ProjectJson)')">true</PrereleaseResolveNuGetPackages>
 
-    <!-- 
+    <!--
         For now, prevent built-in task (if available) from running.
         More changes are needed to light up on their availability
         and use them instead of what we have here. See buildtools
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -124,18 +124,36 @@
...
+      <_TargetingPackPrefix>Microsoft.TargetingPack</_TargetingPackPrefix>
+    </PropertyGroup>
+
+    <!--
+      Add the core libraries and windows to the reference set by default to avoid a lot of duplication in projects
+      They only act as a filter so if they aren't present in the packages references it will not impact anything.
+    -->
+    <ItemGroup Condition="'$(ExcludeDefaultTargetingPackReferences)' != 'true'">
+      <TargetingPackReference Include="mscorlib" />
+      <TargetingPackReference Include="System.Private.CoreLib" />
```
mellinoe: Are we sure we want to default include this one? Most of the assemblies in the .NET Native targeting pack are actually contract-based, and a lot of the .NET Native facades use the targeting pack without referencing System.Private.CoreLib. See System.Reflection.Extensions, System.IO, System.Resources.ResourceManager, others. I would guess that those stop building after this change unless we added the Exclude property to them.

I think it's okay to include Windows.winmd because it's in it's own targeting pack.

weshaggard: Good point. That will cause issues for System.Private.CoreLib because of the mix of contracts and not in that targeting pack. I will remove this one. The biggest reason I added these defaults were for mscorlib and windows as with my change to always filter it will break projects because there are a lot of them that don't have a TargetingPackReference for those as it wasn't required before because we weren't filtering those 2 packages in most cases.

mellinoe: Sounds good. mscorlib by default sounds good as any project using the targeting pack will reference it as far as I can tell.

mellinoe:
LGTM

PR #418, opened 2016-01-28
Update licensing headers

stephentoub:
cc: @richlander, @weshaggard, @martinwoodward


weshaggard:
I only took a quick glance but they looked correct to me.

PR #419, opened 2016-01-28
Update TestWithLocalLibraries support to understand AnyOS

weshaggard:
When TestWithLocalLibraries=true it will find local copied binaries
from AnyOS as well as from the matching OsGroup binary directory.

@stephentoub @ellismg


stephentoub:
LGTM

ellismg:
LGTM

PR #420, opened 2016-01-28
Updating nuget package for perf xunit runner

dsgouda:
Updating nugget package dependency for build tools so we can run the latest runner version and related tools
@ianhays @davmason


ianhays:
LGTM from the buildtools side of things.

PR #421, opened 2016-01-29
Add support for uploading supplemental payloads.

jhendrixMSFT:
In order to support arbitrary workloads folks need the ability to upload
arbitrary payloads.  A new item group, SupplementalPayload, has been added
that is to contain content to be uploaded to the root of the container.
Added property RunnerScript that allows one to override the default runner
script.
Switch to using ';' char for array delimiter as it aligns with how msbuild
writes out an expanded item group.


jhendrixMSFT:
@ChadNedzlek @MattGal  @dsgouda

PR #422, opened 2016-01-30
Add support for NETStandard and baseline references

ericstj:
This change adds a couple opt-in features to packaging:
1. Using NETStandard instead of DotNet as the portable generation moniker.
2. Raising dependencies to a minimal baseline version.

We need both of these to switch our packages to NETStandard.  
#2 is required because the old packages do not have NETStandard
assets for compile time and NETStandard is not compatible with 
dotnet assets.

This also updates the default minClientVersion to 3.4.0 since we
depend on changes in 3.4.0: the latest generation mapping of
dotnet5.6 > dnxcore50 & UAP and NETStandard support.

/cc @weshaggard


src/Microsoft.DotNet.Build.Tasks.Packaging/src/Generations.cs
```diff
@@ -185,29 +195,29 @@ public Version DetermineGenerationFromSeeds(string assemblyName, Version assembl
 
             if (lowerGeneration == null && upperGeneration == null)
             {
-                log.LogMessage(LogImportance.Low, "Assembly {0} is not tracked by generations.", assemblyName);
+                log.LogMessage(LogImportance.Low, $"Assembly {assemblyName} is not tracked by generations.");
```
weshaggard: C# 6.0 features :) Hopefully that doesn't break anything, I don't think it will.

weshaggard:
LGTM - Please squash as the commits seem to be going back and forth :)

When do you plan to enable this?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/Generations.cs
ericstj: Buildtools supports c#6.  I'm already using it in the same file.

ericstj:
Sure I will squash.  I am hoping to enable it this week, but I need to get some other things in place first.

ericstj:
Had to update the baseline versions to account for https://github.com/dotnet/corefx/pull/5845

PR #423, opened 2016-01-30
Add check to GenFacades command so that if DebugType is Portable we disable the rewrite in genfacades.

joperezr:
Given that we have moved to using .Net Core MSBuild in non-Windows corefx builds, we are now able to produce Portable PDBs as part of the build. The problem is that CCI doesn't support Portable PDB reading yet, so GenFacades can't rewrite them as part of the build. This change is to disable the PDB rewriting for now when the PDBs are portable and will be removed once CCI supports it.


joperezr:
cc @weshaggard

mellinoe:
LGTM. To clarify this is just disabling the re-write of the PDB. The assembly itself will still be re-written as expeccted.

joperezr:
Correct, the assembly is being rewritten, but not the PDB. @weshaggard mentioned that the PDBs might actually work like this, given that we only add type-forwards so theoretically the PDB will still be good, but we haven't test that yet.

weshaggard:
LGTM - We should still file a issue tracking the testing of this and potentially needing to add support for this one way or another.

joperezr:
I created issue dotnet/corefx#5825 to track this testing work.

PR #427, opened 2016-02-02
Add a custom helper command to speed up restores

ellismg:
In CoreFX we spend a bunch of time restoring packages, because we end up invoking
dotnet restore 300+ times to work around DNX bugs. This tool helps us batch up
these invocations which decreases the number of processes we have to launch and
should help dotnet restore do some additional caching

These problems may go away when we switch to a dotnet CLI which restores using
nuget, but for now this has a measurable impact on package restore.

In my measurements, package restore where the "packages" folder did not exist (but
the NuGet/DNX caches had not been cleared) dropped from ~20 mins to ~4.5 and a
package restore where everything had already been restored previously (e.g. calling
sync.cmd /p twice and measuring the second run) went from ~17 mins to ~50 seconds.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
@@ -1,0 +1,79 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GatherDirectoriesToRestore : Task
+    {
+        [Required]
+        public string[] RootDirectories { get; set; }
+
+        [Output]
+        public string[] DirectoriesToRestore { get; set; }
+
+        private int directoryListLength = 7000;
+
+        public int DirectoryListLength
+        {
+            get { return directoryListLength; }
+            set { directoryListLength = value; }
+        }
+
+        public override bool Execute()
+        {
+            HashSet<string> directoriesToRestore = new HashSet<string>();
+
+            foreach (string rootDirectory in RootDirectories)
+            {
+                AddDirectoriesToRestore(rootDirectory, rootDirectory, directoriesToRestore);
+            }
+
+            List<string> resultBuilder = new List<string>();
+            StringBuilder sb = new StringBuilder(directoryListLength);
+
+            foreach (string directoryPath in directoriesToRestore)
+            {
+                sb.Append('"');
+                sb.Append(directoryPath);
+                sb.Append("\" ");
+
+                if (sb.Length >= directoryListLength)
+                {
+                    resultBuilder.Add(sb.ToString());
+                    sb.Clear();
+                }
+            }
+
+            if (sb.Length > 0)
+            {
+                resultBuilder.Add(sb.ToString());
+            }
+
+            DirectoriesToRestore = resultBuilder.ToArray();
+
+            return true;
+        }
+
+        public void AddDirectoriesToRestore(string directoryToExplore, string directoryToAdd, HashSet<string> directoriesToRestore)
+        {
+            bool hasProjectJson = File.Exists(Path.Combine(directoryToExplore, "project.json"));
+
+            if (hasProjectJson)
+            {
+                directoriesToRestore.Add(directoryToAdd);
+            }
+
+            foreach (string childDirectory in Directory.GetDirectories(directoryToExplore))
+            {
+                AddDirectoriesToRestore(childDirectory, hasProjectJson ? childDirectory : directoryToAdd, directoriesToRestore);
+            }
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
@@ -19,6 +19,7 @@
   <ItemGroup>
     <Compile Include="Delegates.cs" />
     <Compile Include="ExecWithMutex.cs" />
+    <Compile Include="GatherFoldersToRestore.cs" />
     <Compile Include="GenerateAssemblyList.cs" />
     <Compile Include="GenerateResourcesCode.cs" />
     <Compile Include="GenerateEncodingTable.cs" />
--- a/src/Microsoft.DotNet.Build.Tasks/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/project.json
@@ -9,6 +9,7 @@
     "System.Xml.XDocument": "4.0.11-rc2-23712",
     "System.IO.Compression": "4.0.1-rc2-23712",
     "System.IO.Compression.ZipFile": "4.0.1-rc2-23712",
+    "System.IO.FileSystem": "4.0.1-rc2-23712",
     "System.Xml.ReaderWriter": "4.0.11-rc2-23712",
     "System.Xml.XPath.XmlDocument": "4.0.0",
     "System.Reflection": "4.1.0-rc2-23712",


ellismg:
@dagood I know you were looking at some longer term improvements here (maybe tied to getting back on NuGet on dotnet CLI), but I'm wondering what you think of the above.  I banged this out because I was getting really frustrated with how long restore was taking, especially in the lab.

Do you think doing something like this is worth while?

If so, I will validate that this also works on *NIX (I expect it should, but it's also likely I've screwed up when editing `init-tools.sh`).  Happy to see this whole thing go away once you land your improvements.

/cc @weshaggard @stephentoub

stephentoub:
Thank you, @ellismg!!

src/BatchRestore/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                return RunDotnetRestore(argumentBuilder);
+            }
+
+            return 0;
+        }
+
+        private static int RunDotnetRestore(StringBuilder argumentBuilder)
+        {
+            string DotnetCommandName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "dotnet.exe" : "dotnet";
```
stephentoub: I'd have expected "dotnet" would work in either case.  It doesn't on Windows?

src/BatchRestore/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            if (argumentBuilder.Length > argumentPrefix.Length)
+            {
+                return RunDotnetRestore(argumentBuilder);
+            }
+
+            return 0;
+        }
+
+        private static int RunDotnetRestore(StringBuilder argumentBuilder)
```
stephentoub: Nit: since no arguments are being added in this method, seems like this could just be ```string arguments```.

src/BatchRestore/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+            StringBuilder argumentPrefixBuilder = new StringBuilder();
+
+            argumentPrefixBuilder.Append("restore");
+
+            if (s_packages != null)
+            {
+                argumentPrefixBuilder.AppendFormat(" --packages \"{0}\"", s_packages);
+            }
+
+            if (s_sources.Length > 0)
```
stephentoub: Is this check necessary?

dagood:
Good idea! Yeah, the switch to dotnet-using-nuget will make this unnecessary since it does recursive directory restore in the more expected way, but this looks like it could help sooner.

Using this (if it works xplat) would just mean removing https://github.com/dotnet/corefx/blob/master/dir.props#L127-L131 and changing the command we generate, right?

src/BatchRestore/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+                p.WaitForExit();
+                return p.ExitCode;
+            }
+        }
+
+        private static void VisitDirectory(string directory, bool shouldAddDirectoryToRestoreList)
+        {
+            if (shouldAddDirectoryToRestoreList)
+            {
+                s_directoriesToRestore.Add(directory);
```
dagood: Does this add *all* subdirs to the command where a project.json is in a dir? Or is there some filter I'm not seeing for directories without any project.jsons in them? (Since those directories don't need to be restored.) I don't think it makes a difference for dnu restore time, but it might let us fit more on a single command. It might not be worth the extra complexity though.

src/BatchRestore/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+        private static string[] s_sources = Array.Empty<string>();
+        private static string[] s_rootDirectories = Array.Empty<string>();
+        private static string s_packages;
+
+        private static void ParseCommandLine(string[] args)
+        {
+            CommandLineParser.ParseForConsoleApplication((parser) =>
+            {
+                parser.DefineOptionalQualifier("source", ref s_sources, "--source argument to pass to dotnet restore.");
+                parser.DefineOptionalQualifier("packages", ref s_packages, "--package argument to pass to dontet restore.");
```
dagood: Nit: `dontet` -> `dotnet`

weshaggard:
Thanks for taking the initiative here @ellismg, I like the idea but I instead of making a wrapper exe why not just make a build task that will give back an item group with a chunked set of command lines? That seems like it would be more general purpose and would get you out of the business of starting processes directly, and would eliminate one more exe startup.

src/BatchRestore/Program.cs
ellismg: Didn't try.  I like Wes's idea of making the generation a build task instead, but I'll keep this in mind for the future.

src/BatchRestore/Program.cs
ellismg: There's no filter.  I can go add one, probably worth while to cut down more on work.

src/BatchRestore/Program.cs
ellismg: Nope.  Left over from a different way of doing this I was trying earlier.

ellismg:
That's a really good idea, Wes.  I'll do that.

bartonjs:
My experiment was going to be to try merging all of the relevant project.json parts into one uber-project.json; so nuget only had to do everything once.  It has the drawback of needing to understand the project.json files; but seemed like it might take out enough redundancy to be worth it.

(And I'm jealous that you're only having 20 minutes in the slow case, Matt; I'm on minute 63 of package restore right now)

ellismg:
@stephentoub @dagood @weshaggard PTAL.  I have moved to the task approach that you outlined, Wes.

It's integrated into corefx like this (I'll send a PR after this merges):

```patch
diff --git a/build.proj b/build.proj
index 3e66066..0324f88 100644
--- a/build.proj
+++ b/build.proj
@@ -41,11 +41,17 @@
     </TraversalBuildDependsOn>
   </PropertyGroup>
 
+  <UsingTask TaskName="GatherDirectoriesToRestore" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
+
   <Target Name="BatchRestorePackages">
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Restoring all packages..." />
-    <!-- restore all project.jsons in one pass for perf & to avoid concurrency problems with dnu -->
-    <!-- include ToolsDir to restore test-runtime\project.json as well -->
-    <Exec Command="$(DnuRestoreCommand) &quot;%(DnuRestoreDir.Identity)&quot;" StandardOutputImportance="Low" CustomErrorRegularExpression="(^Unable to locate .*)|(^Updating the invalid lock file with .*)" ContinueOnError="WarnAndContinue" />
+
+    <GatherDirectoriesToRestore RootDirectories="@(DnuRestoreDir)">
+      <Output TaskParameter="DirectoriesToRestore"
+              ItemName="ExpandedDnuRestoreDir" />
+    </GatherDirectoriesToRestore>
+
+    <Exec Command="$(DnuRestoreCommand) &quot;%(ExpandedDnuRestoreDir.Identity)&quot;" StandardOutputImportance="Low" CustomErrorRegularExpression="(^Unable to locate .*)|(^Updating the invalid lock file with .*)" ContinueOnError="WarnAndContinue" />
 
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Restoring all packages...Done." />
   </Target>
@@ -131,13 +131,6 @@
   <ItemGroup>
     <DnuRestoreDir Include="$(MSBuildProjectDirectory)/src" />
     <DnuRestoreDir Include="$(ToolsDir)" />
-    <!-- workaround to address issue where DNX won't recurse directories under a project.json
-         https://github.com/aspnet/dnx/commit/0dda8bf86863364cc20421f1af7494f1b2a3256f -->
-    <DnuRestoreDir Include="$(MSBuildProjectDirectory)/src/*/src/**/project.json" />
-    <DnuRestoreDir Include="$(MSBuildProjectDirectory)/src/*/ref/*/project.json" />
-    <DnuRestoreDir Include="$(MSBuildProjectDirectory)/src/*/tests/*/project.json" />
-    <DnuRestoreDir Include="$(MSBuildProjectDirectory)/src/*/*/netcore50aot/project.json" />
-    <DnuRestoreDir Include="$(MSBuildProjectDirectory)/src/Common/tests/**/project.json" />
   </ItemGroup>
 
   <PropertyGroup>
```

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public class GatherDirectoriesToRestore : Task
+    {
+        [Required]
+        public string[] RootDirectories { get; set; }
+
+        [Output]
+        public string[] DirectoriesToRestore { get; set; }
+
+        public int DirectoryListLength = 7000;
```
weshaggard: I would make this a configurable property as well. Just encase we need to adjust it. It is fine to default it still if it isn't set. Why did you pick 7000? Do you expect that to be different on different OS's?

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+
+            DirectoriesToRestore = resultBuilder.ToArray();
+
+            return true;
+        }
+
+        public void AddDirectoriesToRestore(string directoryToExplore, string directoryToAdd, HashSet<string> directoriesToRestore)
+        {
+            bool hasProjectJson = File.Exists(Path.Combine(directoryToExplore, "project.json"));
```
weshaggard: NiceToHave: might make the file name configurable via a property as well, and as a bonus perhaps accept wildcards.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            if (hasProjectJson)
+            {
+                directoriesToRestore.Add(directoryToAdd);
+            }
+
+            foreach (string childDirectory in Directory.GetDirectories(directoryToExplore))
+            {
+                AddDirectoriesToRestore(childDirectory, hasProjectJson ? childDirectory : directoryToAdd, directoriesToRestore);
```
weshaggard: I must be missing something (perhaps I should go to bed) but why is the directoryToAdd ever different then the directoryToExplore? Are you trying to find root directories or something? If that is so I think you might need to be caution about depending on dotnet restore doing a restore of anything beyond a single project.json file in the same directory.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public int DirectoryListLength = 7000;
+
+        public override bool Execute()
+        {
+            HashSet<string> directoriesToRestore = new HashSet<string>();
+
+            foreach (string rootDirectory in RootDirectories)
+            {
+                AddDirectoriesToRestore(rootDirectory, rootDirectory, directoriesToRestore);
```
weshaggard: I realize just taking the directories is going to shorten the overall arguments but I wonder if that is overall faster or slower then directly passing the project.json file.

weshaggard:
Your patch for corefx looks along the lines as I was expecting.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
ellismg: Yeah the idea was to find root directories.  If there's a project.json, it means that whatever parent directory got us here needed to be restored.  If there isn't one, restoring the parent directory is fine.

For example:

```
Foo
Foo\Bar\project.json
Foo\Bar\Baz\project.json
Foo\Baz\project.json
```

In this case, we call the function with DirectoryToAdd as "Foo".  There's no project.json, so we recur to both Bar and Baz but keep DirectoryToAdd as "Foo"

When we Explore Foo\Bar, we see a project.json, so we add "Foo" (since DNU's recursive walk on Foo will find the project.json in the same directory) and when we recur in to the child directories we see "directoryToAdd" to the new child path (in this case Foo\Bar\Baz).

So we'd end up with `Foo` and `Foo\Bar\Baz` and the directories to restore.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
ellismg: I meant to make this configurable, but forgot to make it a property.  *NIX OS's appear to have no fixed limit on the command line.  Note today with a limit of 7000 we only generate one command line invocation.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
ellismg: I can measure this.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
ellismg: (Note in the above example, we walk into Foo\Baz with DirectoryToAdd as "Foo", there's a project.json there and we do add it to the HashSet of things to restore, but it was already there for when we hit the project.json while walking Foo\Bar)

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
ellismg: That would break the directory walking invariant I want to exploit (to keep command lines small).  If we aren't going to need this, I'd rather just keep what we have...

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
weshaggard: This logic seems like it is a little too bound to the exact issue with dnu today for my taste. I would much prefer a simplier model were we just give all directories with a project.json so we aren't depending on a particular behavior from dnu to discover things.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
ellismg: This is a bit slower.  I think the major issue is when we pass in directories and let `dotnet restore` do the walking we only have to invoke `dotnet restore` more times (5 times vs 1 time).  There's a bit of overhead to actually boot the whole thing up and have it reload all of the state about the feeds, etc.  It's maybe ~15-20s slower in the case where there's no work to be done.  I'd rather keep what I propose.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
ellismg: If we expect to keep this long term, I'm happy to move to that model. I initially added it to work around the exact DNU issues were were seeing, with the understanding that we would remove this once we were off DNU and on to NuGet.

There's a bit of a cost in moving away from this model, since it means that we basically just take all the project.json's and then split them in to groups (this task becomes conceptually simpler since it can remove all the file system logic and just take in an item group of paths and a limit and return an item group of strings of paths where they are combined up to that limit).

If it didn't make the common case slower (see my comment a little later about having to invoke dotnet restore five times instead of once), I'd be all for it.  As is, I'm a little hesitant, but I'll do whatever you and @dagood would prefer.

ellismg: I just realized that minimizing the number of DNU invocations may also help in our internal TFS where we use the file share, since I expect DNU would only have to enumerate it once.

dagood:
This looks good to me.

For the future I think a `UseDnuRecurseWorkaround` arg might be useful, but since xplat nuget supports restore without the strange dnu recurse behavior this won't be necessary once we upgrade in corefx/wcf. Later on if we need to list individual project.jsons it would be good to have this available to make sure we're safe on line length, but I think we should avoid individual project.json restore anyway so that work could be done when we need it.

I wonder if that arg would be useful in other repos that use buildtools, though: is anyone restoring individual project.jsons with xplat nuget dotnet restore? Those repos might appreciate the speed/safety gain from this task without the dnu workaround.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
weshaggard: I'm cool with keeping what you have.

src/Microsoft.DotNet.Build.Tasks/GatherFoldersToRestore.cs
weshaggard: I'm good keeping what you have given this is hopefully just a temporary measure.

weshaggard:
LGTM thanks @ellismg

Issue #430, opened 2016-02-02
Fix no Revision Number on buildtools packages after moving away from nuget.exe


joperezr:
With [my change](https://github.com/dotnet/buildtools/pull/416) for moving away from nuget.exe and using NugetPack instead, I introduced a bug where the package version does not contain a revision number any more.

bryanar:
Test

joperezr:
This was fixed by adding nuget.exe back. For the time being this is the plan since Nuget.exe is still required for the 'pushing' of the packages. Hopefully dotnet will support push command, so we don't depend on nuget.exe anymore.

PR #431, opened 2016-02-02
Falling back to using NuGet.exe for package creation temporary to fix no revision in packages bug.

joperezr:
This is a temporary fix to unblock folks. The right fix is comming which is adding support to overriding Versions in NugetPack task.

### DIFF
--- a/src/dirs.proj
+++ b/src/dirs.proj
@@ -31,7 +31,9 @@
   <Import Project="$(ToolsDir)CodeCoverage.targets" Condition="Exists('$(ToolsDir)CodeCoverage.targets')" />
   <Import Project="$(ToolsDir)PerfTesting.targets" Condition="Exists('$(ToolsDir)PerfTesting.targets')" />
 
-  <!-- Overriding BuildPackages Target to use NuGet-independent version instead -->
+<!--  Temporary Removing this since NugetPack doesn't support passing in a version number, so falling back to Nuget.exe temporarily.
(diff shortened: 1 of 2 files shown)


joperezr:
cc @jhendrixMSFT @weshaggard @ericstj

joperezr:
This change is to temporary unblock @jhendrixMSFT since he is hitting issue #430 introduced by my buildtools update. I'm also working on a real fix which is adding support to passing in a version to the nugetpack task.

src/dirs.proj
```diff
@@ -31,6 +31,8 @@
   <Import Project="$(ToolsDir)CodeCoverage.targets" Condition="Exists('$(ToolsDir)CodeCoverage.targets')" />
   <Import Project="$(ToolsDir)PerfTesting.targets" Condition="Exists('$(ToolsDir)PerfTesting.targets')" />
 
+<!--  Temporary Removing this since NugetPack doesn't support passing in a version number, so falling back to Nuget.exe temporarily.
+      Issue: https://github.com/dotnet/buildtools/issues/430
   <!-- Overriding BuildPackages Target to use NuGet-independent version instead -->
```
mellinoe: Doesn't this nested XML comment mess things up?

joperezr: yup, I didn't actually test it until now and it did. I already pushed the fix

joperezr:
Given that this change is just temporary since it is required for the actual fix to come, I'll go ahead and merge it.

PR #434, opened 2016-02-04
Updating MSBuild package references

joperezr:
This new versions of the packages contain a fix for Microsoft/MSBuild#459 that we need in order to enable Portable PDB generation in non-Windows builds of corefx.


joperezr:
cc: @weshaggard

weshaggard:
LGTM please make sure these packages are on our buildtools feed.

joperezr:
Done, also I have removed the coreclr feed as it was causing errors and 404 thrown because of the redirection to dotnet.myget.org

PR #435, opened 2016-02-04
Add CentOS support

naamunds:
This change adds support for CentOS to the BuildTools tool runtime.  It
also removes ToolNugetRuntimeId, which is no longer used, and fixes the
parameter validation logic in init-tools.sh.

See https://github.com/dotnet/corefx/issues/4650

cc: @ellismg @joperezr @eerhardt


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -13,13 +13,13 @@ if [ ! -d "$__PROJECT_DIR" ]; then
...
-if [ ! -d "$__TOOLS_DIR" ]; then
-   echo "ERROR: Cannot find tools path at '$__TOOLS_DIR'. Please pass in the tools directory as the 2nd parameter."
+if [ ! -e "$__DOTNET_CMD" ]; then
+   echo "ERROR: Cannot find dotnet.exe at path '$__DOTNET_CMD'. Please pass in the path to dotnet.exe as the 2nd parameter."
    exit 1
 fi
 
-if [ ! -e "$__DOTNET_CMD" ]; then
-   echo "ERROR: Cannot find dotnet.exe at path '$__DOTNET_CMD'. Please pass in the path to dotnet.exe as the 3rd parameter."
+if [ ! -d "$__TOOLRUNTIME_DIR" ]; then
```
joperezr: I wouldn't check if ToolRuntime dir doesn't exist as it is possible that this is the case, what I would check instead is:
```
if [ -z "$__TOOLRUNTIME_DIR" ]; then
```
which checks if the variable is defined, or alternatively you could also check if the variable is empty.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
```diff
@@ -13,7 +13,6 @@
 
     <ToolTargetFramework Condition="'$(ToolTargetFramework)' == ''">DNXCore,Version=v5.0</ToolTargetFramework>
     <ToolArchitecture Condition="'$(ToolArchitecture)' == ''">x64</ToolArchitecture>
```
joperezr: I believe that both ToolTargetFramework and ToolArchitecture shouldn't be needed anymore either. I'm not 100% sure though, but if we don't then perhaps we can clean that out as well.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
naamunds: I was thinking that the Tools directory (i.e. `TOOLRUNTIME_DIR`) existing is a prerequisite of invoking init-tools.sh.  The CoreFx init-tools.sh creates this directory when it does `mkdir "$__PROJECT_JSON_PATH"`.

Also, I'm not changing this logic from the way you had it when you created the script. Is there a reason it needs to change?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
naamunds: I searched the code bases in BuildTools, CoreFx, CoreClr, WCF, and CLI, and you're right that they aren't used in any of them. I've removed them from toolruntime.targets locally.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: Most repos will probably mimic what we have in corefx where, as you mentioned previously, we create the toolRuntime dir before calling this script. However, I don't think this to be a requirement, since you could totally select a different folder as your toolruntime folder, and if it doesn't exist, then dotnet publish will create it for you, so I don't think that we should make it a requirement for the folder to already exist. This is why previous logic wasn't checking if the TOOLRUNTIME_DIR existed.

naamunds: The script does `cp -R $__TOOLS_DIR/* $__TOOLRUNTIME_DIR` before calling dotnet publish.  The `cp` command won't work correctly if `__TOOLRUNTIME_DIR` doesn't already exist, so if we want to support the case where `__TOOLRUNTIME_DIR` doesn't exist, then I think we need to have init-tools.sh explicitly create the directory before copying to it.

The previous logic checked if the `__TOOLS_DIR` existed, but it referred to it as the 2nd parameter, so I assumed that it was intended to be checking the if the `__TOOLRUNTIME_DIR` existed.  Maybe that's not correct, though.  In any case, do you think I should add a `mkdir` before the `cp`?  Personally I'd be fine with either approach.

joperezr: sure, we could do mkdir before cp

weshaggard:
@ellismg should this be centos or rhel or both?

joperezr:
@weshaggard if we choose rhel, does the dotnet cli run on redhat? AFAIK dotnet cli only supports windows, ubuntu, osx, and centos

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
naamunds: Ok, I've made that change.

ellismg:
One small comment based on the new push, LGTM otherwise.  Thanks for fixing this up!

joperezr:
LGTM

PR #439, opened 2016-02-05
Fixes for cloud testing tasks.

jhendrixMSFT:
Add a new property, FailIfExists, to task CreateAzureContainer that when
set to true will cause the task to fail if it attempts to create a
container that already exists.
In the UploadToAzure task, if the operation fails don't simply let an
exception go unhandled.  Instead, log an error message and have the task
return false.
Disable overwriting of upload content by default.  To overwrite, set
property OverwriteOnUpload to true.


jhendrixMSFT:
@MattGal

ericstj:
LGTM

PR #440, opened 2016-02-06
[WIP]Add BinClash detecting build logger

ericstj:
This msbuild logger detects binclashes by recording all projects that
build and their TargetPath.  Once the build is done it logs all
clashes that occurred during the build.

This is a work in progress and I plan to refine the logging and
sequence of the error message.

/cc @weshaggard @mellinoe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
@@ -6,6 +6,7 @@ using Microsoft.Build.Framework;
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text;
 
@@ -15,14 +16,13 @@ namespace Microsoft.DotNet.Build.Tasks
     {
         internal class ProjectState
         {
-
             public ProjectState(ProjectStartedEventArgs args)
             {
-                this.GlobalProperties = args.GlobalProperties;
-                this.ProjectFile = args.ProjectFile;
-                this.ContextId = args.BuildEventContext.ProjectContextId;
-                this.ParentContextId = args.ParentProjectBuildEventContext.ProjectContextId;
-                this.TargetPath = (string)args.Properties?.Cast<DictionaryEntry>().FirstOrDefault(p => ((string)p.Key).Equals("TargetPath")).Value;
+                GlobalProperties = args.GlobalProperties;
+                ProjectFile = args.ProjectFile;
+                ContextId = args.BuildEventContext.ProjectContextId;
+                ParentContextId = args.ParentProjectBuildEventContext.ProjectContextId;
+                TargetPath = (string)args.Properties?.Cast<DictionaryEntry>().FirstOrDefault(p => ((string)p.Key).Equals("TargetPath")).Value;
             }
 
             public bool RanBuild { get; set; }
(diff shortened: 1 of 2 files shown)


ericstj:
Here's sample output from the logger: https://gist.github.com/ericstj/e2be25ee073dfe7bd99d

It gets logged at the very end of the build, then I throw an exception to cause the build to fail.  It isn't great but its the only way I could find to have a logger fail the build.  I could write a file then have a target read the file and fail (emitting proper errors) if folks think that's better.

I could also emit the errors during the build at the point the project is building if folks like that better.

mellinoe:
LGTM. We definitely need automatic validation for this, it is far too hard to get right by hand.

src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
```diff
@@ -0,0 +1,231 @@
...
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class BinClashLogger : ILogger
+    {
+        internal class ProjectState
+        {
+            public ProjectState(ProjectStartedEventArgs args)
+            {
+                GlobalProperties = args.GlobalProperties;
```
weshaggard: Do you know if this is read-only at this point?

src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        /// <summary>
+        /// All projects evaluations that built, indexed by context ID.
+        /// </summary>
+        private Dictionary<int, ProjectState> _projectHistory = new Dictionary<int, ProjectState>();
+        private string _logFile = null;
+        private StreamWriter _fileWriter = null;
+
+        private bool _append = false;
```
weshaggard: Doesn't look like you are parsing _append.

src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+
+            if (_fileWriter != null)
+            {
+                _fileWriter.Dispose();
+            }
+
+            if (_exceptionOnError && (clashes != 0))
+            {
+                throw new Exception($"{clashes} bin clashes were detected during the build.");
```
weshaggard: Is throwing an exception from the logger the right way to force a failure?

weshaggard:
This is awesome! Thanks for digging into this and doing it, I should have dug into this approach earlier while writing my script.

src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
ericstj: I spent quite a lot of time trying to find a way to gracefully error-out but I couldn't come up with anything.  I tried forwardinglogger/nodelogger which can let me generate new events but those events only flow to my central logger, not the default loggers.  I could have created some task that runs at the end of the build and communicate with it, but that felt less general purpose.

src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
ericstj: It looks to me like its a detached copy of the properties, created with ToDictionary().  Also, this is a logger, it's purpose is to observe.  If it was possible to change the collection that'd be a bug in MSBuild and not something we should rely on.

src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs
ericstj: Good catch.  I added it with the intent to make it configurable but forgot to parse the param string.

PR #441, opened 2016-02-10
Split references by FrameworkAssembly/Dependency

ericstj:
Previously we'd try to determine if an assembly was mscorlib-based
or System.Runtime-based.  If mscorlib-based we assumed all
dependencies were framework assemblies.  This was incorrect, partial
facades may still reference packages and those packages may have an
OOB implementation.  This was recently the case for
System.IO.Compression's reference to System.Buffers.
https://github.com/dotnet/corefx/issues/5941
/cc @sokket @weshaggard @chcosta

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -91,6 +91,7 @@
 
   <UsingTask TaskName="ApplyPreReleaseSuffix" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
   <UsingTask TaskName="GetAssemblyReferences" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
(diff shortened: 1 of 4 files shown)


ericstj:
Merging now, need to get this moving.  I'll take feedback and follow up.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -442,18 +443,18 @@
...
-
+    <SplitReferences References="@(_FileReferencedAssemblies)"
+                     TargetFramework="$(_TargetFramework)"
+                     FrameworkListsPath="$(FrameworkListsPath)">
+      <Output TaskParameter="FrameworkReferences" ItemName="_FileFrameworkReference"/>
+      <Output TaskParameter="PackageReferences" ItemName="_FilePackageReference"/>
+    </SplitReferences>
+    
+    <ItemGroup Condition="'@(_FilePackageReference)' != ''">
+      <_FilePackageReference Remove="corefx;mscorlib;System;System.Core;System.Xml;Windows" />
```
weshaggard: Who would ever have a reference to "corefx"?

Issue #402, opened 2016-01-19
Remove project.json * dependencies and lockfiles


dagood:
Like what's been done for corefx (https://github.com/dotnet/corefx/pull/4894) and wcf, we need to remove floating dependencies and lockfiles in buildtools, replacing them with version validation.

dagood:
Merged in #442.

PR #442, opened 2016-02-11
Update prerelease versions, add pre-restore dependency validation

dagood:
NuGet restore couldn't find rhel.7-x64 runtimes for old packages, so I updated the tool-runtime project.json.

After that, to finish https://github.com/dotnet/buildtools/issues/402, I added the dependency validation task to the build and upgraded the other project.jsons in `src\`.

/cc @ericstj

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -59,6 +59,21 @@
     <TestRuntimeProjectLockJson Condition="'$(TestRuntimeProjectLockJson)' == ''">$(SourceDir)Microsoft.DotNet.Build.Tasks/PackageFiles/test-runtime/project.lock.json</TestRuntimeProjectLockJson>
   </PropertyGroup>
 
+  <!-- Package dependency validation -->
+  <PropertyGroup>
+    <ValidatePackageVersions>true</ValidatePackageVersions>
+    <ProhibitFloatingDependencies>true</ProhibitFloatingDependencies>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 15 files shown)


ericstj:
There are other projects in this repo that use rc2-23712.  Can we please update them all at once?

dagood:
Sure. I'll take care of https://github.com/dotnet/buildtools/issues/402 now to make sure they stay in sync.

dagood:
Added validation and updated project.jsons, with the unfortunate (as far as the diff's concerned) side effect of formatting them.

ericstj:
LGTM

dir.props
```diff
@@ -59,6 +59,21 @@
...
 
+  <!-- Package dependency validation -->
+  <PropertyGroup>
+    <ValidatePackageVersions>true</ValidatePackageVersions>
+    <ProhibitFloatingDependencies>true</ProhibitFloatingDependencies>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ValidationPattern Include="^((System\..%2A)|(Microsoft\.CSharp)|(Microsoft\.NETCore.%2A)|(Microsoft\.Win32\..%2A)|(Microsoft\.VisualBasic))(?&lt;!TestData)$">
+      <ExpectedPrerelease>rc3-23811</ExpectedPrerelease>
```
weshaggard: Do we actually need these to be RC3 yet? I ask because that might put RC2 at risk if we need to get a buildtools update and some packages now conflict and are newer coming BuildTools.

dagood: The RC2 version being used before didn't pass nuget restore validation because rhel.7 didn't have implementations. I could try a new RC2 version, or we could just keep the old version and ignore the ToolRuntime restore errors. I'll see if rc2-23816 works without restore errors.

PR #445, opened 2016-02-12
Add logic for local Content to correctly binplace to the tests directory

ianhays:
Content files from the test csproj weren't considered as special in the publishing to the tests directory. For all dependencies we take the file name and extension, append that to the tests directory and call it a day. Sub-directories are never considered, regardless of the type of the item (content, none, etc.).

We do something special for restoring supplemental test data from a nuget package (i.e. testdata), but do not do anything along those lines for content within the csproj directly.

This commit adds logic to respect the subdirectories of Content items.

@mellinoe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -102,7 +102,8 @@
         <NugetPackageId>%(TestCopyLocal.NugetPackageId)</NugetPackageId>
         <SourcePath>%(Identity)</SourcePath>
       </_TestCopyLocalByFileName>
-    </ItemGroup>      
+    </ItemGroup>
+
     <RemoveDuplicatesWithLastOneWinsPolicy Inputs="@(_TestCopyLocalByFileName)">
       <Output TaskParameter="Filtered" ItemName="_TestCopyLocalByFileNameWithoutDuplicates" />
     </RemoveDuplicatesWithLastOneWinsPolicy>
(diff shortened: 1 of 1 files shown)


mellinoe:
This approach looks fine to me. I think there might be a way to do it without adding a separate copying target, but I'm not 100% sure how to do it and it might not be worth trying to figure that out as this file is already complicated enough.

Anyways, I think we could add something like a `RelativeDestinationPath` metadata property to the ItemGroup that gets passed to Copy, and then set it to the relative subdirectory only for Content files. Basically, instead of Exclude'ing the Content files from the ItemGroup, just add some metadata item into them that stores the relative path. And then pass that into the end of the DestinationFolder parameter of Copy. It would end up being the same DestinationFolder for most things because they won't have a relative path stored in their metadata.

@weshaggard  Did the above make any sense? Is something like that possible? I forget the MSBuild syntax for this sort of thing.

ianhays:
>This approach looks fine to me. I think there might be a way to do it without adding a separate copying target

If that's something we would want, then yes we can do that. I can't add it to "CopySupplementalTestData" since that depends on SupplementalTestData being defined (i.e. has nuget test assets), but I could merge it into the CopyTestToTestDirectory Copy target if I change it to use DestinationFiles.

>instead of Exclude'ing the Content files from the ItemGroup, just add some metadata item into them that stores the relative path.

That I do not know the syntax for. I was thinking something along the lines of:
```
    <ItemGroup>
      <SourcesWithoutContent Include="@(_TestCopyLocalByFileNameWithoutDuplicates)" Exclude="@(ContentWithTargetPath->'%(FileName)%(Extension)')" />
      <SourcesToCopyToTestDir Include="@(SourcesWithoutContent->'%(SourcePath)');@(ContentWithTargetPath)" />
      <DestinationsForTestDir Include="@(SourcesWithoutContent->'$(TestPath)$(TestTargetFrameworkFolder)%(Filename)%(Extension)');@(ContentWithTargetPath->'$(TestPath)$(TestTargetFrameworkFolder)%(TargetPath)')" />
    </ItemGroup>
...

    <Copy
      SourceFiles="@(SourcesToCopyToTestDir)"
      DestinationFiles="@(DestinationsForTestDir)"
...
```

weshaggard:
I would like to avoid creating another task for this. I would prefer we fold it into the CopyTestToTestDirectory itself. I think what you suggest should be about right. One question Have is how do you expect someone to set ContentWithTargetPath? Do you expect them to explicitly list the full output path?

ianhays:
>One question I have is how do you expect someone to set ContentWithTargetPath? Do you expect them to explicitly list the full output path?

I expect content to be defined as relative paths e.g.
```
    <Content Include="NetworkFiles\rawint">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </Content>
```

ianhays:
Done. Changes pass with CoreFX when we modify the expected content paths for System.Net.NetworkInformation and System.Reflection.TypeExtensions.CoreCLR (the only projects that are affected by these changes). 

Those updates will come in a CoreFX PR after we get a new buildtools built.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -102,8 +102,8 @@
         <NugetPackageId>%(TestCopyLocal.NugetPackageId)</NugetPackageId>
         <SourcePath>%(Identity)</SourcePath>
       </_TestCopyLocalByFileName>
-    </ItemGroup>      
-    <RemoveDuplicatesWithLastOneWinsPolicy Inputs="@(_TestCopyLocalByFileName)">
+    </ItemGroup>
+      <RemoveDuplicatesWithLastOneWinsPolicy Inputs="@(_TestCopyLocalByFileName)">
```
mellinoe: nit: Whitespace

mellinoe:
LGTM

ianhays:
Fixed and rebased to one commit. Thanks, Eric.

ianhays:
included in build tools version 169

PR #446, opened 2016-02-13
Add IgnoreForCI xunit trait

ianhays:
There's a bug in Jenkins that blocked https://github.com/dotnet/corefx/pull/5639 that has to do with non-ascii file names (https://issues.jenkins-ci.org/browse/JENKINS-12610).

This commit adds a new xunit trait ```IgnoreForCI``` that allows us to specify which tests should be run everywhere <i>but</i> Jenkins.

@ellismg @mellinoe


stephentoub:
Thanks for doing this, Ian.

I do wonder whether for a trait this simple whether a new attribute and associated discoverer is actually warranted.  For PlatformSpecific and ActiveIssue, there's logic associated with whether to apply the trait or not, but here, that's not the case: every test it's applied to gets the trait.  Given that, should we instead just have a public string constant exposed that can be used with the built-in TraitAttribute? e.g. ```[Trait(XunitConstants.Category, XunitConstants.IgnoreForCI)]```

ianhays:
Sure, I could do that. I just copied what we do for Outerloop, but IgnoreForCI is going to be far more seldom used. I can remove the discoverer and Attribute and just make XunitConstants.IgnoreForCI and XunitConstants.Category publicly accessible.

stephentoub:
It's up to you... I was just asking the question :smile:

ellismg:
LGTM.  I like the updated approach.

weshaggard:
LGTM

PR #448, opened 2016-02-15
Upgrade code coverage tooling versions

stephentoub:


weshaggard:
@stephentoub we should try and think about a way to eliminate this version number bump to not require a BuildTools revision every time. At a minimim we should allow the version number to be overridable so that individual repo's can specify the version they want to use.

stephentoub:
@weshaggard, absolutely, I'd love to not have to update build tools at all for this. :smile:

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -5,8 +5,8 @@
     Code coverage package versions go here and in the test-runtime-packages.config
   -->
   <PropertyGroup>
-    <OpenCoverVersion>4.6.476-rc</OpenCoverVersion>
-    <ReportGeneratorVersion>2.4.0</ReportGeneratorVersion>
+    <OpenCoverVersion>4.6.519</OpenCoverVersion>
```
weshaggard: For the next update can you add conditions to these something like:

```
 <OpenCoverVersion Condition="'$(OpenConverVersion)'==''">4.6.519</OpenCoverVersion>
```

That way we can set version in corefx repo and only need to update build tools version when there is a reasonable milestone build or we require changes to our targets files. This is the approach we have taken already with test-runtime/project.json file as well, we override it in corefx repo.

stephentoub: Sure, sounds fine.  I've simply been following what I thought was the desired approach.  Happy to change to something that doesn't require touching buildtools.

weshaggard: While is great to fully encapsulate everything into BuildTools I think it has some drawbacks when it comes to all these version numbers. I think BuildTools needs to have a default version defined that works with the tools and targets that we use from BuildTools itself but all the individual specific versions seem more like repo level configuration. So I think making them extensible in BuildTools and overriding them in the repo dir.props seems like the right thing to do for these cases.

PR #450, opened 2016-02-16
Depend on RC2 packages rather than RC3

dagood:
Makes sure packages don't conflict, discussed here: https://github.com/dotnet/buildtools/pull/442#discussion_r52824388. These RC2 packages work fine with the NuGet restore compat checks.

Changes from `rc3-23811` to `rc2-23816` (newest).

/cc @weshaggard


dagood:
Also added direct dependencies in tool-runtime\project.json to System.IO.FileSystem and System.Security.Cryptography to fix linux build. Thanks @joperezr for the fix!

joperezr:
LGTM. @dagood I've verified that issue dotnet/corefx#6128 is fixed by this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -11,32 +11,34 @@
...
     "NuGet.Client": "3.4.0-beta-488",
-    "System.IO.Compression": "4.1.0-rc3-23811",
-    "System.Dynamic.Runtime": "4.0.11-rc3-23811",
-    "System.Linq.Expressions": "4.0.11-rc3-23811",
-    "System.Diagnostics.TextWriterTraceListener": "4.0.0-rc3-23811"
+    "System.IO.Compression": "4.1.0-rc2-23816",
+    "System.Dynamic.Runtime": "4.0.11-rc2-23816",
+    "System.Linq.Expressions": "4.0.11-rc2-23816",
+    "System.Diagnostics.TextWriterTraceListener": "4.0.0-rc2-23816",
+    "System.IO.FileSystem": "4.0.1-rc2-23816",
```
weshaggard: @ericstj when you are back I want to better understand why we needed to directly reference these in order to get the correct runtime packages. Without directly referencing FileSystem and Algorithms we were getting the correct contract package but we were getting older runtime packages for those guys. The older runtime packages were comming from the MSBuild and Compiler package references but I cannot understand why it picked a mis-match between the contract package and runtime package versions.

PR #458, opened 2016-02-18
Add a simple preprocessor task

ellismg:
Add a very simple task that understands the limited form of #if/#endif
used by mscorlib.txt used for mscorlib's resources. This will allow us
to preprocess the file to generate a new .txt file that has only the
resources mscorlib needs (based on what FEATURE_ flags are defined)
and then use that as the .txt file MSBuild will use to generate
resources.

This will be used to resolve dotnet/coreclr#3090.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
@@ -1,0 +1,86 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class PreprocessFile : Task
+    {
+        [Required]
+        public string SourceFile { get; set; }
+
+        [Required]
+        public string OutputFile { get; set; }
+
+        public string[] Defines { get; set; }
+
+        public override bool Execute()
+        {
+            HashSet<string> defineSet = new HashSet<string>(Defines ?? new string[] { }, StringComparer.Ordinal);
+            Stack<bool> preprocessorStack = new Stack<bool>();
+
+            preprocessorStack.Push(true);
+
+            using (StreamWriter outputStream = new StreamWriter(File.Open(OutputFile, FileMode.Create, FileAccess.Write, FileShare.None), Encoding.UTF8))
+            {
+                foreach (string line in File.ReadLines(SourceFile))
+                {
+                    string trimmedLine = line.Trim();
+
+                    if (trimmedLine.StartsWith("#if ", StringComparison.Ordinal))
+                    {
+                        string defineName = trimmedLine.Substring(3).Trim();
+
+                        if (defineName == "")
+                        {
+                            Log.LogError("Malformed #if, missing symbol name");
+                            return false;
+                        }
+
+                        if (defineName[0] != '!')
+                        {
+                            preprocessorStack.Push(preprocessorStack.Peek() && defineSet.Contains(defineName));
+                        }
+                        else
+                        {
+                            preprocessorStack.Push(preprocessorStack.Peek() && !defineSet.Contains(defineName.Substring(1)));
+                        }
+                    }
+                    else if (trimmedLine.StartsWith("#endif", StringComparison.Ordinal))
+                    {
+                        preprocessorStack.Pop();
+
+                        if (preprocessorStack.Count == 0)
+                        {
+                            Log.LogError("Extra #endif detected.");
+                            return false;
+                        }
+                    }
+                    else if (preprocessorStack.Peek())
+                    {
+                        outputStream.WriteLine(line);
+                    }
+                    else
+                    {
+                        // Exclued by defines.
+                    }
+                }
+
+                if (preprocessorStack.Count != 1)
+                {
+                    Log.LogError("Missing #endif detected.");
+                    return false;
+                }
+            }
+
+            return true;
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
@@ -31,6 +31,7 @@
     <Compile Include="GetDoItemsIntersect.cs" />
     <Compile Include="OpenSourceSign.cs" />
     <Compile Include="ParseTestCoverageInfo.cs" />
+    <Compile Include="PreprocessFile.cs" />
     <Compile Include="RemoveDuplicatesWithLastOneWinsPolicy.cs" />
     <Compile Include="PrereleaseResolveNuGetPackageAssets.cs" />
     <Compile Include="UpdatePackageDependencyVersion.cs" />


ellismg:
@weshaggard Please take a look.  The diff for corefx to enable this (and remove the mono dependency) will look like this:

```patch
diff --git a/build.sh b/build.sh
index 74880ae..8fc7f29 100755
--- a/build.sh
+++ b/build.sh
@@ -17,7 +17,7 @@ usage()
     echo "configureonly - do not perform any builds; just configure the build."
     echo "skipconfigure - skip build configuration."
     echo "skipnative - do not build native components."
-    echo "skipmscorlib - do not build mscorlib.dll even if mono is installed."
+    echo "skipmscorlib - do not build mscorlib.dll."
     echo "skiptests - skip the tests in the 'tests' subdirectory."
     echo "cmakeargs - user-settable additional arguments passed to CMake."
 
@@ -223,8 +223,6 @@ build_mscorlib()
        return
     fi
 
-    hash mono 2> /dev/null || { echo >&2 "Skipping mscorlib.dll build since Mono is not installed."; __SkipMSCorLib=1; return; }
-
     # Restore buildTools
 
     restoreBuildTools
diff --git a/src/mscorlib/GenerateSplitStringResources.targets b/src/mscorlib/GenerateSplitStringResources.targets
index 4b69ba6..f4081a4 100644
--- a/src/mscorlib/GenerateSplitStringResources.targets
+++ b/src/mscorlib/GenerateSplitStringResources.targets
@@ -1,33 +1,33 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ResGenCommand Condition="'$(ResGenCommand)'=='' and '$(OsEnvironment)'=='Windows_NT'">$(SDK40ToolsPath)ResGen.exe</ResGenCommand>
-    <ResGenCommand Condition="'$(ResGenCommand)'=='' and '$(OsEnvironment)'=='Unix'">resgen</ResGenCommand>
     <PrepareResourcesDependsOn>GenerateSplitStringResources;$(PrepareResourcesDependsOn)</PrepareResourcesDependsOn>
   </PropertyGroup>
     
+  <UsingTask TaskName="PreprocessFile" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
+
   <Target Name="GenerateSplitStringResources"
-          Inputs="$(MSBuildThisFileFullPath);$(MSBuildProjectFile);$(BclSourcesRoot)\mscorlib.txt"
-          Outputs="@(SplitTextStringResource->'$(IntermediateOutputPath)%(Filename).resources')">
-         
-    <ItemGroup Condition="'$(OsEnvironment)'=='Windows_NT'">
+          Inputs="@(SplitTextStringResource)"
+          Outputs="@(SplitTextStringResource->'$(IntermediateOutputPath)%(Filename).txt')">
+
+    <ItemGroup>         
       <Internal_ResGenDefines Remove="" />
       <Internal_ResGenDefines Include="INCLUDE_DEBUG" />
       <Internal_ResGenDefines Include="INCLUDE_RUNTIME" />
       <Internal_ResGenDefines Include="%(SplitTextStringResource.ResGenDefines)" />
     </ItemGroup>    
 
-    <Exec Command="&quot;$(ResGenCommand)&quot; &quot;%(SplitTextStringResource.Identity)&quot; &quot;$(IntermediateOutputPath)%(SplitTextStringResource.Filename).resources&quot; @(Internal_ResGenDefines->'/D:%(Identity)', ' ')" />
+    <PreprocessFile SourceFile="%(SplitTextStringResource.Identity)" OutputFile="$(IntermediateOutputPath)%(SplitTextStringResource.Filename).txt" Defines="@(Internal_ResGenDefines)" />
 
     <ItemGroup>
-      <EmbeddedResource Include="@(SplitTextStringResource->'$(IntermediateOutputPath)%(Filename).resources')">
+      <EmbeddedResource Include="@(SplitTextStringResource->'$(IntermediateOutputPath)%(Filename).txt')">
         <WithCulture>false</WithCulture>           
         <ManifestResourceName>%(Filename)</ManifestResourceName>
       </EmbeddedResource>
-      <Clean Include="$(IntermediateOutputPath)%(SplitTextStringResource.Filename).resources" />
+      <Clean Include="$(IntermediateOutputPath)%(SplitTextStringResource.Filename).txt" />
     </ItemGroup>
 
     <ItemGroup>
-        <Internal_ResGenDefines Remove="" />
+      <Internal_ResGenDefines Remove="" />
     </ItemGroup>
   </Target>
 </Project>
diff --git a/src/mscorlib/src/mscorlib.txt b/src/mscorlib/src/mscorlib.txt
index 4ae2269..f435a7b 100644
--- a/src/mscorlib/src/mscorlib.txt
+++ b/src/mscorlib/src/mscorlib.txt
@@ -8,7 +8,9 @@
 
 ; For resource info, see the ResourceManager documentation and the ResGen tool,
 ; which is a managed app using ResourceWriter.
-; ResGen now supports C++ & C# style #ifdef's, like #ifndef FOO and #if BAR
+; ResGen now supports limited preprocessing of txt files, you can use
+; #if SYMBOL and #if !SYMBOL to control what sets of resources are included in
+; the resulting resources.
 
 ; The naming scheme is: [Namespace.] ExceptionName _ Reason
 ; We'll suppress "System." where possible.
```

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
```diff
@@ -0,0 +1,85 @@
...
+        public string SourceFile { get; set; }
+
+        [Required]
+        public string OutputFile { get; set; }
+
+        public string[] Defines { get; set; }
+
+        public override bool Execute()
+        {
+            HashSet<string> defineSet = new HashSet<string>(Defines ?? new string[] { }, StringComparer.Ordinal);
```
ellismg: Can't use `Array<string>.Empty` here because of the .NET 4.5 version.

ellismg:
(Note we'd use this for both Windows and Cross-Plat and stop picking up resgen.exe from the system wide SDK, which is in alignment with our long-term goals).

weshaggard:
cc @tarekgh @AlexGhiondea 

Yes this this is along the lines of what I was expecting to see. One potential issue to make sure works is building mscorlib on the TFS side, I'm not entirely sure if BuildTools gets restored on the Razzle side.  If it doesn't however we should either consider restoring it or checking in the build tasks directly into razzle.

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    public class PreprocessFile : Task
+    {
+        [Required]
+        public string SourceFile { get; set; }
+
+        [Required]
+        public string OutputFile { get; set; }
+
+        public string[] Defines { get; set; }
```
tarekgh: should the defines be required too?

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
tarekgh: do we care do case insensitive too? I mean use OrdinalIgnoreCase?

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
```diff
@@ -0,0 +1,85 @@
...
+            HashSet<string> defineSet = new HashSet<string>(Defines ?? new string[] { }, StringComparer.Ordinal);
+            Stack<bool> preprossorStack = new Stack<bool>();
+
+            preprossorStack.Push(true);
+
+            using (StreamWriter outputStream = new StreamWriter(File.Open(OutputFile, FileMode.Create, FileAccess.Write, FileShare.None), Encoding.UTF8))
+            {
+                foreach (string line in File.ReadLines(SourceFile))
+                {
+                    if (line.StartsWith("#if ", StringComparison.Ordinal))
```
tarekgh: what happen if there is spaces before #if? May be we need to trim the spaces before the processing

tarekgh:
added minor comments, but in general LGTM

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
ellismg: It's reasonable to imagine no defines were present when pre-processing.  For mscorlib.txt we always have `INCLUDE_RUNTIME` and `INCLUDE_DEBUG` defined, but it seemed bad to bake such policy into the preprocessor.

tarekgh: I would say if there is no defines present we shouldn't call the task at all. the target file should handle this as a condition. it would be a waste of time processing a file without any need.

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
ellismg: Thanks, I will add the Trim().

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
ellismg: I don't think we want to enforce the policy that FEATURE_FOO == feature_foo here.  ResGen didn't either, FWIW.

tarekgh: ok. that make sense.

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
ellismg: > I would say if there is no defines present we shouldn't call the task at all.

If the file you want to pre-process has `#if` directives in it, then calling this task with no defines will strip out all the stuff guarded by the `#if`'s.  Skipping the task does something different (i.e. it keeps all the resources that should have been removed in the file).

tarekgh: right. I was misunderstood this part. so now it make sense too.

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        [Required]
+        public string OutputFile { get; set; }
+
+        public string[] Defines { get; set; }
+
+        public override bool Execute()
+        {
+            HashSet<string> defineSet = new HashSet<string>(Defines ?? new string[] { }, StringComparer.Ordinal);
+            Stack<bool> preprossorStack = new Stack<bool>();
```
AlexGhiondea: typo (preprossor) ?

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            preprossorStack.Push(true);
+
+            using (StreamWriter outputStream = new StreamWriter(File.Open(OutputFile, FileMode.Create, FileAccess.Write, FileShare.None), Encoding.UTF8))
+            {
+                foreach (string line in File.ReadLines(SourceFile))
+                {
+                    if (line.Trim().StartsWith("#if ", StringComparison.Ordinal))
+                    {
+                        string defineName = line.Trim().Substring(3).Trim();
```
AlexGhiondea: We are doing line.Trim() a couple of times here -- maybe capture that in a local?

AlexGhiondea:
LGTM baring minor comments.

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
ellismg: I will fix this.

src/Microsoft.DotNet.Build.Tasks/PreprocessFile.cs
ellismg: Sure.

gkhanna79:
@ellismg What is the CoreCLR change required to remove the dependency on resgen and use this?

ellismg:
@gkhanna79 see dotnet/coreclr#3214.

Issue #451, opened 2016-02-16
Add Microsoft.CCI to dependency version validation rules


dagood:
Currently the dependency validation rules don't match Microsoft.CCI, so package upgrades using the auto-update target don't affect dependencies on that package. Microsoft.CCI should probably have the same prerelease version as the other dependenices on the dotnet-core feed.

PR #459, opened 2016-02-18
Check Microsoft.Cci dependency version with validation

dagood:
Validate Microsoft.Cci package with the rest of the prerelease packages. Fixes https://github.com/dotnet/buildtools/issues/451.

Also minor fix of changing validation regex to case-insensitive.

/cc @joperezr

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -57,10 +57,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ValidationPattern Include="^((System\..%2A)|(Microsoft\.CSharp)|(Microsoft\.NETCore.%2A)|(Microsoft\.Win32\..%2A)|(Microsoft\.VisualBasic))(?&lt;!TestData)$">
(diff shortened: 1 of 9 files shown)


joperezr:
Did you try running a corefx build with the changes on the version for CCI? I don't see why it would break, but it's always better to check first.

joperezr:
LGTM

dagood:
I didn't try that, good call. I'll do that before it gets merged.

ericstj:
@dagood can you rebase and test this?

dagood:
@ericstj Sure, thanks for reminding me about this.

dagood:
A corefx build using this worked. CI failure seems to be a build machine error:

> There was an I/O error during the build. This could be caused by a hung process on the target machine, preventing it from cleaning up the workspace, or other issues, such as low disk space.

@dotnet-bot test this please.

I'll see if I can remote into that machine and see if the disk is full.

Issue #460, opened 2016-02-18
DropURI property has incorrect value


rahku:
Below lines in CreateAzureStorage target in cloudtest.targets is not correct:
    
    <CreateProperty Value="$(DropUriRoot)/$(BuildArch)$(BuildType)">
      <Output TaskParameter="Value" PropertyName="DropUri" />

ListURI does not have $(BuildArch)$(BuildType) in its path. It works in corefx because they don't specify any property BuildArch and BuildType. This needs to be fixed. Coreclr does specify these two properties.

jhendrixMSFT:
This has been fixed and will show up in build tools version 00172.

PR #462, opened 2016-02-19
Fixes for cloud testing targets.

jhendrixMSFT:
In the CreateAzureContainer task, append a '/' to the storage URI if one
doesn't exist.
Don't append $(BuildArch) and $(BuildType) properties to the root of the
URI as not all repos define these properties.  It's also redundant as
architecture-specific content is already copied to the appropriate
subdirectories in the container.

This fixes the bug mentioned in issue https://github.com/dotnet/buildtools/issues/460.


jhendrixMSFT:
@MattGal @rahku

PR #463, opened 2016-02-19
Workaround to copy native artifacts to the tools dir (for CLI upgrade)

dagood:
In init-tools.sh we expect dotnet publish to copy the native artifacts in `tool-runtime/project.json` packages into the Tools dir. If we upgrade the version of dotnet CLI that we use, native artifact copy doesn't happen due to ~~https://github.com/dotnet/cli/issues/710~~ https://github.com/dotnet/cli/issues/1528.

This works around it by manually copying the expected native artifacts to the Tools dir.

**Open question**: hardcoding the path feels really bad. Another option is to restore `tool-runtime/project.json` *again* into a folder under the tool runtime dir and copy the files out from there. This also would mean that we can use better wildcards rather than hardcoding the path, because we know only the packages we want are in the folder. This would slow down the init though. Is it worth it?

(A reason to use `~/.nuget/packages` in the first place is that dotnet CLI doesn't support specifying a packages dir for publish. https://github.com/dotnet/cli/issues/1075 / https://github.com/dotnet/cli/issues/261)

/cc @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -26,6 +26,12 @@ cd "%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\"
 call "%DOTNET_CMD%" restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/
 call "%DOTNET_CMD%" publish -f dnxcore50 -r %BUILDTOOLS_TARGET_RUNTIME% -o "%TOOLRUNTIME_DIR%"
 
+:: Temporary workaround for CLI #1528: copy native artifacts that weren't during dotnet publish.
+set TOOL_RUNTIME_PACKAGE_DIR=%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\packages
+call "%DOTNET_CMD%" restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/ --packages "%TOOL_RUNTIME_PACKAGE_DIR%"
+copy "%TOOL_RUNTIME_PACKAGE_DIR%\Microsoft.Build.Targets\0.1.0-preview-00017\runtimes\any\native\*" "%TOOLRUNTIME_DIR%"
+copy "%TOOL_RUNTIME_PACKAGE_DIR%\MSBuild\0.1.0-preview-00017\runtimes\any\native\*" "%TOOLRUNTIME_DIR%"
+
 :: Copy Portable Targets Over to ToolRuntime
 mkdir "%BUILDTOOLS_PACKAGE_DIR%\portableTargets"
 echo %MSBUILD_CONTENT_JSON% > "%BUILDTOOLS_PACKAGE_DIR%\portableTargets\project.json"
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -68,6 +68,12 @@ if [ -n "$BUILDTOOLS_OVERRIDE_RUNTIME" ]; then
     cp -R $BUILDTOOLS_OVERRIDE_RUNTIME/* $__TOOLRUNTIME_DIR
 fi
 
+# Temporary workaround for CLI #1528: copy native artifacts that weren't during dotnet publish.
+__TOOL_RUNTIME_PACKAGE_DIR="$__TOOLS_DIR/tool-runtime/packages"
+"$__DOTNET_CMD" restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/ --packages "$__TOOL_RUNTIME_PACKAGE_DIR"
+cp "${__TOOL_RUNTIME_PACKAGE_DIR}/Microsoft.Build.Targets/0.1.0-preview-00017/runtimes/any/native/"* "$__TOOLRUNTIME_DIR"
+cp "${__TOOL_RUNTIME_PACKAGE_DIR}/MSBuild/0.1.0-preview-00017/runtimes/any/native/"* "$__TOOLRUNTIME_DIR"
+
 # Copy Portable Targets Over to ToolRuntime
 mkdir "$__TOOLS_DIR/portableTargets"
 echo $__MSBUILD_CONTENT_JSON > "$__TOOLS_DIR/portableTargets/project.json"


weshaggard:
cc @joperezr 

Why is this just an issue on x-plat? Wouldn't we need a similar change in init-tools.cmd? 

As for the workaround, yes this will break every time we update tool-runtime\project.json, I can already see it. 

Looking at https://github.com/dotnet/cli/issues/710 one of the comments seems to indicate that this works if is a package reference so I would expect this to work for tool-runtime. Are you sure it is the same underlying issue?

joperezr:
>Why is this just an issue on x-plat? Wouldn't we need a similar change in init-tools.cmd?

It must be happening both in x-plat and Windows. The difference is that in Windows we don't use .Net Core MSBuild which I assume that most of the packages with native artifacts are comming from MSBuild dependencies, which is why even though this problem is also in Windows, it is not currently breaking the build. The only thing we use in Windows from the Tools directory are the Build tasks and its dependencies, which don't have native artifacts at all.

I don't love the change specially because I think anybody can change the nuget home directory, but I'm not sure what to do instead. I will say that if we are making this change, we **must** make it as well for init-tools.cmd because even though corefx is not using .NetCore MSBuild in windows, other buildtools consumers might.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -75,6 +75,14 @@ cd "$__TOOLS_DIR/portableTargets"
 "$__DOTNET_CMD" restore --source https://www.myget.org/F/dotnet-buildtools/ --packages "$__TOOLS_DIR/portableTargets/packages/"
 cp -R "$__TOOLS_DIR/portableTargets/packages/Microsoft.Portable.Targets/${__PORTABLETARGETS_VERSION}/contentFiles/any/any/." "$__TOOLRUNTIME_DIR/."
 
+# Temporary workaround for CLI #710: copy native artifacts that weren't during dotnet publish.
+# This must be in sync with tool-runtime/project.json.
+__PACKAGE_DIR="$HOME/.nuget/packages"
+cp "$__PACKAGE_DIR/Microsoft.Build.Targets/0.1.0-preview-00017/runtimes/any/native/"* "$__TOOLRUNTIME_DIR"
+cp "$__PACKAGE_DIR/Microsoft.Net.Compilers.netcore/1.2.0-beta1-20160122-02/runtimes/any/native/"* "$__TOOLRUNTIME_DIR"
```
joperezr: Linux will be case-sensitive, and I believe the right path would be: Microsoft.Net.Compilers.**N**et**C**ore. That is what the project.lock.json says, but you should make sure that is correct.

dagood: My OSX testing has been that the package is installed with the actual package ID as the path, not what's requested, so it's `Microsoft.Net.Compilers.netcore` for me even though it shows up in the lockfile as `NetCore`. But I don't know if this is the same on each platform, etc.

This seems like another reason to avoid this workaround.

dagood:
@weshaggard I'm not seeing a comment saying that it should work... what I'm observing on my machine is that the natives exist in the package dir and aren't being put into publish targets by the publish command.

I'm leaning more towards double-restore, once globally for publish and again to a controlled dir to copy the natives with more wildcards. The extra restore time and disk space seem worth it for the stability. (The nuget http cache should help here.) Once we get https://github.com/dotnet/cli/issues/261 and can publish from a specified dir we will want to be restoring to a controlled dir anyway.

weshaggard:
OK the only issue with double restore is that we will have another subdirectory of packages correct? That is probably OK to get us unblocked.

cc @ericstj

dagood:
Yes, double restore should just be another package dir, another restore, and a copy for something like `[second restore dir]/*/*/runtimes/any/native/*` -> tool runtime dir. Working on that now.

dagood:
Tested OSX-side using a package built on windows and the new workaround works.

Also tested on Windows and it worked. (`MSBuild.exe`, `Microsoft.CSharp.targets`, etc. are in `Tools`.) Ugly nested `for`s in `init-tools.cmd` because cmd doesn't support `*` in paths.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -26,6 +26,15 @@ cd "%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\"
 call "%DOTNET_CMD%" restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/
 call "%DOTNET_CMD%" publish -f dnxcore50 -r %BUILDTOOLS_TARGET_RUNTIME% -o "%TOOLRUNTIME_DIR%"
 
+:: Temporary workaround for CLI #710: copy native artifacts that weren't during dotnet publish.
+set TOOL_RUNTIME_PACKAGE_DIR=%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\packages
+call "%DOTNET_CMD%" restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/ --packages "%TOOL_RUNTIME_PACKAGE_DIR%"
+for /f %%p in ('dir "%TOOL_RUNTIME_PACKAGE_DIR%" /B /A:D') do (
```
ericstj: This is actually only a problem for the MSBuild and Microsoft.Build.Targets packages, can we constrain the workaround for only those packages?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -26,6 +26,12 @@ cd "%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\"
 call "%DOTNET_CMD%" restore --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/
 call "%DOTNET_CMD%" publish -f dnxcore50 -r %BUILDTOOLS_TARGET_RUNTIME% -o "%TOOLRUNTIME_DIR%"
 
+:: Temporary workaround for CLI #710: copy native artifacts that weren't during dotnet publish.
```
ericstj: Update the issue number to https://github.com/dotnet/cli/issues/1528.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -68,6 +68,13 @@ if [ -n "$BUILDTOOLS_OVERRIDE_RUNTIME" ]; then
     cp -R $BUILDTOOLS_OVERRIDE_RUNTIME/* $__TOOLRUNTIME_DIR
 fi
 
+# Temporary workaround for CLI #710: copy native artifacts that weren't during dotnet publish.
```
ericstj: Same, update to https://github.com/dotnet/cli/issues/1528.

ericstj:
LGTM with the minor update to issue number.

dagood:
Fixed issue number, rebased, made sure it worked on OSX + Windows, merged. Thanks!

PR #465, opened 2016-02-19
Fixes to cloud test targets.

jhendrixMSFT:
Added property OtherRunnerScriptArgs allowing additional runner script
args to be specified at build time.
Replaced hard-coded Windows-specific Helix python root path environment
variable with a dynamic property supporting Windows and Unix variants.
Change assembly list file name to all lower case to avoid issues on case-sensitive
file systems.


jhendrixMSFT:
@MattGal @rahku @karajas

PR #466, opened 2016-02-20
Adding TargetOS to the Container Name

venkat-raman251:
@jhendrixMSFT Please have a look.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
@@ -21,7 +21,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)</ContainerName>
+    <ContainerName Condition="'$(ContainerName)' == ''">$(TestProduct)-$(Branch)-$(BuildMoniker)-$(TargetOS)</ContainerName>
     <ContainerName>$(ContainerName.ToLower())</ContainerName>
     <FuncTestListFilename>FuncTests.$(OSGroup).$(Platform)$(ConfigurationGroup).json</FuncTestListFilename>
     <PerfTestListFilename>PerfTests.$(OSGroup).$(Platform)$(ConfigurationGroup).json</PerfTestListFilename>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -21,7 +21,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)</ContainerName>
+    <TargetOS>$(TargetQueue.Replace('.', '-'))</TargetOS>
+    <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)-$(TargetOS)</ContainerName>
```
jhendrixMSFT: Could you please add a condition to ContainerName that allows it to be overridden (see RunnerScript for an example)?

venkat-raman251:
@jhendrixMSFT : Done.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -21,7 +21,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)</ContainerName>
+    <TargetOS>$(TargetQueue.Replace('.', '-'))</TargetOS>
```
jhendrixMSFT: This depends on the target queue having the same name as the target OS, which while true today might not be in the future.  It also adds drags in the full name of the queue which isn't really relevant (e.g. Windows.PerfInfraTest).  If we don't want to impose people putting the OS name in the BuildMoniker property I would rather add a dedicated TargetOS property to be passed explicitly.

venkat-raman251: Yep, agreed.

venkat-raman251:
@jhendrixMSFT : Have just added TargetOS as a parameter that must be passed to the build

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -21,7 +21,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <ContainerName>$(TestProduct)-$(Branch)-$(BuildMoniker)</ContainerName>
+    <ContainerName Condition="'$(ContainerName)' == ''">$(TestProduct)-$(Branch)-$(BuildMoniker)-$(TargetOS)</ContainerName>
```
jhendrixMSFT: You will also need to update target VerifyInputs to verify that the property was provided.

jhendrixMSFT:
LGTM

PR #470, opened 2016-02-22
Revert "Change Debian.8.2 RID to Debian.8 to match the RID returned b…

weshaggard:
Revert change https://github.com/dotnet/coreclr/blob/master/src/classlibnative/bcltype/decimal.cpp#L115.

@chcosta Thinking about this more I think we jumped the gun on this change because the tool-runtime is isolated from the rest of the repo restore. In fact we should only make this change when we update the packages in tool-runtime\project.json itself which I don't want to do yet because I want to keep those packages at RC2 for now.


chcosta:
@weshaggard, should we revert https://github.com/dotnet/corefx/pull/6221 as well until we're ready to update the packages?

PR #472, opened 2016-02-22
More build info to identify official builds

dsgouda:
Added meta info so that we can identify the build for which we ran tests, and the user who triggered the test

@brianrob


brianrob:
LGTM

PR #475, opened 2016-02-23
Specify specific version of S.R.Extensions

ellismg:
In CI, the OS X builds are failing intermittently because restoring the tool
runtime is pulling down an old version of the runtime assemblies for
System.Runtime.Extensions. This version expects some exports from
System.Native which have been renamed.

Setting the correct version of System.Runtime.Extensions seems to
address this issue (I tried restoring the tool-runtime from this
project.json a bunch of times and each time the right thing happened,
whereas without this fix it only worked about a quarter of the time).

Obviously there is some larger issue at play here as I would not expect
this to be needed, but I would really like to unblock our CI runs so
here's a pragmatic change which should do that.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
@@ -20,11 +20,12 @@
     "Newtonsoft.Json": "7.0.1",
     "NuGet.Versioning": "3.4.0-beta-488",
     "NuGet.Client": "3.4.0-beta-488",
-    "System.IO.Compression": "4.1.0-rc2-23816",
-    "System.Dynamic.Runtime": "4.0.11-rc2-23816",
-    "System.Linq.Expressions": "4.0.11-rc2-23816",
     "System.Diagnostics.TextWriterTraceListener": "4.0.0-rc2-23816",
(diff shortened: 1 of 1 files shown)


ellismg:
/cc @dagood @joperezr @ericstj @weshaggard 

This appears to address the issue in my limited testing.  I have a build machine removed from the pool if folks want to investigate this more to understand the real root cause, but for now I really want to unblock the CI builds.

weshaggard:
LGTM - @ericstj can you help us understand why this would fail only sometimes and not others. Is the lock file produced in parallel on multiple threads?

ericstj:
This was a bug in nuget that was fixed.  What happened was they would merge in the runtime.json from the eclipsed package.  I'm not sure the of the exact details that would cause this to be non-deterministic, but I believe it has to do with how long it takes to download the dependencies before they can open up the package and see the lower versioned dependency.  If they've already determined that the package will be eclipsed they won't download it and won't go down that path that was merging in the old runtime.json.
https://github.com/NuGet/Home/issues/1418

ericstj:
The reason this fixes it is that it promotes S.R.E to a root dependency where previous the references where cousin dependencies (subject to version resolution).  When a package is root it will always win over indirects.  This fix shouldn't be required once everyone is on the latest nuget codebase.

PR #479, opened 2016-02-23
Remove workaround for dotnet publish not copying native artifacts

dagood:
Reverses https://github.com/dotnet/buildtools/pull/463 because https://github.com/dotnet/cli/issues/1528 has been fixed. Native artifacts like `MSBuild.exe` and `Microsoft.CSharp.targets` are now correctly copied by dotnet publish.

(The fix is available starting with CLI `1.0.0.001504`.)

/cc @ericstj @weshaggard @joperezr


davidfowl:
I approve :smile:

dagood:
I'm glad this turned out so temporary, hah! Thanks for the quick fix.

Validated with a built package on OSX and a simple test case on Windows.

ericstj:
LGTM

Issue #413, opened 2016-01-27
Remove tool-runtime project.json and move all of those dependencies into the package declared dependencies.


joperezr:
Currently we restore the runtime for the build tools package by restoring a specific project.json inside the package. What we want to do, is to change this such that the package itself should declare the runtime dependencies and we should remove that project.json.

The steps required for this are:
- [x] In BuildTools repo: Move dependencies from tool-runtime.project.json into the BuildTools package dependencies list. (Available in branch AddBuildToolsPackageDependencies)
- [ ] In Corefx repo: Update the version of BuildTools consumed.
- [ ] In Corefx repo: Change init-tools.sh to publish the runtime of the buildtools dependencies.

am11:
@joperezr, is this issue still up for grabs? I looked a bit, but not sure if it is already resolved in CoreFX. Related to this. are all the `project.json` files going to be converted to `csproj` from CoreFX and CoreCLR repos? For example, the SOS debug extension project (`coreclr/src/ToolBox/SOS/NETCore/project.json`) and some other projects are still using `project.json` files.

joperezr:
hello @am11, this is not yet resolved in CoreFx as we still need to eventually move all of buildtools dependencies into package dependencies, as opposed to having to restore/publish a project.json as we do today. I did mark it as up for grabs, but it is not a trivial issue, as I believe it does require some background in order to complete.

> Related to this. are all the project.json files going to be converted to csproj from CoreFX and CoreCLR repos?

CoreFx has already removed most of its project.json files, and now we are strictly only restoring external dependencies. If you sync to latest in master, you'll see how there are hardly any left.
For CoreClr we haven't fully moved to this model (called 'Build from Source') but just like in corefx, we will still need to keep some project.jsons for things that are built externally

am11:
Thanks for the answer @joperezr!

> some project.jsons for things that are built externally

I was wondering, will those project.json be replaced by the latest SDK/CLI csproj equivalents at some point. For instance: 

```json
  "frameworks": {
    "netstandard1.4": {
      "dependencies": {
        "System.Collections": "4.3.0",
        "System.Collections.Concurrent": "4.3.0",
        "System.Runtime.Serialization.Primitives": "4.3.0"
    }
  }
```

will become:

```xml
<!-- csproj -->
<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' ">
  <PackageReference Include="System.Collections" Version="4.3.0" />
[...]

PR #488, opened 2016-02-25
Moving runtime dependencies to the BuildTools package itself

joperezr:
This are the changes required on the build tools side in order to move all dependencies from the previous tool-runtime into the package itself instead. This will also make buildtools package to follow the nuget package conventions, and will make it really easy to restore since now you can simply add the reference to buildtools in a project.json, and run dotnet restore - dotnet publish to get all the runtime deployed correctly.

A new version of the dotnet cli will be required to consume this, it has been tested with version 1.0.0.001519.

cc: @ericstj @weshaggard


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -1,39 +1,29 @@
...
-set TOOLRUNTIME_DIR=%~3
-IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
+set DOTNET_CMD=%~1
+set TOOLRUNTIME_DIR=%~2
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
-set PORTABLETARGETS_VERSION=0.1.1-dev
 set ROSLYNCOMPILERS_VERSION=1.2.0-beta1-20160202-02
-set MSBUILD_CONTENT_JSON={"dependencies": {"Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%"},"frameworks": {"dnxcore50": {},"net46": {}}}
+set NuProj_Version=0.10.4-beta-gf7fc34e7d8
+set MSBUILD_CONTENT_JSON={"dependencies": { "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%", "NuProj": "%NuProj_Version%" },"frameworks": {"dnxcore50": {},"net46": {}}}
```
joperezr: These two packages cannot be published using dotnet cli because their format is:
```
<PackageRoot>
    -  Tools
         - ......
```
And dotnet cli doesn't support publishing tools folder. To fix this, we would need the packages to follow the contentFiles convention.

joperezr:
Fixes #413

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -1,39 +1,29 @@
 @echo off
 setlocal
 
-set PROJECT_DIR=%~1
-set DOTNET_CMD=%~2
-set TOOLRUNTIME_DIR=%~3
-IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
+set DOTNET_CMD=%~1
```
weshaggard: I would prefer we keep the arugments the same otherwise we will require everyone to update their usage of init-tools.cmd in their repo which I don't think it is worth it just to eliminate PROJECT_DIR which we may end up needing again in the future. Lets just keep letting them pass it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -1,39 +1,29 @@
...
+set TOOLRUNTIME_DIR=%~2
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
-set PORTABLETARGETS_VERSION=0.1.1-dev
 set ROSLYNCOMPILERS_VERSION=1.2.0-beta1-20160202-02
```
weshaggard: What kind of strategy can we use to ensure this version number is in-sync with what is restored from the BuildTools package dependencies?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: ok, I'll add it back then.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: mm we don't really have one really other than monitoring PR's and trying to make sure that both the Net Core compilers and the full framework ones stay in sync. Do you have any specific idea in mind?

weshaggard: Unfortunately I don't have any good ideas here which is why I would really like to eliminate this entire custom publishing step. For now I guess we will just have to live with it.

joperezr: yeah, the only way to control this would be to have the Microsoft.Net.Compilers package to follow the contentFiles convention, and then we could move this dependency out of here, and put it in the nuspec as a net45 dependency, and we would get it restored there. we would still have the two versions on the nuspec, but at least they would be on the same file.

joperezr:
@jhendrixMSFT I've updated my PR on top of your work so just adding you here as an FYI.

@weshaggard I've also addressed your feedback after the rebase.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd.rej
```diff
@@ -0,0 +1,13 @@
+diff a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd	(rejected hunks)
+@@ -6,9 +6,10 @@
+ set TOOLRUNTIME_DIR=%~3
+ IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
+ set BUILDTOOLS_PACKAGE_DIR=%~dp0
++set MICROBUILD_VERSION=0.2.0
+ set PORTABLETARGETS_VERSION=0.1.1-dev
+ set ROSLYNCOMPILERS_VERSION=1.2.0-beta1-20160202-02
+-set MSBUILD_CONTENT_JSON={"dependencies": {"Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%"},"frameworks": {"dnxcore50": {},"net46": {}}}
++set MSBUILD_CONTENT_JSON={"dependencies": { "MicroBuild.Core": "%MICROBUILD_VERSION%", Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%"},"frameworks": {"dnxcore50": {},"net46": {}}}
```
jhendrixMSFT: There is a missing " here, @ericstj checked in a fix for it earlier today so be sure to pick it up.

joperezr: I actually don't understand where this file came from, the real cmd file has that fix, I'll remove this .rej file.

joperezr:
Looks like this is not yet ready to get merged. Restoring the runtime works fine in windows but Linux is not working, I'll take a look at that and update this PR.

ericstj:
@joperezr is this still needed?

joperezr:
I think the idea still remains to have BuildTools as a valid package, it is just not high priority for now which is why I haven't worked on this PR. I have this mostly working except that for some reason I can't get this to work with corefx in Linux. Once I'm done with my current issues, I'll come back to this and finish the work.

weshaggard:
Agreed I still want to do this at some point.

joperezr:
Closing this old PR as #813 takes care of that

PR #489, opened 2016-02-25
Add a GenFacades build task

### LINKED ISSUE #482: GenFacades should have a dedicated MSBuild task
I did a quick experiment to see how much faster a dedicated task would be for GenFacades rather than having to launch an entirely new process. Right now, when building System.Runtime.builds in corefx (which builds 3 facades), my machine takes about 3500 ms on the FillPartialFacade task (from a diagnostic log) out of a total 16000 ms. After switching to a dedicated build task, that number dropped to around 700 ms. This seems like a relatively easy thing to change for a decent speedup. Since we are building so many facades individually now (instead of running GenFacades once in batch), it probably makes sense to make this optimization.

My rough work on this is here: https://github.com/mellinoe/buildtools/commit/e26d50f22dbca41455b2ad2b6524d7c3dfe8b3a0. Note that this could be even faster because I am just passing all of the arguments from MSBuild in a single string and parsing them again in the GenFacades code. Instead, we could just pass them as real task properties because they are already stored in the form we need them in the MSBuild target. That probably represents a not-insignificant speedup, as well.

Ideally GenFacades would be split into three components:
- A shareable library with the actual logic in it, accepting strongly-typed parameters; does no parsing of any kind.
- A Console program which parses command-line arguments and passes them to the shareable library.
- An MSBuild task which accepts pre-processed arguments as task properties and passes them to the shareable library.

@weshaggard

mellinoe:
Fixes #482 

This change moves the core logic of GenFacades into a standalone library called "GenFacades.Core". On top of that are two interfaces, the standard GenFacades.exe, accepting the same arguments as before, and a new MSBuild task which accepts similar parameters and which can be used directly from
MSBuild without the need to launch an entirely new process. This is significantly faster and also avoids the need to re-interpret the command line arguments in the console program, which are already known in the MSBuild target which calls GenFacades.

I'm seeing a pretty substantial speedup for the FillPartialFacades target, it takes around 1/5th the time it did before. Note that all of the "business logic" code in GenFacades is identical to before. The only thing done here is abstract how arguments are passed into GenFacades, and then add two projects on top of it that actually pass the arguments in (a console program and an MSBuild task). GenFacades.exe functions identically to before if we still need to use it for some other parts of the build (like contracts in TFS for the time being).

@weshaggard


src/GenFacades.Core.Desktop/GenFacades.Core.Desktop.csproj
```diff
@@ -0,0 +1,40 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
```
weshaggard: I don't understand the "Core.Desktop" extension. LOL It would ideally be one or the other.

weshaggard: OK I understand the naming convention now but I would like to better understand if we actually need both projects still. Can we get away with just the one?

src/Microsoft.DotNet.Build.Tasks/GenFacadesTask.cs
```diff
@@ -0,0 +1,83 @@
...
+                    contractLoadErrorTreatment = (ErrorTreatment)Enum.Parse(typeof(ErrorTreatment), ContractLoadErrorTreatment);
+                }
+
+                string[] seedTypePreferencesUnsplit = null;
+                if (SeedTypePreferencesUnsplit != null)
+                {
+                    seedTypePreferencesUnsplit = SeedTypePreferencesUnsplit.Select(iti => iti.ItemSpec).ToArray();
+                }
+
+                Generator.GenerateFacades(
```
weshaggard: How are we detecting if errors occur? We should make sure the exceptions do not flow out because that will derail our builds but instead just have some return code that indicated success or failure.

src/GenFacades.Core/Generator.cs
```diff
@@ -0,0 +1,704 @@
...
+                            }
+
+                            OutputFacadeToFile(facadePath, seedHost, facade, contract);
+                        }
+                    }
+                }
+            }
+            catch (FacadeGenerationException ex)
+            {
+                Trace.TraceError(ex.Message);
```
weshaggard: We should return false here so we can pass it along to the build task.

weshaggard:
Overall I like the direction just a few comments in the review.

@venkat-raman251 this is an example of what I was thinking we should do for ApiCompat as well.

venkat-raman251:
@weshaggard : sure, I am going to send out the first review in a few minutes. While that's being reviewed, i will work on making this change.

src/GenFacades.Core.Desktop/GenFacades.Core.Desktop.csproj
mellinoe: I realize now that this is a bit confusing :laughing: . My reasoning for this was: 1) I wanted the console application to still be called "GenFacades.exe" so that we could still drop it in as a replacement for places still using it and 2) I wanted it to describe the fact that it has the "core" logic of GenFacades. I will think of a better name.

mellinoe: Regarding having just one project: we could have GenFacades.exe expose the functionality publically and use that from the build task. It's a bit awkward in my opinion, though, and doesn't feel very clean (the console app part would be unused, the build task calling a method from an .exe...). This feels like the more proper thing to do.

src/Microsoft.DotNet.Build.Tasks/GenFacadesTask.cs
mellinoe: I'm not 100% sure how the exceptions are flowing around here. I assumed that MSBuild had some upper-level exception handling for tasks that it launches. I believe that I did see a couple of failures while I was writing this and the build seemed to handle it in a sensible way (i.e. it detected that it failed and printed out the errors, but didn't completely halt the build). I'll double check that we do the right thing here.

src/GenFacades.Core/Generator.cs
mellinoe: Yes that sounds like a good idea.

src/GenFacades.Core.Desktop/GenFacades.Core.Desktop.csproj
weshaggard: I wasn't implying we use the exe project. I'm OK having the wrapper exe separate from the dll I was more interested if we can eliminate the 2 builds of the dll.

mellinoe: Oh okay, I misunderstood you. I will try to see if we can just have one single GenFacades.Core library and reference it from both. I think it should be possible, since we are doing that for Cci.Extensions which has roughly the same dependencies.

mellinoe:
@weshaggard I addressed the issues you mentioned above, but I hit that blocker on the TFS build. It would be quite a bit more work to split this into separate real separate projects for net45/.NET Core. We don't have a separate project for Cci.Extensions, and I don't think there's a net45 NuGet package for Cci, either, so I'd need to do both of those things first. I'm assuming the error is coming from a reference inside one of those assemblies.

weshaggard:
I'm good with keeping these split for now. I think someone also has a dependency on System.IO.FileSystem which doesn't load inside TFS builds. I added a couple minor comments to your last commit otherwise LGTM

mellinoe:
@weshaggard @venkat-raman251 I ended up adding the fall-back mechanism for GenFacades.exe. I am in the process of testing this in TFS now, and will also test it out on Ubuntu to make sure it works there.

mellinoe:
@weshaggard This works in TFS, but I'm not sure how to set this UseGenFacadesExe property from our TFS build. Do we have any mechanism for setting properties in our wrappers? Properties defined there don't seem to flow through.

weshaggard:
@mellinoe set it in fxcore\dir.props

mellinoe:
Okay that works, i was just trying to add it to the wrong targets files. Thanks!

mellinoe:
@dotnet-bot Test this please

weshaggard:
@mellinoe thinking about this a little more for windows builds I believe we always use the net45 versions of the tools (please verify that) because we are using msbuild 14, so for most of our developers I'm not sure how much this approach will actually help with :(

mellinoe:
Wes brought up an additional issue: we are planning on cross-gen'ing at least parts of the tool runtime, and this will cause some dangling references in some cases. I added a couple more commits that move the GenFacades build task into a competely separate assembly to avoid problems with that, but the workarounds are getting a bit too complicated now, so I'm just going to put this on hold until we can figure out how to solve some of these problems more cleanly.

mellinoe:
This is currently not going to work, due to constaints in our different build systems. Perhaps we can revisit this at a later date when we have a converged build system.

PR #493, opened 2016-02-27
Adding ApiCompat to the build

venkat-raman251:
@weshaggard Please have a look. I still need to  
- Move it into a buildTask
- Look for ApiCompat Baselines in ToF.

This is also currently looking for a project.lock.json in the reference folders and throws a warning if it doesn't find one.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,134 @@
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="LocatePreviousContract" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <PropertyGroup Condition="'$(AssemblyName)' != ''">
```
weshaggard: Do we have any cases where the AssemblyVersion isn't set? That seems like a bug if there are any (well at least for product libs and refs).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,134 @@
...
+    <AssemblyVersion Condition="'$(AssemblyVersion)'==''">4.0.0.0</AssemblyVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <ApiCompatBaseline>$(MSBuildProjectDirectory)\ApiCompatBaseline.txt</ApiCompatBaseline>
+    <ImplementationOutputPath>$(BaseOutputPath)$(OSGroup).$(Platform).$(ConfigurationGroup)\$(AssemblyName)\$(targetGroup)</ImplementationOutputPath>
```
weshaggard: Why not use OutputPath https://github.com/dotnet/corefx/blob/master/dir.props#L386?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,134 @@
...
+    <ImplementationProjectJson Condition="'$(ImplementationProjectJson)'=='' and Exists('$(MSBuildProjectDirectory)/project.json')">$(MSBuildProjectDirectory)/project.json</ImplementationProjectJson>
+    <ImplementationProjectLockJson Condition="Exists('$(ImplementationProjectJson)') and '$(ImplementationProjectLockJson)'==''">$(MSBuildProjectDirectory)/project.lock.json</ImplementationProjectLockJson>
+    <ReferenceAssembly>$(BaseOutputPath)ref\$(AssemblyName)\$(AssemblyVersion)\$(AssemblyName).dll</ReferenceAssembly>
```
weshaggard: Use ReferenceAssemblyOutputPath.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,134 @@
...
+    <ReferenceProjectLockJson Condition="Exists('$(ReferenceProjectJson)') and '$(ReferenceProjectLockJson)'==''">$(ReferenceProjectDirectory)/project.lock.json</ReferenceProjectLockJson>
+  </PropertyGroup>
+  
+  <!-- ApiCompat for Implementation Assemblies  -->
+  <Target Name="ValidateApiCompatForSrc"
+          AfterTargets="Build;CopyFilesToOutputDirectory"
+          DependsOnTargets="GetDependencyDirectories;GetContractDependencies" 
+          Condition="Exists('$(ReferenceProject)')" >
+    
+    <PropertyGroup Condition="$(MSBuildProjectDirectory.EndsWith('src')) AND Exists('$(ReferenceProjectLockJson)')">
```
weshaggard: Can you capture `"$(MSBuildProjectDirectory.EndsWith('src')) AND Exists('$(ReferenceProjectLockJson)'` in another property and use it as part of the condition for the target as a whole.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,134 @@
...
+      <_ApiCompatCmd>"$(ToolsDir)CoreRun.exe" "$(ToolsDir)ApiCompat.exe" "$(ReferenceAssembly)" </_ApiCompatCmd>
+      <_ApiCompatCmd>$(_ApiCompatCmd) -contractDepends:"@(_ContractDependencies);"</_ApiCompatCmd>
+      <_ApiCompatCmd>$(_ApiCompatCmd) -implDirs:"$(ImplementationOutputPath);@(_DependencyDirectories);"</_ApiCompatCmd>
+      <_ApiCompatCmd Condition="Exists('$(ApiCompatBaseline)')">$(_ApiCompatCmd) -baseline:"$(ApiCompatBaseline)"</_ApiCompatCmd>
+    </PropertyGroup>
+    <Exec Condition="$(MSBuildProjectDirectory.EndsWith('src')) AND Exists('$(ReferenceProjectLockJson)')" 
+          Command="$(_ApiCompatCmd) "
+          CustomErrorRegularExpression="^[a-zA-Z]+ :" />
+    <Warning Condition="!Exists('$(ReferenceProjectLockJson)')" Text="The project.lock.json was not found for : $(ReferenceProject)" />
```
weshaggard: Is this working in the right location? It seems like the call to ApiCompat will be completely invalid if this doesn't exist.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,134 @@
...
+    <ItemGroup>
+      <_DependencyDirectories Include="@(Reference->'%(RootDir)%(Directory)')" />      
+    </ItemGroup>
+  </Target>
+  
+  <!-- We build the ref project and then get its dependencies -->
+  <Target Name="GetContractDependencies"
+          Returns="@(_ContractDependencies)"
+          Condition="Exists('$(ReferenceProject)')">
+    <PrereleaseResolveNuGetPackageAssets Condition="Exists('$(ReferenceProjectLockJson)')"
```
weshaggard: This seems like the better place to put the warning. You have above.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,134 @@
...
+    <Warning Condition="!Exists('$(PreviousContractProjectLockJson)')" Text="The project.lock.json was not found for : $(_PreviousContractProjectPath)" />      
+  </Target>
+  
+  <!-- Find the full path to the .csproj file of the previous version of the current contract -->
+  <Target Name="GetPreviousContractProjectPath"
+          Returns="$(_PreviousContractProjectPath);$(_PreviousContractVersion)"
+          >
+    <LocatePreviousContract CurrentContractProjectPath="$(MSBuildProjectFullPath)" AssemblyVersion="$(AssemblyVersion)">
+      <Output TaskParameter="PreviousContractProjectPath" PropertyName="_PreviousContractProjectPath" />
+      <Output TaskParameter="PreviousContractVersion" PropertyName="_PreviousContractVersion" />
```
weshaggard: I don't see you handling this if this comes back empty.

src/Microsoft.DotNet.Build.Tasks/LocatePreviousContract.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                // Appending back the .0 that was initially trimmed to fit src folder name.
+                PreviousContractVersion = PreviousContractVersion + ".0";
+
+                PreviousContractProjectPath = Path.Combine(
+                    directoryPath,
+                    Path.GetFileName(CurrentContractProjectPath));
+                if (!File.Exists(PreviousContractProjectPath))
+                {
+                    PreviousContractProjectPath = Directory.GetFiles(directoryPath,
```
weshaggard: Should we build this path to the project via the version just like we are doing in the targets file for the output ref dll?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
mellinoe: This should always be set

mellinoe:
If you do move this into a build task I suspect you will hit the same issue I hit in #489 when running in our internal TFS build. Some of the dependencies (probably CCI or CCI.Extensions) aren't compatible with net45, which that build uses. The type-forwarder for System.IO.IOException is missing from the System.Runtime facade (added in net46) which gives me a type load exception when I try to run my task.

Let me know what you find when you end up getting that far.

src/Microsoft.DotNet.Build.Tasks/LocatePreviousContract.cs
venkat-raman251: If by convention, all previous contracts are called AssemblyName.depproj, then we can just construct the path to that project similar to the targets. The only reason i wrote this was to find the project regardless of the extension.

weshaggard: I like keeping this task focused on just the one task which is finding the previous version. We can keep the path building and existences checks happening in the targets files.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
venkat-raman251: Removed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
venkat-raman251: Used OutputPath, ReferenceAssemblyOutputPath and removed unused properties.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
venkat-raman251: Reworked the conditions and the warnings appear when the contract projects aren't built( no project.lock.json not found)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
venkat-raman251: Added.

venkat-raman251:
@weshaggard :Could you please have a look. I am still getting a couple of errors which i will fix, but is this structurally what you wanted for ApiCompat?

venkat-raman251:
I have also fixed the first commit based off all the comments.

src/Microsoft.DotNet.Build.Tasks/RunApiCompat.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class RunApiCompat : Task
+    {
+        [Required]
+        public string Arguments { get; set; }
+
+
+        public override bool Execute()
+        {
+            ApiCompatRunner.ValidateApiCompat(Arguments);
```
weshaggard: We should check for non-zero return code and return false if we don't get zero.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -16,10 +16,13 @@
     "Microsoft.NETCore.Platforms": "1.0.1-rc2-23816",
     "Microsoft.NETCore.Runtime": "1.0.1-rc2-23816",
     "Microsoft.Cci": "4.0.0-rc3-23811",
+    "Microsoft.Composition": "1.0.30",
```
weshaggard: The tool-runtime is in the process of being removed.

cc @joperezr

src/ApiCompat/ApiCompatRunner.cs
```diff
@@ -34,11 +34,12 @@ public ExportCciSettings()
...
     }
 
-    public class Program
+    public static class ApiCompatRunner
     {
-        public static int Main(string[] args)
+        public static int ValidateApiCompat(string args)
         {
-            ParseCommandLine(args);
+            string[] argsArray = args.Split(' ');
```
weshaggard: I suspect this will actually break if there are any quoted paths with a space in them.

src/Microsoft.DotNet.Build.Tasks/RunApiCompat.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using ApiCompat;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class RunApiCompat : Task
+    {
+        [Required]
+        public string Arguments { get; set; }
```
weshaggard: I'm not sure that a command line arg string is the best way to pass the parameters. It is probably worth splitting these options into individual properties.

src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
```diff
@@ -55,6 +57,16 @@
     <None Include="project.json" />
     <None Include="PackageFiles\**\*.*" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\ApiCompat\ApiCompat.csproj">
```
weshaggard: Just so it is clear by only adding it here we will only use it when running on .NET Core msbuild which we currently only do for x-plat. We will need to have a solution for the net45/full msbuild runs as well.

I was chatting with @mellinoe about this and we think perhaps having both the task and the exe and toggle between them as necessary. In otherwords when we can use this task library use the task but in places we cannot call the exe.

weshaggard:
@venkat-raman251 have you tried running this on corefx yet? What issues have you found there? At a minimum I expect you will need to port some baseline files.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
joperezr: I'll add this to my open PR and verify that the assets get dropped correctly.

src/Microsoft.DotNet.Build.Tasks/RunApiCompat.cs
venkat-raman251: Yes, I will switch to passing them all in as parameters.

venkat-raman251:
@weshaggard : Yes, I ran it on corefx and got a few errors. But i haven't checked yet whether they have a ApicompatBaseline.txt. That is my next step.

weshaggard:
@venkat-raman251 OK please make sure you can cleanly build all of corefx before this gets merged.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,136 @@
...
+          Command="$(_ApiCompatCmd)"
+          CustomErrorRegularExpression="^[a-zA-Z]+ :" />    
+  </Target>
+
+  <!-- Get the set of directories where the dependencies can be found -->
+  <Target Name="GetDependencyDirectories"
+          Returns="@(_DependencyDirectories)"
+          Condition="Exists('$(ReferenceProjectLockJson)')">
+    <ItemGroup>
+      <_DependencyDirectories Include="@(Reference->'%(RootDir)%(Directory)')" />
```
weshaggard: This should use ReferencePath which is the fully resolved paths used for compiling. We shouldn't be trying to map them from Reference.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,136 @@
...
+      <_PreviousContractProjectPath>$(MSBuildProjectDirectory)/$(_PreviousContractVersion)</_PreviousContractProjectPath>
+      <_PreviousContractProject>$(_PreviousContractProjectPath)/$(AssemblyName).depproj</_PreviousContractProject>
+    </PropertyGroup>
+  </Target>
+
+  <!-- Get the set of directories where the current contract's dependencies can be found -->
+  <Target Name="GetCurrentContractDependencies"
+          Returns="@(_CurrentContractDependencies)">          
+    <ItemGroup>
+      <_CurrentContractDependencies Include="@(Reference->'%(RootDir)%(Directory)')" />
```
weshaggard: Reference is likely wrong here as well. ReferencePath is what I think you want.

venkat-raman251:
@weshaggard : Updated to latest target and fixed a bug in Apicompat last week.

weshaggard:
@venkat-raman251 If you think you have all the necessary changes here please go ahead and get it merged. However before doing that please add a condition that disables ApiCompat by default and when repo's want to enable it they can set it in their root dir.props. 

I'm also still interested in some performance numbers for the overall corefx build with and without running ApiCompat.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,137 @@
...
+    <ReferenceOutputPath>$(ReferenceAssemblyOutputPath)$(AssemblyName)/$(AssemblyVersion)</ReferenceOutputPath>
+    <ReferenceAssembly>$(ReferenceOutputPath)\$(AssemblyName).dll</ReferenceAssembly>
+    <ReferenceProjectJson Condition="'$(ReferenceProjectJson)'=='' and Exists('$(ReferenceProjectDirectory)/project.json')">$(ReferenceProjectDirectory)/project.json</ReferenceProjectJson>
+    <ReferenceProjectLockJson Condition="Exists('$(ReferenceProjectJson)') and '$(ReferenceProjectLockJson)'==''">$(ReferenceProjectDirectory)/project.lock.json</ReferenceProjectLockJson>    
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <RunApiCompatForSrc Condition="$(MSBuildProjectDirectory.EndsWith('src'))">true</RunApiCompatForSrc>
+    <RunApiCompatForRef Condition="$(MSBuildProjectDirectory.EndsWith('ref'))">true</RunApiCompatForRef>
+    <RunApiCompat Condition="'$(RunApiCompat)'==''">true</RunApiCompat>
```
venkat-raman251: @weshaggard : If RunApicompat is set to false, it can be skipped.

venkat-raman251: i also ran builds with and without ApiCompat to gather data. It seems to be adding roughly around a minute to the previous build time on corefx. It takes about 7-8 minutes without apicompat and 8-9 with it. I will give you more exact numbers if you want.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -0,0 +1,137 @@
...
+    <ReferenceOutputPath>$(ReferenceAssemblyOutputPath)$(AssemblyName)/$(AssemblyVersion)</ReferenceOutputPath>
+    <ReferenceAssembly>$(ReferenceOutputPath)\$(AssemblyName).dll</ReferenceAssembly>
+    <ReferenceProjectJson Condition="'$(ReferenceProjectJson)'=='' and Exists('$(ReferenceProjectDirectory)/project.json')">$(ReferenceProjectDirectory)/project.json</ReferenceProjectJson>
+    <ReferenceProjectLockJson Condition="Exists('$(ReferenceProjectJson)') and '$(ReferenceProjectLockJson)'==''">$(ReferenceProjectDirectory)/project.lock.json</ReferenceProjectLockJson>    
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <RunApiCompatForSrc Condition="$(MSBuildProjectDirectory.EndsWith('src'))">true</RunApiCompatForSrc>
+    <RunApiCompatForRef Condition="$(MSBuildProjectDirectory.EndsWith('ref'))">true</RunApiCompatForRef>
+    <RunApiCompat Condition="'$(RunApiCompat)'==''">false</RunApiCompat>
```
venkat-raman251: @weshaggard : Updated to false by default.

venkat-raman251:
Thanks @weshaggard

PR #495, opened 2016-02-28
Xamarin FrameworkList.xml fixes

akoeplinger:
- Fix typo in Xamarin TVOS and WatchOS Frameworklist.xml, they still referred to Xamarin.iOS.
- Fix Redist and Name in the Xamarin FrameworkList.xml. Name is supposed to be the display friendly string, but some of the profiles had that one in Redist instead.

/cc @ericstj


ericstj:
LGTM.  FWIW none of this is matters since all we read is the file section https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/Framework.cs#L72.  The main reason those attributes exist is because they were copied from Xamarin's original targeting pack, but they've been hand modified since then.

akoeplinger:
Yep, I figured that but correcting it felt like the right thing to do nonetheless :smile:

ericstj:
Yeah, thanks for doing so.

PR #496, opened 2016-02-29
Suppress implementation dependencies

ericstj:
Use Exclude:Compile for implementation-only dependencies so that implementation dependencies don't leak to the compiler & folks can't take a dependency on our implementation referencing a particular package.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
@@ -202,7 +202,9 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
                                {
                                    Id = d.ItemSpec,
                                    Version = d.GetVersion(),
-                                   TargetFramework = d.GetTargetFramework()
+                                   TargetFramework = d.GetTargetFramework(),
+                                   Include = d.GetValueList("Include"),
+                                   Exclude = d.GetValueList("Exclude")
                                };
 
             return (from dependency in dependencies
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PromoteReferenceDependencies.cs
```diff
@@ -0,0 +1,99 @@
...
+                    {
+                        // if this implementation assembly fx  has a matching reference fx skip promotion
+                        continue;
+                    }
+
+                    var nearestReferenceFx = FrameworkUtilities.GetNearest(fx, candidateFxs);
+
+                    if (nearestReferenceFx == null)
+                    {
+                        // This should ever happen and indicates a bug in the package.  If a package contains references,
```
weshaggard: ever -> never

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
```diff
@@ -202,7 +202,9 @@ private List<PackageDependencySet> GetDependencySets()
                                {
                                    Id = d.ItemSpec,
                                    Version = d.GetVersion(),
-                                   TargetFramework = d.GetTargetFramework()
+                                   TargetFramework = d.GetTargetFramework(),
+                                   Include = d.GetValueList("Include"),
```
weshaggard: Do we actually use Include anywhere yet or is this just for completion?

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
ericstj: Just for completeness.

ericstj:
fixing typo in comment and rebasing.

PR #500, opened 2016-02-29
ApplyBaseLine should not downgrade dependencies

ericstj:
The ApplyBaseLine task was downgrading dependencies. This was an
oversight, I meant for this task to only lift up the dependencies to a
baseline version but didn't implement it that way.

I added some tests (which I should have done initially) to validate this
behavior.


weshaggard:
LGTM

PR #504, opened 2016-03-02
Clean up required flags

ericstj:
Cleaning up the feature flags I added for nestandard rename, baselining,
and runtime.json exclusion now that all branches set these flags.

To summarize:
1. Repos no longer need to set UseNetPlatform to false in order to build/validate as `NETStandard` instead of `dotnet`.  `NETStandard` is now the default.  Set UseNetPlatform=true to use `dotnet`.
2. Baselining is now on by default.  To opt out of package baselining folks can set BaseLinePackageDependencies=false.
3. Runtime.jsons will be excluded by default from normal packages.  Only lineup packages (IsLineupPackage=true) or packages explicitly opting in (IncludeRuntimeJson=true) will get runtime.jsons.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
@@ -3,8 +3,6 @@
   <PropertyGroup>
     <PackagingTaskDir Condition="'$(PackagingTaskDir)' == ''">$(MSBuildThisFileDirectory)</PackagingTaskDir>
     <GenerationDefinitionFile Condition="'$(GenerationDefinitionFile)' == ''">$(MSBuildThisFileDirectory)Generations.json</GenerationDefinitionFile>
-    <!-- turn off dotnet > netstandard rename until ready -->
-    <UseNetPlatform Condition="'$(UseNetPlatform)' == ''">true</UseNetPlatform>
   </PropertyGroup>
 
   <UsingTask TaskName="ValidatePackageTargetFramework" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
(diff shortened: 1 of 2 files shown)


ericstj:
/cc @weshaggard 
I need this for some of the fat-package preparation work I am doing.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -3,8 +3,6 @@
   <PropertyGroup>
     <PackagingTaskDir Condition="'$(PackagingTaskDir)' == ''">$(MSBuildThisFileDirectory)</PackagingTaskDir>
     <GenerationDefinitionFile Condition="'$(GenerationDefinitionFile)' == ''">$(MSBuildThisFileDirectory)Generations.json</GenerationDefinitionFile>
-    <!-- turn off dotnet > netstandard rename until ready -->
```
weshaggard: Do you think it is too soon to clean this out? I'm OK with it but this change still hasn't flushed through the system fully.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: We can still turn it off, I'm not cleaning up any of the code, just changing the default.  If we need to roll back we just set the inverse property values.

ericstj:
I noticed that one of the changes wasn't working because we had two copies of the GetNuGetPackageDependencies target.

This was the result of a previous commit that merged incorrectly:
https://github.com/dotnet/buildtools/commit/79d09b0b3797ef4162c5f4e34777e9f6c3d37b9e
https://github.com/dotnet/buildtools/commit/127c3d106e2bb121616c97bbb4a5e3b58e236102 <-- duplicated the target

PR #506, opened 2016-03-02
Don't promote reference depencies if inbox

ericstj:
We were already doing this similar filtering in SplitReferences.

The only package impacted by this change is System.Net.WinHttpHandler.


ericstj:
Merging this now because I need a buildtools build with the change.  @weshaggard @chcosta PTAL.

weshaggard:
LGTM

PR #507, opened 2016-03-03
Reading escaped csv strings

dsgouda:
Changes include:
- Update to the xunit.performance.* packages which will generate escaped csv strings in the csv results
- Modifications to the csvtojsonconvertor to read the escaped csv strings properly

@davmason @brianrob


dsgouda:
Need to redo the commit history

PR #509, opened 2016-03-04
Update tool-runtime dependencies from rc2-23816 to rc2-23903

stephentoub:
cc: @weshaggard, @ericstj


weshaggard:
I think there are some issues with build 23903 so you might want to wait. It is worth validating a private version of your BuildTools changes against corefx before we pull this trigger.

weshaggard:
Also it looks like package validation is failing for BuildTools. @dagood we probably should exclude the tool-runtime from validation.

ericstj:
Why should we exclude it?  Isn't the tool-runtime acting as a merged set of dependencies for all the things we build in this branch?  Seems like we should be running on a consistent set of dependencies.

PR #510, opened 2016-03-04
Exclude the PackageFiles project.json files from dependency validation

dagood:
Validation had been set up to keep every project.json in the repository in sync, but tool-runtime and test-runtime need to change independently.

Fixes build error in https://github.com/dotnet/buildtools/pull/509

/cc @stephentoub @weshaggard

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -126,7 +126,7 @@
   <!-- Create a collection of all project.json files for dependency updates. -->
   <ItemGroup>
     <!-- Exclude PackageFiles tool-runtime and test-runtime project.json files. -->
-    <ProjectJsonFiles Include="$(SourceDir)**/project.json"
+    <ProjectJsonFiles Include="$(SourceDir)**\project.json"
                       Exclude="$(SourceDir)Microsoft.DotNet.Build.Tasks\PackageFiles\**\project.json" />
   </ItemGroup>
 
--- a/dir.props
+++ b/dir.props
@@ -125,7 +125,9 @@
 
   <!-- Create a collection of all project.json files for dependency updates. -->
   <ItemGroup>
-    <ProjectJsonFiles Include="$(SourceDir)**/project.json" />
+    <!-- Exclude PackageFiles tool-runtime and test-runtime project.json files. -->
+    <ProjectJsonFiles Include="$(SourceDir)**/project.json"
+                      Exclude="$(SourceDir)Microsoft.DotNet.Build.Tasks\PackageFiles\**\project.json" />
   </ItemGroup>
 
   <PropertyGroup Condition="'$(BuildAllProjects)'=='true'">


dir.props
```diff
@@ -125,7 +125,9 @@
 
   <!-- Create a collection of all project.json files for dependency updates. -->
   <ItemGroup>
-    <ProjectJsonFiles Include="$(SourceDir)**/project.json" />
+    <!-- Exclude PackageFiles tool-runtime and test-runtime project.json files. -->
+    <ProjectJsonFiles Include="$(SourceDir)**/project.json"
+                      Exclude="$(SourceDir)Microsoft.DotNet.Build.Tasks\PackageFiles\**\project.json" />
```
stephentoub: Why do we have ```\``` here but ```/``` in the previous line?

stephentoub:
One nit/question, otherwise LGTM.

dir.props
dagood: Well, independently, `Include` uses `/` because what I saw when I added it were a bunch of properties with `/` like https://github.com/dagood/buildtools/blob/exclude-tool-runtime-validation/dir.props#L37-L50. `Exclude` uses `\` because MSBuild expands the `**` with `\`, and paths with `/` won't exclude anything.

I found https://github.com/dotnet/corefx/pull/1310 that explains why there are so many `/` in the build, but I don't know what has happened since then (if we got a better way to transform paths or not). It seems like any path that might find its way to an xplat `Exec` needs to be `/`, but MSBuild-only stuff uses `\`.

I'm open to changing just this instance of `/` to `\` so the different usages at least aren't in the same item.

weshaggard: @dagood I can definitely see msbuild being picky about but what I don't understand is that I wouldn't expect Include and Exclude to be different. We are doing wildcard in both places and so if it is broken I would expect it to be broken by one or the other here. At any rate just be sure to test this and I'm OK with it. I wouldn't be surprised if this is different on non-windows as well, but for BuildTools in particular we don't build on anything except windows right now.

dagood: What I've been seeing from my tests is that any dir separators generated as part of `**` are being created as `\` by MSBuild, and the `/` in `/project.json` isn't preserved. (Maybe `**/` is conceptually what's being picked up.) It ends up with paths like `E:\git\buildtools\src/Microsoft.DotNet.Build.Tasks.Packaging\src\project.json`.

Making both use `\` for consistency. `ProjectJsonFiles` isn't currently used for any `Exec` calls, so even if buildtools were xplat I think this would be safe.

dagood:
Holding off on this for https://github.com/dotnet/buildtools/pull/509#issuecomment-192356470 (@ericstj's comment)

stephentoub:
> Holding off on this for #509 (comment) (@ericstj's comment)

We need to move msbuild to updated packages.  How/when can we do that?  Our outer loop Unix builds in corefx will continue to fail until that happens (hopefully that's the last issue, but there may be something further lurking behind it).

dagood:
With all project.jsons using a single set of validation rules, upgrading the entire repo's version level is the way to go. That can happen anytime--unless there's some issue using the newer packages with the other projects in the repo that I don't know about.

stephentoub:
Thanks, @dagood. Would you mind doing so, then?

dagood:
Sure. Is 23903 best, or should I go for 23907 (newest) or something else?

stephentoub:
> Sure. Is 23903 best, or should I go for 23907 (newest) or something else?

Thanks.  I'm guessing 23907; I think 23903 had some problems.  I'm about to move corefx to that, as well.

weshaggard:
Yes lets try 23907.

dagood:
Closing in favor of https://github.com/dotnet/buildtools/pull/510. We're keeping all project.jsons in sync across the repo.

PR #515, opened 2016-03-07
[WIP] Upgrade prerelease dependencies to rc2-23907.

dagood:
Update to latest packages based on discussion in https://github.com/dotnet/buildtools/pull/510/.

I'm getting some restore errors with 23907:

	Errors in E:\git\buildtools\src\Microsoft.DotNet.Build.Tasks.Packaging\src.Desktop\project.json
	  System.Runtime.InteropServices.RuntimeInformation 4.0.0-rc2-23907 provides a compile-time reference assembly for System.Runtime.InteropServices.RuntimeInformation on .NETFramework,Version=v4.5, but there is no run-time assembly compatible with win.
	  Some packages are not compatible with .NETFramework,Version=v4.5 (win).

Looking in the package, the runtime.json that I think previously let the restore find the runtime package is no longer there. Not sure if this is a bug or if there's some new way to get to runtimes that I'm not aware of. I can work around this by depending on the win runtime package manually, but that's nasty.

	Errors in E:\git\buildtools\src\Microsoft.DotNet.Build.Tasks\project.json
	  Microsoft.Win32.Registry 4.0.0-rc2-23907 provides a compile-time reference assembly for Microsoft.Win32.Registry on DNXCore,Version=v5.0, but there is no run-time assembly compatible with osx.10.10-x64.
	  Some packages are not compatible with DNXCore,Version=v5.0 (osx.10.10-x64).
	  Microsoft.Win32.Registry 4.0.0-rc2-23907 provides a compile-time reference assembly for Microsoft.Win32.Registry on DNXCore,Version=v5.0, but there is no run-time assembly compatible with ubuntu.14.04-x64.
	  Some packages are not compatible with DNXCore,Version=v5.0 (ubuntu.14.04-x64).
	Errors in E:\git\buildtools\src\Microsoft.DotNet.Build.Tasks\PackageFiles\tool-runtime\project.json
	  Microsoft.Win32.Registry 4.0.0-rc2-23907 provides a compile-time reference assembly for Microsoft.Win32.Registry on DNXCore,Version=v5.0, but there is no run-time assembly compatible with centos.7-x64.
	  Some packages are not compatible with DNXCore,Version=v5.0 (centos.7-x64).
	  Microsoft.Win32.Registry 4.0.0-rc2-23907 provides a compile-time reference assembly for Microsoft.Win32.Registry on DNXCore,Version=v5.0, but there is no run-time assembly compatible with debian.8.2-x64.
[...]

PR #516, opened 2016-03-07
Allow dependency value to be object with version property instead of only a version string

dagood:
This allows a restore compatibility error workaround needed for https://github.com/dotnet/buildtools/pull/515.

Verified basic use cases manually: updated a version to another version, validated, and upgraded versions that were strings as well as inside objects.

/cc @ericstj


ericstj:
LGTM

PR #519, opened 2016-03-08
Add infrastructure to support IsNotSupportedAssembly 

ericstj:
This adds infrastructure to buildtools to create NotSupported assemblies.

These assemblies expose API described by the contract and throw PlatformNotSupported from every method.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+
+  <PropertyGroup Condition="'$(GeneratePlatformNotSupportedAssembly)' == 'true'">
+    <!-- Tell ResolveMatchingContract to run and resolve contract to project reference -->
+    <ResolveMatchingContract>true</ResolveMatchingContract>
+    <NotSupportedSourceFile>$(IntermediateOutputPath)$(TargetName).notsupported.cs</NotSupportedSourceFile>
+    <CoreCompileDependsOn>$(CoreCompileDependsOn);GenerateNotSupportedSource</CoreCompileDependsOn>
+  </PropertyGroup>
+
+  <!-- GenerateNotSupportedSource
+       Inputs:
+         * A contract assembly
+         * Reference assemblies
+
+       Generates source for the contract that throws PlatformNotSupportedException
+  -->
+  <Target Name="GenerateNotSupportedSource" 
+          DependsOnTargets="ResolveMatchingContract"
+          Inputs="@(ReferencePath);@(ResolvedMatchingContract)"
+          Outputs="$(NotSupportedSourceFile)">
+
+    <ItemGroup>
+      <!-- build out a list of directories where dependencies are located -->
+      <_referencePathDirectoriesWithDuplicates Include="@(ReferencePath->'%(RootDir)%(Directory)'->TrimEnd('\'))" />
+      <!-- strip metadata, removing duplicates -->
+      <_referencePathDirectories Include="%(_referencePathDirectoriesWithDuplicates.Identity)" />
+    </ItemGroup>
+
+    <Error Text="No single matching contract found." Condition="'@(ResolvedMatchingContract->Count())' != '1'" />
+
+    <PropertyGroup>
+      <GenAPIArgs>"%(ResolvedMatchingContract.Identity)"</GenAPIArgs>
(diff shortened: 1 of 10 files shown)


ericstj:
/cc @weshaggard @KrzysztofCwalina @mellinoe 
I'd like your feedback on this.

stephentoub:
@ericstj, just curious, what's our plan for this?  Are we going to start building Unix "not supported" assemblies for every Windows-specific assembly, for example?

ericstj:
Right now I did a mockup of a single "compat" package that would let folks bring in all of the not-supported assemblies.  https://github.com/ericstj/corefx/commit/b569f96ede171e412b245d58f2c2192b2797a54e

I want the default behavior to alert folks at dev time that they are using a package that doesn't work on a platform.
Today there is no way for folks to unblock themselves if they are OK using unsupported API (eg: a porting effort, MSBuild for example still uses registry).

mellinoe:
The code changes look good to me.

> I want the default behavior to alert folks at dev time that they are using a package that doesn't work on a platform.

Don't they already get an error if they try to restore packages for the specific unsupported runtime (or shouldn't it work that way) ? Although it's better than getting a `FileNotFoundException` at runtime, it's still a pretty roundabout way of just saying "No, this isn't gonna work on that platform". Are we expecting that we would ship these packages into the standard library such that they would already be available (to throw) in the general case?

stephentoub:
I see.  Thanks for the explanation.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
```
weshaggard: nit: extra line

src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
```diff
@@ -0,0 +1,52 @@
...
+
+  <PropertyGroup Condition="'$(IsNotSupportedAssembly)' == 'true'">
```
weshaggard: Can we call this property "AssemblyThrowsPlatformNotSupportedForAllAPIs"? Or something a little more detailed explaining what this actually implies.

weshaggard:
Code changes look reasonable. However I want to have more conversations before we use it for anything.

ericstj:
> Don't they already get an error if they try to restore packages for the specific unsupported runtime (or shouldn't it work that way) ?

Yes but there is no way to workaround that today, the error will block restore.  For upstack consumers of a library they can use `"exclude": "compile"` like we did for MSBuild [here](https://github.com/dotnet/buildtools/commit/00d07dccff6fd1c71c6f42b02a3e3f2c03b7411f#diff-0a4e59b10a58cc85410daf0411ce94c8R8), but there is no way for someone who actually wants static dependencies to code against the not supported API.  This is providing a way for folks to still restore and build once they get that error.  By putting it in a separate package on the side folks have to explicitly opt into getting the NotSupported behavior.

weshaggard:
@ericstj we can get into this conversation more but I would still like to push folks to cross-compile in those instances.

ericstj:
@weshaggard good place for such a conversation would be here: https://github.com/dotnet/corefx/issues/6743

src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
ericstj: IsDummyAssembly, IsThrowAssembly, IsMockAssembly, IsTurdAssembly?

bartonjs: GenerateAsStubAssembly?  GenerateAsNotSupportedAssembly?  GenerateAsPNSEAssembly?

I mean, Turd is poetic and all, but still... :smile:

weshaggard: GeneratePlatformNotSupportedAssembly=true

stephentoub:
FWIW, I used to be in the camp of thinking we should fail a compile-time, specifically because we want to help folks really think through when they're taking a dependency on something that's platform-specific and they may not realize it.  But light-up at run-time is a valid approach that we're effectively blocking with the current scheme; even if we want to push build-per-platform as the desired approach doesn't mean we should block the other.  As long as there's a mechanism for someone to be warned at a compile-time about what they're doing, I don't see a problem with letting them override/suppress that warning and allowing them to use a version of the assembly that'll just throw PlatformNotSupportedException from everything if it's actually used.

PR #520, opened 2016-03-08
Update baseline

ericstj:
/cc @bartonjs @saurabh500


ericstj:
ping can someone please have a look today?  This needs to go in tonight's build.

saurabh500:
@ericstj what is the role of this file.
I am curious about the implications of this change. 
The changes for Sql Client look good based on the version change.

ericstj:
The purpose of this is to ensure that package dependencies we emit have a consistent version.  Prior to this change a package building in CoreFx that referenced SQL client would emit a dependency to a package version 4.0.0-... which wouldn't exist.  For SqlClient this is a noop because no other libraries in CoreFx/WCF actually depend on it.  It's more important for Crypto since we have dependencies in corefx today that are creating invalid package refs.

saurabh500:
Thanks. For System.Data LGTM

PR #521, opened 2016-03-08
Fix dependency versions after automatic prerelease upgrade

dagood:
The auto-upgrade in https://github.com/dotnet/buildtools/pull/515 resulted in some dependencies not existing due to the minor/patch versions changing. Corrected throughout the repository.

/cc @ericstj @stephentoub


ericstj:
LGTM.

PR #522, opened 2016-03-09
Fixes to support real-signed builds in CoreFx.

jhendrixMSFT:
By default skip signing when building depproj projects.
Renamed property UseOpenSourceSign to SignType to better align with real
signing requirements.
Disable Authenticode signing when doing OSS signing.
For real-signed builds that don't use the "open" key add build metadata so
the assembly gets a real strong name.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets
@@ -2,10 +2,10 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <!-- metadata used for Authenticode and strong-name signing in official VSO builds -->
-  <ItemGroup Condition="'$(IsTestProject)' != 'true'">
+  <ItemGroup Condition="'$(SkipSigning)' != 'true' and '$(IsTestProject)' != 'true' and '$(SignType)' != 'oss'">
     <FilesToSign Include="$(TargetPath)">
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
@weshaggard @joperezr can you guys take a look.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets
```diff
@@ -34,7 +35,7 @@
   </PropertyGroup>
 
   <Target Name="OpenSourceSign"
-          Condition="'$(DelaySign)' == 'true' and '@(IntermediateAssembly)' != '' and '$(SkipSigning)' != 'true' and '$(UseOpenSourceSign)' == 'true'"
+          Condition="'$(DelaySign)' == 'true' and '@(IntermediateAssembly)' != '' and '$(SkipSigning)' != 'true' and '$(SignType)' == 'oss'"
```
joperezr: Can we keep the old argument as well as a condition? The reason is that the changes I made in coreclr depend on this target not getting executed depending on this specific property getting passed in.

joperezr: I've moved coreclr to the latest buildtools version (193) and it looks like everything builds correctly, so then there is no need to address my previous comment. Once you merge this and we have a new buildtools package, I'll use that from coreclr.

joperezr:
LGTM

weshaggard:
LGTM

Issue #523, opened 2016-03-09
Errors in restore for init-tools scripts


joperezr:
When using the new version of the dotnet cli (which in fact is the one used by our repos) the init-tools.log show the following errors:
```
error: Microsoft.DotNet.BuildTools 1.0.25-prerelease-00193 is not compatible with DNXCore,Version=v5.0.
error: Some packages are not compatible with DNXCore,Version=v5.0.
error: Microsoft.DotNet.BuildTools 1.0.25-prerelease-00193 is not compatible with DNXCore,Version=v5.0 (win7-x64).
error: Some packages are not compatible with DNXCore,Version=v5.0 (win7-x64).
error: Microsoft.DotNet.BuildTools 1.0.25-prerelease-00193 is not compatible with DNXCore,Version=v5.0 (win7-x86).
error: Some packages are not compatible with DNXCore,Version=v5.0 (win7-x86).
info : Committing restore...
log  : Restore failed in 5912ms.
Errors in C:\Users\joperezr\Desktop\repo\coreclr\Tools\1.0.25-prerelease-00193\project.json
    Microsoft.DotNet.BuildTools 1.0.25-prerelease-00193 is not compatible with DNXCore,Version=v5.0.
    Some packages are not compatible with DNXCore,Version=v5.0.
    Microsoft.DotNet.BuildTools 1.0.25-prerelease-00193 is not compatible with DNXCore,Version=v5.0 (win7-x64).
    Some packages are not compatible with DNXCore,Version=v5.0 (win7-x64).
    Microsoft.DotNet.BuildTools 1.0.25-prerelease-00193 is not compatible with DNXCore,Version=v5.0 (win7-x86).
    Some packages are not compatible with DNXCore,Version=v5.0 (win7-x86).
```

For now, it doesn't really have an impact since restore is being done despite the errors, but we should address that by making BuildTools package dnxcore compliant.

cc: @weshaggard @dagood

dagood:
Would this be as simple as adding `netstandard1.0\_._` or `dnxcore50\_._` or something as a placeholder? They work, but I don't know if we wanted something other than a placeholder.

/cc @ericstj

ericstj:
Yeah, just add a placeholder to make nuget happy is a easy fix.

Alternatively if we can make this thing a "real" package (eg a dotnet CLI-friendly library or app) it would just work.

dagood:
Closing this because the immediate issue has a merged fix.

[...]

PR #526, opened 2016-03-09
Add placeholder netstandard1.0 compat to package to avoid restore errors

dagood:
Stops `restore` from complaining about buildtools not being compatible. Workaround for https://github.com/dotnet/buildtools/issues/523 until we make buildtools a "real" package.

Tested with dotnet CLI 1.0.0.001504 win.

/cc @joperezr @ericstj


joperezr:
I know dnxcore50 was replaced already, but since this is the one we use in [the runtime project.json](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json#L37) and [the init tools command](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd#L28) should we just add a place holder for dnxcore50 as well?

dagood:
I'm not sure when it would be useful (restoring with old versions of nuget/dnx that don't know about netstandard?) but it won't hurt, so why not. Added!

joperezr:
LGTM

PR #527, opened 2016-03-10
Support type.member parsing in ConditionalFact and ConditionalTheory

Priya91:
cc @stephentoub @ellismg 

EDIT: Add support to conditionalfact, to search for type.member in assembly

### DIFF
--- a/src/xunit.netcore.extensions/Discoverers/ConditionalFactDiscoverer.cs
+++ b/src/xunit.netcore.extensions/Discoverers/ConditionalFactDiscoverer.cs
@@ -26,9 +26,22 @@ namespace Xunit.NetCore.Extensions
             MethodInfo testMethodInfo = testMethod.Method.ToRuntimeMethod();
 
             string conditionMemberName = factAttribute.GetConstructorArguments().FirstOrDefault() as string;
+            Type declaringType = testMethodInfo.DeclaringType;
+            string[] symbols = conditionMemberName.Split('.');
+
+            if (symbols.Length == 2)
+            {
+                conditionMemberName = symbols[1];
+                ITypeInfo type = testMethod.TestClass.Class.Assembly.GetTypes(false).Where(t => t.Name.Contains(symbols[0])).FirstOrDefault();
+                if (type != null)
+                {
+                    declaringType = type.ToRuntimeType();
+                }
+            }
+            
             MethodInfo conditionMethodInfo;
             if (conditionMemberName == null ||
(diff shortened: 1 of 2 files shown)


stephentoub:
@Priya91, thanks for working on this... I'm not sure this really belongs on ConditionalFact/Theory, though.  Why not build it into PlatformSpecific instead?  For example:
```C#
[PlatformSpecific(PlatformID.Windows, "WindowsNano")]
[PlatformSpecific(PlatformID.Linux, "Ubuntu14.04")]
...
```
?

If it's better / easier at least in the short term to use ConditionalFact/Theory, that's fine, but I don't think we should modify the attribute for it.  I'd prefer we just have a shared file in Common\tests in corefx that contains the logic for checking whether we're on Windows Nano, and then you can use that in any tests that need it.  If you wanted to extend ConditionalFact/Theory in some way, you could do so by allowing it to use a member of another class rather than just the current class, and that way it'd be easier for you to use your common logic, e.g.
```C#
[ConditionalFact(nameof(NanoDetection) + "." + nameof(IsNanoServer))]
```

Priya91:
@stephentoub I initially worked on getting it included in PlatformSpecific, but it's a trait attribute, which means we need to add new categories to xunit commandline, making it not that user friendly. This will also lead to build script changes to support WithCategories/WithoutCategories, etc, Another option was to make PlatformSpecific derive from FactAttribute, like the ConditionalFact, so we can do the filtering at Fact level itself, but then, the consumers of xunit.netcore.extensions currently have [Fact][PlatformSpecific] on tests, and there can be only one fact on the test methods, this approach will result in cleaning up all the repos (not just corefx) to remove [Fact] whereever [PlatformSpecific] is used. This approach also didnt seem feasible.

And yes the other option was to use the existing string constructor itself, with type.member, and then modify the logic in ConditionalFact to search for the member in the assembly. I can either have this nano logic in corefx and share it, or i can have the nano logic in xunit.netcore.extensions, and have a special string for Nano, and skip the reflection overhead for this platform sku, say,
```
[ConditionalFact("IsWindowsNano")] 
```

Or the other option is to have this logic in corefx and share it as linked file, as you mentioned.

Priya91:
Synced with @stephentoub offline, and we decided it is best to have the windows nano filtering in specific repos, and out of buildtools for now.

Priya91:
@stephentoub Made the changes to enable type.member parsing in xunit.netcore.extensions.

Priya91:
@stephentoub Ping..

stephentoub:
LGTM. We can always make it smarter in the future as necessary.

PR #528, opened 2016-03-10
Fix signing of build tools binaries.

jhendrixMSFT:
In commit eca81d0 I changed how property SkipSigning is evaluated so that
when it's true no signing is performed.  Unfortunately we set this
property in most of the build tools projects and as a result they are no
longer Authenticode signed.  I have removed the SkipSigning property from
the offending projects.
Set the default strong name signing key to the "open" one.
The PublishFilesToRuntime task was running before real signing happens, as
a result we do not sign the renamed binary (e.g. GenAPI.exe ->
GenAPI.dll).  To fix this I sequenced this target after the SignFiles
target which is where the official signing happens.  To ensure this works
for dev builds I stubbed out this target in sign.targets.  It also
required me to re-order the import of the MicroBuild bootstrapper in
Build.Common.targets.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -148,10 +148,10 @@
     <DebugSymbols>false</DebugSymbols>
     <DebugType>none</DebugType>
     <!--
-      Delay signing with the ECMA key currently doesn't work.
-      https://github.com/dotnet/roslyn/issues/2444
+      Full signing with Open key doesn't work with Portable Csc.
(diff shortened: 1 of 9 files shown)


jhendrixMSFT:
@weshaggard @ericstj Can you guys please take a look.

weshaggard:
LGTM

Issue #487, opened 2016-02-25
Enable crossgen'ing of the tool-runtime


weshaggard:
Now that our tools, including Roslyn, are running on top of .NET Core we need a version of the tool-runtime that is crossgen'ed. Especially for the exe's that are started many times like csc and genfacades.

PR #529, opened 2016-03-11
Adding Crossgen to the toolRuntime folder

joperezr:
Even though I don't like this approach, we need crossgen to be dropped in the toolruntime folder as well if we want to crossgen the msbuild to make it build faster.

related: #487 

cc: @weshaggard


ericstj:
Weren't we going to put crossgen in a separate package?

joperezr:
When I was doing this work, the decision to have a new package was not made yet, so I started working through this approach. After @weshaggard talked to Gaurav, he asked me to stop this work since the new package was probably comming. I didn't close this to keep the work in case that the package wasn't created, but if you want me to I can close it and keep the work in a private branch in my fork.

joperezr:
Closing this old PR since we are not doing this anymore

PR #530, opened 2016-03-11
Fix dependency projection

ericstj:
Fixes https://github.com/dotnet/corefx/issues/6813

The CreateTrimDependencyGroups task was incorrectly projecting
implementation dependencies to dependency groups that had a different
implementation assembly.

For example:
	NETStandard1.0 implementation dll
	NET45 facade dll

We were copying the portable implementation's dependencies into the
facade's dependency group.

This was happening because the task wasn't considering all existing
dependency goups.  Instead it was just considering the placeholder
groups "specially" via the TrimFrameworks parameter.

I removed this parameter and refactored the algorithm a bit to
consider any dependency group, which includes the placeholders
as well as any dependency group created by harvesting the references.

This fixed the projection problem, but now we no longer were getting
compile references copied.  We need to preserve compile references
because they are part of the package's compat contract.  If someone
builds a portable library against one of our packages they may depend on
the packages transitive dependencies rather than directly referencing
those.

For example:
 - System.Data.SqlClient's reference assembly depends on S.Data.Common
 - Neither S.Data.SqlClient nor S.Data.Common are inbox on desktop.
 - S.Data.SqlClient's desktop implementation does not depend on
S.Data.Common since it is a facade that only depends on S.Data and
mscorlib
 - A PCL might reference just S.Data.SqlClient but will have assembly
refs to both S.Data.SqlClient and S.Data.Common.
 - We must deploy the S.Data.Common facade on desktop.

To fix this problem I made sure that PromoteReferenceDependencies always
copies the references, even for empty dependency groups (as would be the
case for a facade).  My previous change to add this task simply
overlooked the fact that an empty dependency group needed to be treated
as a supported framework to promote to as opposed to one to skip over.

/cc @weshaggard @chcosta

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/CreateTrimDependencyGroups.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/CreateTrimDependencyGroups.cs
@@ -27,12 +27,6 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
             get;
             set;
         }
-        /* These are frameworks with placeholders (inbox, external, not supported) */
-        public string[] TrimFrameworks
-        {
-            get;
-            set;
-        }
 
         [Required]
         public ITaskItem[] Files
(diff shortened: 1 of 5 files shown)


weshaggard:
LGTM

PR #533, opened 2016-03-15
NuGet update to 3.4 and update baseline

ericstj:
The latest NuGet 3.4 maps UWP to netstandard1.4 so we no longer need to build pre-versioned packages specifically for UWP.  Instead we'll build a single set of packages and keep UWP at 1.4, repackaging the RTM bits inside the new packages.

NuGet 3.4 now has pack support I can remove our fork.

This requires https://github.com/ericstj/corefx/tree/UWP-netstandard1.4 to updated CoreFx.


src/Microsoft.DotNet.Build.Tasks.Packaging/test.Desktop/Microsoft.DotNet.Build.Tasks.Packaging.Tests.Desktop.nuget.props
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
```
weshaggard: We should not commit these nuget.props files.

weshaggard:
Yeah... I love seeing a bunch of code deleted. 

LGTM

src/Microsoft.DotNet.Build.Tasks.Packaging/test.Desktop/Microsoft.DotNet.Build.Tasks.Packaging.Tests.Desktop.nuget.props
ericstj: Good catch.

ericstj:
@weshaggard what are your thoughts on the timing of this, including the corefx part?

weshaggard:
@ericstj is there any reason to not get them in ASAP? I would like to have these in in before the end of week.

ericstj:
Potential compat risk due to decrementing assembly version.  Only an issue between different builds of prerelease.

weshaggard:
If we are going to do it I would much prefer we do it for RC2 as opposed to after.

ericstj:
Agreed.  Ok, let's pull the trigger on BuildTools and I'll get the CoreFx PR ready.  /cc @joperezr FYI we need to line this one up with changes in CoreFx.

PR #534, opened 2016-03-15
Move signing to a post-build step so it batch signs.

jhendrixMSFT:
The current signing implementation signs files one by one which is
incredibly slow.  We can get better throughput by submitting all files for
signing in one item group.
If a file requires signing it will create a ".requires_signing" marker
file that will be read during a post build step; this post build step will
process all of the marker files and submit the signing job in one group.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -2,8 +2,14 @@
 <Project ToolsVersion="12.0" DefaultTargets="BuildAndTest" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="dir.props" />
 
+  <!-- required to build the projects in their specified order -->
+  <PropertyGroup>
+    <SerializeProjects>true</SerializeProjects>
+  </PropertyGroup>
+
   <ItemGroup>
     <Project Include="src\dirs.proj" />
+    <Project Include="src\sign.builds" />
   </ItemGroup>
 
   <Import Project="dir.targets" />
--- a/publishexe.targets
+++ b/publishexe.targets
@@ -9,8 +9,7 @@
     <BaseNuGetRuntimeIdentifier>win7</BaseNuGetRuntimeIdentifier>
   </PropertyGroup>
 
-  <!-- this must happen after the binaries have been signed -->
-  <Target Name="PublishFilesToRuntime" AfterTargets="SignFiles"
+  <Target Name="PublishFilesToRuntime" AfterTargets="CopyFilesToOutputDirectory"
           Condition="'$(OutputType)' == 'Exe' ">
 
     <ItemGroup>
@@ -38,6 +37,10 @@
   <Target Name="FixUpCoreConsoleNames">
     <Move SourceFiles="$(Outdir)$(AssemblyName).exe" DestinationFiles="$(Outdir)$(AssemblyName).dll" />
     <Move SourceFiles="$(Outdir)CoreConsole.exe" DestinationFiles="$(Outdir)$(AssemblyName).exe" />
+    <!-- update TargetPath with the renamed file -->
+    <PropertyGroup>
+      <TargetPath>$(Outdir)$(AssemblyName).dll</TargetPath>
+    </PropertyGroup>
   </Target>
 
 </Project>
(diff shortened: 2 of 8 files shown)


jhendrixMSFT:
@weshaggard this is the change to support batch signing.

build.proj
```diff
@@ -2,8 +2,14 @@
...
   <Import Project="dir.props" />
 
+  <!-- required to build the projects in their specified order -->
+  <PropertyGroup>
+    <SerializeProjects>true</SerializeProjects>
+  </PropertyGroup>
+
   <ItemGroup>
     <Project Include="src\dirs.proj" />
+    <Project Include="src\sign.builds" />
```
weshaggard: Should this be part of the post.builds.

jhendrixMSFT: I thought about that however I think for signing we want to be explicit where in the build process it happens, i.e. it must happen after building sources and before packaging.

src/sign.builds
```diff
@@ -0,0 +1,39 @@
...
+    </ItemGroup>
+    <ReadSigningRequired MarkerFiles="@(SignMarkerFile)">
+      <Output TaskParameter="SigningMetadata" ItemName="FilesToSign" />
+    </ReadSigningRequired>
+  </Target>
+
+  <Target Name="Build"
+          Condition="'$(SkipSigning)' != 'true' and '$(SignType)' != 'oss'"
+          DependsOnTargets="GetFilesToSignItems">
+    <CallTarget Targets="SignFiles" />
```
weshaggard: Where is the SignFiles target?

publishexe.targets
```diff
@@ -38,6 +37,11 @@
   <Target Name="FixUpCoreConsoleNames">
     <Move SourceFiles="$(Outdir)$(AssemblyName).exe" DestinationFiles="$(Outdir)$(AssemblyName).dll" />
     <Move SourceFiles="$(Outdir)CoreConsole.exe" DestinationFiles="$(Outdir)$(AssemblyName).exe" />
+    <!-- create a marker indicating that the file was renamed, this will be used during signing to sign the correct file -->
+    <Touch Files="$(Outdir)$(AssemblyName).publishedexe" AlwaysCreate="true" />
+    <ItemGroup>
```
weshaggard: Why do these need to be special? Can we not just put the correct marker file and include the actual file extension as part of the marker file name? So to get the right file signed we just rename the signing file as well.

src/sign.builds
jhendrixMSFT: Oops I deleted it from sign.targets, I'll add it back to satisfy the case when the MicroBuild signing package isn't installed.

publishexe.targets
jhendrixMSFT: The marker file is renamed in GetFilesToSignItems in sign.builds (e.g. GenAPI.exe -> GetAPI.dll).  We cannot rename it in FixUpCoreConsoleNames as the marker hasn't been written yet.  We could do the renaming here (avoiding this other marker) however we would have to ensure that this target is always executed after the marker file has been written.

weshaggard: OK perhaps we can update TargetPath or setup another property that could we could use for writing the signing file. I would prefer that we don't need to special cases things like this directly in GetFilesToSignItems and instead just have the extensibility point via some property into the target that creates the marker.

src/sign.builds
```diff
@@ -0,0 +1,39 @@
...
+
+  <PropertyGroup>
+    <OutDir>$(BinDir)</OutDir>
+  </PropertyGroup>
+
+  <UsingTask AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" TaskName="ReadSigningRequired" />
+
+  <!-- populates item group FilesToSign with the list of files to sign -->
+  <Target Name="GetFilesToSignItems">
+    <!-- find all of the "published" exe files -->
```
weshaggard: Yes ideally this target wouldn't need to do anything but look for requires_signing files.

publishexe.targets
jhendrixMSFT: Good idea I'll give this a whirl.

src/sign.builds
jhendrixMSFT: This has been removed.

publishexe.targets
jhendrixMSFT: I've removed this marker and updated the TargetPath property (I swear I tried this earlier and it didn't work, glad it's working now as it's much simpler).

weshaggard:
LGTM

PR #541, opened 2016-03-15
Split package building into its own step.

jhendrixMSFT:
Builds packages happens when building src\dirs.proj; this doesn't work
with the recent switch to batch signing as the binaries are now signed
after the packages are created.
Move the package build step into its own project packages.builds file to
be built after signing.

### DIFF
--- a/src/packages.builds
+++ b/src/packages.builds
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="BuildAndTest" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
@weshaggard @joperezr PTAL

src/packages.builds
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="BuildAndTest" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+
+  <!-- packages.targets requires build values -->
+  <Import Project="BuildValues.props" />
+  <Import Project="Microsoft.DotNet.Build.Tasks\PackageFiles\packages.targets" Condition="'$(ImportGetNuGetPackageVersions)' != 'false'" />
```
weshaggard: Do we need ImportGetNugetPackageVersion condition any longer?

weshaggard:
LGTM

PR #542, opened 2016-03-16
Add Versioning control for Repos.

joperezr:
CC: @weshaggard ~~The PR is not ready to be merged yet, but I wanted to send this out to you today so that if you have some time to take a look I can get early feedback. I plan to have this ready by tomorrow.~~ <- PR is ready to merge after feedback now.

FYI: @ericstj ~~We are still considering on adding another optional target that will generate a props file in the root of the repo, so that this doesn't need to be calculated over and over, which I know was more or less what you wanted.~~ <- I've changed the behavior now such that we will create a BuildVersion.props file which will contain the Build Version.

I'll send out a [sample corefx PR](https://github.com/dotnet/corefx/pull/6982) soon on how to consume this.

CC: @gkhanna79 
related: dotnet/coreclr#3559 dotnet/coreclr#3133

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
@@ -1,0 +1,103 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Globalization;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public sealed class GenerateCurrentVersion : Task
+    {
+        /// <summary>
+        /// The passed in date that will be used to generate a version. (yyyy-MM-dd format)
+        /// </summary>
+        [Required]
+        public string SeedDate { get; set; }
+
+        /// <summary>
+        /// Optional parameter that sets the Padding for the version number. Must be 5 or bigger.
+        /// </summary>
+        public int Padding { get; set; }
+
+        /// <summary>
+        /// If basing off of internal builds version format is not required, this optional parameter lets you pass in a comparison date.
+        /// </summary>
+        public string ComparisonDate { get; set; }
+
+        /// <summary>
+        /// The Major Version that will be produced given a SeedDate.
+        /// </summary>
+        [Output]
+        public string GeneratedVersion { get; set; }
+
+        private const string DateFormat = "yyyy-MM-dd";
+        private const string LastModifiedTimeDateFormat = "yyyy-MM-dd HH:mm:ss.FFFFFFF";
+        private CultureInfo enUS = new CultureInfo("en-US");
+
+        public override bool Execute()
+        {
+            if (Padding == 0)
+            {
+                Padding = 5;
+            }
+            else if (Padding < 5)
+            {
+                Log.LogWarning("The specified Padding '{0}' has to be equal to or greater than 5. Using 5 as a default now.", Padding);
+                Padding = 5;
+            }
+            DateTime date;
+            GeneratedVersion = string.Empty;
+            if (!(DateTime.TryParseExact(SeedDate, DateFormat, enUS, DateTimeStyles.AssumeLocal, out date)))
+            {
+                // Check if the timestamp matches the LastModifiedTimeDateFormat
+                if (!(DateTime.TryParseExact(SeedDate, LastModifiedTimeDateFormat, enUS, DateTimeStyles.AssumeLocal, out date)))
+                {
+                    Log.LogError("The seed date '{0}' is not valid. Please specify a date in the short format.({1})", SeedDate, DateFormat);
+                    return false;
+                }
+            }
+            //Convert Date to UTC to converge
+            date = date.ToUniversalTime();
+            GeneratedVersion = GetCurrentVersionForDate(date, ComparisonDate);
+            if (string.IsNullOrEmpty(GeneratedVersion))
+            {
+                Log.LogError("The date '{0}' is not valid. Please pass in a date after {1}.", SeedDate, ComparisonDate);
+                return false;
+            }
+            return true;
+        }
+
+        public string GetCurrentVersionForDate(DateTime seedDate, string comparisonDate)
+        {
+            DateTime compareDate;
+            if (string.IsNullOrEmpty(comparisonDate))
+            {
+                /*
+                 * We need to ensure that our build numbers are higher that what we used to ship internal builds so this date
+                 * will make that possible.
+                 */
+                compareDate = new DateTime(1996, 4, 1, 0, 0, 0, DateTimeKind.Utc);
+            }
+            else
+            {
+                bool isValidDate = DateTime.TryParseExact(comparisonDate, DateFormat, enUS, DateTimeStyles.AssumeLocal, out compareDate);
+                if (!isValidDate)
+                {
+                    Log.LogError("The comparison date '{0}' is not valid. Please specify a date in the short format.({1})", comparisonDate, DateFormat);
+                }
+                //Convert to UTC to converge
+                compareDate = compareDate.ToUniversalTime();
+            }
+            int months = (seedDate.Year - compareDate.Year) * 12 + seedDate.Month - compareDate.Month;
+            if (months > 0) //only allow dates after comparedate
+            {
+                return string.Format("{0}{1}", months.ToString("D" + (Padding -2)), seedDate.Day.ToString("D2"));
+            }
+            return string.Empty;
+        }
+
+    }
+}
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
@@ -1,0 +1,16 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  
+  <PropertyGroup>
+    <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
+    <BuildVersionFilePath>$(BaseIntermediateOutputPath)</BuildVersionFilePath>
+    <BuildVersionFile Condition="'$(BuildVersionFile)'==''">$(BuildVersionFilePath)BuildVersion-$(TodayTimeStamp).props</BuildVersionFile>
+  </PropertyGroup>
+
+  <!-- If BuildVersion.props exists already then import it to get BuildNumberMajor, else generate it and override props values. -->
+  <Import Condition="Exists('$(BuildVersionFile)')" Project="$(BuildVersionFile)" />
+
+  <PropertyGroup Condition="!Exists('$(BuildVersionFile)')">
+    <ShouldCreateVersionFileDuringBuild>true</ShouldCreateVersionFileDuringBuild>
+  </PropertyGroup>
+
+</Project>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -1,13 +1,15 @@
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <!-- Setup the default file version information -->
+  <UsingTask TaskName="GenerateCurrentVersion" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
+
+  <!-- Setup the default file version information -->
   <PropertyGroup>
     <MajorVersion Condition="'$(MajorVersion)' == ''">1</MajorVersion>
     <MinorVersion Condition="'$(MinorVersion)' == ''">0</MinorVersion>
-    
+
     <!-- These should be set by importing the targets below but initializing to 0 for consistency -->
     <BuildNumberMajor Condition="'$(BuildNumberMajor)' == ''">0</BuildNumberMajor>
-    <BuildNumberMinor Condition="'$(BuildNumberMinor)' == ''">0</BuildNumberMinor>
+    <BuildNumberMinor Condition="'$(BuildNumberMinor)' == ''">00</BuildNumberMinor>
   </PropertyGroup>
 
   <!-- Import a build target that includes the build numbers -->
@@ -21,7 +23,7 @@
     <CLSCompliant Condition="'$(CLSCompliant)'==''">false</CLSCompliant>
     <AssemblyFileVersion Condition="'$(AssemblyFileVersion)'==''">$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)</AssemblyFileVersion>
   </PropertyGroup>
-  
+
   <PropertyGroup>
     <GenerateAssemblyInfo Condition="'$(GenerateAssemblyInfo)'==''">true</GenerateAssemblyInfo>
     <BuiltByString Condition="'$(BuiltByString)'==''">%20built by: $(COMPUTERNAME)-$(USERNAME)</BuiltByString>
@@ -40,7 +42,8 @@
   <Target Name="GenerateAssemblyInfo"
       Inputs="$(MSBuildProjectFile)"
       Outputs="$(AssemblyInfoFile)"
-      Condition="'$(GenerateAssemblyInfo)'=='true'">
+      Condition="'$(GenerateAssemblyInfo)'=='true'"
+      DependsOnTargets="CreateVersionFileDuringBuild">
 
     <Error Condition="!Exists('$(IntermediateOutputPath)')" Text="GenerateAssemblyInfo failed because IntermediateOutputPath isn't set to a valid directory" />
 
@@ -58,7 +61,7 @@
       <AssemblyInfoLines Include="[assembly:AssemblyInformationalVersion(@&quot;$(AssemblyFileVersion)$(BuiltByString)&quot;)]" />
       <AssemblyInfoLines Condition="'$(CLSCompliant)'=='true'" Include="[assembly:CLSCompliant(true)]" />
       <AssemblyInfoLines Condition="'$(AssemblyComVisible)'!=''" Include="[assembly:System.Runtime.InteropServices.ComVisible($(AssemblyComVisible))]" />
-      <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'" 
+      <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'"
         Include="[assembly:System.Reflection.AssemblyMetadata(&quot;%(AssemblyMetadata.Identity)&quot;, &quot;%(AssemblyMetadata.Value)&quot;)]" />
     </ItemGroup>
 
@@ -76,7 +79,7 @@
       <AssemblyInfoLines Include="&lt;Assembly:AssemblyInformationalVersion(&quot;$(AssemblyFileVersion)$(BuiltByString)&quot;)&gt;" />
       <AssemblyInfoLines Condition="'$(CLSCompliant)'=='true'" Include="&lt;Assembly:CLSCompliant(True)&gt;" />
       <AssemblyInfoLines Condition="'$(AssemblyComVisible)'!=''" Include="&lt;Assembly:System.Runtime.InteropServices.ComVisible($(AssemblyComVisible))&gt;" />
-      <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'" 
+      <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'"
           Include="&lt;assembly:System.Reflection.AssemblyMetadata(&quot;%(AssemblyMetadata.Identity)&quot;, &quot;%(AssemblyMetadata.Value)&quot;)&gt;" />
     </ItemGroup>
 
@@ -95,10 +98,10 @@
       Lines="@(AssemblyInfoUsings);@(AssemblyInfoLines)"
       Overwrite="true" />
 
-    <ItemGroup>  
-      <Compile Include="$(AssemblyInfoFile)" />  
-      <FileWrites Include="$(AssemblyInfoFile)" />  
-    </ItemGroup>  
+    <ItemGroup>
+      <Compile Include="$(AssemblyInfoFile)" />
+      <FileWrites Include="$(AssemblyInfoFile)" />
+    </ItemGroup>
 
   </Target>
 
@@ -125,7 +128,7 @@
       Inputs="$(MSBuildProjectFile)"
       Outputs="$(NativeVersionHeaderFile)"
       Condition="'$(NativeVersionHeaderFile)'!='' and '$(GenerateVersionHeader)'=='true'">
-      
+
     <ItemGroup>
       <NativeVersionLines Include="#define VER_COMPANYNAME_STR         &quot;Microsoft Corporation&quot;" />
       <NativeVersionLines Include="#define VER_FILEDESCRIPTION_STR     &quot;$(AssemblyName)&quot;" />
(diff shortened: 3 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                Log.LogError("The date '{0}' is not valid. Please pass in a date after April 1996.", CurrentDate);
+                return false;
+            }
+            return true;
+        }
+
+        public static string GetCurrentVersionForDate(DateTime date)
+        {
+            DateTime compareDate = new DateTime(1996,4,1);
```
joperezr: Using this Date, we will Match the internal build version numbering.

weshaggard:
Cc @jhendrix

joperezr:
Ping! :) 
Can somebody take a look please? Ideally we want to check this in today to start moving repos into using it as soon as possible.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// The Major Version that will be used for this build.
+        /// </summary>
+        [Output]
+        public string ResultCurrentVersion { get; set; }
+
+        public override bool Execute()
+        {
+            DateTime date;
+            ResultCurrentVersion = string.Empty;
+            bool isValidDate = DateTime.TryParse(CurrentDate, out date);
```
weshaggard: If we are planning to do parsing here we should specific an exact date pattern given we are round tripping through a string.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public sealed class GenerateCurrentVersion : Task
+    {
+        /// <summary>
+        /// The path to the file that contains the date of the last comit in EPOCH time.
+        /// </summary>
+        [Required]
+        public string CurrentDate { get; set; }
```
weshaggard: CurrentDate seems to be a little miss-leading. How about something like SeedDate or something like that.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// The Major Version that will be used for this build.
+        /// </summary>
+        [Output]
+        public string ResultCurrentVersion { get; set; }
```
weshaggard: I think I would call this GeneratedVersion

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
weshaggard: You should add a comment in the code. Something along the lines of "We need to ensure our build numbers are higher then what we used to ship from our internal builds so we are using a date to align with that number". I also think this should be an optional parameter which we can default to this as I imagine there will be folks using this that don't have that constraint and will want to start the with a lower date.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,60 @@
...
+                return false;
+            }
+            return true;
+        }
+
+        public static string GetCurrentVersionForDate(DateTime date)
+        {
+            DateTime compareDate = new DateTime(1996,4,1);
+            int months = (date.Year - compareDate.Year) * 12 + date.Month - compareDate.Month;
+            if (months > 0) //only allow dates after comparedate
```
weshaggard: We should make sure this is padded, to at least 5 '0's.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
```diff
@@ -0,0 +1,17 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  
+  <PropertyGroup>
+    <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
```
weshaggard: Doesn't there need to be quotes around "yyyMMdd"?

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
joperezr: Makes sense, I'll add the comment and make this settable.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
```diff
@@ -0,0 +1,17 @@
...
+    <BuildVersionFile Condition="'$(BuildVersionFile)'==''">$(BaseIntermediateOutputPath)/$(TodayTimeStamp)/BuildVersion.props</BuildVersionFile>
+  </PropertyGroup>
+
+  <!-- If BuildValues.props exists already then import it to get BuildNumberMajor, else generate it and override props values. -->
+  <Import Condition="Exists('$(BuildVersionFile)')" Project="$(BuildVersionFile)" />
+
+  <PropertyGroup Condition="!Exists('$(BuildVersionFile)')">
+    <GenerateAssemblyInfoDependsOn>CreateVersionFileDuringBuild;$(GenerateAssemblyInfoDependsOn)</GenerateAssemblyInfoDependsOn>
+    <GetNuGetPackageDependenciesDependsOn>CreateVersionFileDuringBuild;$(GetNuGetPackageDependenciesDependsOn)</GetNuGetPackageDependenciesDependsOn>
```
weshaggard: We should avoid mutating these dependsons from various other targets files here. In stead we should provide a common target they can depend on to get the build numbers and we do different things in that target based one the different modes we have.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
ericstj: :+1:

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
ericstj: :+1:

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Xml;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public sealed class GenerateCurrentVersion : Task
+    {
+        /// <summary>
+        /// The path to the file that contains the date of the last comit in EPOCH time.
```
bartonjs: "commit"

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
joperezr: No this is not required in msbuild.

joperezr:
@weshaggard thanks for the feedback! I've addresed it with my latest commit, let me know if you see anything else.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,96 @@
...
+        private CultureInfo enUS = new CultureInfo("en-US");
+        private int padding;
+
+        public override bool Execute()
+        {
+            if (string.IsNullOrEmpty(Padding))
+            {
+                Padding = "5";
+            }
+            bool isValidPadding = int.TryParse(Padding, out padding);
```
weshaggard: You can actually just make the property an int because msbuild supports that type.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    Log.LogError("The comparison date '{0}' is not valid. Please specify a date in the short format.(YYYY-MM-DD)", comparisonDate);
+                }
+            }
+            int months = (seedDate.Year - compareDate.Year) * 12 + seedDate.Month - compareDate.Month;
+            if (months > 0) //only allow dates after comparedate
+            {
+                return string.Format("{0}{1}", months.ToString("D" + (padding-2)), seedDate.Day.ToString("D2"));
+            }
+            return string.Empty;
```
weshaggard: Instead of empty here I think I would prefer we just return a padded set of zeros.

weshaggard: Actually I just noticed you error on empty above so you can ignore this comment. I'm fine with it being an error.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,96 @@
...
+            bool isValidDate = DateTime.TryParseExact(SeedDate, DateFormat, enUS, DateTimeStyles.AssumeLocal, out date);
+            if (!isValidDate)
+            {
+                Log.LogError("The seed date '{0}' is not valid. Please specify a date in the short format.(YYYY-MM-DD)", SeedDate);
+                return false;
+            }
+            GeneratedVersion = GetCurrentVersionForDate(date, ComparisonDate);
+            if (string.IsNullOrEmpty(GeneratedVersion))
+            {
+                Log.LogError("The date '{0}' is not valid. Please pass in a date after April 1996.", SeedDate);
```
weshaggard: "April 1996" -> ComparisonDate

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
 
+  <!-- #################################### -->
+  <!-- Generate BuildVersion.props -->
+  <!-- #################################### -->
+  <Target Name="CreateOrUpdateCurrentVersionFile" Condition="'$(SkipVersionGeneration)'!='true'" DependsOnTargets="GenerateCurrentVersion">
+    <Error Condition="'$(OverrideBuildNumberMajor)'==''" Text="Error when generating current version." />
+
+    <ItemGroup>
+      <CurrentVersionLines Include="%3C%3Fxml version=%221.0%22 encoding=%22utf-8%22%3F%3E" />
+      <CurrentVersionLines Include="%3C!-- This is a generated file. To re-generate it, run build.cmd --%3E" />
```
weshaggard: I think I would exclude the "regenerate it" step because it might be different across repos.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
```diff
@@ -0,0 +1,15 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  
+  <PropertyGroup>
+    <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
+    <BuildVersionFile Condition="'$(BuildVersionFile)'==''">$(BaseIntermediateOutputPath)/$(TodayTimeStamp)/BuildVersion.props</BuildVersionFile>
```
weshaggard: Why does the TodayTimeStamp need to be a directory? Does that make it easier in some way? As best I can tell it might be simpiler to just have it as part of the filename.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+      <CurrentVersionLines Include="%3C/Project%3E" />
+      <BuildVersionFileItem Include="$(BuildVersionFile)" />
+    </ItemGroup>
+
+    <PropertyGroup>
+      <BuildVersionSemaphoreFile>%(BuildVersionFileItem.RootDir)%(BuildVersionFileItem.Directory)semaphore.txt</BuildVersionSemaphoreFile>
+      <ShouldLockFileWrite Condition="!Exists('$(BuildVersionSemaphoreFile)')">true</ShouldLockFileWrite>
+    </PropertyGroup>
+
+    <MakeDir Condition="!Exists('%(BuildVersionFileItem.RootDir)%(BuildVersionFileItem.Directory)')" Directories="%(BuildVersionFileItem.RootDir)%(BuildVersionFileItem.Directory)" />
```
weshaggard: See my other comment I think if we put the date in the file name we don't need to worry about making all these directories.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
     </ItemGroup>
   </Target>
 
+  <!-- #################################### -->
+  <!-- Generate BuildVersion.props -->
+  <!-- #################################### -->
+  <Target Name="CreateOrUpdateCurrentVersionFile" Condition="'$(SkipVersionGeneration)'!='true'" DependsOnTargets="GenerateCurrentVersion">
+    <Error Condition="'$(OverrideBuildNumberMajor)'==''" Text="Error when generating current version." />
```
weshaggard: Why is this an error? Is it because we already did it? It might be worth explaining why this is en error.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+      Lines="@(CurrentVersionLines)"
+      Overwrite="true" />
+
+  </Target>
+
+  <Target Name="GenerateCurrentVersion" DependsOnTargets="GetLatestCommitDate">
+    <Error Condition="'$(DateToUse)'==''" Text="A date is needed to Generate the current version." />
+
+    <GenerateCurrentVersion SeedDate="$(DateToUse)">
+      <Output PropertyName="OverrideBuildNumberMajor" TaskParameter="GeneratedVersion" />
```
weshaggard: when reading OverrideBuildNumberMajor I didn't at all pick-up that it was the one you are generating until just now. I think it might be best to call it GeneratedBuildNumberMajor.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+      ContinueOnError="WarnAndContinue"
+      Condition="'$(ShouldLockFileWrite)'=='true'"
+      File="$(BuildVersionFile)"
+      Lines="@(CurrentVersionLines)"
+      Overwrite="true" />
+
+  </Target>
+
+  <Target Name="GenerateCurrentVersion" DependsOnTargets="GetLatestCommitDate">
+    <Error Condition="'$(DateToUse)'==''" Text="A date is needed to Generate the current version." />
```
weshaggard: We should rename DateToUse to be a little clearer. How about VersionSeedDate.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+  </Target>
+
+  <Target Name="GetLatestCommitDate" Condition="'$(DateToUse)'==''">
+    <Exec Command="git show -s --format=^%25%25cd --date=short HEAD" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="DateToUse" />
+    </Exec>
+  </Target>
+
+  <Target Name="GetTargetsFileLastModifiedDate" Condition="'$(DateToUse)'==''" AfterTargets="GetLatestCommitDate">
+    <ItemGroup>
```
weshaggard: Why not just do this in the same Target above? Just put a condition on on the item/property group. There isn't any real reason to have two targets here. That way we can also give it a more general name like "GetVersionSeedDate" and people could override it if they wanted to compute it in a different way.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+  <!-- #################################### -->
+  <Target Name="CreateOrUpdateCurrentVersionFile" Condition="'$(SkipVersionGeneration)'!='true'" DependsOnTargets="GenerateCurrentVersion">
+    <Error Condition="'$(OverrideBuildNumberMajor)'==''" Text="Error when generating current version." />
+
+    <ItemGroup>
+      <CurrentVersionLines Include="%3C%3Fxml version=%221.0%22 encoding=%22utf-8%22%3F%3E" />
+      <CurrentVersionLines Include="%3C!-- This is a generated file. To re-generate it, run build.cmd --%3E" />
+      <CurrentVersionLines Include="%3CProject xmlns=%22http://schemas.microsoft.com/developer/msbuild/2003%22%3E" />
+      <CurrentVersionLines Include="%3CPropertyGroup%3E" />
+      <CurrentVersionLines Include="%3CBuildNumberMajor Condition=%22%27%24(BuildNumberMajor)%27==%27%27%22%3E$(OverrideBuildNumberMajor)%3C/BuildNumberMajor%3E" />
```
weshaggard: Do we expect that this file will always be imported above versioning.targets? Otherwise BuildNumberMajor might be 0 all the time because we set it in there.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+      <CurrentVersionLines Include="%3C/PropertyGroup%3E" />
+      <CurrentVersionLines Include="%3C/Project%3E" />
+      <BuildVersionFileItem Include="$(BuildVersionFile)" />
+    </ItemGroup>
+
+    <PropertyGroup>
+      <BuildVersionSemaphoreFile>%(BuildVersionFileItem.RootDir)%(BuildVersionFileItem.Directory)semaphore.txt</BuildVersionSemaphoreFile>
```
weshaggard: In this case I don't think you actually even need a sempahore file, as the buildversion.props can be the sempahore file.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
```diff
@@ -0,0 +1,15 @@
...
+  </PropertyGroup>
+
+  <!-- If BuildValues.props exists already then import it to get BuildNumberMajor, else generate it and override props values. -->
+  <Import Condition="Exists('$(BuildVersionFile)')" Project="$(BuildVersionFile)" />
+
+  <PropertyGroup Condition="!Exists('$(BuildVersionFile)')">
```
weshaggard: This isn't a very good way to decide whether or not to generate this file or not. As you will always get trying to do it in all the initial set of projects you build. Instead we should have inputs and outputs on the target that generates the file itself.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
weshaggard: Actually if you do switch to inputs/outputs then you will need a semaphore file but I'm not sure there is much value in doing having it here. I think I would just do the exists check on the target condition itself for the buildversion.props file and if it doesn't exist execute the target. Just make sure the WriteLinesToFile correctly handles errors. I will tell you WarnAndContinue will cause build warnings from time to time and we will need to explain. I wonder if just doing nothing is better and simply check for existence of the file after the write and if it doesn't exist give an error.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+    <Touch AlwaysCreate="true" Files="$(BuildVersionSemaphoreFile)" Condition="'$(ShouldLockFileWrite)'=='true'" />
+
+    <WriteLinesToFile
+      ContinueOnError="WarnAndContinue"
+      Condition="'$(ShouldLockFileWrite)'=='true'"
+      File="$(BuildVersionFile)"
+      Lines="@(CurrentVersionLines)"
+      Overwrite="true" />
+
+  </Target>
```
weshaggard: It is probably worth deleting all other BuildVersion.target files here that aren't the one you just wrote. That will help keep this clean.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+  <Target Name="GenerateCurrentVersion" DependsOnTargets="GetLatestCommitDate">
+    <Error Condition="'$(DateToUse)'==''" Text="A date is needed to Generate the current version." />
+
+    <GenerateCurrentVersion SeedDate="$(DateToUse)">
+      <Output PropertyName="OverrideBuildNumberMajor" TaskParameter="GeneratedVersion" />
+    </GenerateCurrentVersion>
+  </Target>
+
+  <Target Name="GetLatestCommitDate" Condition="'$(DateToUse)'==''">
+    <Exec Command="git show -s --format=^%25%25cd --date=short HEAD" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
```
weshaggard: `^%25%25cd` correct me if I'm wrong but that is the escaped version of `^%%cd` what is the `^` and extra `%` for?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,75 @@
...
+      <DateToUse>%(VersionTargetsFile.ModifiedTime)</DateToUse>
+    </PropertyGroup>
+  </Target>
+
+  <!-- This target will only be executed if BuildValues.props doesn't exist yet -->
+  <Target Name="CreateVersionFileDuringBuild" Condition="'$(SkipVersionGeneration)'!='true' AND '$(ShouldCreateVersionFileDuringBuild)'=='true'" DependsOnTargets="CreateOrUpdateCurrentVersionFile">
+    <PropertyGroup>
+      <BuildNumberMajor>$(OverrideBuildNumberMajor)</BuildNumberMajor>
+      <AssemblyFileVersion>$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)</AssemblyFileVersion>
+      <VersionSuffix Condition="'$(PreReleaseLabel)' != ''">-$(PreReleaseLabel)</VersionSuffix>
```
weshaggard: Is this one needed?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
```diff
@@ -0,0 +1,15 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
weshaggard: Where do you expect this file to be imported? Is there any reason to not do this work directly in version.targets?

weshaggard: Or move the build number generation targets into this file as well?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
joperezr: Yes, BuildValues.targets is one of the first files to load which sets the correct version, and in the case where i hasn't been written yet, then we call the CreateVersionFileDuringBuild target which will override the defaults for buildnumbermajor

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
joperezr: We need this because of DOS escaping, the reason is that the exec task secretly runs everything in a shell instead of directly calling createprocess, so an extra escape is required. I still have to test this in Mac to make sure that this also works there.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
joperezr: Yes, We basically need to reconstruct the versionsuffix since by the time that this target executes, the VersionSuffix is already defined, so we need to override it with the new BuildNumberMajor we have calculated.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
joperezr: > Is there any reason to not do this work directly in version.targets

Yes, We want this file to be one of the first imports in dir.props on the root of each repo, and by this it would set the right buildnumbermajor before other target files are loaded. That is why I have that as a separate file now.

> Or move the build number generation targets into this file as well?

I can do this if you prefer, the reason why I have it as it is now, is because I thought that everything that actually calculates what the version is should go under versioning.targets, whereas the actual file that needs to be imported first to calculate all of this needs to be separate to be loaded before other properties are set.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
weshaggard: Yes I understand the appending the BuildNumberMajor but I was asking about the part that just puts the PreReleaseLabel. Either way it isn't a big deal just something that caught my eye.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -158,65 +158,74 @@
...
       ContinueOnError="WarnAndContinue"
-      Condition="'$(ShouldLockFileWrite)'=='true'"
+      Condition="!Exists('$(BuildVersionFile)')"
       File="$(BuildVersionFile)"
       Lines="@(CurrentVersionLines)"
       Overwrite="true" />
+    
+    <!-- Delete old BuildVersion.props files -->
+    <ItemGroup>
+      <OldBuildVersionFiles Include="$(BuildVersionFilePath)*.props" Exclude="$(BuildVersionFilePath)%(BuildVersionFileItem.Filename).props" />
```
weshaggard: To be safe you should use `$(BuildVersionFilePath)BuildVersions-*.props`. Also thinking about this a minute perhaps doing the delete before you write the new file would eliminate the need to exclude it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -158,65 +158,74 @@
...
   <!-- #################################### -->
   <Target Name="CreateOrUpdateCurrentVersionFile" Condition="'$(SkipVersionGeneration)'!='true'" DependsOnTargets="GenerateCurrentVersion">
-    <Error Condition="'$(OverrideBuildNumberMajor)'==''" Text="Error when generating current version." />
+    <!-- Error out if GeneratedBuildNumberMajor is empty since GenerateCurrentVersion should have calculated it already. -->
+    <Error Condition="'$(GeneratedBuildNumberMajor)'==''" Text="Error when generating current version." />
 
     <ItemGroup>
       <CurrentVersionLines Include="%3C%3Fxml version=%221.0%22 encoding=%22utf-8%22%3F%3E" />
-      <CurrentVersionLines Include="%3C!-- This is a generated file. To re-generate it, run build.cmd --%3E" />
+      <CurrentVersionLines Include="%3C!-- This is a generated file. --%3E" />
```
weshaggard: I guess you can remove the `--%3E` as well.

weshaggard: nevermind that is the xml end comment marker :)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -158,65 +158,74 @@
...
       <CurrentVersionLines Include="%3C/PropertyGroup%3E" />
       <CurrentVersionLines Include="%3C/Project%3E" />
       <BuildVersionFileItem Include="$(BuildVersionFile)" />
     </ItemGroup>
-
+    
     <PropertyGroup>
-      <BuildVersionSemaphoreFile>%(BuildVersionFileItem.RootDir)%(BuildVersionFileItem.Directory)semaphore.txt</BuildVersionSemaphoreFile>
-      <ShouldLockFileWrite Condition="!Exists('$(BuildVersionSemaphoreFile)')">true</ShouldLockFileWrite>
+      <BuildVersionFilePath>%(BuildVersionFileItem.RootDir)%(BuildVersionFileItem.Directory)</BuildVersionFilePath>
```
weshaggard: I would define this in the BuildVersion.txt and use it there as well. No need to use metadata traversal for this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,84 @@
...
+    <!-- Windows Exec command requires DOS escaping for the percent sign since it secretly runs the process in a shell instead of calling createprocess. -->
+    <PropertyGroup>
+      <GitCommand Condition="'$(OsEnvironment)'=='Windows_NT'">git show -s --format=^%25%25cd --date=short HEAD</GitCommand>
+      <GitCommand Condition="'$(OsEnvironment)'!='Windows_NT'">git show -s --format=%25cd --date=short HEAD</GitCommand>
+    </PropertyGroup>
+      
+    <Exec Command="$(GitCommand)" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="VersionSeedDate" />
+    </Exec>
+    <PropertyGroup Condition="'$(VersionSeedDate)'==''">
```
weshaggard: Add another property that says where we got the SeedDate so we can add it in the comment for the generated file to aide in debugging.

joperezr:
@weshaggard I added the Build Minor Version with: 8270399a403e271268e34658a8439534363a5987 Could you PTAL?

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,103 @@
...
+                Padding = 5;
+            }
+            DateTime date;
+            GeneratedVersion = string.Empty;
+            if (!(DateTime.TryParseExact(SeedDate, DateFormat, enUS, DateTimeStyles.AssumeLocal, out date)))
+            {
+                // Check if the timestamp matches the LastModifiedTimeDateFormat
+                if (!(DateTime.TryParseExact(SeedDate, LastModifiedTimeDateFormat, enUS, DateTimeStyles.AssumeLocal, out date)))
+                {
+                    Log.LogError("The seed date '{0}' is not valid. Please specify a date in the short format.(yyyy-MM-dd)", SeedDate);
```
weshaggard: To prevent future errors. I would change the error string to use DateFormat instead of copying the value.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,103 @@
...
+                 * will make that possible.
+                 */
+                compareDate = new DateTime(1996, 4, 1, 0, 0, 0, DateTimeKind.Utc);
+            }
+            else
+            {
+                bool isValidDate = DateTime.TryParseExact(comparisonDate, DateFormat, enUS, DateTimeStyles.AssumeLocal, out compareDate);
+                if (!isValidDate)
+                {
+                    Log.LogError("The comparison date '{0}' is not valid. Please specify a date in the short format.(yyyy-MM-dd)", comparisonDate);
```
weshaggard: Use DateFormat in error message.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -0,0 +1,103 @@
...
+                }
+                //Convert to UTC to converge
+                compareDate = compareDate.ToUniversalTime();
+            }
+            int months = (seedDate.Year - compareDate.Year) * 12 + seedDate.Month - compareDate.Month;
```
weshaggard: Sorry I didn't catch this earlier but you should just do `Timespan span = seedDate - comparDate; span.Months` The way you currently have it will not work in all cases.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,93 @@
...
+    <ItemGroup>
+      <CurrentVersionLines Include="%3C%3Fxml version=%221.0%22 encoding=%22utf-8%22%3F%3E" />
+      <CurrentVersionLines Include="%3C!-- This is a generated file. $(VersionSeedSourceComment) Seed Date is $(VersionSeedDate). --%3E" />
+      <CurrentVersionLines Include="%3CProject xmlns=%22http://schemas.microsoft.com/developer/msbuild/2003%22%3E" />
+      <CurrentVersionLines Include="%3CPropertyGroup%3E" />
+      <CurrentVersionLines Include="%3CBuildNumberMajor Condition=%22%27%24(BuildNumberMajor)%27==%27%27%22%3E$(GeneratedBuildNumberMajor)%3C/BuildNumberMajor%3E" />
+      <CurrentVersionLines Include="%3C/PropertyGroup%3E" />
+      <CurrentVersionLines Include="%3C/Project%3E" />
+    </ItemGroup>
+
```
weshaggard: nit: extra empty line

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -151,6 +154,93 @@
...
 
+  <!-- #################################### -->
+  <!-- Generate BuildVersion.props -->
+  <!-- #################################### -->
+  <Target Name="CreateOrUpdateCurrentVersionFile" Condition="'$(SkipVersionGeneration)'!='true'" DependsOnTargets="GenerateCurrentVersion">
+    <!-- Error out if GeneratedBuildNumberMajor is empty since GenerateCurrentVersion should have calculated it already. -->
+    <Error Condition="'$(GeneratedBuildNumberMajor)'==''" Text="Error when generating current version." />
+
+    <PropertyGroup>
+      <VersionSeedSourceComment Condition="'$(VersionSeedSourceComment)'==''">VersionSeedDate was produced by last git commit timestamp.</VersionSeedSourceComment>
```
weshaggard: I would reword this a little. "VersionSeedData was produced by taking the timestamp of the last git commit."

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
joperezr: TimeSpan class doesn't have a `Months` property, which is why I code it as is. I'm not sure I understand why it wouldn't work in all cases the way I have it.

tarekgh: the way to calculate the number of months differentness is 
var d = new DateTime(seedDate.Ticks - compareDate.Ticks);
int numberOfMonths = (d.Year - 1) * 12 + d.Month -1;

weshaggard:
LGTM 
Please squash and get this merged and lets try it out in BuildTools. Also I assume you have a plan for how to plug in the BuildNumberMinor into the targets somehow. Not sure if it requires any change in these core BuildTools targets or not.

PR #547, opened 2016-03-17
Replace built-in source command with cat command for better compatibility for Linux distributions.

leemgs:
The **source**  command is a bash build-in command. It is not a program - like ls or grep.
On up-to-date Ubuntu distribution, /bin/sh calls /bin/dash. Traditionally, /bin/sh
called /bin/bash for supporting compatibility mode. Let's use **cat** command instead of
**source** command to support better compatibility.

Signed-off-by: Geunsik Lim geunsik.lim@samsung.com
Signed-off-by: Prajwal A N an.prajwal@samsung.com
Signed-off-by: MyungJoo Ham myungjoo.ham@samsung.com

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -36,14 +36,13 @@ case $OSName in
         ;;
 
     Linux)
-        source /etc/os-release
-        if [ "$ID" == "centos" ]; then
+        if [ "$(cat /etc/*-release | grep -cim1 centos)" -eq 1 ]; then
             __PUBLISH_RID=centos.7-x64
-        elif [ "$ID" == "rhel" ]; then
+        elif [ "$(cat /etc/*-release | grep -cim1 rhel)" -eq 1 ]; then
             __PUBLISH_RID=rhel.7-x64
-        elif [ "$ID" == "ubuntu" ]; then
+        elif [ "$(cat /etc/*-release | grep -cim1 ubuntu)" -eq 1 ]; then
             __PUBLISH_RID=ubuntu.14.04-x64
-        elif [ "$ID" == "debian" ]; then
+        elif [ "$(cat /etc/*-release | grep -cim1 debian)" -eq 1 ]; then
             __PUBLISH_RID=debian.8.2-x64
         else
             echo "Unsupported Linux distribution '$ID' detected. Downloading ubuntu-x64 tools."


leemgs:
This patch is based on CoreCLR's #issue: https://github.com/dotnet/coreclr/issues/3681

leemgs:
\cc @jkotas , @dotnet-bot, @ericstj , @myungjoo

jasonwilliams200OK:
@leemgs, while we are at it, should we consider upgrading the script from `#!/usr/bin/env bash` to pure `#!/bin/sh` (contribution to https://github.com/dotnet/home/issues/29)?

Defaults:

Many linux and OS X: bash
Some linux: dash
Alpine linux: ash
Many BSDs: csh

What is most portable and the winner? None of the above, it's `sh`!

leemgs:
> What is most portable and the winner? None of the above, it's sh!

@jasonwilliams200OK ,  So, we have to use **cat** command instead of built-in **source** command.
Like the below example, on up-to-date Ubuntu distribution, /bin/sh calls /bin/dash. Let's use **cat** command to support better compatibility mode against the existing **source** command for Linux distributions such Ubuntu, Debian, CentOS, and RHEL. 
```
ls -l `which sh`
invain@target:~$ ls -l `which sh`
lrwxrwxrwx 1 root root 4 12 21 17:01 /bin/sh -> dash
invain@target:~$ 

```

akoeplinger:
@jasonwilliams200OK the script already uses `#!/usr/bin/env bash` as the shebang, so I'm a bit curious why it didn't work?

jasonwilliams200OK:
@leemgs, sure I am agreeing with you and I am just saying that using raw `sh` (with shebang `#!/bin/sh`) and base programs (i.e. `cat`) would work in all Unices without additionally installing `bash` and its constituents. And this script, along with many script across the `dotnet/*` and `aspnet/*` organizations' repos are explicitly relying on `bash` because of majority; whereas `sh` is available everywhere. For example, this is the first cross-platform script I have encountered in MSBuild repo, which is pure sh: https://github.com/Microsoft/msbuild/blob/xplat/cibuild.sh, but then it calls `init-tools.sh` which has a `bash` shebang on the top. That init-tools.sh was borrowed from this (buildtools) repo as is..

@akoeplinger, yup, that is kind of weird that even though bash shebang is there, it is not working in @leemgs's new Ubuntu environment. Instead, if bash is not installed in the system, it should ask the user to do `apt-get install bash` on new Ubuntu (where `dash` is default, like happens on other Unix-likes where `bash` doesn't get installed by default)?.

Using `sh`, IMO, is the true cross-Unix solution (like npm does for every package in for example `node_modules/.bin` directory for shortcut scripts and many other build time scripts etc.). It is a one time big effort, and these scripts don't undergo huge changes very often after couple of dev iterations.

dagood:
Question about changing `source` to `cat`: could we use `.` as a more direct `sh` alternative? Or is there some other reason to use `cat` rather than `.` that I'm not seeing?

leemgs:
> @leemgs's new Ubuntu environment. Instead, if bash is not installed in the system, it should ask the > user to do apt-get install bash on new Ubuntu (where dash is default, like happens on other Unix-
> likes where bash doesn't get installed by default)?.

@jasonwilliams200OK ,  The `ask the user to do apt-get install bash`  is not helpful to handle a lot of Linux distributions. As you mentioned, there are a lot of default shells among the Linux distributions. (Please, refer to http://distrowatch.com/  ). I don't want to handle a repetitive situation whenever the Linux distribution communities change their default shell package? Let's change the function with more effective approach that I have submitted.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -36,14 +36,13 @@ case $OSName in
         ;;
 
     Linux)
-        source /etc/os-release
-        if [ "$ID" == "centos" ]; then
+        if [ "$(cat /etc/*-release | grep -cim1 centos)" -eq 1 ]; then
             __PUBLISH_RID=centos.7-x64
-        elif [ "$ID" == "rhel" ]; then
+        elif [ "$(cat /etc/*-release | grep -cim1 rhel)" -eq 1 ]; then
```
bartonjs: Why `cat file | grep` instead of `grep file`?

bartonjs: Oh, wow, you introduced a wildcard here... that doesn't seem like "removing calls to `source`", but rather "lets do something different"

leemgs: Yes. In order to handle `/etc/os-release` and `/etc/lsb-release` in the Linux distributions, we may use cat `file | grep` more than `grep`.  It's is portable to aware Linux-based most distributions. Could you tell us if you have a better approach?

leemgs:
> could we use . as a more direct sh alternative? 
> Or is there some other reason to use cat rather than . that I'm not seeing?

@dagood , Thank you for your opinion. Is '.' different from 'source' built-in command? I think that '.' script == 'source' script. If I am wrong, correct me.  And, Here is another reason to apply a PR that we have submitted. **Note** that Linux-based distributions have a different default shell (e.g., {original} sh, ash, csh, zsh, bash, dash, and so on)

bartonjs:
@leemgs "dot" is part of the XSI shell command interface (aka the real /bin/sh does it).  BASH has `source` as an alias for that functionality.

http://pubs.opengroup.org/onlinepubs/7908799/xcu/chap2.html#tag_001_014_004

jasonwilliams200OK:
> Note that Linux-based distributions have a different default shell (e.g., {original} sh, ash, csh, zsh, bash, dash, and so on)

Yes and `{original} sh` is common in all Unices.

dagood:
A reason for `cat` over `.` (after seeing what @bartonjs pointed out on the diff) ***might*** be that you want to output multiple files, since `.` seems to only source from the first file that matches the wildcard. I don't know if `*-release` will ever match multiple files though.

Thanks @bartonjs for info about `.`, I just took my cue from https://wiki.ubuntu.com/DashAsBinSh#source

leemgs:
@bartonjs , :+1:  Actually, I have often used "dot" command in the shell such as "$> .  /etc/profile", even though I have frequently used built-in "source" command in a script file. 

> Yes and {original} sh is common in all Unices.
@jasonwilliams200OK , Right, However, the `sh` command has been using as a symbolic file in some Linux distributions. This is why we may not use `sh` command. 

One more, the below official release-note have mentioned Ubunt 14.04 LTS as supported platform. As we all know, many users are utilizing Ubuntu all over the world. Let's care the default shell environment of Ubuntu 14.04 LTS with `cat` command instead of asking the user to do `apt-get install bash`. 

```
https://github.com/dotnet/core/tree/master/release-notes
RC1 released 11/18/2015
Supported platforms for this release are Ubuntu 14.04 LTS, OS X 10.10, and Windows 7+.
```

bartonjs:
Yeah, I'd be comfortable changing `source /etc/os-release` to `. /etc/os-release`, but I don't personally like the wildcard (what if other, misleading -release files appear?) or that nothing in the PR title or description talked about wanting to read lsb-release (which, assuming \* gets sorted, would be preferred).

So, my personal stance: source => . : Good.
source => cat-with-wildcard: Overreaching, requires more thought.

leemgs:
@bartonjs , @jasonwilliams200OK , Thank you for sharing the technical comments to find better technique instead of the built-in `source` command. I will submit a PR again after doing more build test with "dot" and "cat" command.

leemgs:
@bartonjs , @jasonwilliams200OK , 
I have executed the below sample script to compare a difference "dot" command and that "source" command on Ubuntu 14.04 LTS X64 PC. 

Here is sample script. 
```
lgs@u1404:/work/dotnet/source-builtin-test$ cat ./test.sh 
---- test.sh: Start -----------------------------------
#!/usr/bin/env bash

a="shebang-a"
b="shebang-b"

if [ "$shebang-a" == "$shebang-b" ]
then
  echo "equal"
fi
---- test.sh: End ------------------------------------
```

Below is test result with "dot" command in a dash shell, a default shell script of up-to-date Ubuntu Linux distribution (e.g. LTS 14.04). 
```

lgs@u1404:/work/dotnet/source-builtin-test$ ls -al `which /bin/sh`
lrwxrwxrwx 1 root root 4  1\uc6d4  1 09:09 /bin/sh -> dash
lgs@u1404:/work/dotnet/source-builtin-test$ dash
$ .  ./test.sh
dash: 6: [: -a: unexpected operator
$ exit
lgs@u1404:/work/dotnet/source-builtin-test$ dash ./test.sh
./test.sh: 6: [: -a: unexpected operator
lgs@u1404:/work/dotnet/source-builtin-test$ 
lgs@u1404:/work/dotnet/source-builtin-test$ 
```

Do we have to still  keep the "dot" command or the existing "source" built-in command of a shell.? 
How about we change the existing "source" command with "cat" command such as CoreCLR. 
( Refer to https://github.com/dotnet/coreclr/blob/master/init-tools.sh )

leemgs:
\CC: @bartonjs , @jasonwilliams200OK ,
We have again updated PR with 'squash' command to keep a single commit.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
ellismg: We used `/etc/os-release` because that's what downstream tools like `cli` use to compute the RID.  If we start inspecting other files like `/etc/lsb-release` we can end up in a place where the thing we compute in this script isn't going to match what `cli` might compute (for example on a system that has an `lsb-release` file but on `os-release` file).  That may not be great,.

ellismg:
My preference is we do not do this (at least not now).  The `source` issues can be solved by not invoking this shell script with a non bash shell (which is a reasonable statement to make, given we have the shebang at the top of the file that says we want to use bash).  The place we were doing this incorrectly was in coreclr's init-tools.sh by explicitly using `sh` to call into this script instead of just invoking it and letting the shebang do it's thing.

Assuming we fix coreclr to not invoke this script with the wrong command interpreter (and we will) all this does is add additional sets of files to probe for a distro from.  I would prefer the logic for computing the rid here matches what cli is going to do.  cli isn't going to look at `/etc/lsb-release` at all.  It probes `/etc/os-release` for `ID` and `VERSION_ID` full stop.

If we want to make changes to CLI to also consider /etc/lsb-release and then bring them into here, that's fine, but if we do let's make sure we mimic the logic from CLI (which will very likely not be what we have here).  Note that cat/grep logic of this proposed patch when run on Ubuntu:

```
matell@MATELL-UBUNTU:~/git/coreclr$ cat /etc/*-release | grep -cim1 debian
1
```

This happens because of the `ID_LIKE` field in `/etc/os-release`.

This shows how the simple grep case is pretty dangerous.  Now the OS I detect depends on how I order my if/else statements.

danmosemsft:
I'm going to close this at this point since it seems not going to happen.

PR #551, opened 2016-03-17
Copy over locally built native shim for testing.

Priya91:
cc @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -72,8 +72,8 @@
       <TestCopyLocal Include="@(RunTestsForProjectInputs)" Exclude="@(PackagesConfigs)" />
     </ItemGroup>
 
-    <!-- Test using locally built shim libraries by default. -->
-    <ItemGroup>
+    <!-- Test using locally built shim libraries. -->
+    <ItemGroup Condition="'$(TestWithLocalLibraries)'=='true'">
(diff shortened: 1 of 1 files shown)


ericstj:
@jhendrixMSFT should look at this.  Are you sure this is going to give the correct results?  We don't always build the IL on the same machine as we run it.  Will those native bits on the IL build machine be the right OS?

Priya91:
@ericstj 
> We don't always build the IL on the same machine as we run it

Yes true, but on such OSes, we use a different script called run-test.sh, that does the same thing as build.sh, overlays the native libs on that box over the existing test bits. 
This flow will only be lit up on OSes which use build.cmd/build.sh to do both build and run tests.

Moreover if building for Linux on Windows, the native libs that are pulled down by nuget restore, will not be replaced by windows native libs (different file names). Similarly for Linux and OSX cross builds. Now, if building for RHEL on Ubuntu, there is no way currently to specify build for specific linux distros. Then, when building for RHEL on Ubuntu, will anyways pull down only ubuntu native libs from nuget, and replacing local ubuntu native shims, is not going to break the already broken scenario.

Priya91:
@dotnet-bot test this please

jhendrixMSFT:
LGTM
@weshaggard as he knows this part better than I.

weshaggard:
I'm not sure always depending on the live native shims is the correct thing to do here. We need to make sure we can still correctly run against a particular version of our packages which might have dependencies on older shims. 

cc @chcosta

Priya91:
@weshaggard Should we do this based on a msbuild property, similar to ```TestWithLocalLibraries```, we can have ```TestWithLocallyBuildShim```

weshaggard:
Yes I think this makes since for the TestWithLocalLibraries property set.

PR #552, opened 2016-03-18
Fixed a property for running ApiCompat on ref assemblies

venkat-raman251:
ApiCompat needed a minor fix that needed to be checked in. The property _PreviousContractVersion was not referenced correctly in a few conditions


venkat-raman251:
@weshaggard

weshaggard:
LGTM

PR #553, opened 2016-03-18
Force explicit qualifier usage of required parameter in GenAPI

mellinoe:
CommandLineParser chokes on unix-style paths for positional parameters.

@Priya91 , @ericstj 

After this fix is published to NuGet, we need to update the version grabbed by the tool runtime, and update notsupported.targets slightly (to pass this qualifier).


mellinoe:
I manually verified that this fixes the corefx build issues.

dagood:
I don't know GenAPI, but the plan sounds good to me. Thanks!

mellinoe:
@ericstj I changed GenAPI.targets as well. Is anyone consuming that nuget package currently?

Priya91:
Thanks @mellinoe

PR #554, opened 2016-03-18
Update notsupported.targets to use required qualifier

mellinoe:
Didn't realize that this program was included directly in buildtools. We need this change to use GenAPI.exe, which is already merged.

@Priya91


Priya91:
LGTM

mellinoe:
Thanks, going ahead and merging this.

PR #557, opened 2016-03-19
Add support for ValidationPatternsFile.

eerhardt:
In order to support easier maintenance of the valid dependency versions, allow the version information to be specified in a text file. This allows automated tools to update the version numbers easier than updating MSBuild files. It also removes having to put regular expressions in MSBuild files (which requires non-trivial work arounds).

@dagood @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/ValidateProjectDependencyVersions.cs
@@ -21,7 +21,15 @@ namespace Microsoft.DotNet.Build.Tasks
 
             public ValidationPattern(ITaskItem item, TaskLoggingHelper log)
             {
-                _idPattern = new Regex(item.ItemSpec);
+                string idRegex = item.GetMetadata("IdentityRegex");
+                if (string.IsNullOrEmpty(idRegex))
+                {
+                    // Temporarily support reading the regex from the Include/ItemSpec for backwards compatibility
+                    // when the IdentityRegex isn't specified. This can be removed once all consumers are using IdentityRegex.
+                    idRegex = item.ItemSpec;
+                }
+
+                _idPattern = new Regex(idRegex);
                 _expectedVersion = item.GetMetadata("ExpectedVersion");
                 _expectedPrerelease = item.GetMetadata("ExpectedPrerelease");
                 _log = log;


eerhardt:
To view a usage of this new feature, please see: https://github.com/dotnet/corefx/pull/7056

ellismg:
LGTM, FWIW.

weshaggard:
@eerhardt I'm not opposed to this change but I want to understand it a little further first. Why do you think the txt file makes it easier for tools to update the versions? They are both just text files after all. I agree the msbuild file requires us to do some extra escaping of the regex patterns but to me that isn't a huge deal. 

Another thing to consider is that we have various other places that we will likely start needing to know this version number and keeping it in the msbuild targets file will make those scenarios easier. See https://github.com/dotnet/coreclr/pull/3642/files as an example.

eerhardt:
> Why do you think the txt file makes it easier for tools to update the versions?

The way it currently stands in corefx, what approach would you take to update the corefx prerelease build number?  https://github.com/dotnet/corefx/blob/master/dir.props#L83-L94

```XML
  <ItemGroup>
    <ValidationPattern Include="^(?i)((System\..%2A)|(NETStandard\.Library)|(Microsoft\.CSharp)|(Microsoft\.NETCore.%2A)|(Microsoft\.TargetingPack\.Private\.(CoreCLR|NETNative))|(Microsoft\.Win32\..%2A)|(Microsoft\.VisualBasic))(?&lt;!TestData)$">
      <ExpectedPrerelease>rc3-23918</ExpectedPrerelease>
    </ValidationPattern>
    <ValidationPattern Include="^(?i)Microsoft\.TargetingPack\.(NetFramework.%2A|Private\.WinRT)$">
      <ExpectedVersion>1.0.1</ExpectedVersion>
    </ValidationPattern>
    <ValidationPattern Include="^(?i)xunit$">
      <ExpectedVersion>2.1.0</ExpectedVersion>
    </ValidationPattern>
    <ValidationPattern Include="^(?i)xunit\.netcore\.extensions$">
      <ExpectedVersion>1.0.0-prerelease-00123</ExpectedVersion>
    </ValidationPattern>
```

You would have to parse the xml, look under all `ItemGroup` elements, and then what?  Hard-code the Regex in the tool?  Assume it is the only one using an ExpectedPrerelease? Assume it is the only one using an ExpectedPrerelease with format `rcX-XXXXX`?  All of those seem rather fragile.

> I agree the msbuild file requires us to do some extra escaping of the regex patterns but to me that isn't a huge deal.

In doing this work, I found an issue with the current MSBuild approach that it is only validating the first 2 patterns. Take a look at this commit:  https://github.com/dotnet/corefx/commit/3cfd81e7369628d052da6367b28c6f5b5839f55d. You can see that the xunit.netcore.extensions version was updated from `00123` to `00187`.  However, the dir.props file wasn't updated, and no build error happened. These kinds of issues do seem like a big deal to me. Why have this tool if it isn't working?
I haven't investigated why it is broke, because I am actually fixing it with this txt file approach, so no reason to know why the old thing was broke.

> Another thing to consider is that we have various other places that we will likely start needing to know this version number and keeping it in the msbuild targets file will make those scenarios easier. 

@dagood and I discussed this, and we both decided that is isn't that hard for those places to read the .txt file. This also helps other scenarios outside of MSBuild that need to know these version numbers. Since those places won't need to parse MSBuild files.

weshaggard:
>Assume it is the only one using an ExpectedPrerelease with format rcX-XXXXX? All of those seem rather fragile.

We could also include some information about providing a Name for these versions, not too different from your txt file naming scheme. Something like "corefx_rcX-XXXX" as an example.

>However, the dir.props file wasn't updated, and no build error happened. These kinds of issues do seem like a big deal to me. Why have this tool if it isn't working?

That is just a bug which could also happen with the txt file version as well, nothing specific to msbuild just something we need to fix.

>@dagood and I discussed this, and we both decided that is isn't that hard for those places to read the .txt file. This also helps other scenarios outside of MSBuild that need to know these version numbers. Since those places won't need to parse MSBuild files.

For my information how would you imagine that we would do that to be able to plug these version numbers in to paths or other files?

All this said I'm not trying to block this approach I just want to be sure we are doing it for the right reasons and not just because we don't like msbuild. I really would prefer we not define our own format for this information and have everyone start parsing it if isn't necessary.

dagood:
> >However, the dir.props file wasn't updated, and no build error happened. These kinds of issues do seem like a big deal to me. Why have this tool if it isn't working?

> That is just a bug which could also happen with the txt file version as well, nothing specific to msbuild just something we need to fix.

Unless I'm mistaken, this is an issue with putting regex in msbuild. In the [coreclr dir.props](https://github.com/dotnet/coreclr/blob/master/tests/dir.props#L49) I'm pretty sure I worked around the same issue that Eric's mentioning in corefx by adding that dummy value. I haven't traced down the issue but it seems that MSBuild is trying to find files based on the regex.

With a raw file I don't think we'd need to worry about anything trying to interpret it, and as a bonus we don't need escaping unless newlines start appearing in package names. :smile:

I agree that it makes it more difficult to use the valid versions in MSBuild. One way could be a target that reads lines from file, and puts them into items (using some complex expressions no doubt). Later another target could make a property containing a version by batching on the items with a condition on identity. (Or more simply we could use a new custom task.)

I also agree that the least-change way to allow tooling to modify the file would be a `Name`. I think there are other benefits to moving to a simpler file though.

(I'm wondering now if a different way to work around the "dummy value" thing would be to move the regex to the metadata... like
```
<ValidationPattern Name="corefx package versions">
  <IdentityRegex>^(?i)...$</IdentityRegex>
  <ExpectedPrerelease>rc3-99999</ExpectedPrerelease>
</ValidationPattern>
```
I'm not sure what the escaping rules are in contents though.)

eerhardt:
Since I don't have the background, context, or vision for this tool, I will let you guys decide the design of where this information goes.

Here are my reasons for thinking pulling this information out of the .props file makes sense:

1. There is no need for escaping or to work around MSBuild idiosyncrasies.
2. For processes/tools outside of MSBuild, it is easier to consume a plain text file than it is to try to parse/edit the current file. (Same reason why the build tools and CLI version numbers aren't inside a MSBuild .props file)
3. Having the versions in a separate file allows us to easily hook when that separate file changes to add other automation. (Ex. When the CoreFX dependency version changes, we could go make that version the LKG out in the version repo, which would signal CLI to uptake the new Corefx build.)

Another point to note is that this code supports version numbers as they currently are, and now in a separate file, so if some numbers make more sense to be in a .props file, we can leave them there.

Let me know what you decide and I will implement it.

weshaggard:
BuildTools and CLI version numbers are special because they are about boot-strapping and I actually expect them to be the only build numbers needed outside of msbuild. Everything else should be contained inside msbuild as that is the build engine we are using. Currently I think we are putting too much information in other scripts/tools which we should clean-up and use the one language (i.e. msbuild) for the meat of our build engineering system. 

From my prospective I think we should just change the dir.props file to include some sort of name so that we can update that directly from our automation.

dagood:
I looked into what I'm running into with MSBuild with those items not showing up is actually because of the `?` in the rule. For example, the rule `<R Include="^(?i)foo$" />` disappears, but `<R Include="^(%3fi)foo$" />` works. (Adding a `|` presumably makes MSBuild decide that it couldn't be a path so it doesn't try to find one.)

Storing the regex as metadata works even better though--don't need to escape `?` or even `*`.

The way forward seems to be to change validation patterns to:

    <ValidationPattern Include="CoreFxVersions">
      <IdentityRegex>^(?i)((System\..*)|(NETStandard\.Library)|(Microsoft\.CSharp)|(Microsoft\.NETCore.*)|(Microsoft\.TargetingPack\.Private\.(CoreCLR|NETNative))|(Microsoft\.Win32\..*)|(Microsoft\.VisualBasic))(?&lt;!TestData)$</IdentityRegex>
      <ExpectedPrerelease>rc3-23921</ExpectedPrerelease>
    </ValidationPattern>
    <ValidationPattern Include="TargetingPackVersions">
      <IdentityRegex>^(?i)Microsoft\.TargetingPack\.(NetFramework.*|Private\.WinRT)$</IdentityRegex>
      <ExpectedVersion>1.0.1</ExpectedVersion>
    </ValidationPattern>
    <ValidationPattern Include="xUnitStableVersions">
      <IdentityRegex>^(?i)xunit$</IdentityRegex>
      <ExpectedVersion>2.1.0</ExpectedVersion>
    </ValidationPattern>
    <ValidationPattern Include="xUnitExtensionsVersions">
      <IdentityRegex>^(?i)xunit\.netcore\.extensions$</IdentityRegex>
      <ExpectedVersion>1.0.0-prerelease-00123</ExpectedVersion>
    </ValidationPattern>

Instead of adding versions by feeding a property into `ExpectedVersion` like https://github.com/dotnet/coreclr/pull/3642/files#diff-6178796b5428b89bb33ad7bf80143a79R26, we can use this in the target where the version is needed so the validation items stay self-contained:

    <PropertyGroup>
      <ValidVersion Condition="'%(ValidationPattern.Identity)'=='TargetingPackVersions'">%(ValidationPattern.ExpectedVersion)</ValidVersion>
    </PropertyGroup>

Does that sound reasonable @weshaggard?

eerhardt:
OK @dagood and @weshaggard - I've changed this to follow @dagood's pattern above.  Please re-take a look. Basically now it supports an IdentityRegex metadata property. If it isn't there, it falls back to using the Include for backwards compat. We can remove that once all the consumers are moved over to using IdentityRegex metadata.

dagood:
Cool, LGTM.

weshaggard:
LGTM. Thanks @eerhardt and @dagood.

@dagood as an FYI you cannot do `'%(ValidationPattern.Identity)'=='TargetingPackVersions'"` outside of a target so there might still be a need to define the some versions as individual properties perhaps. In either case I think we could use both if we have the need.

As another way to make our automation easier we could do something like. 

```
<PropertyGroup>
  <CoreFxPrereleaseVersion>rc3-23921</CoreFxPrereleaseVersion>
</PropertyGroup>

<ValidationPattern Include="CoreFxVersions">
  <IdentityRegex>^(?i)((System\..*)|(NETStandard\.Library)|(Microsoft\.CSharp)|(Microsoft\.NETCore.*)|(Microsoft\.TargetingPack\.Private\.(CoreCLR|NETNative))|(Microsoft\.Win32\..*)|(Microsoft\.VisualBasic))(?&lt;!TestData)$</IdentityRegex>
  <ExpectedPrerelease>$(CoreFxPrereleaseVersion)</ExpectedPrerelease>
</ValidationPattern>
```

That would make automation easier for updating and still enable us to use that version in properties without requiring us to be inside a target.

dagood:
Yeah, that's what I meant by "we can use this in the target" but I should have added "only". :+1: Good point that we can still make a property tool-friendly. (I'd suggest `CoreFxVersionsPrerelease` so it's a simple append though.)

PR #561, opened 2016-03-21
Allow a compatible version match between ref/impl

ericstj:
Previously we validated for strict version match when a package contains
both reference and implementation assembly.

This change adds an override PermitHigherCompatibleImplementationVersion
to be specificed in case a package needs to ship a higher implemetnation
version that reference.  This is needed for
https://github.com/dotnet/corefx/issues/7114

/cc @weshaggard


ericstj:
Merging to unblock https://github.com/dotnet/corefx/issues/7114

PR #564, opened 2016-03-23
Get runtime OS info to add xunit categories.

### LINKED ISSUE #563: PlatformSpecific tests for AnyOS assemblies running incorrectly
We're seeing tests being run for platforms they shouldn't be run for.

Specifically, it appears that the xunit command-line arguments like "-notrait category=nonwindowstests" are not being output when running the tests for platform-agnostic assemblies.  But some such assemblies do have platform-specific tests, like System.Globalization.  As a result, running those tests results in failures, because, for example, tests not intended for Windows are being run on Windows.

For example, when I do ```msbuild /t:rebuildandtest``` for System.Globalization.Extensions (which has a Windows and a Unix build), I correctly get this command-line:
```
RunTestsForProject:
  c:\Users\stoub\Source\repos\corefx\bin/tests/Windows_NT.AnyCPU.Debug/System.Globalization.Extensions.Tests/dnxcore50/CoreRun.exe xunit.console.netcore.exe System.Globalization.Extensions.Tests.dll  -xml testResults.xml -notrait category=nonwindowstests -notrait Benchmark=true  -notrait category=OuterLoop -notrait  category=failing
```
but when I do the same for System.Globalization, which is a facade for types in mscorlib that are platform-specific and which has platform-specific tests, I see this:
```
RunTestsForProject:
  c:\Users\stoub\Source\repos\corefx\bin/tests/AnyOS.AnyCPU.Debug/System.Globalization.Tests/dnxcore50/CoreRun.exe xunit.console.netcore.exe System.Globalization.Tests.dll  -xml testResults.xml -notrait Benchmark=true  -notrait category=OuterLoop -notrait category=failing
```
(no "-notrait nonwindowstests") and if I look at the tests getting run (with -showProgress), I see this test:
```
   TextInfoTests.EnUsPosixIsNotATurkishCasingLocale [STARTING]
   TextInfoTests.EnUsPosixIsNotATurkishCasingLocale [FINISHED] Time: 0.0031722s
```
which as shown here:
https://github.com/dotnet/corefx/blob/d55bead1604399a0e0ad17a2569a566340e0cba0/src/System.Globalization/tests/TextInfo.cs#L154
is attributed as being for Unix-only.

cc: @ericstj, @weshaggard, @dagood, @ellismg, @tarekgh

Priya91:
fixes #563

cc @stephentoub @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.net45/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/project.json
@@ -1,11 +1,16 @@
 {
-  "dependencies": {
-    "Newtonsoft.Json": "7.0.1",
-    "NuGet.Versioning": "3.4.0-beta-488",
-    "System.Collections.Immutable": "1.1.37",
-    "System.Reflection.Metadata": "1.1.0"
-  },
-  "frameworks": {
-    "net45": { }
-  }
+    "dependencies": {
+        "Newtonsoft.Json": "7.0.1",
+        "NuGet.Versioning": "3.4.0-beta-488",
+        "System.Collections.Immutable": "1.1.37",
+        "System.Reflection.Metadata": "1.1.0",
+        "Microsoft.NETCore.Platforms": "1.0.1-rc2-23907",
+        "System.Runtime.InteropServices.RuntimeInformation": "4.0.0-rc2-23907"
+    },
+    "frameworks": {
+        "net46": { }
+    },
+    "runtimes": {
+        "win": { }
+    }
 }
--- a/src/Microsoft.DotNet.Build.Tasks/GetTargetOS.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GetTargetOS.cs
@@ -1,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Runtime.InteropServices;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetTargetOS : Task
+    {
+        [Output]
+        public string TargetOS { get; set; }
+
+        public override bool Execute()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                TargetOS = "Windows_NT";
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                TargetOS = "Linux";
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                TargetOS = "OSX";
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Create("FREEBSD")))
+                TargetOS = "FreeBSD";
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Create("NETBSD")))
+                TargetOS = "NetBSD";
+
+            return true;
+        }
+    }
+}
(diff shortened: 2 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks.net45/project.json
```diff
@@ -1,11 +1,16 @@
...
-  }
-}
\ No newline at end of file
+    "dependencies": {
+        "Newtonsoft.Json": "7.0.1",
+        "NuGet.Versioning": "3.4.0-beta-488",
+        "System.Collections.Immutable": "1.1.37",
+        "System.Reflection.Metadata": "1.1.0",
+        "Microsoft.NETCore.Platforms": "1.0.1-rc2-23907",
+        "System.Runtime.InteropServices.RuntimeInformation": "4.0.0-rc2-23907"
```
weshaggard: I suspect this may actually break when trying to run on the framework in TFS because of the dependencies for it aren't supported there. Can you please verify that before merging this?

weshaggard:
FYI @sokket 

Thanks for doing this @Priya91 this looks good to me beyond my one concern about it working in our internal builds. We need to make sure that task library doesn't break there.

src/Microsoft.DotNet.Build.Tasks.net45/project.json
Priya91: The Microsoft.Dotnet.Build.Tasks.Packaging src.Desktop project.json also is same. Will verify on TFS too.

src/Microsoft.DotNet.Build.Tasks.net45/project.json
```diff
@@ -1,11 +1,16 @@
...
+    },
+    "frameworks": {
+        "net46": { }
```
weshaggard: Actually to help with making sure the dependencies are correct can you change this from net46 to net45 which is what we currently run on in TFS.

weshaggard: Actually I see that it was net45 and you changed it to net46, that is definitely a red flag that this is going to fail in TFS.

Priya91: Yeah, with net45 i get build errors, that the runtime lib is not found for win7-x86, win7-x64.

weshaggard: @ericstj looks like you did the same trick in the packaging project.json. Is it the RID inference you are trying to workaround here or do we need to actually fix something in our packages.

weshaggard: @Priya91 Does it still fail if you have "net45" as the framework but keep "win" as the runtime?

Priya91: yeah..
```
>EXEC : error : System.Runtime.InteropServices.RuntimeInformation 4.0.0-rc2-23907 provides a compile-time reference assembly for System.Runtime.InteropServices.RuntimeInformation on .NETFramework,Version=v4.5, but there is no run-
       time assembly compatible with win. [E:\buildtools\src\Microsoft.DotNet.Build.Tasks.net45\Microsoft.DotNet.Build.Tasks.net45.csproj]
     1>EXEC : error : Some packages are not compatible with .NETFramework,Version=v4.5 (win). [E:\buildtools\src\Microsoft.DotNet.Build.Tasks.net45\Microsoft.DotNet.Build.Tasks.net45.csproj]
     1>E:\buildtools\Tools\packageresolve.targets(30,5): error MSB3073: The command ""E:\buildtools\Tools/dotnetcli/bin/dotnet.exe" restore --packages "E:\buildtools\packages" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index
       .json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://www.myget.org/F/nugetbuild/api/v3/index.json --source https://api.nuget.org/v3/index.json "E:\buildtools\src\Microsoft.DotNet.Build.Ta
       sks.net45/project.json"" exited with code 1. [E:\buildtools\src\Microsoft.DotNet.Build.Tasks.net45\Microsoft.DotNet.Build.Tasks.net45.csproj]
```

Priya91:
@weshaggard I also verified by building and running tests on TFS, it didnt cause build break.

src/Microsoft.DotNet.Build.Tasks.net45/project.json
ericstj: That's because the old build conditioned the lineup on net46.  The new builds use a common lineup which fixes this.  Can you please either upgrade the packages and change this to net45 or file an issue to do so.

PR #565, opened 2016-03-23
Escape invalid chars when writing test output to XML

jamesqo:
Fixes dotnet/corefx#7166, where this test will 'fail' because invalid characters are being written to `testResults.xml`:

```cs
public static IEnumerable<object[]> Equal_TestData()
{
    yield return new object[] { "\uD800", "\uD800" };
}

[Theory]
[MemberData(nameof(Equal_TestData))]
public void Equal(string expected, string actual)
{
    Assert.Equal(expected, actual);
}
```

Related xUnit PRs upstream: xunit/xunit#804 and xunit/assert.xunit#1

cc @weshaggard @stephentoub

### DIFF
--- a/src/xunit.runner.uwp/XmlTestExecutionVisitor.cs
+++ b/src/xunit.runner.uwp/XmlTestExecutionVisitor.cs
@@ -273,10 +273,30 @@ namespace Xunit
             value = Escape(value);
             var escapedValue = new StringBuilder(value.Length);
             for (var idx = 0; idx < value.Length; ++idx)
-                if (value[idx] < 32)
-                    escapedValue.Append($"\\x{((byte)value[idx]).ToString("x2")}");
+            {
+                char ch = value[idx];
+                if (ch < 32)
+                {
+                    escapedValue.AppendFormat(@"\x{0}", (+ch).ToString("x2"));
+                }
+                else if (char.IsSurrogatePair(value, idx)) // handles the case when ch is the last char
+                {
+                    // Append the chars like normal if
+                    // they're valid surrogates
+                    escapedValue.Append(ch);
+                    escapedValue.Append(value[++idx]);
+                }
+                // Check for invalid chars, including
+                // stray surrogates, U+FFFE, and U+FFFF
+                else if (char.IsSurrogate(ch) || ch == '\uFFFE' || ch == '\uFFFF')
+                {
+                    escapedValue.AppendFormat(@"\x{0}", (+ch).ToString("x4"));
+                }
                 else
-                    escapedValue.Append(value[idx]);
+                {
+                    escapedValue.Append(ch);
+                }
+            }
 
             return escapedValue.ToString();
         }


src/common/XmlTestExecutionVisitor.cs
```diff
@@ -274,10 +274,24 @@ protected static string XmlEscape(string value)
             value = Escape(value);
             var escapedValue = new StringBuilder(value.Length);
             for (var idx = 0; idx < value.Length; ++idx)
-                if (value[idx] < 32)
-                    escapedValue.AppendFormat("\\x{0}", ((byte)value[idx]).ToString("x2"));
+            {
+                char ch = value[idx];
+                if (ch < 32)
+                    escapedValue.AppendFormat(@"\x{0}", (+ch).ToString("x2"));
```
stephentoub: Nit: All of these if/else blocks should have braces around them.

stephentoub: What's the purpose of the '+' before "ch"?

jamesqo: '+' converts the char to a number: `Console.WriteLine(+'A')` prints 65.

stephentoub:
Thanks for submitting a fix for this.

I'm concerned that we're diverging from the xunit source, though.  The change you ended up making for xunit... if we upgrade to a newer version of the other xunit components, won't we get that change as part of it, such that no changes are needed here?  If so, I'd prefer we do that, even if it means waiting for some time until we upgrade, rather than making competing changes here.

jamesqo:
@stephentoub The thing is, the PR that ended up getting merged doesn't actually seem to affect `XmlTestExecutionVisitor`; I tried running the console program again with the new changes, and it still has the same issues with invalid chars as it did previously. (In fact, I'm not even sure it calls the function where I made changes.) I'm still [waiting for a response](https://github.com/xunit/xunit/pull/804#issuecomment-198988357) from the repo owner about this, but I think we may have to end up changing the code here either way.

stephentoub:
Ok.  I'd like to keep this code in sync, so if the real xunit code is changed, we should make the identical change here, and if it's not, we should leave this as-is.

jamesqo:
OK. I'll close this temporarily then, and reopen if the corresponding change gets made upstream.

stephentoub:
Sounds good, thanks.

JonHanna:
This seems the wrong approach. It's failing because it's writing non–well-formed XML, and this fix makes it not choke on non–well-formed XML, but it's still producing incorrect XML that other software will (indeed *should*) choke on. I think a better approach would be to escape non-characters and other characters that aren't allowed in XML before it gets expressed as such.

mellinoe:
Agreed that whatever change we make should be in sync with the Xunit sources.

jamesqo:
This has been merged upstream: xunit/xunit#807

@stephentoub @mellinoe PTAL at reopening this.

hughbe:
Love it. Does this work with [InlineData] too?

jamesqo:
@hughbe Unfortunately invalid surrogate chars won't work with `InlineData`, since strings passed to attributes are represented as UTF-8 in IL (and since invalid surrogate chars cannot be represented as valid UTF-8, they are converted to U+FFFD).

See [this](http://stackoverflow.com/questions/36104766/xunit-net-why-do-these-2-equivalent-tests-have-different-results) StackOverflow question (and the blog post linked to in the answer) for more on this. Basically, it's a limitation of intermediate language itself, not xUnit.

hughbe:
Ach. Seems like a frustrating limitation, thanks

hughbe:
@jamesqo I think your changes have been merged in xunit, right? Do you know if this means this PR is able to be merged too? 

Cheers

jamesqo:
Yep, they have. I'm waiting for a response from @stephentoub / @mellinoe on this.

hughbe:
@jamesqo Do you have plans to update this one, I think XmlTestExecutionVisitor has been moved to https://github.com/dotnet/buildtools/blob/8b2d04969113983588e0a3eea5adda7ae54f2374/src/xunit.runner.uwp/XmlTestExecutionVisitor.cs which is causing your merge conflicts

jamesqo:
Thanks for reminding me about this, @hughbe. I just rebased so this no longer has merge conflicts.

hughbe:
Awesome!

jamesqo:
@hughbe Sounds good to me.

hughbe:
@ericstj PTAL if you have the time

ericstj:
LGTM, though I wonder why X-Linq doesn't do this automatically /cc @krwq

jamesqo:
@ericstj Thank you for merging this!

hughbe:
Thanks! @ericstj for me to update dotnet/corefx removing these workarounds, we need to update xunit.runner.uwp. Do we need to trigger an official build or something?

ericstj:
I believe you can just commit an update to https://github.com/dotnet/corefx/blob/master/src/Common/test-runtime/project.json and submit a PR.  The official build happens automatically when changes are merged, for this one that would be 1.0.3-prerelease-00903-02.

PR #572, opened 2016-03-24
Fix package versioning break

ericstj:
Change https://github.com/dotnet/buildtools/commit/daac863e9ea367338d6295c42264a58f4eb5bf1e
broke package build.  Two problems: BuildNumberMinor was not defined,
and NuGet didn't like a dash followed by a 2-part version (eg -00001.00)

For now I've changed to use a hyphen instead of a dash.  We may want to
file a bug on NuGet for this.

/cc @joperezr @weshaggard @Priya91


joperezr:
LGTM. Sorry for breaking this.

PR #573, opened 2016-03-24
Convert CloudTestTasks to run on Core

karajas:
This is to convert CloudTestTasks to run on Core. This will allow us to run full matrix test runs from any platform. 

/cc: @Chrisboh. @sokket @ChadNedzlek

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks.net45/Microsoft.DotNet.Build.CloudTestTasks.net45.csproj
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks.net45/Microsoft.DotNet.Build.CloudTestTasks.net45.csproj
@@ -1,0 +1,26 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <NuGetRuntimeIdentifier>win</NuGetRuntimeIdentifier>
+    <ImportedProjectRelativePath>..\Microsoft.DotNet.Build.CloudTestTasks\</ImportedProjectRelativePath>
+    <ProjectGuid>{17C66BCE-EB35-44CA-893C-8AAFB67708E9}</ProjectGuid>
+  </PropertyGroup>
+  <Import Project="$(ImportedProjectRelativePath)Microsoft.DotNet.Build.CloudTestTasks.csproj" />
+  <ItemGroup>
+    <Reference Include="Microsoft.Build.Tasks.v4.0" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="mscorlib" />
+    <Reference Include="Microsoft.Build.Framework" />
+    <Reference Include="Microsoft.Build.Utilities.v4.0" />
+    <Reference Include="System" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.IO.Compression" />
+    <Reference Include="System.IO.Compression.FileSystem" />
+    <Reference Include="System.Windows.Forms" />
+    <Reference Include="System.Xml" />
+    <Reference Include="System.Xml.Linq" />
+  </ItemGroup>
+</Project>
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
@@ -1,0 +1,282 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    public static class AzureHelper
+    {
+        /// <summary>
+        ///     The storage api version.
+        /// </summary>
+        public static readonly string StorageApiVersion = "2015-04-05";
+
+        public const string DateHeaderString = "x-ms-date";
+
+        public const string VersionHeaderString = "x-ms-version";
+
+        public const string AuthorizationHeaderString = "Authorization";
+
+        public enum SasAccessType
+        {
+            Read, 
+            Write,
+        };
+
+        public static string AuthorizationHeader(
+            string storageAccount, 
+            string storageKey, 
+            string method, 
+            DateTime now,
+            HttpRequestMessage request, 
+            string ifMatch = "", 
+            string contentMD5 = "", 
+            string size = "", 
+            string contentType = "")
+        {
+            string stringToSign = string.Format(
+                "{0}\n\n\n{1}\n{5}\n{6}\n\n\n{2}\n\n\n\n{3}{4}", 
+                method, 
+                (size == string.Empty) ? string.Empty : size, 
+                ifMatch, 
+                GetCanonicalizedHeaders(request), 
+                GetCanonicalizedResource(request.RequestUri, storageAccount), 
+                contentMD5, 
+                contentType);
+
+            byte[] signatureBytes = Encoding.UTF8.GetBytes(stringToSign);
+            string authorizationHeader;
+            using (HMACSHA256 hmacsha256 = new HMACSHA256(Convert.FromBase64String(storageKey)))
+            {
+                authorizationHeader = "SharedKey " + storageAccount + ":"
+                                      + Convert.ToBase64String(hmacsha256.ComputeHash(signatureBytes));
+            }
+
+            return authorizationHeader;
+        }
+
+        public static string CreateContainerSasToken(
+            string accountName, 
+            string containerName, 
+            string key, 
+            SasAccessType accessType, 
+            int validityTimeInDays)
+        {
+            string signedPermissions = string.Empty;
+            switch (accessType)
+            {
+                case SasAccessType.Read:
+                    signedPermissions = "r";
+                    break;
+                case SasAccessType.Write:
+                    signedPermissions = "wdl";
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(accessType), accessType, "Unrecognized value");
+            }
+
+            string signedStart = DateTime.UtcNow.ToString("O");
+            string signedExpiry = DateTime.UtcNow.AddDays(validityTimeInDays).ToString("O");
+            string canonicalizedResource = "/blob/" + accountName + "/" + containerName;
+            string signedIdentifier = string.Empty;
+            string signedVersion = StorageApiVersion;
+
+            string stringToSign = ConstructServiceStringToSign(
+                signedPermissions, 
+                signedVersion, 
+                signedExpiry, 
+                canonicalizedResource, 
+                signedIdentifier, 
+                signedStart);
+
+            byte[] signatureBytes = Encoding.UTF8.GetBytes(stringToSign);
+            string signature;
+            using (HMACSHA256 hmacSha256 = new HMACSHA256(Convert.FromBase64String(key)))
+            {
+                signature = Convert.ToBase64String(hmacSha256.ComputeHash(signatureBytes));
+            }
+
+            string sasToken = string.Format(
+                "?sv={0}&sr={1}&sig={2}&st={3}&se={4}&sp={5}", 
+                WebUtility.UrlEncode(signedVersion), 
+                WebUtility.UrlEncode("c"), 
+                WebUtility.UrlEncode(signature), 
+                WebUtility.UrlEncode(signedStart), 
+                WebUtility.UrlEncode(signedExpiry), 
+                WebUtility.UrlEncode(signedPermissions));
+
+            return sasToken;
+        }
+
+        public static string GetCanonicalizedHeaders(HttpRequestMessage request)
+        {
+            StringBuilder sb = new StringBuilder();
+            List<string> headerNameList = (from headerName in request.Headers
+                                           where
+                                               headerName.Key.ToLowerInvariant()
+                                               .StartsWith("x-ms-", StringComparison.Ordinal)
+                                           select headerName.Key.ToLowerInvariant()).ToList();
+            headerNameList.Sort();
+            foreach (string headerName in headerNameList)
+            {
+                StringBuilder builder = new StringBuilder(headerName);
+                string separator = ":";
+                foreach (string headerValue in GetHeaderValues(request.Headers, headerName))
+                {
+                    string trimmedValue = headerValue.Replace("\r\n", string.Empty);
+                    builder.Append(separator);
+                    builder.Append(trimmedValue);
+                    separator = ",";
+                }
+
+                sb.Append(builder);
+                sb.Append("\n");
+            }
+
+            return sb.ToString();
+        }
+
+        public static string GetCanonicalizedResource(Uri address, string accountName)
+        {
+            StringBuilder str = new StringBuilder();
+            StringBuilder builder = new StringBuilder("/");
+            builder.Append(accountName);
+            builder.Append(address.AbsolutePath);
+            str.Append(builder);
+            Dictionary<string, HashSet<string>> queryKeyValues = ExtractQueryKeyValues(address);
+            Dictionary<string, HashSet<string>> dictionary = GetCommaSeparatedList(queryKeyValues);
+
+            foreach (KeyValuePair<string, HashSet<string>> pair in dictionary.OrderBy(p => p.Key))
+            {
+                StringBuilder stringBuilder = new StringBuilder(string.Empty);
+                stringBuilder.Append(pair.Key + ":");
+                string commaList = string.Join(",", pair.Value);
+                stringBuilder.Append(commaList);
+                str.Append("\n");
+                str.Append(stringBuilder);
+            }
+
+            return str.ToString();
+        }
+
+        public static List<string> GetHeaderValues(HttpRequestHeaders headers, string headerName)
+        {
+            List<string> list = new List<string>();
+            IEnumerable<string> values;
+            headers.TryGetValues(headerName, out values);
+            if (values != null)
+            {
+                list.Add((values.FirstOrDefault() ?? string.Empty).TrimStart(null));
+            }
+
+            return list;
+        }
+
+        private static string ConstructServiceStringToSign(
+            string signedPermissions, 
+            string signedVersion, 
+            string signedExpiry, 
+            string canonicalizedResource, 
+            string signedIdentifier, 
+            string signedStart, 
+            string signedIP = "", 
+            string signedProtocol = "", 
+            string rscc = "", 
+            string rscd = "", 
+            string rsce = "", 
+            string rscl = "", 
+            string rsct = "")
+        {
+            // constructing string to sign based on spec in https://msdn.microsoft.com/en-us/library/azure/dn140255.aspx
+            var stringToSign = string.Join(
+                "\n", 
+                signedPermissions, 
+                signedStart, 
+                signedExpiry, 
+                canonicalizedResource, 
+                signedIdentifier, 
+                signedIP, 
+                signedProtocol, 
+                signedVersion, 
+                rscc, 
+                rscd, 
+                rsce, 
+                rscl, 
+                rsct);
+            return stringToSign;
+        }
+
+        private static Dictionary<string, HashSet<string>> ExtractQueryKeyValues(Uri address)
+        {
+            Dictionary<string, HashSet<string>> values = new Dictionary<string, HashSet<string>>();
+            Regex newreg = new Regex(@"\?(\w+)\=([\w|\=]+)|\&(\w+)\=([\w|\=]+)");
+            MatchCollection matches = newreg.Matches(address.Query);
+            foreach (Match match in matches)
+            {
+                string key, value;
+                if (!string.IsNullOrEmpty(match.Groups[1].Value))
+                {
+                    key = match.Groups[1].Value;
+                    value = match.Groups[2].Value;
+                }
+                else
+                {
+                    key = match.Groups[3].Value;
+                    value = match.Groups[4].Value;
+                }
+
+                HashSet<string> setOfValues;
+                if (values.TryGetValue(key, out setOfValues))
+                {
+                    setOfValues.Add(value);
+                }
+                else
+                {
+                    HashSet<string> newSet = new HashSet<string> { value };
+                    values.Add(key, newSet);
+                }
+            }
+
+            return values;
+        }
+
+        private static Dictionary<string, HashSet<string>> GetCommaSeparatedList(
+            Dictionary<string, HashSet<string>> queryKeyValues)
+        {
+            Dictionary<string, HashSet<string>> dictionary = new Dictionary<string, HashSet<string>>();
+
+            foreach (string queryKeys in queryKeyValues.Keys)
+            {
+                HashSet<string> setOfValues;
+                queryKeyValues.TryGetValue(queryKeys, out setOfValues);
+                List<string> list = new List<string>();
+                list.AddRange(setOfValues);
+                list.Sort();
+                string commaSeparatedValues = string.Join(",", list);
+                string key = queryKeys.ToLowerInvariant();
+                HashSet<string> setOfValues2;
+                if (dictionary.TryGetValue(key, out setOfValues2))
+                {
+                    setOfValues2.Add(commaSeparatedValues);
+                }
+                else
+                {
+                    HashSet<string> newSet = new HashSet<string> { commaSeparatedValues };
+                    dictionary.Add(key, newSet);
+                }
+            }
+
+            return dictionary;
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
@@ -3,107 +3,159 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Globalization;
+using System.IO;
+using System.Net;
+using System.Threading.Tasks;
+using System.Net.Http;
 using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
-using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Blob;
+
+using Task = Microsoft.Build.Utilities.Task;
 
 namespace Microsoft.DotNet.Build.CloudTestTasks
 {
+
     public sealed class CreateAzureContainer : Task
     {
         /// <summary>
-        /// The Azure account name used when creating the connection string.
-        /// </summary>
-        [Required]
-        public string AccountName { get; set; }
-
-        /// <summary>
-        /// The Azure account key used when creating the connection string.
+        ///     The Azure account key used when creating the connection string.
         /// </summary>
         [Required]
         public string AccountKey { get; set; }
 
         /// <summary>
-        /// The name of the container to create.  The specified name must be in the correct format, see the
-        /// following page for more info.  https://msdn.microsoft.com/en-us/library/azure/dd135715.aspx
+        ///     The Azure account name used when creating the connection string.
+        /// </summary>
+        [Required]
+        public string AccountName { get; set; }
+
+        /// <summary>
+        ///     The name of the container to create.  The specified name must be in the correct format, see the
+        ///     following page for more info.  https://msdn.microsoft.com/en-us/library/azure/dd135715.aspx
         /// </summary>
         [Required]
         public string ContainerName { get; set; }
 
         /// <summary>
-        /// The read-only SAS token created when ReadOnlyTokenDaysValid is greater than zero.
+        ///     When false, if the specified container already exists get a reference to it.
+        ///     When true, if the specified container already exists the task will fail.
+        /// </summary>
+        public bool FailIfExists { get; set; }
+
+        /// <summary>
+        ///     The read-only SAS token created when ReadOnlyTokenDaysValid is greater than zero.
         /// </summary>
         [Output]
         public string ReadOnlyToken { get; set; }
 
         /// <summary>
-        /// The number of days for which the read-only token should be valid.
+        ///     The number of days for which the read-only token should be valid.
         /// </summary>
         public int ReadOnlyTokenDaysValid { get; set; }
 
         /// <summary>
-        /// The URI of the created container.
+        ///     The URI of the created container.
         /// </summary>
         [Output]
         public string StorageUri { get; set; }
 
         /// <summary>
-        /// The write-only SAS token create when WriteOnlyTokenDaysValid is greater than zero.
+        ///     The write-only SAS token create when WriteOnlyTokenDaysValid is greater than zero.
         /// </summary>
         [Output]
         public string WriteOnlyToken { get; set; }
 
         /// <summary>
-        /// The number of days for which the write-only token should be valid.
+        ///     The number of days for which the write-only token should be valid.
         /// </summary>
         public int WriteOnlyTokenDaysValid { get; set; }
 
-        /// <summary>
-        /// When false, if the specified container already exists get a reference to it.
-        /// When true, if the specified container already exists the task will fail.
-        /// </summary>
-        public bool FailIfExists { get; set; }
-
         public override bool Execute()
         {
-            Log.LogMessage(MessageImportance.High, "Creating container named '{0}' in storage account {1}.", ContainerName, AccountName);
+            return ExecuteAsync().GetAwaiter().GetResult();
+        }
 
-            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(string.Format("DefaultEndpointsProtocol=https;AccountName={0};AccountKey={1}", AccountName, AccountKey));
-            CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
-            CloudBlobContainer newContainer = blobClient.GetContainerReference(ContainerName);
+        public async Task<bool> ExecuteAsync()
+        {
+            Log.LogMessage(
+                MessageImportance.High, 
+                "Creating container named '{0}' in storage account {1}.", 
+                ContainerName, 
+                AccountName);
+            DateTime dt = DateTime.UtcNow;
+            string url = string.Format(
+                "https://{0}.blob.core.windows.net/{1}?restype=container", 
+                AccountName, 
+                ContainerName);
+            StorageUri = string.Format(
+                "https://{0}.blob.core.windows.net/{1}/", 
+                AccountName, 
+                ContainerName);
 
-            if (FailIfExists && newContainer.Exists())
+            Log.LogMessage(MessageImportance.Normal, "Sending request to create Container");
+            using (HttpClient client = new HttpClient())
             {
-                Log.LogError("The container '{0}' already exists.", ContainerName);
+                using (HttpRequestMessage req = new HttpRequestMessage(HttpMethod.Put, url))
+                {
+                    req.Headers.Add(AzureHelper.DateHeaderString, dt.ToString("R", CultureInfo.InvariantCulture));
+                    req.Headers.Add(AzureHelper.VersionHeaderString, AzureHelper.StorageApiVersion);
+                    req.Headers.Add(AzureHelper.AuthorizationHeaderString, AzureHelper.AuthorizationHeader(
+                            AccountName,
+                            AccountKey,
+                            "PUT",
+                            dt,
+                            req));
+                    byte[] bytestoWrite = new byte[0];
+                    int bytesToWriteLength = 0;
+
+                    Stream postStream = new MemoryStream();
+                    postStream.Write(bytestoWrite, 0, bytesToWriteLength);
+                    req.Content = new StreamContent(postStream);
+
+                    using (HttpResponseMessage response = await client.SendAsync(req))
+                    {
+                        this.Log.LogMessage(
+                            MessageImportance.Normal,
+                            "Received response to create Container {0}: Status Code: {1} {2}",
+                            this.ContainerName, response.StatusCode, response.Content);
+                    }
+                }
+            }
+
+            try
+            {
+                if (ReadOnlyTokenDaysValid > 0)
+                {
+                    ReadOnlyToken = AzureHelper.CreateContainerSasToken(
+                        AccountName,
+                        ContainerName,
+                        AccountKey,
+                        AzureHelper.SasAccessType.Read,
+                        ReadOnlyTokenDaysValid);
+                }
+                else
+                {
+                    Log.LogWarning("{0} will not allow read permissions - ReadOnlyTokenDaysValid must be greater than 0.", StorageUri);
+                }
+
+                if (WriteOnlyTokenDaysValid > 0)
+                {
+                    WriteOnlyToken = AzureHelper.CreateContainerSasToken(
+                        AccountName,
+                        ContainerName,
+                        AccountKey,
+                        AzureHelper.SasAccessType.Write,
+                        WriteOnlyTokenDaysValid);
+                }
+                else
+                {
+                    Log.LogWarning("{0} will not allow write permissions - WriteOnlyTokenDaysValid must be greater than 0.", StorageUri);
+                }
+            }
+            catch (ArgumentOutOfRangeException e)
+            {
+                Log.LogError("ArgumentOutOfRangeException: Invalid AzureHelper.SasAccessType :"+e.Message);
                 return false;
-            }
-            else
-            {
-                newContainer.CreateIfNotExists();
-            }
-
-            StorageUri = newContainer.Uri.ToString();
-
-            // to keep with the msbuild convention of having a path end with
-            // a directory separator char append a '/' to the end of the URI
-            if (!StorageUri.EndsWith("/"))
-                StorageUri = string.Format("{0}/", StorageUri);
-
-            if (ReadOnlyTokenDaysValid > 0)
-            {
-                var sasRO = new SharedAccessBlobPolicy();
-                sasRO.SharedAccessExpiryTime = DateTime.UtcNow.AddDays(ReadOnlyTokenDaysValid);
-                sasRO.Permissions = SharedAccessBlobPermissions.Read;
-                ReadOnlyToken = newContainer.GetSharedAccessSignature(sasRO);
-            }
-
-            if (WriteOnlyTokenDaysValid > 0)
-            {
-                var sasWO = new SharedAccessBlobPolicy();
-                sasWO.SharedAccessExpiryTime = DateTime.UtcNow.AddDays(WriteOnlyTokenDaysValid);
-                sasWO.Permissions = SharedAccessBlobPermissions.Write;
-                WriteOnlyToken = newContainer.GetSharedAccessSignature(sasWO);
             }
 
             return true;
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
@@ -4,47 +4,29 @@
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProductVersion>8.0.30703</ProductVersion>
-    <SchemaVersion>2.0</SchemaVersion>
     <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.DotNet.Build.CloudTestTasks</RootNamespace>
     <AssemblyName>Microsoft.DotNet.Build.CloudTestTasks</AssemblyName>
-    <CopyNuGetImplementations>true</CopyNuGetImplementations>
-    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
+    <CLSCompliant>false</CLSCompliant>
     <AssemblyVersion>1.0.0.0</AssemblyVersion>
-    <ResolveNuGetPackages>false</ResolveNuGetPackages>
     <ProjectGuid>{F01FAA6C-4EF1-4283-B9B5-4C1416FA7F50}</ProjectGuid>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
   </PropertyGroup>
+  <PropertyGroup>
+    <StartupObject />
+  </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\common\AssemblyResolver.cs" />
+    <Compile Include="AzureHelper.cs" />
     <Compile Include="CreateAzureContainer.cs" />
-    <Compile Include="CreateAzureFileShare.cs" />
     <Compile Include="GetPerfTestAssemblies.cs" />
-    <Compile Include="SendJsonToDocumentDb.cs" />
     <Compile Include="SendToEventHub.cs" />
+    <Compile Include="UploadClient.cs" />
     <Compile Include="UploadToAzure.cs" />
     <Compile Include="WriteBuildStatsJson.cs" />
     <Compile Include="WriteItemsToJson.cs" />
-  </ItemGroup>
-  <ItemGroup>
-    <Reference Include="Microsoft.Build.Tasks.v4.0" />
-    <Reference Include="Microsoft.CSharp" />
-    <Reference Include="mscorlib" />
-    <Reference Include="Microsoft.Build.Framework, Version=4.0.0.0" />
-    <Reference Include="Microsoft.Build.Utilities.v4.0" />
-    <Reference Include="System.Core" />
-    <Reference Include="System.IO.Compression" />
-    <Reference Include="System" />
-    <Reference Include="System.Data" />
-    <Reference Include="System.Runtime.Serialization" />
-    <Reference Include="System.ServiceModel" />
   </ItemGroup>
   <ItemGroup>
     <None Include="RunnerScripts\**\*.*" />
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
@@ -4,50 +4,125 @@
 
 using System;
 using System.IO;
+using System.Net;
+using System.Security.Cryptography;
 using System.Text;
+using System.Threading.Tasks;
+
 using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
-using Microsoft.ServiceBus.Messaging;
+using System.Globalization;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using Task = Microsoft.Build.Utilities.Task;
 
 namespace Microsoft.DotNet.Build.CloudTestTasks
 {
-    public sealed class SendToEventHub : Task
+
+    public class SendToEventHub : Task
     {
-        /// <summary>
-        /// The Service Bus connection string.
-        /// </summary>
-        [Required]
-        public string ConnectionString { get; set; }
 
         /// <summary>
-        /// The path to the Event Hub.
-        /// </summary>
-        [Required]
-        public string EventHubPath { get; set; }
-
-        /// <summary>
-        /// The event data used to form the body stream.
+        ///     The event data used to form the body stream.
         /// </summary>
         [Required]
         public string EventData { get; set; }
 
         /// <summary>
-        /// The partition key for the event.
+        ///     The Event Hub Url 
+        /// </summary>
+        [Required]
+        public string EventHubPath { get; set; }
+
+        /// <summary>
+        ///     The Shared Access Key Name of the event hub
+        /// </summary>
+        [Required]
+        public string EventHubSharedAccessKey { get; set; }
+
+        /// <summary>
+        ///     The Shared Access Key Name of the event hub
+        /// </summary>
+        [Required]
+        public string EventHubSharedAccessKeyName { get; set; }
+
+        /// <summary>
+        ///     The partition key for the event.
         /// </summary>
         public string PartitionKey { get; set; }
 
+        /// <summary>
+        ///     Gets or sets the content bytes.
+        /// </summary>
+        internal byte[] contentBytes { get; set; }
+
         public override bool Execute()
         {
-            using (var streamReader = new StreamReader(EventData))
+            return ExecuteAsync().GetAwaiter().GetResult();
+        }
+
+        public async Task<bool> ExecuteAsync()
+        {
+            using (FileStream fs = new FileStream(EventData, FileMode.Open, FileAccess.Read))
             {
-                EventHubClient client = EventHubClient.CreateFromConnectionString(ConnectionString, EventHubPath);
-                client.Send(new EventData(Encoding.UTF8.GetBytes(streamReader.ReadToEnd()))
+                using (StreamReader streamReader = new StreamReader(fs))
                 {
-                    PartitionKey = PartitionKey
-                });
-                Log.LogMessage(MessageImportance.Normal, "Successfully sent notification to event hub path {0}.", EventHubPath);
+                    using (HttpClient client = new HttpClient())
+                    {
+                        string url = String.IsNullOrEmpty(PartitionKey) ? EventHubPath + "/messages" : EventHubPath + "/partitions/" + PartitionKey +"/messages";
+                        using (HttpRequestMessage req = new HttpRequestMessage(HttpMethod.Post, url))
+                        {
+                            req.Headers.Add(AzureHelper.AuthorizationHeaderString, ConstructSharedAccessToken());
+                            
+                            contentBytes = Encoding.UTF8.GetBytes(streamReader.ReadToEnd());
+                            using (Stream postStream = new MemoryStream())
+                            {
+                                postStream.Write(this.contentBytes, 0, this.contentBytes.Length);
+                                postStream.Seek(0, SeekOrigin.Begin);
+                                StreamContent contentStream = new StreamContent(postStream);
+                                req.Content = contentStream;
+
+                                Log.LogMessage(MessageImportance.High, "Sending {0} to event hub {1}", EventData, url);
+
+                                using (HttpResponseMessage response = await client.SendAsync(req))
+                                {
+                                    this.Log.LogMessage(MessageImportance.Normal, "Received response to send event to event hub");
+
+                                    if (response.StatusCode != HttpStatusCode.Created)
+                                    {
+                                        this.Log.LogError(
+                                            "Failed to send event to event hub: StatusCode:{0} Response:{1}",
+                                            response.StatusCode,
+                                            await response.Content.ReadAsStringAsync());
+                                        return false;
+                                    }
+                                }
+                                Log.LogMessage(
+                                    MessageImportance.Normal,
+                                    "Successfully sent notification to event hub path {0}.",
+                                    EventHubPath);
+                            }
+
+                        }
+                    }
+                }
             }
+
             return true;
+        }
+
+        private string ConstructSharedAccessToken()
+        {
+            //Time in milliseconds from Jan 1 1970 is expected by the API, and we want to allow the ability for this task to send for up to 20 minutes from now
+            int expiry = (int)DateTime.UtcNow.AddMinutes(20).Subtract(new DateTime(1970, 1, 1)).TotalSeconds;
+            string stringToSign = WebUtility.UrlEncode(this.EventHubPath) + "\n" + expiry;
+            string signature;
+            using (HMACSHA256 hmac = new HMACSHA256(Encoding.UTF8.GetBytes(this.EventHubSharedAccessKey)))
+            {
+                signature = Convert.ToBase64String(hmac.ComputeHash(Encoding.UTF8.GetBytes(stringToSign)));
+            }
+            string sharedAccessToken = "SharedAccessSignature sr=" + WebUtility.UrlEncode(this.EventHubPath) + "&sig="
+                           + WebUtility.UrlEncode(signature) + "&se=" + expiry + "&skn=" + this.EventHubSharedAccessKeyName;
+            return sharedAccessToken;
         }
     }
 }
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
@@ -1,0 +1,182 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections;
+using System.Globalization;
+using System.IO;
+using System.Net;
+using System.Text;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using Microsoft.Build.Utilities;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+
+
+    public class UploadClient
+    {
+        private TaskLoggingHelper log;
+
+        public UploadClient(TaskLoggingHelper loggingHelper)
+        {
+            log = loggingHelper;
+        }
+
+        public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
+        {
+            string numberOfBlocksString = numberOfBlocks.ToString("D" + lengthOfId);
+
+            byte[] bytes = Encoding.UTF8.GetBytes(numberOfBlocksString);
+            return Convert.ToBase64String(bytes);
+        }
+
+        public async Task UploadBlockBlobAsync(
+            CancellationToken ct,
+            string AccountName,
+            string AccountKey,
+            string ContainerName,
+            string filePath,
+            string destinationBlob)
+        {
+           
+            string resourceUrl = string.Format("https://{0}.blob.core.windows.net/{1}", AccountName, ContainerName);
+
+            string fileName = destinationBlob;
+            fileName = fileName.Replace("\\", "/");
+            string blobUploadUrl = resourceUrl + "/" + fileName;
+            int size = (int)new FileInfo(filePath).Length;
+            int blockSize = 4 * 1024 * 1024; //4MB max size of a block blob
+            int bytesLeft = size;
+            List<string> blockIds = new List<string>();
+            int numberOfBlocks = (size / blockSize) + 1;
+            int countForId = 0;
+            using (FileStream fileStreamTofilePath = new FileStream(filePath, FileMode.Open))
+            {
+                int offset = 0;
+
+                while (bytesLeft > 0)
+                {
+                    int nextBytesToRead = (bytesLeft < blockSize) ? bytesLeft : blockSize;
+                    byte[] fileBytes = new byte[blockSize];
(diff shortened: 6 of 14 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+using System.Linq;
+using System.Net;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    public class AzureHelper
```
ChadNedzlek: This class contains only static members, so should, itself, be static (you're never mean to instantiate an instance of "AzureHelper", it's just a useful container of methods)

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+    {
+        public const string STORAGE_API_VERSION = "2015-04-05";
```
ChadNedzlek: This is the not the style guideline for const naming. Should just be StorageApiVersion.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        public static string createContainerSASToken(string AccountName, string ContainerName, string key,
```
ChadNedzlek: Should capitalize "Create" and Pascal "Sas".

Parameters should be "accountName"

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            string accessType, int tokenValue)
```
ChadNedzlek: The magic string of "accessType" should probably be an enum, since literally the only valid answers are "write" and anything else (which we interpret as "read")

ChadNedzlek: tokenValue is a bad name for this parameter, since it really validityTimeInDays

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            string signedpermissions = "r";
+            if (accessType == "write")
+            {
+                signedpermissions += "wdl";
+            }
+            string signedstart = DateTime.UtcNow.ToString("O");
+            string signedexpiry = DateTime.UtcNow.AddDays(tokenValue).ToString("O");
+            string canonicalizedresource = "/blob/" + AccountName + "/" + ContainerName;
+            string signedidentifier = "";
+            string signedversion = "2015-04-05";
```
ChadNedzlek: Use the const-that-shouldn't-be-a-const

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            string rscc = "";
+            string rscd = "";
+            string rsce = "";
+            string rscl = "";
+            string rsct = "";
+
```
ChadNedzlek: What in the world are these values?

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+               rscd + "\n" +
+               rsce + "\n" +
+               rscl + "\n" +
+               rsct;
+
+            byte[] SignatureBytes = System.Text.Encoding.UTF8.GetBytes(StringToSign);
+            System.Security.Cryptography.HMACSHA256 SHA256 = new System.Security.Cryptography.HMACSHA256(Convert.FromBase64String(key));
+            string sig = Convert.ToBase64String(SHA256.ComputeHash(SignatureBytes));
+
+            string sasURL = string.Format("?sv={0}&sr={1}&sig={2}&st={3}&se={4}&sp={5}",
```
ChadNedzlek: Url

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+               "" + "\n" +
+               signedversion + "\n" +
+               rscc + "\n" +
+               rscd + "\n" +
+               rsce + "\n" +
+               rscl + "\n" +
+               rsct;
+
+            byte[] SignatureBytes = System.Text.Encoding.UTF8.GetBytes(StringToSign);
+            System.Security.Cryptography.HMACSHA256 SHA256 = new System.Security.Cryptography.HMACSHA256(Convert.FromBase64String(key));
```
ChadNedzlek: using this namespace (the line is way to long and noisy without it)

ChadNedzlek: Also "using" this object, since hmacs are IDisposable

ChadNedzlek: Also, this object is an "hmacSha256", not a "sha256" (there's a huge difference between the two)

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                WebUtility.UrlEncode(signedpermissions)
+                );
+
+            return sasURL;
+        }
+
+
+        public static string AuthorizationHeader(string StorageAccount, string StorageKey, string method, DateTime now,
+            HttpWebRequest request, string ifMatch = "", string md5 = "", string size="",string contentType="")
+        {
```
ChadNedzlek: See previous comment about method/parameter casing

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                (size == String.Empty) ? string.Empty : size,
+                ifMatch,
+                GetCanonicalizedHeaders(request),
+                GetCanonicalizedResource(request.RequestUri, StorageAccount),
+                md5,
+                contentType
+                );
+
+            var SignatureBytes = Encoding.UTF8.GetBytes(MessageSignature);
+            var SHA256 = new HMACSHA256(Convert.FromBase64String(StorageKey));
```
ChadNedzlek: Previous comment about variable names (especially here, since you used var)

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            var AuthorizationHeader = "SharedKey " + StorageAccount + ":" +
+                                      Convert.ToBase64String(SHA256.ComputeHash(SignatureBytes));
+            return AuthorizationHeader;
+        }
+
+        public static string GetCanonicalizedHeaders(HttpWebRequest request)
+        {
+            var headerNameList = new ArrayList();
```
ChadNedzlek: ArrayList is bad, especially when you are force casting things in it to strings. Use List<string>

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                    builder.Append(trimmedValue);
+                    separator = ",";
+                }
+                sb.Append(builder);
+                sb.Append("\n");
+            }
+            return sb.ToString();
+        }
+
+        public static ArrayList GetHeaderValues(WebHeaderCollection headers, string headerName)
```
ChadNedzlek: List<string>

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            var list = new ArrayList();
+            var values = headers[headerName];
```
ChadNedzlek: No var here, I have no idea what the return type of that indexer is.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        public static string GetCanonicalizedResource(Uri address, string accountName)
+        {
+            var str = new StringBuilder();
+            var builder = new StringBuilder("/");
+            builder.Append(accountName);
+            builder.Append(address.AbsolutePath);
+            str.Append(builder);
+            var values2 = new Dictionary<string, HashSet<string>>();
```
ChadNedzlek: values2?

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            if (true)
```
ChadNedzlek: Uh...

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            var list = new ArrayList();
+            var values = headers[headerName];
+            if (values != null)
+            {
+                list.Add(values.TrimStart(null));
+            }
+            return list;
+        }
+
+        public static string GetCanonicalizedResource(Uri address, string accountName)
```
ChadNedzlek: This method is 100 lines long. It might be a bit easier to digest if it the giant foreach loops were factored into private methods. I can't remember what the types of all the local variables are with the method so large.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            var values2 = new Dictionary<string, HashSet<string>>();
+            if (true)
+            {
+                var values = new Dictionary<string, HashSet<string>>();
+                var newreg = new Regex(@"\?(\w+)\=([\w|\=]+)|\&(\w+)\=([\w|\=]+)");
+                var matches = newreg.Matches(address.Query);
+                foreach (Match match in matches)
+                {
+                    string key, value;
+                    if (match.Groups[1].Value != String.Empty)
```
ChadNedzlek: I'd use "String.IsNullOrEmpty", the reference check to that const is probably going to be valid (I think we ALWAYS intern the empty string), but the other handles null, doesn't depend on that particular implementation detail, and is pretty standard "I don't want non-values" for strings.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                        key = match.Groups[1].Value;
+                        value = match.Groups[2].Value;
+                    }
+                    else
+                    {
+                        key = match.Groups[3].Value;
+                        value = match.Groups[4].Value;
+                    }
+
+                    if (!values.ContainsKey(key))
```
ChadNedzlek: This should just be TryGet, so you don't "get" twice. (Otherwise the second one shouldn't be "try" get, since you already checked it)

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                    }
+                    else
+                    {
+                        HashSet<string> setOfValues = null;
+                        values.TryGetValue(key, out setOfValues);
+                        setOfValues.Add(value);
+                        setOfValues.Remove(key);
+                        values.Add(key, setOfValues);
+                    }
+                    //values.Add(match.Groups[1].Value, match.Groups[2].Value);
```
ChadNedzlek: Remove this

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                        HashSet<string> newSet = new HashSet<string>();
+                        newSet.Add(value);
+                        values.Add(key, newSet);
+                    }
+                    else
+                    {
+                        HashSet<string> setOfValues = null;
+                        values.TryGetValue(key, out setOfValues);
+                        setOfValues.Add(value);
+                        setOfValues.Remove(key);
```
ChadNedzlek: We add the value and then remove the key? So if I have "?b=b&b=c", then in the end we only have "c", since I saw the key a time after I saw it's value? I'm like 90% that's unintentional.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                        values.Add(key, setOfValues);
```
ChadNedzlek: This is going to crash. You called "add" on a thing that you know is already in the dictionary. You don't have to readd it, because it's already in there. Just remove this line.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                    }
+                    //values.Add(match.Groups[1].Value, match.Groups[2].Value);
+                }
+                foreach (string str2 in values.Keys)
+                {
+                    HashSet<string> setOfValues = null;
+                    values.TryGetValue(str2, out setOfValues);
+                    var list_array = setOfValues.ToArray();
+                    ArrayList list = new ArrayList();
```
ChadNedzlek: No ArrayList

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                    {
+                        HashSet<string> setOfValues = null;
+                        values.TryGetValue(key, out setOfValues);
+                        setOfValues.Add(value);
+                        setOfValues.Remove(key);
+                        values.Add(key, setOfValues);
+                    }
+                    //values.Add(match.Groups[1].Value, match.Groups[2].Value);
+                }
+                foreach (string str2 in values.Keys)
```
ChadNedzlek: Don't use the "Keys" enumeration if you are just going to immediately call "GetValue" on it, just enumerate the dictionary and you'll get key/value pairs.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                {
+                    HashSet<string> setOfValues = null;
+                    values.TryGetValue(str2, out setOfValues);
+                    var list_array = setOfValues.ToArray();
+                    ArrayList list = new ArrayList();
+                    list.AddRange(list_array);
+                    list.Sort();
+                    var builder2 = new StringBuilder();
```
ChadNedzlek: These 10 lines are literally String.Join(",", list), just call that

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                    foreach (var obj2 in list)
+                    {
+                        if (builder2.Length > 0)
+                        {
+                            builder2.Append(",");
+                        }
+                        builder2.Append(obj2);
+                    }
+                    
+                    string key = (str2 == null) ? str2 : str2.ToLowerInvariant();
```
ChadNedzlek: Key's can't be null

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+            var list2 = new ArrayList(values2.Keys);
+            list2.Sort();
+            foreach (string str3 in list2)
+            {
+                var builder3 = new StringBuilder(string.Empty);
+                builder3.Append(str3);
+                builder3.Append(":");
+                HashSet<string> setOfValues = null;
+                values2.TryGetValue(str3, out setOfValues);
+                string commavalue = "";
```
ChadNedzlek: Another String.Join

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,234 @@
...
+                        setOfValues2.Remove(key);
+                        values2.Add(key, setOfValues2);
+                    }
+
+                    //values2.Add((str2 == null) ? str2 : str2.ToLowerInvariant(), builder2.ToString());
+                }
+            }
+            var list2 = new ArrayList(values2.Keys);
+            list2.Sort();
+            foreach (string str3 in list2)
```
ChadNedzlek: foreach(var pair in values2.OrderBy(p => p.Key))

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,159 @@
...
+using System.Text;
+using System.Threading;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
-using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Blob;
 
 namespace Microsoft.DotNet.Build.CloudTestTasks
 {
-    public sealed class CreateAzureContainer : Task
```
ChadNedzlek: Why did you unseal this?

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,159 @@
...
 
         /// <summary>
-        /// When false, if the specified container already exists get a reference to it.
-        /// When true, if the specified container already exists the task will fail.
+        ///     When false, if the specified container already exists get a reference to it.
+        ///     When true, if the specified container already exists the task will fail.
         /// </summary>
         public bool FailIfExists { get; set; }
 
+        internal bool FailState { get; set; }
```
ChadNedzlek: No one sets or reads this

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,159 @@
...
+
+        static public byte[] bytestoWrite;
```
ChadNedzlek: This probably shouldn't be either public or static...

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,159 @@
...
             {
-                newContainer.CreateIfNotExists();
+                WriteOnlyToken = AzureHelper.createContainerSASToken(AccountName, ContainerName, AccountKey, "write", WriteOnlyTokenDaysValid);
             }
 
-            StorageUri = newContainer.Uri.ToString();
+            Log.LogMessage(MessageImportance.High, "Sending request to create Container");
+            bytestoWrite = new byte[0];
+            bytesToWriteLength = 0;
+            request.BeginGetRequestStream(GetRequestStreamCallback, request);
```
ChadNedzlek: I strongly suggest using the GetRequestStreamAsync pattern here, it's much easier to use and read than the BeginFoo versions.  I'd also suggest "HttpClient", it's got a nicer interface, if you don't need to do anything complicated.

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,159 @@
...
-            // a directory separator char append a '/' to the end of the URI
-            if (!StorageUri.EndsWith("/"))
-                StorageUri = string.Format("{0}/", StorageUri);
+            return true;
+        }
 
-            if (ReadOnlyTokenDaysValid > 0)
+        private static void GetResponseStreamCallback(IAsyncResult callbackResult)
+        {
+            var request = (HttpWebRequest)callbackResult.AsyncState;
```
ChadNedzlek: Yeah... don't do this.  It DEFINTELY can't be static (/m will kill your build in most horrible ways). So you need to pass yourself as IAsyncResult.State.

But again, just use the Async stuff instead, it's way cleaner and you don't have to deal with all this messy stuff.

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,159 @@
...
             }
-
-            if (WriteOnlyTokenDaysValid > 0)
+            catch (WebException e)
             {
-                var sasWO = new SharedAccessBlobPolicy();
-                sasWO.SharedAccessExpiryTime = DateTime.UtcNow.AddDays(WriteOnlyTokenDaysValid);
-                sasWO.Permissions = SharedAccessBlobPermissions.Write;
-                WriteOnlyToken = newContainer.GetSharedAccessSignature(sasWO);
+                Console.WriteLine(e.Message);
```
ChadNedzlek: Don't use Console, use Log

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,159 @@
...
+                Console.WriteLine(e.Status);
+                var resp = new StreamReader(e.Response.GetResponseStream()).ReadToEnd();
```
ChadNedzlek: You need to using the response, the stream, and the reader all.

src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
```diff
@@ -4,50 +4,137 @@
...
+using System.Security.Cryptography;
 using System.Text;
+using System.Threading;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
-using Microsoft.ServiceBus.Messaging;
 
 namespace Microsoft.DotNet.Build.CloudTestTasks
 {
-    public sealed class SendToEventHub : Task
```
ChadNedzlek: sealed?

karajas:
Offline discussion with @ChadNedzlek. Updated with recommended changes

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        {
+            var signedpermissions = string.Empty;
+            if (accessType == SasAccessType.read)
+            {
+                signedpermissions += "r";
+            }
+
+            if (accessType == SasAccessType.write)
+            {
+                signedpermissions += "wdl";
```
ChadNedzlek: Does "write" not also imply read?

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net;
+    using System.Security.Cryptography;
+    using System.Text;
+    using System.Text.RegularExpressions;
+
+    #endregion
```
ChadNedzlek: I believe the code style for corefx/buildtools is usings on the outside

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+            var canonicalizedresource = "/blob/" + accountName + "/" + containerName;
+            var signedidentifier = string.Empty;
+            var signedversion = StorageApiVersion;
+
+            var stringToSign = signedpermissions + "\n" + signedstart + "\n" + signedexpiry + "\n"
+                               + canonicalizedresource + "\n" + signedidentifier + "\n" + string.Empty + "\n" + string.Empty + "\n"
+                               + signedversion + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty;
+
+            var signatureBytes = Encoding.UTF8.GetBytes(stringToSign);
+            var hmacSha256 = new HMACSHA256(Convert.FromBase64String(key));
```
ChadNedzlek: using the hmac (they are disposable)

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+
+            var signedstart = DateTime.UtcNow.ToString("O");
+            var signedexpiry = DateTime.UtcNow.AddDays(validityTimeInDays).ToString("O");
+            var canonicalizedresource = "/blob/" + accountName + "/" + containerName;
+            var signedidentifier = string.Empty;
+            var signedversion = StorageApiVersion;
+
+            var stringToSign = signedpermissions + "\n" + signedstart + "\n" + signedexpiry + "\n"
+                               + canonicalizedresource + "\n" + signedidentifier + "\n" + string.Empty + "\n" + string.Empty + "\n"
+                               + signedversion + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty;
```
bartonjs: Adding string.Empty is weird.  If you're using them to signify a placeholder, that'd make a lot more sense if the structure you were building was documented immediately beforehand.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+                    values.Add(key, newSet);
+                }
+                else
+                {
+                    HashSet<string> setOfValues = null;
+                    values.TryGetValue(key, out setOfValues);
+                    if (setOfValues != null)
+                    {
+                        setOfValues.Add(value);
+                        values.Remove(key);
```
ChadNedzlek: You don't need to remove/add the same set, the current set is already in there.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+                if (!values.ContainsKey(key))
+                {
+                    var newSet = new HashSet<string> { value };
+                    values.Add(key, newSet);
+                }
+                else
+                {
+                    HashSet<string> setOfValues = null;
+                    values.TryGetValue(key, out setOfValues);
+                    if (setOfValues != null)
```
ChadNedzlek: And it shouldn't be able to be null (since you control all the adds, and they are always non-null)

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+
+                if (!values.ContainsKey(key))
+                {
+                    var newSet = new HashSet<string> { value };
+                    values.Add(key, newSet);
+                }
+                else
+                {
+                    HashSet<string> setOfValues = null;
+                    values.TryGetValue(key, out setOfValues);
```
ChadNedzlek: This "try" should be promoted up to the "if". At the moment, you test, and then test a second time.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+            var signedexpiry = DateTime.UtcNow.AddDays(validityTimeInDays).ToString("O");
+            var canonicalizedresource = "/blob/" + accountName + "/" + containerName;
+            var signedidentifier = string.Empty;
+            var signedversion = StorageApiVersion;
+
+            var stringToSign = signedpermissions + "\n" + signedstart + "\n" + signedexpiry + "\n"
+                               + canonicalizedresource + "\n" + signedidentifier + "\n" + string.Empty + "\n" + string.Empty + "\n"
+                               + signedversion + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty + "\n" + string.Empty;
+
+            var signatureBytes = Encoding.UTF8.GetBytes(stringToSign);
```
bartonjs: Under the CoreFx style guidelines, every `var` between 139 and 160 (inclusive) is wrong, because it's not a cast, a `new` or an `as`.

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,185 @@
...
-            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(string.Format("DefaultEndpointsProtocol=https;AccountName={0};AccountKey={1}", AccountName, AccountKey));
-            CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
-            CloudBlobContainer newContainer = blobClient.GetContainerReference(ContainerName);
+            return this.ExecuteAsync().GetAwaiter().GetResult();
+        }
 
-            if (FailIfExists && newContainer.Exists())
+        public async Task<bool> ExecuteAsync()
+        {
+            this.Log.LogMessage(
```
ChadNedzlek: I believe we usually avoid "this." unless it's absolutely necessary.

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,185 @@
...
+                this.WriteOnlyToken = AzureHelper.CreateContainerSasToken(
+                    this.AccountName, 
+                    this.ContainerName, 
+                    this.AccountKey, 
+                    AzureHelper.SasAccessType.write, 
+                    this.WriteOnlyTokenDaysValid);
             }
 
-            StorageUri = newContainer.Uri.ToString();
+            this.Log.LogMessage(MessageImportance.High, "Sending request to create Container");
```
ChadNedzlek: Probably include the container name in this log, since it's "High", it's nice to have it tell some extra information

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
bartonjs: Then, the next comment would be

"This is probably easier to read as

```c#
string stringToSign = string.Join(
    "\n",
    signedPermissions,
    signedStart,
    ....);
```
"

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -1,112 +1,185 @@
...
             {
-                var sasWO = new SharedAccessBlobPolicy();
-                sasWO.SharedAccessExpiryTime = DateTime.UtcNow.AddDays(WriteOnlyTokenDaysValid);
-                sasWO.Permissions = SharedAccessBlobPermissions.Write;
-                WriteOnlyToken = newContainer.GetSharedAccessSignature(sasWO);
+                if (!e.Status.Equals("ProtocolError"))
+                {
+                    this.Log.LogError(e.Message);
+                    this.Log.LogError(e.Status.ToString());
+                    var resp = new StreamReader(e.Response.GetResponseStream()).ReadToEnd();
```
ChadNedzlek: There are 2 IDisposable things on this line. (The stream and the reader). You'll have to break it out into multiple lines to using them all correctly, unfortunately.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+            if (accessType == SasAccessType.write)
+            {
+                signedpermissions += "wdl";
+            }
+
+            var signedstart = DateTime.UtcNow.ToString("O");
+            var signedexpiry = DateTime.UtcNow.AddDays(validityTimeInDays).ToString("O");
+            var canonicalizedresource = "/blob/" + accountName + "/" + containerName;
+            var signedidentifier = string.Empty;
+            var signedversion = StorageApiVersion;
```
bartonjs: Naming: `signedStart`, `signedExpiry`, `canonicalizedResource`, `signedIdentifier`, `signedVersion`

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// The now.
+        /// </param>
+        /// <param name="request">
+        /// The request.
+        /// </param>
+        /// <param name="ifMatch">
+        /// The if match.
+        /// </param>
+        /// <param name="md5">
+        /// The md 5.
```
bartonjs: "The MD5 hash of [whatever it's hashing], as [I'm assuming base64, but maybe hex?]"?

src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
```diff
@@ -1,53 +1,141 @@
...
+                using (var streamReader = new StreamReader(fs))
                 {
-                    PartitionKey = PartitionKey
-                });
-                Log.LogMessage(MessageImportance.Normal, "Successfully sent notification to event hub path {0}.", EventHubPath);
+                    var request = (HttpWebRequest)WebRequest.Create(this.EventHubPath);
+                    request.Method = "POST";
+                    request.ContentType = "application/atom+xml;type=entry;charset=utf-8";
+
+                    var expiry = (int)DateTime.UtcNow.AddMinutes(20).Subtract(new DateTime(1970, 1, 1)).TotalSeconds;
```
ChadNedzlek: This is quite a bit of date math, a little comment in english about what it's trying to say (which is, I think "The api expects times in 'number of seconds since 1970', and we want the upload good for 20 minutes from now)

src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
```diff
@@ -1,53 +1,141 @@
...
+                    var stringToSign = WebUtility.UrlEncode(this.EventHubPath) + "\n" + expiry;
+                    var hmac = new HMACSHA256(Encoding.UTF8.GetBytes(this.EventHubSharedAccessKey));
```
ChadNedzlek: using hmac

src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
```diff
@@ -1,53 +1,141 @@
...
+                        this.Log.LogMessage(
+                            MessageImportance.Normal,
+                            "Successfully sent notification to event hub path {0}.",
+                            this.EventHubPath);
+                    }
+                    catch (WebException e)
+                    {
+                        this.Log.LogError(e.Message);
+                        this.Log.LogError(e.Status.ToString());
+                        var resp = new StreamReader(e.Response.GetResponseStream()).ReadToEnd();
```
ChadNedzlek: Another 2 IDisposable

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,196 @@
...
+//   
+// </copyright>
+// <summary>
+//   The upload client.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    #region
```
ChadNedzlek: move up

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// </param>
+        /// <param name="size">
+        /// The size.
+        /// </param>
+        /// <param name="contentType">
+        /// The content type.
+        /// </param>
```
bartonjs: Nit: We don't really have a style rule on this, but usually if the leading spaces, start tag, payload, and end tag would fit in ~110 chars we just one-line it:

```c#
/// <summary>
/// The authorization header.
/// </summary>
/// <param name="storageAccount">The storage account.</param>
...
```

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,196 @@
...
+            var filename = destinationBlob;
+            filename = filename.Replace("\\", "/");
+            var blobUploadUrl = resourceUrl + "/" + filename;
+            var size = (int)new FileInfo(filePath).Length;
+            var blockSize = 4 * 1024 * 1024;
+            var bytesLeft = size;
+            var blockIds = new ArrayList();
+            var numberOfBlocks = (size / blockSize) + 1;
+            var countForId = 0;
+            var f = new FileStream(filePath, FileMode.Open);
```
ChadNedzlek: using the "f", and probably call it "sourceFileStream" or something

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// The method.
+        /// </param>
+        /// <param name="now">
+        /// The now.
+        /// </param>
+        /// <param name="request">
+        /// The request.
+        /// </param>
+        /// <param name="ifMatch">
+        /// The if match.
```
bartonjs: API documentation should be useful when written.  Just saying "the [nameof(param)]" isn't really very helpful... it should be beefed up to help someone know what to pass in; or perhaps consider not having it.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,196 @@
...
+            var offset = 0;
+
+            while (bytesLeft > 0)
+            {
+                var nextBytes = (bytesLeft < blockSize) ? bytesLeft : blockSize;
+                var fileBytes = new byte[blockSize];
+                var read = f.Read(fileBytes, 0, nextBytes);
```
ChadNedzlek: Technically, "Read" doesn't _have_ to return the number of bytes you requested. Though I don't know of any implementation of FileStream that doesn't.  You should probably validate that "read" and "nextBytes" are the same after this line with some sort of Exception or Error log or something, so if it weirdly fails in the future, we'll see why.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
bartonjs: > I believe the code style for corefx/buildtools is usings on the outside

Yep.  Not sure if we have rules against \#regions or not

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+
+        /// <summary>
+        /// The sas access type.
+        /// </summary>
+        public enum SasAccessType
+        {
+            /// <summary>
+            /// The read.
+            /// </summary>
+            read,
```
bartonjs: enum members should be `PascalCased`.  So `Read` and `Write`

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,196 @@
...
+        /// <param name="ContainerName">
+        /// The container name.
+        /// </param>
+        /// <param name="filePath">
+        /// The file path.
+        /// </param>
+        /// <param name="destinationBlob">
+        /// The destination blob.
+        /// </param>
+        public void UploadBlockBlob(
```
ChadNedzlek: Since this method is doing a _lot_ of web and file work, it would be appropriate to make it UploadBlockBlobAsync, and return a Task, so that users don't their threads blocked if they want to do some other stuff while all this IO is taking place.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        {
+            /// <summary>
+            /// The read.
+            /// </summary>
+            read, 
+
+            /// <summary>
+            /// The write.
+            /// </summary>
+            write
```
bartonjs: It's a good idea to always end in trailing commas (when the language allows), so that appending things in the future doesn't show an edit on this line.  (`git blame` and other "when did this show up?" type tools will show the original addition, instead of "oh, adding a comma")

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
ChadNedzlek: This license shouldn't be deleted (and it should be added to all the other files). Probably drop the <copyright/summary> stuff too.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        public static string AuthorizationHeader(
+            string storageAccount, 
+            string storageKey, 
+            string method, 
+            DateTime now, 
+            HttpWebRequest request, 
+            string ifMatch = "", 
+            string md5 = "", 
+            string size = "", 
+            string contentType = "")
```
bartonjs: Why is everything a string?

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
...
-        [Required]
-        public string AccountName { get; set; }
+        private CancellationTokenSource tokenSource;
+
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="UploadToAzure" /> class.
+        /// </summary>
+        public UploadToAzure()
+        {
+            this.Overwrite = false;
```
ChadNedzlek: False is already the default for booleans (and if you wanted to be explicit anyway, you could just add "= false" to the end of the property line to avoid having to add the constructor)

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
...
         {
-            Log.LogMessage(MessageImportance.High, "Begin uploading blobs to Azure account {0} in container {1}.", AccountName, ContainerName);
+            this.tokenSource = new CancellationTokenSource();
+            var ct = this.tokenSource.Token;
 
-            if (Items.Length == 0)
-                throw new ArgumentException("No items were provided for upload.");
+            var task =
+                System.Threading.Tasks.Task.Factory.StartNew(() => this.ExecuteAsync(ct), ct).GetAwaiter().GetResult();
+            return task.Result;
```
ChadNedzlek: This should just look like all the other ones. You don't need to start a new thread for this.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
...
+        /// <returns>
+        ///     The <see cref="bool" />.
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        /// </exception>
         public override bool Execute()
         {
-            Log.LogMessage(MessageImportance.High, "Begin uploading blobs to Azure account {0} in container {1}.", AccountName, ContainerName);
+            this.tokenSource = new CancellationTokenSource();
+            var ct = this.tokenSource.Token;
```
ChadNedzlek: I'd also make this variable set in the initialize, so that you don't have a race condition of someone cancelling it before this line runs.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+            if (accessType == SasAccessType.read)
+            {
+                signedpermissions += "r";
+            }
+
+            if (accessType == SasAccessType.write)
+            {
+                signedpermissions += "wdl";
+            }
+
```
bartonjs: In the future you add a permission, FlibbertyGibbet.  Someone builds using that, and since it's an enum in their program it just says "2".   They then end up using this library, and pass (effectively) `(SasAccessType)2`.  This sets signedPermissions to `string.Empty`.  Is that desired?  Or, do you want

```c#
switch (accessType)
{
    case SasAccessType.Read:
        signedPermissions = "r";
        break;
    case SasAccessType.Write:
       signedPermissions = "wdl";
       break;
    default:
        throw new ArgumentOutOfRangeException(nameof(accessType), accessType, "Unrecognized value");
}
```
?

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
...
+            }
+            catch (WebException e)
+            {
+                this.Log.LogError(e.Message);
+                this.Log.LogError(e.Status.ToString());
+                var resp = new StreamReader(e.Response.GetResponseStream()).ReadToEnd();
+                this.Log.LogError(resp);
+            }
+
+            this.Log.LogMessage(MessageImportance.High, "Received response to check whether Container blobs exist");
```
ChadNedzlek: "High" is really high. In general, a task should only have one "High" level event, saying the task started. The rest should probably be "Normal" or "Low", so that when building with "minimal" you still get a sense of progress, but messages that you can't really act on (like this one) don't shove off other messages.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
bartonjs: accessType isn't \[Flags\], so why are you using `+=`?

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
...
+
+            var result = true;
+            Parallel.ForEach(
+                this.Items, 
+                (item, loopState) =>
+                    {
+                        if (ct.IsCancellationRequested)
+                        {
+                            this.Log.LogError("Task UploadToAzure cancelled");
```
ChadNedzlek: I don't know that Log is thread safe.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+    /// </summary>
+    public static class AzureHelper
+    {
+        /// <summary>
+        /// The storage api version.
+        /// </summary>
+        public static readonly string StorageApiVersion = "2015-04-05";
+
+        /// <summary>
+        /// The sas access type.
```
bartonjs: Since it's in documentation here, "sas" should be one of "SAS", "Shared Access Signature (SAS)", or "SAS (Shared Access Signature)".

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
...
+                this.Log.LogError(e.Message);
+                this.Log.LogError(e.Status.ToString());
+                var resp = new StreamReader(e.Response.GetResponseStream()).ReadToEnd();
+                this.Log.LogError(resp);
+            }
+
+            this.Log.LogMessage(MessageImportance.High, "Received response to check whether Container blobs exist");
+
+            var result = true;
+            Parallel.ForEach(
```
ChadNedzlek: If UploadBlockBlob was Async, you wouldn't have to use Parallel here (and then wouldn't have to worry about thread safety issues). You could just call:
await Task.WhenAll(Items.Select(UploadSingleItemAsync))

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+            read, 
+
+            /// <summary>
+            /// The write.
+            /// </summary>
+            write
+        };
+
+        /// <summary>
+        /// The authorization header.
```
bartonjs: More doc-comment nits: The summary says what a method does; which means there should be a verb in it somewhere... it's a whole sentence.

```c#
/// <summary>
/// Computes the authorization header according to the current storage API version.
/// </summary>
```

Or something like that.

src/Microsoft.DotNet.Build.CloudTestTasks.net45/Microsoft.DotNet.Build.CloudTestTasks.net45.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <NuGetRuntimeIdentifier>win</NuGetRuntimeIdentifier>
+    <ImportedProjectRelativePath>..\Microsoft.DotNet.Build.CloudTestTasks\</ImportedProjectRelativePath>
+    <ProjectGuid>{17C66BCE-EB35-44CA-893C-8AAFB67708E9}</ProjectGuid>
+  </PropertyGroup>
+  <Import Project="$(ImportedProjectRelativePath)Microsoft.DotNet.Build.CloudTestTasks.csproj" />
+  <ItemGroup>
+    <Reference Include="Microsoft.Build.Tasks.v4.0" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="mscorlib" />
```
bartonjs: If you're trying to make this run on core, why are you referencing mscorlib, et al?

bartonjs: Oh, nevermind; based on the filename I think this is "read all that code, but compile against net45 instead of corefx".

src/Microsoft.DotNet.Build.CloudTestTasks/WriteBuildStatsJson.cs
```diff
@@ -97,7 +97,8 @@ public TestBuildStatsJson(ITaskItem[] correlationIDs, string logUri, int builtCo
 
         public override bool Execute()
         {
-            using (StreamWriter streamWriter = new StreamWriter(OutputFile))
+            using (var fs = new FileStream(OutputFile, FileMode.OpenOrCreate, FileAccess.Read | FileAccess.Write | FileAccess.ReadWrite))
```
ChadNedzlek: Do you really want "OpenOrCreate". It has unexpected behavior if you are writing less data than is currently in the file (it doesn't truncate the existing file, so you'd end up with whatever was left of that file still at the end of yours)
Example: If current file is "aaaaaaa", and you open it and write "bbb", you'd have "bbbaaaa".  I think you want "Create".  Using the "File.Create" helper method should work too

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// </summary>
+        /// <param name="headers">
+        /// The headers.
+        /// </param>
+        /// <param name="headerName">
+        /// The header name.
+        /// </param>
+        /// <returns>
+        /// The <see>
+        ///         <cref>List</cref>
```
bartonjs: I think this was supposed to be `The <see cref="List" />`?  It's gone pretty wonky...

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// </returns>
+        private static Dictionary<string, HashSet<string>> GetCommaSeparatedList(
+            Dictionary<string, HashSet<string>> queryKeyValues)
+        {
+            var dictionary = new Dictionary<string, HashSet<string>>();
+            
+            foreach (var queryKeys in queryKeyValues.Keys)
+            {
+                HashSet<string> setOfValues = null;
+                queryKeyValues.TryGetValue(queryKeys, out setOfValues);
```
bartonjs: You're about to deference `setOfValues`... why wasn't the return value of `TryGetValue` important?

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+        ///     .
+        /// </returns>
+        private static Dictionary<string, HashSet<string>> GetCommaSeparatedList(
+            Dictionary<string, HashSet<string>> queryKeyValues)
+        {
+            var dictionary = new Dictionary<string, HashSet<string>>();
+            
+            foreach (var queryKeys in queryKeyValues.Keys)
+            {
+                HashSet<string> setOfValues = null;
```
bartonjs: This is about to be written by an `out`, you don't need to initialize it.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+                queryKeyValues.TryGetValue(queryKeys, out setOfValues);
+                var listArray = setOfValues.ToArray();
+                var list = new List<string>();
+                list.AddRange(listArray);
```
bartonjs: List.AddRange should be able to take the HashSet... why is there an explicit intermediate array?

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -0,0 +1,373 @@
...
+                list.Sort();
+                var commaSeparatedValues = string.Join(",", list);
+                var key = queryKeys.ToLowerInvariant();
+                if (!dictionary.ContainsKey(key))
```
bartonjs: Rule \#1 of using Dictionary: If you call ContainsKey, you meant to use TryGetValue.
Rule \#2: Very rarely, rule \#1 is wrong.

```c#
HashSet<string> cachedSet;

if (dictionary.TryGetValue(key, out cachedSet))
{
    // whatever is currently in the else;
    // but I'm concerned it just crashes, since you're calling
    // dictionary.Add with a key you already know is present
}
else
{
    cachedSet = new HashSet<string> { commaSeparatedValues };
    dictionary.Add(key, cachedSet);
}
```

bartonjs: Looking, and guessing, I think what you want is

```c#
HashSet<string> keyedSet;

if (!dictionary.TryGetValue(key, out cachedSet))
{
    keyedSet = new HashSet<string> { commaSeparatedValues };
    dictionary.Add(key, keyedSet);
}

keyedSet.Add(commaSeparatedValues);
```

src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
bartonjs: You're taking Now (UTC) and subtracting 1970-01-01 (Local).  That's going to have a variable amount of delta, depending on where it was executed.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,196 @@
...
+    using Microsoft.Build.Utilities;
+
+    #endregion
+
+    /// <summary>
+    /// The upload client.
+    /// </summary>
+    public class UploadClient
+    {
+        private TaskLoggingHelper log;
```
bartonjs: Naming: private fields are `_underscoreCamelCased`, so `_log`.  This mostly obviates the need for `this.`.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,196 @@
...
+        {
+            var resourceUrl = string.Format("https://{0}.blob.core.windows.net/{1}", AccountName, ContainerName);
+
+            var filename = destinationBlob;
+            filename = filename.Replace("\\", "/");
+            var blobUploadUrl = resourceUrl + "/" + filename;
+            var size = (int)new FileInfo(filePath).Length;
+            var blockSize = 4 * 1024 * 1024;
+            var bytesLeft = size;
+            var blockIds = new ArrayList();
```
bartonjs: Why ArrayList?

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,196 @@
...
+                var read = f.Read(fileBytes, 0, nextBytes);
+                var blockId = this.EncodeBlockIds(countForId, numberOfBlocks.ToString().Length);
+
+                blockIds.Add(blockId);
+                var blockUploadUrl = blobUploadUrl + "?comp=block&blockid=" + blockId;
+
+                var request = (HttpWebRequest)WebRequest.Create(blockUploadUrl);
+                var dt = DateTime.UtcNow;
+                request.Method = "PUT";
+                request.Headers["x-ms-date"] = dt.ToString("R", CultureInfo.InvariantCulture);
```
bartonjs: You've used these strings a couple times, consider making them constants.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -1,96 +1,229 @@
...
+        /// The execute async.
+        /// </summary>
+        /// <param name="ct">
+        /// The ct.
+        /// </param>
+        /// <returns>
+        /// The <see cref="Task"/>.
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        /// </exception>
```
bartonjs: In the \<exception\> element you should describe what causes that exception :)

src/Microsoft.DotNet.Build.CloudTestTasks.net45/Microsoft.DotNet.Build.CloudTestTasks.net45.csproj
karajas: Yup this is following the Microsoft.DotNet.Build.Tasks method of having a lib run on both Core and net45.

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -2,111 +2,186 @@
...
 
-using System;
-using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
-using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Blob;
-
 namespace Microsoft.DotNet.Build.CloudTestTasks
 {
+    using System;
```
ChadNedzlek: The usings shouldn't be inside the namespace.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,205 @@
...
+using System.Net;
+using System.Text;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    using Task = System.Threading.Tasks.Task;
```
ChadNedzlek: outside (I'd probably specify the TaskLoggingHelper myself, since that type is used fewer times in this file, but that's just a preference)

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -3,94 +3,194 @@
...
-                var relativeBlobPath = item.GetMetadata("RelativeBlobPath");
-                if (string.IsNullOrEmpty(relativeBlobPath))
+                XmlDocument doc;
+                using (WebResponse response = await request.GetResponseAsync())
                 {
-                    Log.LogError(string.Format("Metadata 'RelativeBlobPath' is missing for item '{0}'.", item.ItemSpec));
-                    result = false;
-                    loopState.Stop();
+                    doc = new XmlDocument();
+                    doc.LoadXml(new StreamReader(response.GetResponseStream()).ReadToEnd());
```
ChadNedzlek: using(var stream = response.GetResponseStream())
{
   doc.Load(stream)
}

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -3,94 +3,194 @@
...
-                    blockBlob.UploadFromStream(fileStream);
+                    Log.LogError(errorStreamReader.ReadToEnd());
                 }
-            });
+            }
+
+            Log.LogMessage(MessageImportance.Normal, "Received response to check whether Container blobs exist");
+
+            bool result = true;
+            Parallel.ForEach(
```
ChadNedzlek: SInce Upload is Async now, you don't actually need to parallel this, because it doesn't block on web requests anymore.

Factor out the body of the lamba (which is good practice anyway, so you can keep track of what values are going to be referenced by multiple callers and which one).

Then ditch the Parallel, and just do:
await Task.WhenAll(Items.Select(item => MyRefactoredMethod(item))

Parallel usually brings a lot more complications to the party than is necessary, and since we aren't actually doing a lot of parallel CPU bound work here, those complications probably outweigh the benefits.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -3,94 +3,194 @@
...
+                            result = false;
+                            loopState.Stop();
+                        }
+
+                        lock (Log)
+                        {
+                            Log.LogMessage("Uploading {0} to {1}.", item.ItemSpec, ContainerName);
+                        }
+                        UploadClient uploadClient = new UploadClient(Log);
+                        await uploadClient.UploadBlockBlobAsync(
```
ChadNedzlek: Pass the cancellation token along, so that the outstanding web requests can be cancelled too, since those requests are things you want to cancel. Essentially, if you have a cancellation token, every Async method you call inside should also be passed the token (unless they are methods that you can't change, and don't accept tokens).

Though HttpWebRequest/Response don't have cancellable methods (HttpClient, however does), so you won't actually be able to cancel outstanding requests currently.

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -3,110 +3,166 @@
...
+                            dt,
+                            req));
+                    byte[] bytestoWrite = new byte[0];
+                    int bytesToWriteLength = 0;
+
+                    Stream postStream = new MemoryStream();
+                    postStream.Write(bytestoWrite, 0, bytesToWriteLength);
+                    req.Content = new StreamContent(postStream);
+                    
+                    HttpResponseMessage response = await client.SendAsync(req);
```
ChadNedzlek: This type is disposable too (for the same reasons HttpWebResponse is, and might cause the weird hangs again)

src/Microsoft.DotNet.Build.CloudTestTasks/SendToEventHub.cs
```diff
@@ -4,50 +4,122 @@
...
+                            req.Headers.Add(AzureHelper.AuthorizationHeaderString, ConstructSharedAccessToken());
+                            
+                            contentBytes = Encoding.UTF8.GetBytes(streamReader.ReadToEnd());
+                            using (Stream postStream = new MemoryStream())
+                            {
+                                postStream.Write(this.contentBytes, 0, this.contentBytes.Length);
+                                postStream.Seek(0, SeekOrigin.Begin);
+                                req.Content = new StreamContent(postStream);
+
+                                HttpResponseMessage response = await client.SendAsync(req);
```
ChadNedzlek: Another using

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                            log.LogMessage("Sending request to upload part {0} of file {1}", countForId, fileName);
+
+                            using (Stream postStream = new MemoryStream())
+                            {
+                                postStream.Write(fileBytes, 0, nextBytesToRead);
+                                postStream.Seek(0, SeekOrigin.Begin);
+                                StreamContent contentStream = new StreamContent(postStream);
+                                req.Content = contentStream;
+                                HttpResponseMessage response = await client.SendAsync(req, ct);
```
ChadNedzlek: One more

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -0,0 +1,179 @@
...
+                            bodyData.Length.ToString(),
+                            ""));
+                    using (Stream postStream = new MemoryStream())
+                    {
+                        postStream.Write(bodyData, 0, bodyData.Length);
+                        postStream.Seek(0, SeekOrigin.Begin);
+                        StreamContent contentStream = new StreamContent(postStream);
+                        req.Content = contentStream;
+
+                        HttpResponseMessage response = await client.SendAsync(req, ct);
```
ChadNedzlek: And a last one

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -3,94 +3,169 @@
...
+                    req.Headers.Add(AzureHelper.VersionHeaderString, AzureHelper.StorageApiVersion);
+                    req.Headers.Add(AzureHelper.AuthorizationHeaderString, AzureHelper.AuthorizationHeader(
+                        AccountName,
+                        AccountKey,
+                        "GET",
+                        dt,
+                        req));
+                    
+                    Log.LogMessage(MessageImportance.Normal, "Sending request to check whether Container blobs exist");
+                    HttpResponseMessage response = await client.SendAsync(req, ct);
```
ChadNedzlek: using

src/Microsoft.DotNet.Build.CloudTestTasks/WriteItemsToJson.cs
```diff
@@ -3,15 +3,16 @@
 // See the LICENSE file in the project root for more information.
```
ChadNedzlek: You can probably revert this whole file. None of the changes here look substantive.

karajas: It needed the FileStream change to run on Core.

karajas:
@dotnet-bot test this please

joperezr:
@dotnet-bot test this please

src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
```diff
@@ -4,49 +4,32 @@
...
     <RootNamespace>Microsoft.DotNet.Build.CloudTestTasks</RootNamespace>
     <AssemblyName>Microsoft.DotNet.Build.CloudTestTasks</AssemblyName>
-    <CopyNuGetImplementations>true</CopyNuGetImplementations>
-    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
+    <CLSCompliant>false</CLSCompliant>
     <AssemblyVersion>1.0.0.0</AssemblyVersion>
-    <ResolveNuGetPackages>false</ResolveNuGetPackages>
+    <SkipSigning>true</SkipSigning>
```
jhendrixMSFT: Why are we not signing this binary?

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
```diff
@@ -3,110 +3,168 @@
...
+using Task = Microsoft.Build.Utilities.Task;
 
 namespace Microsoft.DotNet.Build.CloudTestTasks
 {
+
     public sealed class CreateAzureContainer : Task
     {
         /// <summary>
-        /// The Azure account name used when creating the connection string.
+        ///     The Azure account key used when creating the connection string.
```
jhendrixMSFT: Why are there all of these leading white spaces in the comments now?  They are making the diff incredibly noisy.

src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
karajas: This was an older version of the csproj, before the signing fix was checked in. Will update.

src/Microsoft.DotNet.Build.CloudTestTasks/CreateAzureContainer.cs
karajas: Will resolve.

ChadNedzlek:
This change is blocking a bunch of work, and none of the failures appear to be related to the actual change, and we cannot afford to be blocked any longer.

I'm merging it as is.

PR #579, opened 2016-03-25
Use runtime arch info to locate shim libs on machine.

Priya91:
cc @joshfree @stephentoub @weshaggard 

Tested locally on OSX and Windows, works as expected. Will trigger all CI runs on subsequent PR to corefx to test all scenarios.

Reference: dotnet/corefx#7081


src/Microsoft.DotNet.Build.Tasks/GetTargetArchitecture.cs
```diff
@@ -0,0 +1,30 @@
...
+            if (RuntimeInformation.OSArchitecture == Architecture.X64)
+                TargetArch = "x64";
+            else if (RuntimeInformation.OSArchitecture == Architecture.X86)
+                TargetArch = "x86";
+            else if (RuntimeInformation.OSArchitecture == Architecture.Arm)
+                TargetArch = "arm";
+            else if (RuntimeInformation.OSArchitecture == Architecture.Arm64)
+                TargetArch = "arm64";
+            
+            return true;
```
stephentoub: Nit: just for future-proofing, should this be ```return TargetArch != null;```?

stephentoub:
LGTM, but @weshaggard should review whether this is the right approach.

src/Microsoft.DotNet.Build.Tasks/GetTargetArchitecture.cs
```diff
@@ -0,0 +1,30 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetTargetArchitecture : Task
```
weshaggard: Can we combine this and the GetTargetOS tasks? I don't think there is a much value in splitting them. We can call it GetTargetMachineInfo and have a TargetArch and TargetOS output property.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -65,16 +67,24 @@
...
+      <Output TaskParameter="TargetArch" PropertyName="TargetArch" />
+    </GetTargetArchitecture>
+    <GetTargetOS Condition="'$(TestWithLocalLibraries)'=='true' AND '$(OSGroup)'=='AnyOS'">
+      <Output TaskParameter="TargetOS" PropertyName="TargetOS" />
+    </GetTargetOS>
+
     <!-- Test using locally built shim libraries. -->
     <ItemGroup Condition="'$(TestWithLocalLibraries)'=='true'">
-      <TestCopyLocal Include="$(BaseOutputPath)$(OSGroup).$(Platform).$(ConfigurationGroup)/Native/*.*" />
+      <TestCopyLocal Condition="'$(OSGroup)'!='AnyOS'" Include="$(BaseOutputPath)$(OSGroup).$(TargetArch).$(ConfigurationGroup)/Native/*.*" />
```
weshaggard: Why not just set TargetOS to OSGroup and not need to split the property on this condition.

Priya91: We could be cross-building, and may not want to override the native libs.. Although for now, it's not a big deal, as the file extensions are different across Windows, Linux and OSX. This target just creates the test directory, and doesn't actually run tests like in test.targets, hence kept it so.

src/Microsoft.DotNet.Build.Tasks/GetTargetArchitecture.cs
Priya91: Sure.

src/Microsoft.DotNet.Build.Tasks/GetTargetArchitecture.cs
Priya91: Sounds good.

joshfree:
:+1:  thanks @Priya91

src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
```diff
@@ -26,7 +26,7 @@
     <Compile Include="GetPackageVersion.cs" />
     <Compile Include="GetPackageDependencies.cs" />
     <Compile Include="GetNextRevisionNumber.cs" />
-    <Compile Include="GetTargetOS.cs" />
+    <Compile Include="GetTargetMachineInfo.cs" />
```
weshaggard: I don't see this file in the commit.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -65,16 +67,21 @@
...
-     <!-- We may have an indirect package reference that we want to replace with a project reference.
+    <!-- We may have an indirect package reference that we want to replace with a project reference.
           Those are part of RunTestsForProjectInputs. The order that we append to TestCopyLocal is
           significant, later entries override earlier entries. -->
     <ItemGroup>
       <TestCopyLocal Include="@(RunTestsForProjectInputs)" Exclude="@(PackagesConfigs)" />
     </ItemGroup>
+    
+    <GetTargetMachineInfo Condition="'$(TestWithLocalLibraries)'=='true'">
+      <Output TaskParameter="TargetArch" PropertyName="TargetArch" />
```
weshaggard: Can you also add the condition for TargetArch similar to how you did for TargetOS?

src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
Priya91: Added.

src/Microsoft.DotNet.Build.Tasks/GetTargetMachineInfo.cs
```diff
@@ -29,7 +48,7 @@ public override bool Execute()
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.Create("NETBSD")))
                 TargetOS = "NetBSD";
 
-            return true;
+            return TargetArch != null && TargetOS != null;
```
weshaggard: Probably worth an error message here as well saying which one is unknown or not supported.

weshaggard:
LGTM other then the last comment about the error message.

src/Microsoft.DotNet.Build.Tasks/GetTargetMachineInfo.cs
```diff
@@ -29,6 +48,18 @@ public override bool Execute()
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.Create("NETBSD")))
                 TargetOS = "NetBSD";
 
+            if (TargetArch == null)
+            {
+                Log.LogError("{0} is null", nameof(TargetArch));
```
weshaggard: Minor but when updating this in the future we should make these messages more useful. At least say why it is null otherwise folks will still not know why. I'm thinking something simple like "TargetArch/OS is unknown"

PR #581, opened 2016-03-25
Add symbol package creation

dagood:
Symbol packages include pdb files for each dll and any source files inside the project root that are referenced in `<Compile>` items. The nuspec contains all these files and the `NuGetPack` task filters using the same logic as the `nuget pack` command.

This is enabled by default in corefx and wcf but will require a change in TFS to turn on (and a change to add a placeholder to a manually-created nuspec that has symbols but no source.)

Tested successfully on Windows with corefx, wcf, and TFS.

On OSX with corefx no symbol packages are created, presumably because the source file paths aren't handled correctly. Looking into it. I might create an issue and get to it later because getting symbols published from TFS to MyGet seems like the higher priority.

For an example of source and symbol indexing at work, check out https://www.myget.org/feed/dagood-test-dotnet-core/package/nuget/System.Linq.Parallel. (You will need to log in to see it, but you shouldn't need any special permissions.) MyGet should have a symbol source for the feed that can be used from VS--I don't have the newest dotnet core VS tooling set up so I'm still working on testing that out.

/cc @weshaggard @ericstj @gkhanna79 

(Sorry about the NuGetPack diff--I moved some code out into another method and unindented as part of my changes. :cry:)

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
@@ -17,6 +17,14 @@
       <PackageIncludeDocs Condition="'$(PackageIncludeDocs)' == '' AND '$(DocumentationFile)' != ''">true</PackageIncludeDocs>
     </PropertyGroup>
 
+    <!-- Find a root directory that is an ancestor of every possible source file, including
+         ones in bin and Tools. Allow configuration, otherwise prefer TFS root. -->
+    <PropertyGroup>
+      <ProjectRoot Condition="'$(ProjectRoot)'==''">$(SourcesRootPath)</ProjectRoot>
+      <ProjectRoot Condition="'$(ProjectRoot)'==''">$(ProjectDir)</ProjectRoot>
+      <ProjectRootLength>$(ProjectRoot.Length)</ProjectRootLength>
+    </PropertyGroup>
+
     <!-- XmlDocFileRoot should be defined externally since these are currently not
          part of the corefx repo. -->
     <!-- This isn't a straight mapping to an algorithm using CultureInfo.
@@ -116,8 +124,13 @@
 
     <!-- *** include assets *** -->
     <ItemGroup>
+      <!-- Include symbols output -->
+      <SymbolFileToPackage Include="$(PackagePath.Replace('.dll', '.pdb'))" Condition="Exists($(PackagePath.Replace('.dll', '.pdb')))">
+        <IsSymbolFile>true</IsSymbolFile>
+      </SymbolFileToPackage>
+
       <!-- Include primary output -->
-      <FilesToPackage Include="$(PackagePath)">
+      <FilesToPackage Include="$(PackagePath);@(SymbolFileToPackage)">
         <AssemblyVersion>$(AssemblyVersion)</AssemblyVersion>
         <TargetFramework>%(PackageDestination.TargetFramework)</TargetFramework>
         <TargetPath>%(PackageDestination.Identity)</TargetPath>
@@ -139,6 +152,13 @@
       </FilesToPackage>
       <FilesToPackage Condition="'%(FilesToPackage.SubFolder)' != ''">
         <TargetPath>%(TargetPath)%(FilesToPackage.SubFolder)</TargetPath>
+      </FilesToPackage>
+
+      <!-- Include source files for symbol packages. -->
+      <FilesToPackage Include="@(Compile->'%(FullPath)')">
+        <TargetPath>src</TargetPath>
+        <TargetPath Condition="$([System.String]::new('%(FullPath)').StartsWith('$(ProjectRoot)'))">src\$([System.String]::new('%(FullPath)').Substring('$(ProjectRootLength)'))</TargetPath>
+        <IsSourceCodeFile>true</IsSourceCodeFile>
       </FilesToPackage>
     </ItemGroup>
   </Target>
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -63,6 +63,7 @@
   <!-- Shared properties -->
   <PropertyGroup>
     <PackageOutputPath Condition="'$(PackageOutputPath)' == ''">$(BaseOutputPath)pkg/</PackageOutputPath>
+    <SymbolPackageOutputPath Condition="'$(SymbolPackageOutputPath)' == ''">$(BaseOutputPath)symbolpkg/</SymbolPackageOutputPath>
     <OutputPath>$(PackageOutputPath)</OutputPath>
     <NuSpecOutputPath Condition="'$(NuSpecOutputPath)' == ''">$(PackageOutputPath)specs/</NuSpecOutputPath>
     <NuSpecPath>$(NuSpecOutputPath)$(Id)$(NuspecSuffix).nuspec</NuSpecPath>
@@ -330,6 +331,8 @@
                       '%(File.SkipPackageFileCheck)' != 'true' AND
                       '%(File.FileName)' != '_' AND
                       '%(File.FileName)%(File.Extension)' != 'runtime.json' AND
+                      '%(File.Extension)' != '.cs' AND
+                      '%(File.Extension)' != '.vb' AND
                       !$(ID.Contains('%(File.FileName)'))"
            Text="Package $(ID) contains file with name %(File.FileName).  If this is expected you can disable this filename checking for this item or package by setting SkipPackageFileCheck = true"
            ContinueOnError="ErrorAndContinue" />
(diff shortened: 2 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -140,6 +140,12 @@
       <FilesToPackage Condition="'%(FilesToPackage.SubFolder)' != ''">
         <TargetPath>%(TargetPath)%(FilesToPackage.SubFolder)</TargetPath>
       </FilesToPackage>
+
+      <!-- Include source files for symbol packages. -->
```
ericstj: Why aren't we including pdbs here?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -140,6 +140,12 @@
...
+      <FilesToPackage Include="@(Compile->'%(FullPath)')" Condition="$([System.String]::new('%(FullPath)').StartsWith('$(ProjectDir)'))">
+        <TargetPath>src\$([System.String]::new('%(FullPath)').Replace('$(ProjectDir)', ''))</TargetPath>
```
ericstj: Have you looked at [$([MSBuild]::MakeRelative())](https://msdn.microsoft.com/en-us/library/dd633440.aspx#Anchor_6)?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -411,12 +414,38 @@
...
-      <PackageFile Include="@(File)" />
+      <!-- Include all files except source files. Sources need to be deduplicated. -->
+      <PackageFile Include="@(File)" Condition="'%(File.IsSourceCodeFile)'!='true'" />
       <PackageFile Condition="'%(PackageFile.PackageId)' == ''">
         <PackageId>$(Id)</PackageId>
         <PackageVersion>$(Version)</PackageVersion>
       </PackageFile>
 
+      <!-- Include Symbols -->
+      <PackageFileSymbol Condition="$([System.String]::new('%(PackageFile.Identity)').EndsWith('.dll'))"
```
ericstj: I'd prefer if we had packageLibs.targets do the symbol inclusion.  We can avoid complexity here if we just flow it through from the library project.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -896,6 +940,8 @@
     <NugetPack Nuspecs="$(NuSpecPath)"
                OutputDirectory="$(PackageOutputPath)"
                ExcludeEmptyDirectories="true"
+               PackSymbolPackage="true"
+               SymbolPackageOutputDirectory="$(SymbolPackageOutputPath)"
                Condition="'$(_SkipCreatePackage)' != 'true'"/>
     <!-- Create a marker that records the path to the generated package -->
     <WriteLinesToFile Lines="$(PackageOutputPath)$(Id).$(Version).nupkg"
```
ericstj: We need to write the symbol package path as well so that it gets cleaned up on /t:clean

ericstj:
Nice! Didn't know you were working on this.  I suspect the NuGetPack differences are mostly just a port from https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/PackCommandRunner.cs.  Let me know if there's any significant diffs worth reviewing.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -140,6 +140,12 @@
       <FilesToPackage Condition="'%(FilesToPackage.SubFolder)' != ''">
         <TargetPath>%(TargetPath)%(FilesToPackage.SubFolder)</TargetPath>
       </FilesToPackage>
+
+      <!-- Include source files for symbol packages. -->
+      <FilesToPackage Include="@(Compile->'%(FullPath)')" Condition="$([System.String]::new('%(FullPath)').StartsWith('$(ProjectDir)'))">
```
weshaggard: What about the source code not under the project directory? Is the expectation that the source code will be reference only and not build able?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
dagood: Yeah, I used that for a while but it breaks OSX: https://github.com/Microsoft/msbuild/issues/429. (It also has the risk of including `..\..\..` if there are any code files coming from `ToolRuntime` on TFS, although I changed this before testing TFS so I don't know for sure if the paths lead there.)

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
dagood: I see, yeah, this was part of @chcosta's prototype of making the symbol packages and I didn't think to change it after figuring out `PackageLibs.targets`.

On that note I forgot to cc you on the review @chcosta, sorry!

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
dagood: I only saw this file when trying to add sources, didn't think of moving the pdb include here. I'll see if I can change it.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
dagood: Hmm, for corefx and wcf this means the root of the repo (so it would include sources under `Tools` and `bin`), and I didn't think much about the TFS case. I suppose for that I can find a different property that changes internally to the root of the enlistment and use that as the root here.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
dagood: Ah, I'll add that. Does it matter if the line has a file that doesn't exist? (Wondering if I need to handle the case where `NugetPack` decides not to create a package because there are no symbols or sources.)

dagood:
@ericstj The code is from https://github.com/NuGet/NuGet.Client/blob/afc65ae26b6f79b85f7a8deddab18c2cc3182b70/src/NuGet.Clients/NuGet.CommandLine/Commands/PackCommand.cs#L24, but that part was really just a few lines. (I was able to reuse a class that provided the heavy logic.) I'll see if I can generate a better diff on a new branch so you can see the changes I actually did without the move cluttering things.

dagood:
@ericstj Ok, if you check out https://github.com/dotnet/buildtools/compare/master...dagood:nugetpack-cleardiff you can see two commits: the move and the changes to bring to this PR. The move really doesn't want to show up clear in a whitespace-sensitive diff, but the other changes show up great: https://github.com/dotnet/buildtools/commit/11318e775229fdd38bdc6e9e7ed671844e596a5b

ericstj:
Thanks, that diff looks reasonable.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: It shouldn't, delete task skips missing files.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: Just wish this didn't use string comparison, that feels fragile and the mode of failure is to just drop what isn't understood.  Seems to me we should fix the unix behavior of URI...
Barring that I'd say lets have the inverse of this condition drop the src files in the root src\ directory.  Also, use substring rather than replace since you already know the thing you are removing is at the start of the string.  Can you double check that ProjectDir will always be normalized, otherwise create a new property that you've normalized with GetFullPath.

dagood: Per my comment about source not under project dir, I'm switching to use `SourcesRootPath` or `ProjectDir`, and both of those are full paths. (`SourcesRootPath` manually, and `ProjectDir` is `MSBuildThisFileDirectory` which is a full path.) That should mean that no source files are dropped, but agreed, I'll make it fall back to dropping files directly in `src`. I'll make the substring change too, good catch.

dagood:
Pushed new changes to address comments. To test the project base dir changes I tried forcing a Compile include to a file on a different drive and it ended up in the package's `src` dir. Also, in TFS, I see lines like these:

```XML
<file src="C:\ff\ProjectK\Intermediate\Open\CoreFx\ref\System.Net.Primitives\4.0.10.0\_AssemblyInfo.cs" target="src\Intermediate\Open\CoreFx\ref\System.Net.Primitives\4.0.10.0\_AssemblyInfo.cs" exclude="" />
<file src="C:\ff\ProjectK\src\ndp\FxCore\Open\src\System.Net.Primitives\ref\System.Net.Primitives.cs" target="src\src\ndp\FxCore\Open\src\System.Net.Primitives\ref\System.Net.Primitives.cs" exclude="" />
```

Which suggest that it's working fine there.

@weshaggard @ericstj Could you take another look?

dagood:
Ping! @weshaggard @ericstj Could you take another look? Thanks!

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -140,6 +153,13 @@
       <FilesToPackage Condition="'%(FilesToPackage.SubFolder)' != ''">
         <TargetPath>%(TargetPath)%(FilesToPackage.SubFolder)</TargetPath>
       </FilesToPackage>
+
+      <!-- Include source files for symbol packages. -->
+      <FilesToPackage Include="@(Compile->'%(FullPath)')">
+        <TargetPath>src</TargetPath>
```
weshaggard: Is there ever any case where a source file isn't somewhere under ProjectRoot? If there are is my assumption that they just end up in the root src target folder? Will the pdb indexing of the source code actually be affected by that?

weshaggard: I guess one case where that might happen is for generated files under intermediates, like the assemblyinfo.cs file we generate.

weshaggard:
Is it safe to assume that the these packages will get the same description and the matching library package? If not we should be sure to include metadata like the SHA of the commit associated with the source files and such in this package as well.


Overall LGTM just a couple questions.

dagood:
@weshaggard Yep, there is actually only one `nuspec` created that is used to pack both the normal and symbol package, so the contents are identical except for those file filters declared in `NuGetPack.cs`.

dagood:
@ericstj Did you want to take another look at this? It'd be really neat to get this through so we can start putting symbols on dotnet.myget.org.

weshaggard:
@dagood go ahead and get this merged. @ericstj is out at BUILD this week.

dagood:
Thanks Wes, didn't think about that. :smile: Merging.

PR #582, opened 2016-03-25
Adding BuildNumber Minor since VSO can't pass in the Revision Number

joperezr:
VSO cannot pass in the Revision number directly, so with this we are stripping it out from the BuildNumber

cc: @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
@@ -6,6 +6,7 @@ using System;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using System.Globalization;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
@@ -16,6 +17,11 @@ namespace Microsoft.DotNet.Build.Tasks
         /// </summary>
         [Required]
         public string SeedDate { get; set; }
+
+        /// <summary>
+        /// Optional parameter containing the Official Build Id. We'll use this to get the revision number out and use it as BuildNumberMinor.
+        /// </summary>
+        public string OfficialBuildId { get; set; }
 
         /// <summary>
         /// Optional parameter that sets the Padding for the version number. Must be 5 or bigger.
@@ -33,12 +39,29 @@ namespace Microsoft.DotNet.Build.Tasks
         [Output]
         public string GeneratedVersion { get; set; }
 
+        /// <summary>
+        /// The Revision number that will be produced from the BuildNumber.
+        /// </summary>
+        [Output]
+        public string GeneratedRevision { get; set; }
+
         private const string DateFormat = "yyyy-MM-dd";
         private const string LastModifiedTimeDateFormat = "yyyy-MM-dd HH:mm:ss.FFFFFFF";
         private CultureInfo enUS = new CultureInfo("en-US");
 
         public override bool Execute()
         {
+            // Calculating GeneratedRevision
+            if (string.IsNullOrEmpty(OfficialBuildId))
+            {
+                GeneratedRevision = "00";
+            }
+            else
+            {
+                GeneratedRevision = GetCurrentRevisionForBuildName(OfficialBuildId);
+            }
+
+            // Calculating GeneratedVersion
             if (Padding == 0)
             {
                 Padding = 5;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -18,6 +19,11 @@ public sealed class GenerateCurrentVersion : Task
         public string SeedDate { get; set; }
 
         /// <summary>
+        /// Optional parameter containing the BuildNumber. We'll use this to get the revision number out and use it as BuildNumberMinor.
+        /// </summary>
+        public string BuildNumber { get; set; }
```
weshaggard: Lets call this OfficialBuildId

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -70,6 +93,17 @@ public override bool Execute()
...
 
+        public string GetCurrentRevisionForBuildName(string buildName)
+        {
+            Regex regex = new Regex(@"^\d*\.(\d{2})$");
+            Match match = regex.Match(buildName);
+            if (match.Success && match.Groups.Count > 1)
+            {
+                return match.Groups[1].Value;
+            }
+            return "00";
```
weshaggard: `\.\d+$` that should be enough for the regex. Thinking a little more if this doesn't doesn't exist lets default to just "0" that way we can actually use it as an indicator for something different for local builds.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -234,6 +236,7 @@
   <Target Name="CreateVersionFileDuringBuild" Condition="'$(SkipVersionGeneration)'!='true' AND '$(ShouldCreateVersionFileDuringBuild)'=='true'" DependsOnTargets="CreateOrUpdateCurrentVersionFile">
     <PropertyGroup>
       <BuildNumberMajor>$(GeneratedBuildNumberMajor)</BuildNumberMajor>
```
weshaggard: It doesn't look like this actually respects if someone explicitly sets BuildNumberMajor or BuildNumberMinor. We should probably make these conditional on not already being set in some way, so we can freeze these in our servicing branches if we need to.

joperezr: We can't do that, because this Names will already be set always. They are both set when this file is loaded (versioning.targets) so at the time of the target runs they will always be set already. For servicing, it will be easy enough to just pass in both versions with `msbuild /p:BuildNumberMajor=...` or we could also simply just set `/p:SkipVersionGeneration=true`

weshaggard: We will need to think through this some more but I was kind of hoping we could check-in something for locking the versions in servicing and not require the correct things be passed in to the build.

joperezr: If you want to checkin something, then you can still accomplish this by adding this to dir.props:
```xml
<PropertyGroup>
  <BuildNumberMajor>00001</BuildNumberMajor>
  <BuildNumberMinor>00</BuildNumberMinor>
  <SkipVersionGeneration>true</SkipVersionGeneration>
</PropertyGroup>
```

weshaggard: But if you do that won't it get overridden here?

joperezr: No, the assignments we do here are inside a target, and that target will get skipped if SkipVersionGeneration is true

weshaggard: Ah I see yes if you checked in the setting of the property as well it would work. Thanks.

joperezr: Ok makes sense, with my last commit c5e7804 I added the condition to the property group as well.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -33,12 +39,29 @@ public sealed class GenerateCurrentVersion : Task
...
         private const string DateFormat = "yyyy-MM-dd";
         private const string LastModifiedTimeDateFormat = "yyyy-MM-dd HH:mm:ss.FFFFFFF";
         private CultureInfo enUS = new CultureInfo("en-US");
 
         public override bool Execute()
         {
+            // Calculating GeneratedRevision
+            if (string.IsNullOrEmpty(OfficialBuildId))
+            {
+                GeneratedRevision = "00";
```
weshaggard: We should also update this to be "0" and our targets file IIRC you had an instance of that.

PR #584, opened 2016-03-25
Updating BuildTools to get RevisionNumber

joperezr:
With this last change, Buildtools will me moved to using the new versioning scheme.

cc: @weshaggard

### DIFF
--- a/BuildToolsVersion.txt
+++ b/BuildToolsVersion.txt
@@ -1,1 +1,1 @@
-1.0.25-prerelease-00221
+1.0.25-prerelease-00225-00


BuildToolsVersion.txt
```diff
@@ -1 +1 @@
-1.0.25-prerelease-00221
\ No newline at end of file
+1.0.25-prerelease-00225-00
```
weshaggard: Wouldn't it be "-01"? I thought the build revs always started with 1.

joperezr: It will after I merge this change. Build tools repo was still using the old version of the buildtools package on the previous official build, and in that one BuildVersionMinor didn't work yet so we were defaulting to 00.

PR #585, opened 2016-03-25
Removing BuildValues.props and the targets that modify it.

joperezr:
We don't need this anymore with our new versioning scheme. I have already changed the internal build to not pass UpdateBuildValues anymore.

cc: @weshaggard @ericstj


weshaggard:
cc @jhendrixMSFT

joperezr:
Ping

weshaggard:
LGTM we should also consider cleaning out the supporting infrastructure (i.e. custom targets and tasks), that generates this.

joperezr:
Was this repo the only buildtools consumer that uses the infrastructure, I left it in case there were others that were using revisions for their versioning, but I can remove that later if nobody uses it.

weshaggard:
@jhendrixMSFT would have to say for sure but I believe it was the only repo using it.

PR #587, opened 2016-03-28
Add targets file to be used for publishing build artifacts.

jhendrixMSFT:
The build pipeline requires the ability to publish build artifacts to
various endpoints.  This change adds a targets file that will contain
various publishing targets.  At present the only endpoint we need is Azure
blob storage.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
@@ -1,0 +1,44 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <UsingTask TaskName="CreateAzureContainer" AssemblyFile="$(ToolsDir)net45/Microsoft.DotNet.Build.CloudTestTasks.dll"/>
+  <UsingTask TaskName="UploadToAzure" AssemblyFile="$(ToolsDir)net45/Microsoft.DotNet.Build.CloudTestTasks.dll"/>
+
+  <PropertyGroup>
+    <OverwriteOnPublish Condition="'$(OverwriteOnPublish)' == ''">false</OverwriteOnPublish>
+  </PropertyGroup>
+
+  <!-- gathers the items to be published -->
+  <Target Name="GatherItemsForPattern">
+    <Error Condition="'$(PublishPattern)' == ''" Text="Please specify a value for PublishPattern using standard msbuild 'include' syntax." />
+    <ItemGroup>
+      <ForPublishing Include="$(PublishPattern)" />
+    </ItemGroup>
+    <Error Condition="'@(ForPublishing)' == ''" Text="No items were found matching pattern '$(PublishPattern)'." />
+  </Target>
+
+  <!-- publishes items to blob storage in Azure -->
+  <Target Name="UploadToAzure" DependsOnTargets="GatherItemsForPattern">
+    <Error Condition="'$(ContainerName)' == ''" Text="Missing property ContainerName." />
+    <Error Condition="'$(CloudDropAccountName)' == ''" Text="Missing property CloudDropAccountName." />
+    <Error Condition="'$(CloudDropAccessToken)' == ''" Text="Missing property CloudDropAccessToken." />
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@weshaggard @chcosta Here's the change on the build tools side for publishing artifacts to blob storage.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
```diff
@@ -0,0 +1,40 @@
...
+  <UsingTask TaskName="UploadToAzure" AssemblyFile="$(ToolsDir)net45/Microsoft.DotNet.Build.CloudTestTasks.dll"/>
+
+  <PropertyGroup>
+    <OverwriteOnPublish Condition="'$(OverwriteOnPublish)' == ''">false</OverwriteOnPublish>
+  </PropertyGroup>
+
+  <!-- gathers the items to be published -->
+  <Target Name="GatherItems">
+    <ItemGroup>
+      <ForPublish Include="$(Pattern)" />
```
weshaggard: Where is Pattern coming from? We should probably give it a more specific name. Perhaps GatherItemsPattern or PublishContentPattern or something to connect the dots.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
```diff
@@ -0,0 +1,40 @@
...
+    </ItemGroup>
+    <Error Condition="'@(ForPublish)' == ''" Text="No items were found matching pattern '$(Pattern)'." />
+  </Target>
+
+  <!-- publishes items to blob storage in Azure -->
+  <Target Name="UploadToAzure" DependsOnTargets="GatherItems">
+    <!-- create the container if it doesn't exist -->
```
weshaggard: I assume this is a no-op if it already exists?

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
```diff
@@ -0,0 +1,40 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <UsingTask TaskName="CreateAzureContainer" AssemblyFile="$(ToolsDir)net45/Microsoft.DotNet.Build.CloudTestTasks.dll"/>
```
weshaggard: Minor but something to think about... we are now using tasks from a CloudTestTasks library for non-Test related scenarios. We might want to think about how we factor and name these at some point.

joperezr:
@dotnet-bot test this please

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
jhendrixMSFT: Correct although you can change this behavior (see attr FailIfExists).

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
jhendrixMSFT: I was thinking about that, perhaps as part of Karthik's work we can move these general-purpose tasks into build tools proper.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
jhendrixMSFT: Pattern is provided by the caller, e.g. "/p:Pattern=D:\git\buildtools\bin\packages\**\*.nupkg".  I will tweak these names a bit.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
weshaggard: Either that or we just rename the task library to be something more generic.

weshaggard:
@dotnet-bot test this please

weshaggard:
LGTM

PR #588, opened 2016-03-28
Support Desktop test Build and Run

tarekgh:
This change is to add the necessary test publishing filters and restore the xunit desktop runner pack to allow compiling and running the tests against the desktop

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -67,6 +67,15 @@
       <Output TaskParameter="ResolvedCopyLocalItems" ItemName="TestCopyLocal" />
     </PrereleaseResolveNuGetPackageAssets>
 
+    <ItemGroup Condition="'$(TestWithCore)' != 'true'">
+      <!-- Filter out the copy-local set of references coming from the targeting pack packages -->
+      <TestCopyLocalToRemove Include="@(TestCopyLocal)"
+        Condition="$([System.String]::new('%(TestCopyLocal.NuGetPackageId)').StartsWith('$(_TargetingPackPrefix)'))" />
+      <TestCopyLocal Remove="@(TestCopyLocalToRemove)" />
+      
+      <TestCopyLocal Include="$(PackagesDir)xunit.runner.console/2.1.0/tools/*.*" />
+    </ItemGroup>
+
     <!-- We may have an indirect package reference that we want to replace with a project reference.
           Those are part of RunTestsForProjectInputs. The order that we append to TestCopyLocal is
           significant, later entries override earlier entries. -->
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/test-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/test-runtime/project.json
@@ -1,20 +1,27 @@
 {
   "dependencies": {
-    "Microsoft.NETCore.Platforms": "1.0.1-rc2-23923",
-    "Microsoft.NETCore.TestHost": "1.0.0-rc2-23923",
-    "Microsoft.NETCore.Console": "1.0.0-rc2-23923",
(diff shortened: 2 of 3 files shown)


tarekgh:
@weshaggard

tarekgh:
@weshaggard please have a quick look.

tarekgh:
@JeremyKuhne

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -67,6 +67,15 @@
       <Output TaskParameter="ResolvedCopyLocalItems" ItemName="TestCopyLocal" />
     </PrereleaseResolveNuGetPackageAssets>
 
+    <ItemGroup Condition="'$(TestWithCore)' != 'true'">
+      <!-- Filter out the copy-local set of references coming from the targeting pack packages -->
+      <TestCopyLocalToRemove Include="@(TestCopyLocal)"
+        Condition="$([System.String]::new('%(TestCopyLocal.NuGetPackageId)').StartsWith('$(_TargetingPackPrefix)'))" />
+      <TestCopyLocal Remove="@(TestCopyLocalToRemove)" />
+      
+      <TestCopyLocal Include="$(PackagesDir)xunit.runner.console/2.1.0/tools/*.*" />
```
JeremyKuhne: Can we avoid the hard coded path here or put a define somewhere that will be easier to find when we update to a newer xunit runner? Or at least a comment in the project.json that this needs updated?

tarekgh: This is the xunit runner package which is not referenced in the project. so er don't have the real path. I believe Wes mentioned to me we can keep that for now but he has some idea to change this later.

JeremyKuhne:
lgtm

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
weshaggard: We should file a tracking issue to fix this as I already know the hardcoded path with the version is going to be a pain to keep updated and we are trying really hard to avoid such versions in the targets. If we need it we should parameterize it into a property and share it with the package validation task we have in dir.props.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -20,6 +20,10 @@
   </ItemGroup>
 
   <PropertyGroup>
+    <TestWithCore Condition="'$(TestWithCore)' == '' And '$(TargetGroup)' == 'net45'">false</TestWithCore>
```
weshaggard: We should consider moving these propertyes and the TestFramework ItemGroup below into dir.props so that we can try to keep the TargetGroup conditions localized there and setup the specifics in a common location.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/test-runtime/project.json
```diff
@@ -1,20 +1,27 @@
...
     "dnxcore50": {
+      "dependencies": {
+        "Microsoft.NETCore.Platforms": "1.0.1-rc2-23923",
+        "Microsoft.NETCore.TestHost": "1.0.0-rc2-23923",
+        "Microsoft.NETCore.Console": "1.0.0-rc2-23923",
+        "xunit.console.netcore": "1.0.2-prerelease-00101"
+      },
       "imports": "portable-net45+win8"
+    },
+    "net462": {
```
weshaggard: You used net462 here but you seem to be adding support for net45+, you should change this to net45 as well to ensure it will work on the lower versions also.

weshaggard:
Overall looks good but I've suggested some clean-up.

tarekgh:
@weshaggard thanks for review. I have files the issue https://github.com/dotnet/buildtools/issues/593 to track the clean up in general here. will do this clean up later after finishing the Helix support

PR #589, opened 2016-03-29
Add validation report and doc generation task.

ericstj:
Add functionality to emit some of the results from package validation to
use for generating documentation.

I'm not currently using the emitted AllSupportedFrameworks items but we
could use this in the future to append a list of supported frameworks to
the package description.  Folks have various opinions about what is the
right amount of data to show here, until we have some concensus I'm
going to avoid shoving more data into the publicly facing package
description.

I'll follow up to a pointer of the CoreFx change that consumes this.


weshaggard:
LGTM

PR #590, opened 2016-03-29
Create a common ruleset to be used at a repo level for changing analyzer behavior

AlexGhiondea:
By default, all analyzers are going to be enabled and their default level (error, warning) will be specified by each analyzer as it sees fit.
If you need to change the default behavior at the repo level, you can update the Microsoft.DotNet.CodeAnalysis.ruleset file and modify the level as needed.

If individual projects need to change the defaults for analyzers they can do so in their own ruleset file.
Each project can also import the common ruleset file.

/cc @pallavit @joperezr @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
+++ b/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
@@ -18,6 +18,6 @@
   </ItemGroup>
 
   <PropertyGroup>
-  	<CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSet Condition="'$(CodeAnalysisRuleSet)'==''">$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 </Project>
--- a/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
+++ b/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
@@ -16,4 +16,8 @@
     <AdditionalFiles Include="$(MSBuildProjectDirectory)/*.analyzerdata" />
     <AdditionalFiles Include="$(MSBuildProjectDirectory)/*.analyzerdata.$(Platform)" />
   </ItemGroup>
+
+  <PropertyGroup>
+  	<CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
 </Project>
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -7,4 +7,8 @@
     <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata" />
     <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata.$(Platform)" />
   </ItemGroup>
+
+  <PropertyGroup>
+  	<CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis.ruleset</CodeAnalysisRuleSet>
```
weshaggard: This should be conditional so that individual repo's can override it in their repo.

AlexGhiondea: Individual projects can already override it. I will make it such that this won't overwrite it if something is already set.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -7,4 +7,8 @@
     <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata" />
```
joperezr: Can we switch the backslashes in this file to forward slashes? The reason is that in Linux backslashes are permited as a file name, so sometimes this has unexpected results.

AlexGhiondea: Thanks for pointing that out! I have a different PR #595 that also addresses that!

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
pallavit: Also, would it not be good to have OS specific rule sets?

AlexGhiondea: If we do, they can use the $(OverrideCodeAnalysisRuleSet) to specify which ruleset to use for that OS.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -7,4 +7,10 @@
...
     <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata.$(Platform)" />
   </ItemGroup>
+
+  <PropertyGroup>
+  	<!-- We need this additional property because by default msbuild will populate $(CodeAnalysisRuleSet) to a non-empty value -->
+    <OverrideCodeAnalysisRuleSet Condition="'$(OverrideCodeAnalysisRuleSet)'==''">$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis.ruleset</OverrideCodeAnalysisRuleSet>
```
weshaggard: I would like to avoid creating yet another property here if we can avoid it. I would much perfer we check for empty and only set it if it isn't already set.

weshaggard:
LGTM

PR #592, opened 2016-03-30
Upgrade dotnet cli version

dagood:
See https://github.com/dotnet/corefx/pull/7353.

/cc @eerhardt @ericstj


dagood:
@stephentoub @joshfree This one should be good too.

PR #595, opened 2016-03-30
Make sure code analyzers can be correctly loaded on all supported platforms.

AlexGhiondea:
There were a couple of issues with the way analyzers were enabled:
- The net core version of the analyzer was placed in the root of the Tools directory. This meant that it became part of the TPA (Trusted Platform Assemblies) list when the net core compiler was run with corerun. 
 - This lead to a very mysterious issue where the runtime complained that it could not load the analyzer file because it could not find its dependencies.
 - In reality, this was caused by the fact that the assembly was part of the TPA list and the way the Roslyn compiler loads assemblies.
 - **Fix: Move the analyzer in a separate folder under Tools to prevent it from showing up in the TPA list.**
- The property that was used to decide the path where the assembly should be loaded from was `$(BuildToolsTaskDir)`
 - This property however is fixed to be net45. This was needed at the time because we still depended on Mono to build our binaries on non Windows platform. 
 - This is still required today because we still use Mono to run our tests.
 - **Fix: Temporarily introduce a new property that uses `$(OSEnvironment)` to decide if it should use the net45 version or the net core version for the analyzer**

Note: During the investigation I noticed that the way that we load the `Microsoft.DotNet.Build.Tasks.dll` happens to work because of an implementation detail in how assemblies are being loaded.
What happens is that when we are running on the net core msbuild, we still pass it the path to the net45 assembly. However, because there is already a file with the same simple name next to corerun, we are going to end up loading that assembly instead of the one we specified. 
This is something we need to address in the future.

### DIFF
--- a/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
+++ b/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
@@ -2,7 +2,7 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Condition="'$(EnableDotnetAnalyzers)' == 'true'">
     <!-- Enable the Microsoft.Dotnet.CodeAnalysis.dll analyzer --> 
-    <Analyzer Include="$(BuildToolsTaskDir)Microsoft.Dotnet.CodeAnalysis.dll"/>
+    <Analyzer Include="$(BuildToolsTaskDir)analyzers\Microsoft.Dotnet.CodeAnalysis.dll"/>
 
     <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata" />
     <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata.$(Platform)" />
--- a/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
+++ b/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
@@ -1,10 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- 
+    We should be using $(BuildToolsTaskDir) for this, however we are currently forcing that property to 
+    always be net45 (because we use Mono to run our tests and Mono needs to load the net45 version)
+-->
+  <PropertyGroup>
+    <CodeAnalyzerDir Condition="'$(CodeAnalyzerDir)'=='' and '$(OSEnvironment)'=='Windows_NT'">$(ToolsDir)net45/analyzers/</CodeAnalyzerDir>
+    <CodeAnalyzerDir Condition="'$(CodeAnalyzerDir)'=='' and '$(OSEnvironment)'!='Windows_NT'">$(ToolsDir)analyzers/</CodeAnalyzerDir>
+  </PropertyGroup>
+  
   <ItemGroup Condition="'$(EnableDotnetAnalyzers)' == 'true'">
     <!-- Enable the Microsoft.Dotnet.CodeAnalysis.dll analyzer --> 
-    <Analyzer Include="$(BuildToolsTaskDir)analyzers\Microsoft.Dotnet.CodeAnalysis.dll"/>
+    <Analyzer Include="$(CodeAnalyzerDir)Microsoft.DotNet.CodeAnalysis.dll"/>
 
-    <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata" />
-    <AdditionalFiles Include="$(MSBuildProjectDirectory)\*.analyzerdata.$(Platform)" />
+    <AdditionalFiles Include="$(MSBuildProjectDirectory)/*.analyzerdata" />
+    <AdditionalFiles Include="$(MSBuildProjectDirectory)/*.analyzerdata.$(Platform)" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -1,10 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- 
+    We should be using $(BuildToolsTaskDir) for this, however we are currently forcing that property to 
+    always be net45 (because we use Mono to run our tests and Mono needs to load the net45 version)
+-->
+  <PropertyGroup>
+    <CodeAnalyzerDir Condition="'$(CodeAnalyzerDir)'=='' and '$(OSEnvironment)'=='Windows_NT'">$(ToolsDir)/net45/</CodeAnalyzerDir>
```
joperezr: shouldn't this be `$(ToolsDir)/net45/analyzers/`

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -1,10 +1,19 @@
...
+    <CodeAnalyzerDir Condition="'$(CodeAnalyzerDir)'=='' and '$(OSEnvironment)'!='Windows_NT'">$(ToolsDir)/</CodeAnalyzerDir>
+  </PropertyGroup>
+  
   <ItemGroup Condition="'$(EnableDotnetAnalyzers)' == 'true'">
     <!-- Enable the Microsoft.Dotnet.CodeAnalysis.dll analyzer --> 
-    <Analyzer Include="$(BuildToolsTaskDir)Microsoft.Dotnet.CodeAnalysis.dll"/>
+    <Analyzer Include="$(CodeAnalyzerDir)analyzers/Microsoft.DotNet.CodeAnalysis.dll"/>
```
joperezr: Oh I see, looks like you are going to the analyzer in here, I think it would be better if your CodeAnalyzerDir already points to the analyzers dir instead of pointing to the tools dir and having to reference it here.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
AlexGhiondea: The /analyzers part is added below.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -1,10 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- 
+    We should be using $(BuildToolsTaskDir) for this, however we are currently forcing that property to 
+    always be net45 (because we use Mono to run our tests and Mono needs to load the net45 version)
+-->
+  <PropertyGroup>
+    <CodeAnalyzerDir Condition="'$(CodeAnalyzerDir)'=='' and '$(OSEnvironment)'=='Windows_NT'">$(ToolsDir)/net45/</CodeAnalyzerDir>
+    <CodeAnalyzerDir Condition="'$(CodeAnalyzerDir)'=='' and '$(OSEnvironment)'!='Windows_NT'">$(ToolsDir)/</CodeAnalyzerDir>
```
joperezr: also nit picking I believe ToolsDir already ends with a trailing slash.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
AlexGhiondea: I can do that, but that means duplicating the addition of the /analyzer suffix. And when we change it, we need to change in 2 places. I though that was the 'standard' way of doing things.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
AlexGhiondea: Will change! :)

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
joperezr: You can leave it if you think it's best, I just don't think it makes sense that it is called CodeAnalyzerDir but it doesn't point to the CodeAnalyzer dir. If anybody for any reason needs to reference the analyzers in any other way, they would have to remember that this is not really pointing where the name says.

AlexGhiondea: This should hopefully go away soon and this should revert to being $(BuildToolsTaskDir).

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
weshaggard: I would actually include the /ananlyzers part here as well and that way we can still keep this property later once we can redirect it to BuildToolsTasksDir.

joperezr:
LGTM

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -1,10 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- 
+    We should be using $(BuildToolsTaskDir) for this, however we are currently forcing that property to 
+    always be net45 (because we use Mono to run our tests and Mono needs to load the net45 version)
+-->
```
weshaggard: nit: indention

weshaggard:
LGTM

PR #597, opened 2016-03-30
Add win-api set dependency for win7.

Priya91:
cc @weshaggard @stephentoub @joshfree 

fixes dotnet/corefx#7294

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -5,6 +5,7 @@ set PROJECT_DIR=%~1
 set DOTNET_CMD=%~2
 set TOOLRUNTIME_DIR=%~3
 IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
+IF [%BUILDTOOLS_NET45_TARGET_RUNTIME%]==[] set BUILDTOOLS_NET45_TARGET_RUNTIME=win7-x86
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
 set MICROBUILD_VERSION=0.2.0
 set PORTABLETARGETS_VERSION=0.1.1-dev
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -5,6 +5,7 @@ set PROJECT_DIR=%~1
 set DOTNET_CMD=%~2
 set TOOLRUNTIME_DIR=%~3
 IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
+IF [%BUILDTOOLS_NET45_TARGET_RUNTIME%]==[] set BUILDTOOLS_NET45_TARGET_RUNTIME=win7-x86
```
mellinoe: Doesn't this need to be x64 if you're on a 64-bit machine? Or are always running 32-bit MSBuild?

weshaggard: I think in most cases it is x86 but I'm not certain it is in all cases.

weshaggard:
cc @joperezr 

Looks reasonable to me but it does mean that we are always publishing for both now. I wonder if we can figure out what to publish for and only do one?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
Priya91: I dont think there's a 64 bit MSBuild.. MSBuild and VS are 32bit.

Priya91:
>  I wonder if we can figure out what to publish for and only do one?

Depending on if MSBuild is used from Tools?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -26,6 +27,7 @@ ROBOCOPY "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%" /E
 cd "%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\"
 call "%DOTNET_CMD%" restore --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json
 call "%DOTNET_CMD%" publish -f dnxcore50 -r %BUILDTOOLS_TARGET_RUNTIME% -o "%TOOLRUNTIME_DIR%"
+call "%DOTNET_CMD%" publish -f net45 -r %BUILDTOOLS_NET45_TARGET_RUNTIME% -o "%TOOLRUNTIME_DIR%\net45"
```
joperezr: Will this not conflict with any of the libraries that we already have on that folder?

Priya91: the tool-runtime project.json has only the api-sets as dependency for net45. So this will publish only the api sets.. Verified on win7, is there some path i am missing?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
mellinoe: MSBuild is just a .NET application, though. Can it not be run 64-bit?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: oh I didn't see that it's only the api-sets, that should be fine.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
Priya91: MSBuild is a 32 bit app, so it is run under wow in 64 bit process.

Priya91:
@weshaggard @joperezr @mellinoe Are there any other concerns, before merge?

joperezr:
LGTM

weshaggard:
LGTM 

I do think we might need to revisit the always publishing of net45 at some point but I'm good with it for now.

PR #601, opened 2016-03-31
We should only overwrite the CodeAnalysisRuleSet property if the analyzers are enabled

AlexGhiondea:
When enabling the analyzers I missed a condition that would not overwrite the CodeAnalysisRuleset if the analyzers were not enabled.


weshaggard:
LGTM

joperezr:
LGTM

PR #602, opened 2016-04-01
Modify checks for IsOfficial CloudTest build

karajas:
This is to update the checks on the properties passed in to allow builds to be marked as "official".
@ChadNedzlek

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
@@ -64,9 +64,9 @@
     <Error Condition="'$(EventHubPath)' == ''" Text="Missing required property EventHubPath." />
     <Error Condition="'$(EventHubSharedAccessKey)' == ''" Text="Missing required property EventHubSharedAccessKey." />
     <Error Condition="'$(EventHubSharedAccessKeyName)' == ''" Text="Missing required property EventHubSharedAccessKeyName." />
-    <Error Condition="'$(BuildIsOfficial)' == 'true' and '$(BuildIsOfficialConnection)' == ''" Text="Missing required property BuildIsOfficialConnection." />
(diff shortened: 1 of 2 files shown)


dsgouda:
:+1:  LGTM If you have confirmed this works with CoreFX (Perf and func), feel free to merge

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -64,9 +64,9 @@
     <Error Condition="'$(EventHubPath)' == ''" Text="Missing required property EventHubPath." />
     <Error Condition="'$(EventHubSharedAccessKey)' == ''" Text="Missing required property EventHubSharedAccessKey." />
     <Error Condition="'$(EventHubSharedAccessKeyName)' == ''" Text="Missing required property EventHubSharedAccessKeyName." />
-    <Error Condition="'$(BuildIsOfficial)' == 'true' and '$(BuildIsOfficialConnection)' == ''" Text="Missing required property BuildIsOfficialConnection." />
-    <Error Condition="'$(BuildIsOfficial)' == 'true' and '$(DocumentDbKey)' == ''" Text="Missing required property DocumentDbKey." />
-    <Error Condition="'$(BuildIsOfficial)' == 'true' and '$(DocumentDbUri)' == ''" Text="Missing required property DocumentDbUri." />
+    <Error Condition="'$(BuildIsOfficial)' == 'true' and '$(BuildIsOfficialEventHubPath)' == ''" Text="Missing required property BuildIsOfficialEventHubPath." />
```
dsgouda: What happens here if a build is not official?

karajas:
Yes. This change also followed the current method of running a net45 project, where the overall dependencies are specified by the Microsoft.DotNet.Build.Tasks project.json and only those dlls are copied to the package (as specified by the BuildTasks nuspec) .

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
karajas: Just don't set BuildIsOfficial or BuildIsOfficial = false

PR #605, opened 2016-04-02
Enable the following analyzers

pallavit:
1. PinvokeAnalyzer
   a. Validates that the Pinvoke calls in the list are allowed.
   b. This is enabled only for the non-test binaries.
   c. Also, this is currently enabled only for AnyOS and Windows_NT configurations.

2. ResourceUsageAnalyzer
   Validates that the SR.Format calls are always of the form SR.Format(string, params object[]);


pallavit:
/cc: @AlexGhiondea, @tarekgh, @weshaggard, @joperezr Could you please take a look.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+                    }
+
+                    // 2. Method could be called with variable arguments of type object.
+                    else
+                    {
+                        isValidCall = true;
+                        for (int i = 1; i < memberSymbol.Parameters.Length; i++)
+                        {
+                            if (memberSymbol.Parameters[i].Type.SpecialType != SpecialType.System_Object)
+                                isValidCall = false;
```
tarekgh: when this can happen?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            if (memberSymbol.Name.Equals("Format") && memberSymbol.ContainingType.Equals(SRSymbol))
+            {
+                bool isValidCall = false;
+                if (memberSymbol.Parameters.Length > 1
+                    && memberSymbol.Parameters[0].Type.SpecialType == SpecialType.System_String)
+                {
+                    // The method could be called directly by passing in an object array.
+                    if (memberSymbol.Parameters[1].IsParams
+                    && memberSymbol.Parameters[1].Type is IArrayTypeSymbol
```
tarekgh: nit: could you insert some spaces in the beginning of this line and following line?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+        INamedTypeSymbol SRSymbol { get; set; }
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterCompilationStartAction(StartCompile);
+        }
+
+
+        private void StartCompile(CompilationStartAnalysisContext obj)
+        {
+            SRSymbol = obj.Compilation.GetTypeByMetadataName("System.SR");
```
tarekgh: is it a convention to use "Symbol" in the name? or can we call it SRType?

tarekgh:
added couple of questions, LGTM.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
+using System;
```
AlexGhiondea: They are missing the copyright header. 

You could run the code formatter tool on this

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.IO;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+using System.Diagnostics;
```
AlexGhiondea: Are all of these needed?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class PinvokeAnalyzer : DiagnosticAnalyzer
+    {
+        private const string Title = "Invalid Pinvoke call";
+        private const string MessageFormat = @"{0} is not supported on one\more targeted platforms.{1}";
+        private const string Description = "The Pinvoke call should be removed.";
+        private const string AnalyzerName = "PinvokeAnalyzer";
+
+        private static DiagnosticDescriptor InvalidPinvokeCall = new DiagnosticDescriptor(DiagnosticIds.BCL0100.ToString(), Title, MessageFormat, AnalyzerName, DiagnosticSeverity.Error, isEnabledByDefault: true, description: Description);
```
AlexGhiondea: nit: BCL0020 should be fine. Trying not to bump the number of warnings by too much

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                return _exceptionPinvokes;
+            }
+        }
+
+        private AdditionalText ExceptionFile { get; set; }
+        private AdditionalText AllowedPinvokeFile { get; set; }
+
+        private HashSet<string> _isNotSupportedOnWin7 = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        private HashSet<string> IsNotSupportedOnWin7 { get { return _isNotSupportedOnWin7; } }
```
AlexGhiondea: Are any of these properties used outside of this class? If not, we can probably get away with just using the field directly.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                }
+
+                return _exceptionPinvokes;
+            }
+        }
+
+        private AdditionalText ExceptionFile { get; set; }
+        private AdditionalText AllowedPinvokeFile { get; set; }
+
+        private HashSet<string> _isNotSupportedOnWin7 = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
AlexGhiondea: Should we do case-sensitive instead?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+        private HashSet<string> IsNotSupportedOnWin7 { get { return _isNotSupportedOnWin7; } }
+
+
+        private Dictionary<string, string> ParseAdditionalFile(AdditionalText additionalFile)
+        {
+            Dictionary<string, string> parsedPinvokes = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            if (additionalFile != null)
```
AlexGhiondea: nit: you can reduce the indentation of this method by returning the parsedPinvokes if the additionalFile is null.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                SourceText fileContents = additionalFile.GetText();
+                foreach (TextLine line in fileContents.Lines)
+                {
+                    string lineStr = line.ToString();
+                    if (!string.IsNullOrWhiteSpace(lineStr) && !lineStr.StartsWith("<!--"))
```
AlexGhiondea: What if you have whitespace before `<!--` ?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                    {
+                        string[] splitCount = lineStr.Split('!');
+
+                        if (splitCount.Length == 2 || splitCount.Length == 3)
+                        {
+                            parsedPinvokes[splitCount[1]] = splitCount[0];
+                            if (splitCount.Length == 3)
+                            {
+                                IsNotSupportedOnWin7.Add(splitCount[1]);
```
AlexGhiondea: Should this be splitCount[2] ?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                }
+            }
+
+            return parsedPinvokes;
+        }
+
+        private const string AltMsgString = @"Consider using {0} instead.";
+        private const string NotSupportedOnWin7 = @"{0} is not supported on win7";
+
+        private bool CheckIfMemberPresent(string methodName, string moduleName, ref string altMsg)
```
AlexGhiondea: Suggestion: Add a comment to capture the rules coded below.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            bool isPresent = false;
+            string altMsg = string.Empty;
+            if (data.EntryPointName != null)
+            {
+                isPresent = CheckIfMemberPresent(data.EntryPointName, data.ModuleName, ref altMsg);
+            }
+            else
+            {
+                isPresent = CheckIfMemberPresent(methodSymbol.Name, data.ModuleName, ref altMsg) || CheckIfMemberPresent(methodSymbol.Name + 'W', data.ModuleName, ref altMsg);
```
AlexGhiondea: nit: Split this in 2 lines -- makes it easier to read :)

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        INamedTypeSymbol SRSymbol { get; set; }
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterCompilationStartAction(StartCompile);
+        }
+
+
+        private void StartCompile(CompilationStartAnalysisContext obj)
+        {
```
AlexGhiondea: nit: I would rename `obj` to `context`

src/Microsoft.DotNet.CodeAnalysis/DiagnosticIds.cs
```diff
@@ -12,5 +12,9 @@ internal enum DiagnosticIds
         BCL0010, // AppContextDefaultNotInitializedToTrueDiagnostic
         BCL0011, // AppContextDefaultUsedUnexpectedIfStatement
         BCL0012, // DefaultValueDefinedOutsideIfCondition
+
+        BCL0100, // PinvokeCallCheck
```
AlexGhiondea: Since this is not a flags enum we can probably leave about 5 numbers per analyzer (in case new warnings are added to existing analyzers). I would make these be BCL0015 and BCL0020.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/Microsoft.DotNet.CodeAnalysis_DisablePinvokeAnalyzer.ruleset
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Common diagnostic rules" Description="Enables/disable rules for all projects." ToolsVersion="14.0">
+  <Include Path="$(MSBuildThisFileDirectory)\Microsoft.DotNet.CodeAnalysis.ruleset" Action="Default" />
+  <Rules AnalyzerId="Microsoft.DotNet.CodeAnalysis" RuleNamespace="Microsoft.DotNet.CodeAnalysis.Analyzers">
+    <!-- PinvokeChecker should only be enabled for AnyOS or Windows_NT configurations -->
+    <Rule Id="BCL0100" Action="None" />
```
AlexGhiondea: So there will be a ruleset inside CoreFx that will enable this analyzer and we are going to use the ruleset depending on the OS?

AlexGhiondea: Nevermind -- I see it in the file below :)

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup Condition="'$(OSGroup)'=='Windows_NT'">
+    <!-- This file helps us choose the correct allowed pinvoke file.
+       1. In case we are building for windows and TargetGroup is Netcore50 or netcore50aot (Store profile) we use OneCoreUWPApis.txt
+       2. For all other windows configuration we use OneCoreAPIs.txt file.
+    -->
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'==''">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
```
AlexGhiondea: I think we should have a folder structure where each analyzer gets its folder. I think that way we can manage the number of files that each one needs (and we can easily tell which files are used by which analyzer).

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
```diff
@@ -0,0 +1,24 @@
...
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard13aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard15aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netcore50' or '$(TargetGroup)'=='netcore50aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_UWPApis.txt</ValidPinvokeMappings>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(CodeAnalysisRuleSet)'==''">
+    <CodeAnalysisRuleSet Condition="'$(IsTestProject)' == true or '$(TargetsUnix)' == 'true'">$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis_DisablePinvokeAnalyzer.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSet Condition="'$(TargetsWindows)'=='true' and ($(TargetGroup.Contains('net46')) or $(TargetGroup.Contains('net45')))">$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis_DisablePinvokeAnalyzer.ruleset</CodeAnalysisRuleSet>
```
AlexGhiondea: What about net462?

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
AlexGhiondea: Should we have this in the codeanalysis.targets file instead of a separate file?

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -9,6 +9,8 @@
     <CodeAnalyzerDir Condition="'$(CodeAnalyzerDir)'=='' and '$(OSEnvironment)'!='Windows_NT'">$(ToolsDir)analyzers/</CodeAnalyzerDir>
   </PropertyGroup>
   
+  <Import Condition="'$(EnableDotnetAnalyzers)' == 'true'" Project="$(MSBuildThisFileDirectory)PinvokeAnalyzer.props"/>
```
joperezr: As Alex suggested above, instead of having this import we should just move the contents of your props file here instead.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
pallavit: As of now we could have made the condition simply if Parameters.Length >=2 but I just wanted to make it tighter and use this instead. Do you see any issue with it?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
pallavit: Taken care.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
pallavit: I was trying to be consistent with Roslyn model... since we are at the semantic model I chose to use symbol here.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: Will do. Thanks!

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: Yes :)

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: They aren't used outside of this class, however I would prefer using properties in case we want to add some extra logic.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: No. There is no guarantee that they would be identical, in fact the tool should be resilient to that.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: Taken care.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: The supported list format is pretty rudimentary. I have not handled cases where the comments would be multiline or present in the same line etc. However since I don't expect the files to be changed that often, I don't see any issue with it.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
pallavit: Spoke to Tarek offline, updated the code :)

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: No. SplitCount[2] only contains false to indicate that the method is absent on Win7. The list here is trying to store the methodName and hence is using splitCount[1] instead.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: Sounds good. Taken care.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: Taken care.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/ResourceUsageAnalyzer.cs
pallavit: This is how Roslyn generates it now in the template however I agree that context sounds much better :). Taken care.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
pallavit: As of now, we have only 1 analyzer having more than 1 asset so this might be an overkill. However, if in future we see more analyzers following suite, I agree that would be better.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
pallavit: net462 would be covered in net46 query.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
pallavit: Hmm.. I did that initially however thought it might be better to put pinvoke specific props in a separate props file for better classification. Although I find this preferable, if you feel strongly about it, I will change it :). Let me know.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
AlexGhiondea: In case we need extra logic I think we should add it then. This is going to be called in a hot path during compilation and method calls adds to the overhead. I would prefer we keep these as fields until we need more logic around them.

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
AlexGhiondea: I am ok with that -- in that case, can you add a comment in the file that contains the APIs to call out the structure to follow?

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
AlexGhiondea: Got i!

src/Microsoft.DotNet.CodeAnalysis/Analyzers/PinvokeAnalyzer.cs
pallavit: Umm..k

pallavit:
Please let me know if you have any further comments.

tarekgh:
LGTM.

AlexGhiondea:
LGTM. Please squash the commits :).

Have you tried this in the TFS build?

pallavit:
> Have you tried this in the TFS build?

I did with the initial draft but I will run it once again before merging :). Thanks for pointing out.

pallavit:
Thanks for the review :).

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
```diff
@@ -0,0 +1,24 @@
...
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup Condition="'$(OSGroup)'=='Windows_NT'">
+    <!-- This file helps us choose the correct allowed pinvoke file.
+       1. In case we are building for windows and TargetGroup is Netcore50 or netcore50aot (Store profile) we use OneCoreUWPApis.txt
+       2. For all other windows configuration we use OneCoreAPIs.txt file.
+    -->
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'==''">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard13aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
```
weshaggard: Should these be an itemgroup instead of a propertygroup? Do we expect to every use multiple of them?

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
weshaggard: Given the ties in with other common properties like CodeAnalysisRuleSet I would merge this into common code analyzer targets file. Plus there isn't enough setting in here to keep them separate in my opinion.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
```diff
@@ -0,0 +1,24 @@
...
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard15aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
+    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netcore50' or '$(TargetGroup)'=='netcore50aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_UWPApis.txt</ValidPinvokeMappings>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(CodeAnalysisRuleSet)'==''">
+    <CodeAnalysisRuleSet Condition="'$(IsTestProject)' == true or '$(TargetsUnix)' == 'true'">$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis_DisablePinvokeAnalyzer.ruleset</CodeAnalysisRuleSet>
```
weshaggard: How does this rule set get merged with other rulesets?

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
pallavit: As of now no, but there are scenarios which might require this in future.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
pallavit: It does not get merged automatically, but from what I know that is always the case when one tries to override a rule set. Is there a better way?

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer.props
pallavit: Since I merged the change before your comments, I will start a separate PR with this updated :).

PR #611, opened 2016-04-05
Change "not creating symbol package" from warning to message

dagood:
Not making a symbol package is relatively normal, and the warnings were noisy. Changing them to high-importance messages still lets us track the cause of a missing symbol package.

/cc @ericstj @weshaggard


ericstj:
LGTM

weshaggard:
What is the expectation with this message. Are people supposed to take action on it? If so then it should probably be a warning still. If it is just informational then I'm not even sure it should be High Importance I would say low or normal, so that it doesn't go to the console by default.

dagood:
The message is informational in case a symbol package isn't created and it isn't clear why it wasn't. There are some packages now that already trigger this so there's no expected action.

I think low makes sense, I'll change that. (At least until we expect all symbol packages to always be created.) I hadn't fully thought through the importance level, thanks.

AlexGhiondea:
LGTM

weshaggard:
LGTM

PR #613, opened 2016-04-05
Enable optimization data

AlexGhiondea:
This enables applying profile guilded optimization data to managed assemblies.

/cc @weshaggard @joperezr


src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
```diff
@@ -0,0 +1,48 @@
...
+        public string DestinationDirectory { get; set; }
+
+        /// <summary>
+        /// Indicates if the destination archive should be overwritten if it already exists.
+        /// </summary>
+        public bool OverwriteDestination { get; set; }
+
+        public override bool Execute()
+        {
+            if (Directory.Exists(DestinationDirectory) && OverwriteDestination == true)
```
joperezr: Should we log a warning or error in case of DestinationDirectory already be present and OverwriteDestination = false?

src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                Log.LogMessage(MessageImportance.Low, "'{0}' already exists, trying to delete before unzipping...", DestinationDirectory);
+                Directory.Delete(DestinationDirectory, recursive: true);
+            }
+
+            Log.LogMessage(MessageImportance.High, "Decompressing '{0}' into '{1}'...", SourceArchive, DestinationDirectory);
+            if (!Directory.Exists(Path.GetDirectoryName(DestinationDirectory)))
+                Directory.CreateDirectory(Path.GetDirectoryName(DestinationDirectory));
+
+            ZipFile.ExtractToDirectory(SourceArchive, DestinationDirectory);
```
joperezr: Try catch here and log to false in case of error?

src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// The path of the archive to be created.
+        /// </summary>
+        [Required]
+        public string DestinationDirectory { get; set; }
+
+        /// <summary>
+        /// Indicates if the destination archive should be overwritten if it already exists.
+        /// </summary>
+        public bool OverwriteDestination { get; set; }
```
joperezr: Just wondering, did you try not passing in this argument? I ask because I'm not sure how msbuild treats optional params when they are not type string. On string, it just simply sets them to null, but given that this is not a nullable bool, I'm not sure how that would work.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
```diff
@@ -0,0 +1,81 @@
...
+      <OptimizedAssemblyFile>$(IntermediateOutputPath)/OptimizedAssembly/$(AssemblyName).dll</OptimizedAssemblyFile>
+    </PropertyGroup>
+
+    <!-- Extract the profile data from the archive -->
+    <ExtractZipFile SourceArchive="$(OptimizationDataDir)/$(AssemblyName).zip"
+                    DestinationDirectory="$(OptimizationDataForAssemblyDir)"
+                    OverwriteDestination="true" />
+
+    <!-- Copy the compiled assembly into a folder for further processing -->
+    <MakeDir Directories="$(IntermediateOutputPath)/OptimizedAssembly" />
```
joperezr: Will this task fail if dir already exist?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
```diff
@@ -0,0 +1,81 @@
...
+    <Message Text="Copy @(IntermediateAssembly) to $(IntermediateOutputPath)/OptimizedAssembly" />
```
joperezr: perhaps we can set this message important to low so that it is only present on diagnostic builds

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
```diff
@@ -0,0 +1,81 @@
...
+    <Copy SourceFiles="@(IntermediateAssembly)" DestinationFolder="$(IntermediateOutputPath)/OptimizedAssembly" />
+
+    <!-- Delete existing optimization data from the file that came from with the training data -->
+    <Exec Command="$(IBCMergeToolPath) -f -delete -mo $(OptimizationDataForAssemblyDir)/$(AssemblyName).dll $(OptimizationDataForAssemblyDir)/$(AssemblyName).ibc" />
```
joperezr: Add &quot; surronding IBCMergeToolPath in case of path with spaces.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
```diff
@@ -0,0 +1,81 @@
...
+  </Target>
+
+  <!-- We need the OptimizationData package in order to be able to optimize the assembly -->
+  <Target Name="RestoreOptimizationDataPackage" BeforeTargets="CoreCompile"
+          Condition="'$(EnableProfileGuidedOptimization)'=='true' and !Exists('$(OptimizationDataDir)\project.json')">
+
+    <!-- Dynamically create a project.json file used to restore the optimization data-->
+    <Message Text="Generating project.json for optimization data"  Importance="low" />
+    <ItemGroup>
+      <_OptimizationDataJsonLine Include="{&quot;dependencies&quot;: {" />
```
joperezr: I'm not sure if it will look cleaner or not, but we could make this json to be a one-liner.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
```diff
@@ -0,0 +1,81 @@
...
+      <_OptimizationDataJsonLine Include="&quot;$(OptimizationDataPackageName)&quot; : &quot;$(OptimizationDataVersion)&quot; " />
+      <_OptimizationDataJsonLine Include="},&quot;frameworks&quot;: {&quot;dnxcore50&quot;: {},&quot;net46&quot;: {}}}"/>
+    </ItemGroup>
+
+    <MakeDir Directories="$(OptimizationDataDir)" ContinueOnError="true" />
+    <WriteLinesToFile File="$(OptimizationDataDir)\project.json" Lines="@(_OptimizationDataJsonLine)" Overwrite="true" />
+
+
+    <!-- Restore the OptimizationData package -->
+    <Exec Command="$(DnuRestoreCommand) $(OptimizationDataDir)\project.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json"
```
joperezr: doesn't DnuRestoreCommand already contain --source attributes? Also we should add &quot; here as well.

src/Microsoft.DotNet.Build.Tasks/ZipFileCreateFromDirectory.cs
```diff
@@ -39,7 +39,7 @@ public override bool Execute()
         {
             if (File.Exists(DestinationArchive) && OverwriteDestination == true)
             {
-                Log.LogMessage(MessageImportance.Low, "{0} already existed, deleting before zipping...", SourceDirectory, DestinationArchive);
+                Log.LogMessage(MessageImportance.Low, "{0} already existed, deleting before zipping...", DestinationArchive);
```
joperezr: Thanks for fixing this! :smile:

src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
AlexGhiondea: Just tried. You get 'false' for a boolean. I assume this will give you default(T) for any T when the property is optional.

src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
AlexGhiondea: I can do that, but in case of failure I wonder if the callstack might be more informative.

I based this on the create archive code that did not catch exceptions around this.

src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
AlexGhiondea: We could, but you are going to get an error when that happens anyway... Maybe we can give a nicer error.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
AlexGhiondea: I don't believe it will fail

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
AlexGhiondea: Removed that entirely.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
AlexGhiondea: Done -- good catch!

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
AlexGhiondea: I had it on one line, but then I moved it to separate lines where the line in the middle is the package name. That way we can easily add more packages (if we have to).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
AlexGhiondea: It does, but I don't know if it contains the buildtools feed everytime.

joperezr: It does. Extracted from dir.props:
```xml
  <!-- list of nuget package sources passed to dnu -->
  <ItemGroup>
    <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/nugetbuild/api/v3/index.json" />
    <DnuSourceList Include="https:%2F%2Fapi.nuget.org/v3/index.json" />
  </ItemGroup>
```

AlexGhiondea: Great! In that case I will remove it! Thanks for checking!

src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
weshaggard: Yes I would expect it to be false if not passed.

src/Microsoft.DotNet.Build.Tasks/ExtractZipFile.cs
weshaggard: You followed the bad thing and didn't follow the good thing :) The bad, I agree with @joperezr that it is probably better to handle the exceptions. The good, we should name this task ZipFileExtractToDirectory to match the name of the primary purpose.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
weshaggard: Why not capture this directory in a property as well since you use it in at least 4 places.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
weshaggard: That list should be fixed as it should only contain nuget and dotnet-core for most cases. For your case @AlexGhiondea you should directly depend on buildtools and nuget.

AlexGhiondea:
@briansull Could you take a look at the calls that are made to ibcmerge to make sure they look right?

Here is where the calls start:
https://github.com/dotnet/buildtools/pull/613/files#diff-ca437958d8fd6082537d1364635957d2R31

src/Microsoft.DotNet.Build.Tasks/ZipFileExtractToDirectory.cs
```diff
@@ -0,0 +1,65 @@
...
+                    Directory.Delete(DestinationDirectory, recursive: true);
+                }
+                else
+                {
+                    Log.LogWarning("'{0}' already exists. Did you forget to set '{1}' to true?", DestinationDirectory, nameof(OverwriteDestination));
+                }
+            }
+
+            Log.LogMessage(MessageImportance.High, "Decompressing '{0}' into '{1}'...", SourceArchive, DestinationDirectory);
+            try
```
weshaggard: I would wrap the entire function in this try block. Including the Directory.Delete.

src/Microsoft.DotNet.Build.Tasks/ZipFileCreateFromDirectory.cs
```diff
@@ -37,58 +37,75 @@ public sealed class ZipFileCreateFromDirectory : Task
...
+                    Log.LogWarning("'{0}' already exists. Did you forget to set '{1}' to true?", DestinationArchive, nameof(OverwriteDestination));
+                }
             }
 
-            Log.LogMessage(MessageImportance.High, "Compressing {0} into {1}...", SourceDirectory, DestinationArchive);
-            if (!Directory.Exists(Path.GetDirectoryName(DestinationArchive)))
-                Directory.CreateDirectory(Path.GetDirectoryName(DestinationArchive));
-
-            if (ExcludePatterns == null)
+            try
```
weshaggard: I would wrap the entire function in the try block.

weshaggard:
Couple more nits but LGTM

briansull:
AFAICT the Ibcmerge commands look OK

AlexGhiondea:
Thank you @briansull @weshaggard and @joperezr!

I have rebased and squashed some commits.

PR #615, opened 2016-04-05
Merging PinvokeAnalyzer properties in codeanalysis.targets file

pallavit:
Incorporating pending comments of PR #605


pallavit:
/cc: @weshaggard Could you please take a look at this.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -17,9 +26,14 @@
...
+	
+	<!-- Add the necessary valid Pinvoke mapping file -->
+	<AdditionalFiles Condition="'$(ValidPinvokeMappings)' != ''" Include="$(ValidPinvokeMappings)"/>
   </ItemGroup>
 
-  <PropertyGroup Condition="'$(EnableDotnetAnalyzers)' == 'true'">
-    <CodeAnalysisRuleSet Condition="'$(CodeAnalysisRuleSet)'==''">$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis.ruleset</CodeAnalysisRuleSet>
+  <PropertyGroup Condition="'$(EnableDotnetAnalyzers)' == 'true' and '$(CodeAnalysisRuleSet)'==''">
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis.ruleset</CodeAnalysisRuleSet>
+	<CodeAnalysisRuleSet Condition="'$(IsTestProject)' == true or '$(TargetsUnix)' == 'true'">$(MSBuildThisFileDirectory)Microsoft.DotNet.CodeAnalysis_DisablePinvokeAnalyzer.ruleset</CodeAnalysisRuleSet>
```
AlexGhiondea: nit: indentation

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -17,9 +26,14 @@
 
     <AdditionalFiles Include="$(MSBuildProjectDirectory)/*.analyzerdata" />
     <AdditionalFiles Include="$(MSBuildProjectDirectory)/*.analyzerdata.$(Platform)" />
+	
+	<!-- Add the necessary valid Pinvoke mapping file -->
+	<AdditionalFiles Condition="'$(ValidPinvokeMappings)' != ''" Include="$(ValidPinvokeMappings)"/>
```
AlexGhiondea: nit: indentation

AlexGhiondea:
I am not sure if the ruleset will use the msbuild properties here:
https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/Microsoft.DotNet.CodeAnalysis_DisablePinvokeAnalyzer.ruleset#L3 

I think you can directly reference the file.

weshaggard:
@AlexGhiondea how do you plan on merging the different rulesets for the different rules?

weshaggard:
LGTM beyond @AlexGhiondea's comments.

pallavit:
> I am not sure if the ruleset will use the msbuild properties here. I think you can directly reference the file.

I looked at Roslyn codebase and found the same pattern. So, I think this will work.

AlexGhiondea:
@weshaggard @pallavit I think the way we should be handling different rulesets per platform should not be handled in the buildtools repo.

Instead, each repo should define their own rulesets if they need to do something fancy like this. The ruleset in the buildtools repo should be used for things that we want to turn off completely, for all impacted platforms.

pallavit:
> Instead, each repo should define their own rulesets if they need to do something fancy like this. 

I agree with you in general. However, for Pinvoke analyzer for instance, this would mean every repo that ever consumes Pinvoke analyzer will need to duplicate these properties because this policy is going to be true independent of the repo we run this analyzer.

AlexGhiondea:
@pallavit I am going to integrate this into the work to enable the configuration for analyzers. You don't have to merge this in.

pallavit:
Closing this PR since it has already been incorporated in #618.

PR #616, opened 2016-04-06
Update PrereleaseResolveNugetPackageAssets to support project references, use project reference for test-runtime

mellinoe:
This updates the fork we have of PrereleaseResolveNugetPackageAssets in order to support the use of project references in dependencies. This will allow us to reference the test-runtime project in corefx from the dependencies of each test project. This will allow nuget to do proper resolution of the combined set of dependencies, which is an improvement over the error-prone way we are currently combining the two. It will also allow us to more easily reference a different test-runtime, for example.

Includes the following changes:

* Updated fork of Nuget code. This is mostly verbatim from the nuget sources, but does include some changes to maintain compatibility with our build system. Long-term we need to move off of having a fork of this code.
* Update ValidateProjectDependencyVersions to handle previously-unexpected values in the project.json file.
* Update publishtest.targets. We no longer restore both the test-runtime project.json and the test's project.json. We only restore the test's project.json and resolve assets from it. The test project.json must now have a reference to a test-runtime to use, or more specifically must include references to all of the runtime dependencies that it needs to run tests.

Don't merge this yet, I need to prepare and test changes to go along with this in corefx and WCF.

@ericstj , @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -12,12 +12,6 @@
     <TargetOS Condition="'$(OSGroup)'!='AnyOS'">$(OSGroup)</TargetOS>
   </PropertyGroup>
 
-  <Target Name="RestoreTestRuntimePackage"
-          BeforeTargets="ResolveNuGetPackages"
-          Condition="'$(RestorePackages)'=='true' AND '$(IsTestProject)' == 'true' AND '$(TestRuntimeProjectJson)' != ''">
-    <Exec Command="$(DnuRestoreCommand) &quot;$(TestRuntimeProjectJson)&quot;" StandardOutputImportance="Low" CustomErrorRegularExpression="^Unable to locate .*" />
-  </Target>
-
   <PropertyGroup>
     <PrepareForRunDependsOn Condition="'$(IsTestProject)'=='true'">$(PrepareForRunDependsOn);CopyTestToTestDirectory;CreateTestNuGetPackage;CreateAssemblyListTxt</PrepareForRunDependsOn>
     <TestArchitecture Condition="'$(TestArchitecture)' == ''">x64</TestArchitecture>
@@ -51,18 +45,13 @@
   <Target Name="CopyTestToTestDirectory"
           DependsOnTargets="DiscoverTestInputs" Condition="'$(DisableCopyTestToTestDirectory)'!='true'">
 
-    <ItemGroup>
-      <TestNugetProjectLockFile Include="$(ProjectLockJson)" Condition="Exists($(ProjectLockJson))"/>
-      <TestNugetProjectLockFile Include="$(TestRuntimeProjectLockJson)" Condition="Exists($(TestRuntimeProjectLockJson))"/>
-    </ItemGroup>
-
-    <PrereleaseResolveNuGetPackageAssets Condition="'@(TestNugetProjectLockFile)' != ''"
+    <PrereleaseResolveNuGetPackageAssets Condition="Exists($(ProjectLockJson))"
                                AllowFallbackOnTargetSelection="true"
                                IncludeFrameworkReferences="false"
                                NuGetPackagesDirectory="$(PackagesDir)"
                                RuntimeIdentifier="$(TestNugetRuntimeId)"
                                ProjectLanguage="$(Language)"
-                               ProjectLockFile="%(TestNugetProjectLockFile.FullPath)"
+                               ProjectLockFile="$(ProjectLockJson)"
                                TargetMonikers="@(TestTargetFramework)">
       <Output TaskParameter="ResolvedCopyLocalItems" ItemName="TestCopyLocal" />
     </PrereleaseResolveNuGetPackageAssets>
--- a/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
@@ -8,11 +8,12 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Reflection.Metadata;
-using System.Reflection.PortableExecutable;
 using System.Text;
 using Newtonsoft.Json.Linq;
 using Newtonsoft.Json;
+using System.Security.Cryptography;
+using System.Reflection.PortableExecutable;
+using System.Reflection.Metadata;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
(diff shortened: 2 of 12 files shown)


mellinoe:
WIP version of the corefx changes we will need:
https://github.com/dotnet/corefx/compare/master...mellinoe:test-runtime-projectref

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
```diff
@@ -8,11 +8,12 @@
...
-using System.Reflection.Metadata;
-using System.Reflection.PortableExecutable;
 using System.Text;
 using Newtonsoft.Json.Linq;
 using Newtonsoft.Json;
+using System.Security.Cryptography;
+using System.Reflection.PortableExecutable;
+using System.Reflection.Metadata;
 
 namespace Microsoft.DotNet.Build.Tasks
```
weshaggard: We shouldn't be changing this task as we are trying to break our dependency on it and switch to using the shipped version.

weshaggard:
How does dotnet restore find the "test-runtime" project reference? 

Before we go too much further down this route we should make sure we understand how we are running test can be converged with how we are doing our official helix test runs. @Chrisboh @MattGal I would like to make sure we use the same infrastructure so that we end up maintaining too ways of building, deploying, and running our tests.

mellinoe:
> How does dotnet restore find the "test-runtime" project reference?

A "global.json" file is required which lists the directories in which projects can reside. See [this addition](https://github.com/mellinoe/corefx/blob/2ade9e8470487ebeebc91e1ca8cb1c4274c44ebe/src/global.json).

> Before we go too much further down this route we should make sure we understand how we are running test can be converged with how we are doing our official helix test runs.

In a lot of ways, this change simplifies how we deploy test assets. Most of the complexity here is just coming from updating the nuget-related tasks. By the way, I agree that we should remove our forked version as soon as we can (which be now if Nuget already ships a version that supports our platforms).

This approach should also give us a lot more flexibility than before. For example, we could reference a different test-runtime in each framework target, if we wanted. Or, certain test projects could use their own specialized runtime if they needed special assets that we didn't want to restore everywhere.

My branch above in corefx is more or less all of the changes we should need, and it is limited to adding an extra line to each project.json, as well as the global.json. There's a few warnings I need to clean up related to dotnet restore, but other than that it is the complete set of changes we need.

mellinoe:
The version here should be all of the changes we need. My branch [here](https://github.com/dotnet/corefx/compare/master...mellinoe:test-runtime-projectref) has all of the changes we need to make in corefx to get this working there, as well. I'm rebasing it now on top of master.

mellinoe:
@weshaggard @ericstj 

Could you guys give this another look and let me know if you have any more feedback?

weshaggard:
Can we wait until after our meeting tomorrow? I want to make sure this is aligning with the overall direction first.

mellinoe:
Okay, we can chat at the meeting tomorrow. To be clear, this really doesn't radically change anything about how tests are deployed for us. It changes a few intermediate steps that affect test asset resolution and combination from nuget dependencies. The rest of the process is the same as before. Currently, this is all very convoluted and fragile, and is completely broken on x-plat builds right now, so this needs to be addressed regardless.

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
weshaggard: After talking offline I understand that this is mostly a port of the recent task changes as opposed to another implementation, so I'm less concerned about these changes. @mellinoe is there any good way to create another commit on top of just the port to easily show what you needed to change after the port?

weshaggard:
Thanks for doing this @mellinoe now that I better understand the scope here I like the approach and think we should move forward. We might need to figure out how to transition this change however because it will be a breaking change for any repo's using test-runtime. I know for sure corefx (which you are handling) and wcf will be broken by this change. I don't know if there are other repo's using the test-runtime infrastructure or not, but something to think about, especially with your target file changes.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -12,12 +12,6 @@
     <TargetOS Condition="'$(OSGroup)'!='AnyOS'">$(OSGroup)</TargetOS>
   </PropertyGroup>
 
-  <Target Name="RestoreTestRuntimePackage"
```
weshaggard: We should consider keeping this, at least for a while for anyone that has a TestRuntimeProjectJson file property defined. That will help other repo's transition to your new model over time.

ericstj: I'd actually prefer that we force projects to update.  Validate that they have a package reference to test-runtime unless they explicitly opt out.

weshaggard: I would like them to update as well but I also don't want this to be a blocker for people adopting a new build tools version. Perhaps keeping this target and adding another property that allows people to post-pone that work temporarily would be good.

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
mellinoe: I could rebase with an intermediate commit here, but I think it would be easier to just diff the two locally, you can see the original NuGet version here:

https://github.com/NuGet/NuGet.BuildTasks/blob/b8552204ff93245bb6e36ff9a4cb6091d05a7b94/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
mellinoe: I can try allowing a fallback, but I'm not sure it will 100% work with all of the other changes here. I will see if I can get WCF to work without any changes in that repo. I was planning on having another fix for WCF ready shortly after this goes in. I don't think there will be very much churn there, because there are not very many test projects, compared to corefx.

mellinoe:
I went ahead and fixed up WCF as well. There weren't very many changes needed there:

https://github.com/dotnet/wcf/compare/master...mellinoe:test-runtime-projectref

@weshaggard This should be good to go as long as you have no more reservations. I have changes ready for both corefx and WCF which should account for any breaking changes made here.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
mellinoe: I think it is easier to just update WCF to account for the breaking changes. See https://github.com/dotnet/wcf/compare/master...mellinoe:test-runtime-projectref

mellinoe:
@weshaggard I've actually added the fallback to allow specifying TestRuntimeProjectLockJson. It looks like we still need it for the .NET Framework testing that was added recently. It should function the same as before, but I've added a warning if it is used, because we should move off of it as soon as possible. I will not define that property in our regular corefx builds.

weshaggard:
Can you help me understand why it is needed for the .NET Framework testing? I assume we can still use the test-runtime project reference for that scenario as well correct?

mellinoe:
Yes, that should work. I haven't had time to test it yet, and I'm not familiar with how our net46 testing is done right now (it's only enabled for two projects), so I was hesitant to try it. I will give it a shot, though.

mellinoe:
@weshaggard Okay, that approach works fine. We should be able to just reference the net46 test-runtime from the net46 dependency groups in corefx and we shouldn't need this fallback logic at all.

weshaggard:
LGTM

PR #620, opened 2016-04-07
Adding target that generates version.c

joperezr:
This target generates a version source file which contains one line like:
```c
static char sccsid[] __attribute__((used)) = "@(#)Version 1.0.12345.0";
```
When adding this source file to any compilation, we will be able to get the version of native binaries built in non-Windows platforms by using either `what(1)` or `strings+grep`

related: dotnet/coreclr#3133
cc: @ellismg @weshaggard 
FYI: @gkhanna79

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -155,6 +155,30 @@
     </ItemGroup>
   </Target>
 
+  <!-- Non Windows versioning requires to generate a source file and include it on the compilation. -->
+  <Target Name="GenerateVersionSourceFile"
+      Inputs="$(MSBuildProjectFile)"
+      Outputs="$(NativeVersionSourceFile)"
+      DependsOnTargets="CreateVersionFileDuringBuild"
+      Condition="'$(NativeVersionSourceFile)'!='' and '$(GenerateVersionSourceFile)'=='true'">
+
+    <ItemGroup>
+      <SourceFileLines />
+      <SourceFileLines Include="static char sccsid%5B%5D %5F%5Fattribute%5F%5F%28%28used%29%29 %3D %22%40%28%23%29Version $(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)%22%3B" />
+      <!-- Since this is a source file, compiler will complain if there is no new line at end of file, so adding one bellow. -->
+      <SourceFileLines Include=" " />
+    </ItemGroup>
+
+    <WriteLinesToFile
+      File="$(NativeVersionSourceFile)"
+      Lines="@(SourceFileLines)"
+      Overwrite="true" />
+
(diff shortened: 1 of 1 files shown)


ellismg:
FWIW, I don't like the idea of having to run managed code (in this case this MSBuild task) in order to generate artifacts needed by the native cross platform builds.  This will make boot strapping much harder because we will need the CLR to work in order to compile the native parts of the CLR and CoreFX.

Is there a way this could be done without using managed code (i.e. in build.sh or something)?

It's not the end of the world if it can't, we could add more hacks to generate a dummy version of this file for a stage0 compile then once we had managed code working build all the native components, but as a general principal I think not requiring managed code to run when building the native components will make things easier.

Thoughts?

weshaggard:
@ellismg that is definitely a concern to take under consideration. Part of the trouble is we would either need to implement the version number generation twice or move it to another level of the bootstrapping and maintain it there. Neither is ideal in my mind. I kind of think for now we should simply allow for a way to manually set the build number to seed any new builds and that should enable the scenario. Unless I'm missing some obvious way to enable this. 

We are always going to have some level of bootstrapping problem we simply need to ensure we have a way to override and get past them.

joperezr:
I totally agree with your point. This is only meant to be a stage0 solution, in which for now we run this target in build.sh before starting the native build, but I also think that ideally this should be generated differently. The problem right now is that the logic we have to calculate the right file version is in managed code (in a msbuild task) so in order to accomplish this we would have to have a different way of calculating it which works in both worlds.

If it's ok with you, I would like to get this in as a first stage to get versioning working for native non-Windows binaries, and I'll log an issue to fix this and have it calculate it differently.

ellismg:
> If it's ok with you, I would like to get this in as a first stage to get versioning working for native non-Windows binaries, and I'll log an issue to fix this and have it calculate it differently.

That's fine.  Note that when we try to wire this up to build.sh for CoreFX and CoreCLR we'll need to make sure that we have a fallback for places where we can't run managed code (otherwise we'll break CI which builds both OpenSUSE 13.2 (where we don't have a dotnet cli build) and FreeBSD). So we'll hit this problem "sooner" rather an "later".  In these cases if we just want to dump an empty sccsid[] string we can use for bootstrapping, that's fine.

joperezr:
> (otherwise we'll break CI which builds both OpenSUSE 13.2 (where we don't have a dotnet cli build) and FreeBSD)

I haven't thought of that. I think that for that case what I'll do is I'll add a file here in buildtools that contains the sccsid as empty as you suggest, and when we are building in those platforms I'll simply use that instead. I'll add that to this PR.

weshaggard:
LGTM

ellismg:
I don't see how cdef2f2459ff68e0ca3f8247a921991c2d9c8752 is going to help.  If we are able to restore the build-tools (and hence pull down this file), it means we could run managed code, correct?  If that was the case, we could just run the task.  Am I misunderstanding how you plan to use that file?

joperezr:
you are right, removed.

PR #623, opened 2016-04-08
Instead of caching the disabled analyzers list read it fresh for every analyzer

AlexGhiondea:


src/Microsoft.DotNet.CodeAnalysis/Analyzers/BaseAnalyzer.cs
```diff
@@ -22,14 +20,21 @@ public sealed override void Initialize(AnalysisContext context)
...
         {
-            EnsureConfigFileLoaded(context.Options);
+            var configFile = context.Options.AdditionalFiles.FirstOrDefault(file => file.Path.Contains(ConfigFileName));
 
-            // Disable analyzers if they are in that file
-            if (s_disabledAnalyzers.Contains(GetType().Name))
+            if (configFile == null)
             {
                 return;
             }
```
pallavit: If configFile is null should we not execute OnCompilationStart() instead of return? GenerateAnalyzerConfigFile runs only when DisabledAnalyzers has some value so we wont be having any config file for windows build and so in that case we won't be running any analyzers at all. Isn't it?

AlexGhiondea: Yes. Good catch!

PR #625, opened 2016-04-08
Packaging fixes/features for open meta-packages

ericstj:
This PR does the following:
- Fix perf for building package project references to use all cores
- Preserve supports section in runtime.json (a bug I noticed when examining the generated runtime.json)
- Feature for splitting pkgproj dependencies by minimum netstandard version, needed to fix https://github.com/dotnet/corefx/issues/6561.

For the last feature I initially coded it to use the validation report but this created unnecessary complexity and sequencing restrictions, so I changed to plumb it through as meta-data on the resolved pkgproj reference.

/cc @weshaggard @chcosta

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -106,6 +106,8 @@
   <UsingTask TaskName="NuGetPack" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
   <UsingTask TaskName="ApplyBaseLine" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
   <UsingTask TaskName="PromoteReferenceDependencies" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
+  <UsingTask TaskName="GetMinimumNETStandard" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
+  <UsingTask TaskName="SplitDependenciesBySupport" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
 
   <!-- Determine if we actually need to build for this architecture -->
   <!-- Packages can specifically control their architecture by specifying the PackagePlatforms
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -26,6 +26,8 @@
 
     <!-- By default we'll build libraries referenced by packages -->
     <BuildPackageLibraryReferences Condition="'$(BuildPackageLibraryReferences)' == ''">true</BuildPackageLibraryReferences>
+
+    <BuildInParallel Condition="'$(BuildInParallel)' == '' AND '$(MSBuildNodeCount)' > '1'">true</BuildInParallel>
```
weshaggard: Is the node count check actually interesting?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -342,11 +348,22 @@
...
+                           Frameworks="@(_referenceFrameworks)">
+      <Output TaskParameter="MinimumNETStandard" PropertyName="MinimumNETStandard"/>
+    </GetMinimumNETStandard>
+    
     <ItemGroup>
       <_PackageIdentity Include="$(Id)">
         <Version>$(Version)</Version>
         <TargetFramework Condition="'$(PackageTargetFramework)' != ''">$(PackageTargetFramework)</TargetFramework>
         <TargetRuntime Condition="'$(PackageTargetRuntime)' != ''">$(PackageTargetRuntime)</TargetRuntime>
+        <MinimumNETStandard Condition="'$(MinimumNETStandard)' != ''">$(MinimumNETStandard)</MinimumNETStandard>
```
weshaggard: I assume you are going to use this metadata later in the meta-package references somewhere?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: Yes, BuildInParallel isn't set by default and MSBuild doesn't seem to provide any other indication of wether it's been run in parallel other than this Node count.  So this is really just setting BuildInParallel to whatever the user specified on the command line (/m => true)

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: Yes, it is used by SplitDependenciesBySupport

ericstj:
@dotnet-bot test this please

PR #626, opened 2016-04-08
Download product packages from Azure

maririos:
Download product packages from Azure BLOB.

cc: @chcosta @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
@@ -1,0 +1,133 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using System.Xml;
+
+using Task = Microsoft.Build.Utilities.Task;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    public sealed class DownloadFromAzure : Task
+    {
+        /// <summary>
+        /// The Azure account name used when creating the connection string.
+        /// </summary>
+        [Required]
+        public string AccountName { get; set; }
+
+        /// <summary>
+        /// The Azure account key used when creating the connection string.
+        /// </summary>
+        [Required]
+        public string AccountKey { get; set; }
+
+        /// <summary>
+        /// The name of the container to access.  The specified name must be in the correct format, see the
+        /// following page for more info.  https://msdn.microsoft.com/en-us/library/azure/dd135715.aspx
+        /// </summary>
+        [Required]
+        public string ContainerName { get; set; }
+
+        /// <summary>
+        /// Directory to download blob files to
+        /// </summary>
+        public string DownloadDirectory { get; set; }
+        
+        public override bool Execute()
+        {
+            return ExecuteAsync().GetAwaiter().GetResult();
+        }
+
+        public async Task<bool> ExecuteAsync()
+        {
+            Log.LogMessage(MessageImportance.Normal, "Downloading container {0} from storage account '{1}'.", ContainerName, AccountName);
+
+            DateTime dateTime = DateTime.UtcNow;
+            List<string> blobsNames = null;
+            string urlListBlobs = string.Format("https://{0}.blob.core.windows.net/{1}?restype=container&comp=list", AccountName, ContainerName);
+            
+            Log.LogMessage(MessageImportance.Normal, "Sending request to list blobsNames for container '{0}'.", ContainerName);
+
+            using (HttpClient client = new HttpClient())
+            {
+                using (HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, urlListBlobs))
+                {
+                    try
+                    {
+                        request.Headers.Add(AzureHelper.DateHeaderString, dateTime.ToString("R", CultureInfo.InvariantCulture));
+                        request.Headers.Add(AzureHelper.VersionHeaderString, AzureHelper.StorageApiVersion);
+                        request.Headers.Add(AzureHelper.AuthorizationHeaderString, AzureHelper.AuthorizationHeader(
+                                AccountName,
+                                AccountKey,
+                                "GET",
+                                dateTime,
+                                request));
+
+                        XmlDocument responseFile;
+                        using (HttpResponseMessage response = await client.SendAsync(request))
+                        {
+                            responseFile = new XmlDocument();
+                            responseFile.LoadXml(await response.Content.ReadAsStringAsync());
+                            XmlNodeList elemList = responseFile.GetElementsByTagName("Name");
+
+                            blobsNames = elemList.Cast<XmlNode>()
+                                                        .Select(x => x.InnerText)
+                                                        .ToList();
+                        }
+                    }
+                    catch (Exception e)
+                    {
+                        Log.LogError("Failed to retrieve information.\n" + e.Message);
+                        return false;
+                    }
+                }
+
+                DownloadDirectory = DownloadDirectory ?? Directory.GetCurrentDirectory();
+                foreach (string blob in blobsNames)
+                {
+                    Log.LogMessage(MessageImportance.Normal, "Downloading BLOB - {0}", blob);
+                    string urlGetBlob = string.Format("https://{0}.blob.core.windows.net/{1}/{2}", AccountName, ContainerName, blob);
+
+                    string filename = Path.Combine(DownloadDirectory, blob);
+                    string blobDirectory = blob.Substring(0, blob.LastIndexOf("/"));
+                    string downloadBlobDirectory = Path.Combine(DownloadDirectory, blobDirectory);
+                    if (!Directory.Exists(downloadBlobDirectory))
+                    {
+                        Directory.CreateDirectory(downloadBlobDirectory);
+                    }
+
+                    using (HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, urlGetBlob))
+                    {
+                        request.Headers.Add(AzureHelper.DateHeaderString, dateTime.ToString("R", CultureInfo.InvariantCulture));
+                        request.Headers.Add(AzureHelper.VersionHeaderString, AzureHelper.StorageApiVersion);
+                        request.Headers.Add(AzureHelper.AuthorizationHeaderString, AzureHelper.AuthorizationHeader(
+                                AccountName,
+                                AccountKey,
+                                "GET",
+                                dateTime,
+                                request));
+                        
+                        using (HttpResponseMessage response = await client.SendAsync(request))
+                        {
+                            Stream responseStream = await response.Content.ReadAsStreamAsync();
+                            using (FileStream sourceStream = File.Open(filename, FileMode.Create))
+                            {
+                                responseStream.CopyTo(sourceStream);
+                            }
+                        }
+                    }
+                }
+            }
+            return true;
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/ListAzureContainers.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/ListAzureContainers.cs
@@ -1,0 +1,95 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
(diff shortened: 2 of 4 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// following page for more info.  https://msdn.microsoft.com/en-us/library/azure/dd135715.aspx
+        /// </summary>
+        [Required]
+        public string ContainerName { get; set; }
+
+        /// <summary>
+        /// An item group of blob filenames to download.  
+        /// </summary>
+        [Required]
+        public ITaskItem[] Items { get; set; }
```
chcosta: Probably don't need "[Required]" here, and I don't think the below code handles the case where you actually pass an item that you want to download.  Either remove this property, or update the code below to download a specified Item.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildTestsPipeline.targets
```diff
@@ -0,0 +1,39 @@
...
+
+  <PropertyGroup>
+    <DownloadDirectory Condition="'$(DownloadDirectory)' == ''">$(PackagesDir)/AzureTransfer</DownloadDirectory>
+  </PropertyGroup>
+
+  <!-- Not currently used -->
+  <ItemGroup>
+    <DownloadItems Include="Microsoft.CSharp.4.0.1-rc3-00001-00.nupkg" />
+  </ItemGroup>
+
```
chcosta: This was used for testing, you can remove it.

maririos:
The changes were made by @chcosta. I took a look at them to plug them with the dev workflow

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildTestsPipeline.targets
```diff
@@ -0,0 +1,39 @@
...
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="DownloadBlobsFromAzure" Condition="'$(ContainerName)' != ''">
+    <Message Text="Download from Azure the container $(ContainerName)" Importance="High" />
+    <DownloadFromAzure AccountName="$(CloudDropAccountName)"
+                       AccountKey="$(CloudDropAccessToken)"
+                       ContainerName="$(ContainerName)"
+                       DownloadDirectory="$(DownloadDirectory)"
+                       Items="@(DownloadItems)" />
```
chcosta: Only needed if you actually support this functionality.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildTestsPipeline.targets
```diff
@@ -0,0 +1,39 @@
...
+
+  <!-- Grab latest azure container if a specific azure container is not specified -->
+  <Target Name="ListAzureContainers" Condition="'$(ContainerName)' == ''">
+    <Message Text="List Azur containers:" Importance="High" />
+    <ListAzureContainers AccountName="$(CloudDropAccountName)"
+                         AccountKey="$(CloudDropAccessToken)"
+                         Prefix="$(ContainerNamePrefix)">
+       <Output TaskParameter="ContainerNames" ItemName="AzureContainerName" />
+    </ListAzureContainers>
+    <Message Importance="High" Text="Container names: %(AzureContainerName.Identity)" />
```
chcosta: Is this "Importance" level right?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildTestsPipeline.targets
```diff
@@ -0,0 +1,39 @@
...
+  <!-- Not currently used -->
+  <ItemGroup>
+    <DownloadItems Include="Microsoft.CSharp.4.0.1-rc3-00001-00.nupkg" />
+  </ItemGroup>
+
+  <Target Name="DownloadBlobsFromAzureTargets" DependsOnTargets="ListAzureContainers;DownloadBlobsFromAzure" />
+
+  <!-- Grab latest azure container if a specific azure container is not specified -->
+  <Target Name="ListAzureContainers" Condition="'$(ContainerName)' == ''">
+    <Message Text="List Azur containers:" Importance="High" />
```
chcosta: Is this "Importance" level right?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildTestsPipeline.targets
```diff
@@ -0,0 +1,39 @@
+<Project ToolsVersion="12.0" DefaultTargets="DownloadBlobsFromAzureTargets" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
chcosta: The filename for this target (BuildTestsPipeline.targets) was derived when this applied only to the build pipeline.  Now that this is more general purpose, you can change the name, or move this code somewhere else in the dev workflow.

maririos:
After I updated my branch (it was a month behind :| ) CloudTestTask is now running on Core, so I'll port this changes to target core too.

maririos:
I updated the PR with the changes to DownloadFromAzure and ListAzureContainers to target core

maririos:
FYI: @karajas

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
```diff
@@ -0,0 +1,31 @@
...
+    <DownloadDirectory Condition="'$(DownloadDirectory)' == ''">$(PackagesDir)/AzureTransfer</DownloadDirectory>
+  </PropertyGroup>
+
+  <Target Name="DownloadBlobsFromAzureTargets" DependsOnTargets="ListAzureContainers;DownloadBlobsFromAzure" />
+
+  <!-- Grab latest azure container if a specific azure container is not specified -->
+  <Target Name="ListAzureContainers" Condition="'$(ContainerName)' == ''">
+    <ListAzureContainers AccountName="$(CloudDropAccountName)"
+                         AccountKey="$(CloudDropAccessToken)"
+                         Prefix="$(ContainerNamePrefix)">
```
chcosta: How are you handling the case which includes a "-" character which in validation was not returning results?

src/Microsoft.DotNet.Build.CloudTestTasks/ListAzureContainers.cs
```diff
@@ -0,0 +1,87 @@
...
+        public ITaskItem[] ContainerNames { get; set; }
+
+        public override bool Execute()
+        {
+            return ExecuteAsync().GetAwaiter().GetResult();
+        }
+
+        public async Task<bool> ExecuteAsync()
+        {
+            Log.LogMessage(MessageImportance.High, "List of Azure containers in storage account '{0}'.", AccountName);
```
chcosta: I would downgrade the importance level for the messages in this task, this is a utility task and the info is more useful as diagnostic output than in the typical case.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
```diff
@@ -0,0 +1,31 @@
...
+  <UsingTask TaskName="ListAzureContainers" AssemblyFile="$(ToolsDir)/net45/Microsoft.DotNet.Build.CloudTestTasks.dll"/>
+
+  <PropertyGroup>
+    <DownloadDirectory Condition="'$(DownloadDirectory)' == ''">$(PackagesDir)/AzureTransfer</DownloadDirectory>
+  </PropertyGroup>
+
+  <Target Name="DownloadBlobsFromAzureTargets" DependsOnTargets="ListAzureContainers;DownloadBlobsFromAzure" />
+
+  <!-- Grab latest azure container if a specific azure container is not specified -->
+  <Target Name="ListAzureContainers" Condition="'$(ContainerName)' == ''">
```
chcosta: The target name may be more appropriate as "GetLatestContainerNameFromAzure" than as "ListAzureContainers".

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
```diff
@@ -0,0 +1,31 @@
...
+    <ListAzureContainers AccountName="$(CloudDropAccountName)"
+                         AccountKey="$(CloudDropAccessToken)"
+                         Prefix="$(ContainerNamePrefix)">
+       <Output TaskParameter="ContainerNames" ItemName="AzureContainerName" />
+    </ListAzureContainers>
+    <Message Importance="High" Text="Container names: %(AzureContainerName.Identity)" />
```
chcosta: Change importance to "Normal" or "Low"

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
maririos: In the ListAzureContainer class I'm manually looking at the prefix in the container names.

maririos:
Thanks @chcosta, I submitted another commit with the suggested changes.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
chcosta: I see. cool.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
```diff
@@ -0,0 +1,31 @@
+<Project ToolsVersion="12.0" DefaultTargets="DownloadBlobsFromAzureTargets" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="DownloadFromAzure" AssemblyFile="$(ToolsDir)net45/Microsoft.DotNet.Build.CloudTestTasks.dll"/>
```
karajas: Instead of $(ToolsDir), use $(BuildToolsTaskDir) which will use the net45 version of the dll for Windows the build and Core compiled version on non-Windows platform builds.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
```diff
@@ -0,0 +1,31 @@
...
+  <UsingTask TaskName="ListAzureContainers" AssemblyFile="$(ToolsDir)/net45/Microsoft.DotNet.Build.CloudTestTasks.dll"/>
```
joperezr: Same comment that @karajas had applies here.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
```diff
@@ -0,0 +1,31 @@
...
+  <Target Name="DownloadBlobsFromAzureTargets" DependsOnTargets="GetLatestContainerNameFromAzure;DownloadBlobsFromAzure" />
+
+  <!-- Grab latest azure container if a specific azure container is not specified -->
+  <Target Name="GetLatestContainerNameFromAzure" Condition="'$(ContainerName)' == ''">
+    <ListAzureContainers AccountName="$(CloudDropAccountName)"
+                         AccountKey="$(CloudDropAccessToken)"
+                         Prefix="$(ContainerNamePrefix)">
+       <Output TaskParameter="ContainerNames" ItemName="AzureContainerName" />
+    </ListAzureContainers>
+    <Message Importance="Normal" Text="Container names: %(AzureContainerName.Identity)" />
```
joperezr: Do we need this message? If so, should we set the importance to low so that it only shows on verbose logs?

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/SyncCloudContent.targets
chcosta: You need to import dir.props before using these properties.

joperezr: No need to be done here, the projects that are consuming this have already imported their dir.props, or else they would not be able to use buildtools at all :smile:

chcosta: Ok, that's fine.  I selfishly wanted something I place because I am currently hacking a project with pieces of @maririos change, and I am directly calling this targets file rather than utilizing a wrapping script. Once this is all part of the clean / sync / build workflow, my hack is unnecessary.  I will continue to hack around this in the meantime. :)

maririos: @karajas thanks! I didn't know about that property!

chcosta:
LGTM

PR #627, opened 2016-04-08
Fixes for project reference to package dependency conversion in test builds

chcosta:
Enable project reference to package conversion for test builds
        - support multiple dependencies sections in project.json
        - support multiple frameworks passed as target groups
        - flow target group to generation
        - Propagate OSGroup and TargetGroup for creating per config project.json files.  OSGroup
          project.json's are typically identical but we generate separate json's to prevent a 
          "binclash" scenario

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -22,9 +22,11 @@ namespace Microsoft.DotNet.Build.Tasks
         // Additional Dependencies to add to the project.json. May Optionally contain a version.
         // Will Override dependencies present in the project if there is a conflict.
         // AdditionalDependencies required metadata:  Name, Version
-
         [Required]
         public ITaskItem[] AdditionalDependencies { get; set; }
+
+        // Framework section which the additional dependencies apply to.  Empty is the default dependencies section.
+        public string[] Frameworks { get; set; }
 
         public string BuildNumberOverrideStructureRegex { get; set; }
 
@@ -39,8 +41,8 @@ namespace Microsoft.DotNet.Build.Tasks
         [Required]
         public string OutputProjectJson { get; set; }
 
-        private Regex versionStructureRegex;
-        private Regex buildNumberOverrideStructureRegex;
+        private Regex _versionStructureRegex;
+        private Regex _buildNumberOverrideStructureRegex;
 
         public override bool Execute()
         {
@@ -55,8 +57,8 @@ namespace Microsoft.DotNet.Build.Tasks
                 {
                     Log.LogError("Missing required parameter.  BuildNumberOverrideStructureRegex must be provided if PackageBuildNumberOverride is provided.");
                 }
-                buildNumberOverrideStructureRegex = new Regex(BuildNumberOverrideStructureRegex);
-                if(!buildNumberOverrideStructureRegex.IsMatch(PackageBuildNumberOverride))
+                _buildNumberOverrideStructureRegex = new Regex(BuildNumberOverrideStructureRegex);
+                if(!_buildNumberOverrideStructureRegex.IsMatch(PackageBuildNumberOverride))
                 {
                     Log.LogError("Invalid package version format: '{0}'", PackageBuildNumberOverride);
                     return false;
@@ -64,13 +66,57 @@ namespace Microsoft.DotNet.Build.Tasks
             }
 
             JObject projectRoot = ReadProject(ProjectJson);
+            var invalidFramework = AreValidFrameworkPaths(projectRoot);
+            if(invalidFramework != string.Empty)
+            {
+                OutputProjectJson = ProjectJson;
+                Log.LogError("Unable to find framework section '{0}' in '{1}'", invalidFramework, ProjectJson);
+                return false;
+            }
+            _versionStructureRegex = new Regex(VersionStructureRegex);
 
-            versionStructureRegex = new Regex(VersionStructureRegex);
+            // No PackageBuildNumberOverride was specified, so try to find one to associate with our AdditionalDependencies
+            PackageBuildNumberOverride = PackageBuildNumberOverride ?? DeriveBuildNumber(projectRoot);
+
+            // No Frameworks were specified, apply AdditionalDependencies to all framework groups in the project.json
+            if (Frameworks == null || Frameworks.Length == 0)
+            {
+                Frameworks = projectRoot.SelectTokens("frameworks").SelectMany(f => f.Children().Select(c => ((JProperty)c).Name)).ToArray();
+            }
+
+            // Update default dependencies section
             JObject dependencies = GenerateDependencies(projectRoot);
-            projectRoot = UpdateProperty("dependencies", projectRoot, dependencies);
+            projectRoot = UpdateDependenciesProperty(projectRoot, dependencies);
+
+            // Update framework dependencies sections
+            for (int i = 0; i < Frameworks.Length; i++)
+            {
+                dependencies = GenerateDependencies(projectRoot, Frameworks[i]);
+                projectRoot = UpdateDependenciesProperty(projectRoot, dependencies, Frameworks[i]);
+            }
             WriteProject(projectRoot, OutputProjectJson);
 
             return true;
+        }
+
+        private string AreValidFrameworkPaths(JObject projectRoot)
+        {
+            if(Frameworks == null ||
+                Frameworks.Length == 0)
+            {
+                return string.Empty;
+            }
+            // Check for a valid path, if invalid, exit
+            for (int i = 0; i < Frameworks.Length; i++)
+            {
+                var _frameworkPath = "frameworks." + Frameworks[i];
+                var validFramework = projectRoot.SelectToken(_frameworkPath);
+                if (validFramework == null)
+                {
+                    return _frameworkPath;
+                }
+            }
+            return string.Empty;
         }
 
         private static JObject ReadProject(string projectJsonPath)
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -55,24 +57,65 @@ public override bool Execute()
...
+            }
+            _versionStructureRegex = new Regex(VersionStructureRegex);
+
+            // No PackageBuildNumberOverride was specified, so try to find one to associate with our AdditionalDependencies
+            PackageBuildNumberOverride = PackageBuildNumberOverride ?? DeriveBuildNumber(projectRoot);
 
-            versionStructureRegex = new Regex(VersionStructureRegex);
-            JObject dependencies = GenerateDependencies(projectRoot);
-            projectRoot = UpdateProperty("dependencies", projectRoot, dependencies);
+            if (Frameworks == null || Frameworks.Count() == 0)
```
weshaggard: Frameworks.Length == 0. In general perfer concrete Lenght or Count properties over the Count() extension method.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -55,24 +57,65 @@ public override bool Execute()
...
+            {
+                JObject dependencies = GenerateDependencies(projectRoot);
+                projectRoot = UpdateProperty(projectRoot, dependencies);
+            }
+            else
+            {
+                for (int i = 0; i < Frameworks.Count(); i++)
```
weshaggard: Dido... Frameworks.Length

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -55,24 +57,65 @@ public override bool Execute()
...
+            }
             WriteProject(projectRoot, OutputProjectJson);
 
             return true;
         }
 
+        private string AreValidFrameworkPaths(JObject projectRoot)
+        {
+            if(Frameworks == null ||
+                Frameworks.Count() == 0)
```
weshaggard: Dido... Frameworks.Length

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -55,24 +57,65 @@ public override bool Execute()
...
+            {
+                return string.Empty;
+            }
+            // Check for a valid path, if invalid, exit
+            for (int i = 0; i < Frameworks.Count(); i++)
```
weshaggard: Dido... Frameworks.Length

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -19,32 +21,68 @@
...
+      <GeneratedProjectLockJson>$(ProjectsGeneratedProjectJsonDir)project.lock.json</GeneratedProjectLockJson>
       <!-- Not all projects generate a new project.json, so only update the project.lock.json if it has been restored. -->
       <ProjectLockJson Condition="Exists('$(GeneratedProjectLockJson)')">$(GeneratedProjectLockJson)</ProjectLockJson>
     </PropertyGroup>
     <Message Condition="Exists('$(ProjectLockJson)')" Text="Using '$(ProjectLockJson)'" Importance="Low"/>
   </Target>
 
+  <!-- The msbuild project file may be an ".builds" file, if it is, then we need to evaluate the "Project" items in the builds file
+       to generate the project.json's for each of those items. -->
+  <Target Name="EvaluateBuildsProjects"
```
weshaggard: I'm not sure why you have this here and in the dir.traversal.targets. Shouldn't only one or the other be necessary?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -155,13 +193,14 @@
...
         <PackageBuildNumberOverride Condition="'$(PackageBuildNumberOverride)' == '' and '$(_DerivedPackageNumber)' != ''">$(_DerivedPackageNumber)</PackageBuildNumberOverride>
-        <VersionStructureRegex Condition="'$(VersionStructureRegex)' == ''">(\d+\.\d+\.\d+)-((beta|rc2|rc3)-?(\d+)?)</VersionStructureRegex>
+        <VersionStructureRegex Condition="'$(VersionStructureRegex)' == ''">(\d+\.\d+\.\d+)-((beta|rc2|rc3)-?(\d+(-\d+)?)?)</VersionStructureRegex>
         <BuildNumberOverrideStructureRegex Condition="'$(BuildNumberOverrideStructureRegex)' == ''">([^-]+)?-?(\d+(-\d+)?)?</BuildNumberOverrideStructureRegex>
     </PropertyGroup>
     <AddDependenciesToProjectJson Condition="'$(ErrorNoVersion)' == ''"
                                   VersionStructureRegex="$(VersionStructureRegex)"
                                   AdditionalDependencies="@(_InjectProjectReferenceDependency)"
                                   ProjectJson="$(OriginalProjectJson)"
+                                  Frameworks="$(TargetGroup)"
```
weshaggard: Should this be GeneratedTargetGroup?

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -22,10 +22,12 @@ public class AddDependenciesToProjectJson : Task
         // Additional Dependencies to add to the project.json. May Optionally contain a version.
         // Will Override dependencies present in the project if there is a conflict.
         // AdditionalDependencies required metadata:  Name, Version
-
         [Required]
         public ITaskItem[] AdditionalDependencies { get; set; }
 
+        // Framework section which the additional dependencies apply to.  Empty is the default dependencies section.
```
weshaggard: Unfortunately I think the assumption of "empty" means the default dependency section is not correct. For a project.json file the default dependencies section is a shared section between all the different target frameworks and our TargetGroup being empty may map to many different target frameworks so they aren't really equated.

chcosta: @weshaggard , can you provide an example of what your expectation for an empty framework is.  It's not clear to me what the difference you are describing is.

The current change says that, if not Frameworks (TargetGroups) are specified, then add the dependencies to the shared dependencies section...

{
  "Dependencies": {
    "new_dependency": "1.0.0-release-major-minor"
  }
  "Frameworks": {
    // blah
  }
}

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
chcosta: Yes

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
chcosta: When using sync.cmd or build.cmd or build-tests.cmd, in dir.traversal.targets, we end up calling GenerateTestProjectJson for each of the test ".builds" files.  When we try to determine the generated project.json path, we have to then iterate on the proj files described in the ".builds" files, so we call "EvaluateBuildsProjects".  

When building a project directly, we don't end up calling EvaluteBuildsProjects from buildagainstpackages.targets (because of the Condition in EvaluateBuildsProjects).  

Essentially, EvaluateBuildsProjects is used to iterate on the ".builds" files project files.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
weshaggard: See https://github.com/dotnet/corefx/blob/master/src/System.Globalization/tests/project.json which is an example where it has 2 target frameworks and a set of shared dependencies.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -83,41 +126,51 @@ private static JObject ReadProject(string projectJsonPath)
            } 
         }
 
+        private JToken GetFrameworkDependenciesSection(JObject projectJsonRoot, string framework = null)
+        {
+            if(string.IsNullOrWhiteSpace(framework))
```
weshaggard: Here is an example of where you are assuming empty framework means the shared dependencies section. Where what should happen here is that we should always add this set of dependencies to every framework.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
weshaggard: The entire point of dir.traversals.targets is to iterate on the ".builds" files. So I guess I'm still missing the point of why we are doing it here and there. So what is the difference between this and https://github.com/dotnet/corefx/pull/7604/files#diff-cd0fc5e0bad8102e1a45aa7575bdd102R175.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: In that example there are no Project References to convert, but your point is taken.  System.Globalization.Calendars appears to be applicable.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
weshaggard: @chcosta I still don't see that you have updated this to merged the different set of dependencies.

chcosta: Perhaps I'm still missing something you are describing.  The way it works, in this code, is that if you do not specify a framework, then we grab all of the frameworks sections, and add the dependencies to each of those...

            if (Frameworks == null || Frameworks.Length == 0)
            {
                Frameworks = projectRoot.SelectTokens("frameworks").SelectMany(f => f.Children().Select(c => ((JProperty)c).Name)).ToArray();
            }
            for (int i = 0; i < Frameworks.Length; i++)
            {
                JObject dependencies = GenerateDependencies(projectRoot, Frameworks[i]);
                projectRoot = UpdateProperty(projectRoot, dependencies, Frameworks[i]);
            }

The method you are referencing is called from GenerateDependencies, so we update each framework with the new dependencies.

weshaggard: I think there are still a couple things that don't add up for me. One are there any cases where framework will ever be null in this method? If I'm reading the code correctly then I don't think there are any cases.

Second when you are looking for existing dependencies below so you can update add them if they don't exist you aren't taking into account the shared dependencies in the project.json file that aren't under the framework section itself.

chcosta: To your first point, correct.  It's more important to me that I understand the scenario you were describing for the default (empty) case, but this is good cleanup.

To your second point, is this an issue?  Is this also just cleanup?  I haven't noticed any problem with duplicate entries in the generated json.  

{
  "dependencies": {
    "Microsoft.NETCore.Platforms": "1.0.1-rc3-24011-00",
    "System.Globalization": "4.0.11-rc3-24011-00",
    "System.Globalization.Calendars": "4.0.1-rc3-24011-00",
    "System.Linq.Expressions": "4.0.11-rc3-24011-00",
    "System.ObjectModel": "4.0.12-rc3-24011-00",
    "System.Runtime.Extensions": "4.1.0-rc3-24011-00",
    "System.Text.RegularExpressions": "4.0.12-rc3-24011-00",
    "System.Threading.Tasks": "4.0.11-rc3-24011-00",
    "xunit": "2.1.0",
    "xunit.netcore.extensions": "1.0.0-prerelease-00231-07"
  },
  "frameworks": {
    "dnxcore50": {
      "imports": "portable-net45+win8",
      "dependencies": {
        "System.Globalization.Calendars": "4.0.1-rc3-24011-00"
      }
    },
    "net46": {
      "dependencies": {
        "Microsoft.TargetingPack.NETFramework.v4.6": "1.0.1",
        "System.Globalization.Calendars": "4.0.1-rc3-24011-00"
      }
    }
  }
}

weshaggard:
LGTM

PR #628, opened 2016-04-08
Work around to MSBuild issue were Exec task doesn't read exit code correctly

joperezr:
Working around the issue in .Net Core MSBuild Exec task where the return code is not processed correctly. This will basically check if there were any failures in the run, and if so, then it will fail the exec call.

related: dotnet/corefx#7362 , microsoft/msbuild#566

cc: @weshaggard 
FYI: @mellinoe


mellinoe:
LGTM

weshaggard:
LGTM

PR #630, opened 2016-04-08
Fixing warnings that were introduced in coreclr

joperezr:
When adding native versioning to Windows binaries, a few warnings were introduced because some of the variables that we define in version.h were previously defined by a macro. To fix this, we should ensure that we first undefine the variables before assigning the values to them.

cc: @weshaggard 

FYI: @BruceForstall @AndyAyersMS 

related: dotnet/coreclr#4208

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -131,6 +131,20 @@
       Condition="'$(NativeVersionHeaderFile)'!='' and '$(GenerateVersionHeader)'=='true'">
 
     <ItemGroup>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -131,6 +131,20 @@
       Condition="'$(NativeVersionHeaderFile)'!='' and '$(GenerateVersionHeader)'=='true'">
 
     <ItemGroup>
+      <!-- Some of this variables might be defined by macros, so undefine then first to avoid build warnings. -->
```
BruceForstall: grammar: this => these, then => them

weshaggard:
LGTM

PR #631, opened 2016-04-10
Add support for microsoft.netcore.targets lineup package

chcosta:
Porting change to support building lineup package in the open.  I considered changing the name of this target to reflect that the meta packages are no longer part of the work performed by this target, but I left it as-is so that this target (imported before our internal definition) will be overridden when building internally.

I need to verify that this does not have any unexpected side-affects internally.  Please review, but don't merge.  I'll merge this change after further validation and sign-off.

/cc @weshaggard @ericstj


ericstj:
Lets not put this in buildtools.  See https://github.com/dotnet/corefx/pull/7610

PR #633, opened 2016-04-11
Add Inbox Validation for PCL profiles

ericstj:
PCL profiles support many of our packages "inbox" via reference
assemblies in the targeting pack.  Validate for these profiles.

CoreFx changes: https://github.com/ericstj/corefx/commit/80c6626eab4c4974a4fa4bc9fa31125884f207ab


weshaggard:
What made you chose just those specific portable profiles? I notice from your changes in corefx that it only appears to be the "Inbox" type placeholders. Do we expect any of our packages (beyond the pure PCL metadatareader/immutable/etc) to have accesses in the PCL folders?

ericstj:
I chose those folders as they are the broadest to support all PCL profiles.

There are packages that already support PCL with implementation assemblies as you point out.  This change addresses the assemblies in PCL which are part of the targeting packs.  That entire set is inbox.  Today we do not oob any of these assemblies with newer versions than in the ta rt hetung pack due to inability to do so on phone/store (which all of these profiles support).

weshaggard:
LGTM

PR #634, opened 2016-04-13
Project Reference to package dependency conversion fixes for corefx and build pipeline

chcosta:
- Move _BuildAgainstPackages property into buildtools common targets

- Project Reference to package dependency fixes
    - Use ParsePackageNames task to determine assembly version (I know this is "bad" and is on 
      the list of things to fix)
    - A couple of tasks were eliminated in favor of the ParsePackageNames before previous PR's 
      were merged, one fallout was that determining a version number broke because it wasn't 
      updated correctly to use the ParsePackageNames task
    - PackageNameRegEx explicitly looks for the "buildnumber-buildminor" instead of just grabbing 
      the rest of the package name as the Prerelease version because I noticed a bug in some of 
      the packages being produced on xplatforms where the package names 
      where "buildnumber.buildminor" instead of "-"
    - Move RegEx up to the top of the file where it is easy to see / edit / fix.

/cc @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
@@ -16,7 +16,6 @@
       a project is localized in German, then it is localized in all languages.
     -->
     <ExcludeLocalizationImport Condition="'$(ExcludeLocalizationImport)'=='' And !Exists('$(MSBuildProjectDirectory)/MultilingualResources/$(MSBuildProjectName).de.xlf')">true</ExcludeLocalizationImport>
-    
   </PropertyGroup>
 
   <!-- Assembly metadata indicating that an assembly is a framework (as opposed to user) assembly:
@@ -98,6 +97,12 @@
 
   <Import Project="$(MSBuildThisFileDirectory)Packaging.targets" Condition="'$(ExcludePackagingImport)'!='true' AND '$(MSBuildProjectExtension)' == '.pkgproj'" />
 
+  <!-- Property which signals whether to import buildagainstpackages.targets.  This target contains 
+       the targets necessary to build tests with project references converted to package dependencies.
+  -->
+  <PropertyGroup>
+    <_BuildAgainstPackages Condition="'$(KeepAllProjectReferences)' != 'true' and '$(IsTestProject)' == 'true' and '$(MSBuildProjectExtension)' != '.builds'">$(BuildTestsAgainstPackages)</_BuildAgainstPackages>
+  </PropertyGroup>
   <!-- 
     Import the default package restore and resolve targets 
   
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
```diff
@@ -98,6 +98,12 @@
 
   <Import Project="$(MSBuildThisFileDirectory)Packaging.targets" Condition="'$(ExcludePackagingImport)'!='true' AND '$(MSBuildProjectExtension)' == '.pkgproj'" />
 
+  <!-- Property which signals whether to import buildagainstpackages.targets.  This target contains 
+       the targets necessary to build tests with project references converted to package dependencies.
+  -->
+  <PropertyGroup>
+    <_BuildAgainstPackages Condition="'$(KeepAllProjectReferences)' != 'true' and $([System.Text.RegularExpressions.Regex]::IsMatch('$(MSBuildProjectFullPath)', '\\tests\\', System.Text.RegularExpressions.RegexOptions.IgnoreCase)) and '$(MSBuildProjectExtension)' != '.builds'">$(BuildTestsAgainstPackages)</_BuildAgainstPackages>
```
weshaggard: I think we should fold the IsTest part of this condition in with the IsTestProject property above in this file.

src/Microsoft.DotNet.Build.Tasks/ParsePackageNames.cs
```diff
@@ -31,7 +31,7 @@ public override bool Execute()
                     Log.LogWarning("PackageDrop does not exist - '{0}'", packageDrop);
                     continue;
                 }
-                IEnumerable<ITaskItem> packages = Directory.GetFiles(packageDrop)?.Select(f => new TaskItem(f));
+                IEnumerable<ITaskItem> packages = Directory.GetFiles(packageDrop)?.Select(f => new TaskItem(Path.GetFileNameWithoutExtension(f)));
```
weshaggard: coding feedback. You should try to avoid using "?." in these type of contexts. I don't think that return value can ever be null and if it were you would just be propagating your error further by now having a packages that is null and would nullref later. While it seems like a good idea to use everywhere I think you really need to think about when you use "?.", and the implications of the different code paths if the property is null.

weshaggard:
Looks reasonable beyond the couple comments.

src/Microsoft.DotNet.Build.Tasks/ParsePackageNames.cs
chcosta: Thanks Wes.  For some reason I was thinking that Directory.GetFiles returned null for an empty file set, which I think would be valid.  Actually taking a moment to looking at the API, clearly shows that it returns an empty array.  I'll update this and keep your feedback in mind.

PR #635, opened 2016-04-13
Fix small versioning issue on packages name

joperezr:
Fixing issue where sometimes the package name will have a '.' in the name instead of a '-'

cc: @jhendrixMSFT @maririos @chcosta


jhendrixMSFT:
LGTM

PR #638, opened 2016-04-14
Update package baseline to reflect API changes

ericstj:
/cc @weshaggard @chcosta 

Http has already changed, Linq.Expressions and Regex are changing in https://github.com/ericstj/corefx/commit/ecdca800cc7cf0d2a014615d021137a7588a6cf0

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/baseline.packages.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/baseline.packages.targets
@@ -146,7 +146,7 @@
       <Version>4.0.1</Version>
     </BaseLinePackage>
     <BaseLinePackage Include="System.Linq.Expressions">
-      <Version>4.0.11</Version>
+      <Version>4.1.0</Version>
     </BaseLinePackage>
     <BaseLinePackage Include="System.Linq">
       <Version>4.1.0</Version>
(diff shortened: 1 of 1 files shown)


weshaggard:
LGTM

ericstj:
I wish we had a better way to update this.  Right now it has to be in buildtools because it is used by both corefx and WCF.  Any chance we're going to have some small common repo that is shared between the two?

chcosta:
LGTM

weshaggard:
>I wish we had a better way to update this. Right now it has to be in buildtools because it is used by both corefx and WCF. Any chance we're going to have some small common repo that is shared between the two?

BuildTools is the repo that is shared between them :) 

What about putting them in an independent package, such as the line-up package, that could contain this and the line-up file that we can build in corefx and consume in WCF or other repo' that need this information.

chcosta:
Thanks for updating these Eric.  This (the corresponding corefx change) should fix an error we were seeing when restoring test projects with project reference converted to package dependencies.

ericstj:
> What about putting them in an independent package, such as the line-up package

That's an interesting idea.  I like it.  CoreFx will consume the file directly, and it doesn't need to know about WCF.  WCF will get this package updated when it updates CoreFx which is exactly what we want.  WCF can list its own baselines locally.  Once we do that we can actually validate that the baseline is in-sync with what is being built.

PR #643, opened 2016-04-18
Open source parts of the PCL targeting pack:

rchande:
- Tasks/targets for all versions of the targeting pack
- The v5.0 framework definitions and framework list

### DIFF
--- a/src/Portable/Frameworks/v5.0/.NET Framework 4.6.xml
+++ b/src/Portable/Frameworks/v5.0/.NET Framework 4.6.xml
@@ -1,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Framework
+    Identifier=".NETFramework"
+    Profile="*"
+    MinimumVersion="4.6"
+    Family=".NETFramework"
+    MinimumVisualStudioVersion="14.0"
+    DisplayName=".NET Framework"
+    MinimumVersionDisplayName="4.6"
+    PlatformArchitectures="AnyCPU;x86;x64"
+    IsBuiltIn="true"
+    IsHidden="false">
+    <RetargetData
+        ProjectLanguage="CSharp">
+        <NuGet
+            Identifier="net46.app"
+            PortableTargetMoniker="net452">
+            <Package
+                Identifier="Microsoft.NETCore"
+                Version="5.0.0" />
+            <Package
+                Identifier="Microsoft.NETCore.Portable.Compatibility"
+                Version="1.0.0" />
+        </NuGet>
+    </RetargetData>
+    <RetargetData
+        ProjectLanguage="VisualBasic">
+        <NuGet
+            Identifier="net46.app"
+            PortableTargetMoniker="net452">
+            <Package
+                Identifier="Microsoft.NETCore"
+                Version="5.0.0" />
+            <Package
+                Identifier="Microsoft.NETCore.Portable.Compatibility"
+                Version="1.0.0" />
+        </NuGet>
+    </RetargetData>
+</Framework>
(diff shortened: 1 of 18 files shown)


rchande:
Couple of questions:
Should add a license to these files? These are all msbuild tasks/targets or autogenerated XML files
Is it okay to add a top-level "Portable" directory to src?

weshaggard:
cc @ericstj 

We haven't been adding license headers to targets/props files so don't worry about that. 

Yes I'm fine with a top level Portable folder.

The title talks about targeting packages but I only see targets/props files in the review. What is your plan for the reference assemblies?

rchande:
@weshaggard: This is being done in response to a request from Xamarin. They only needed these files, no reference assemblies.

Pilchie:
Also, there *aren't* reference assemblies for Portable v5.0, we just depend on the packages.  @rchande will we need the pre-canned project.lock.json for doing DTAR however?

rchande:
@Pilchie: Per our conversation, that project.lock.json is for .NETCore, so not applicable here.

akoeplinger:
The `Copyright (C) Microsoft Corporation. All rights reserved.` in the targets/props looks wrong to me, shouldn't that either be .NET Foundation or removed?

rchande:
Good point @akoeplinger. From talking to @jaredpar and others, it sounds like we don't care about the copyright text on targets files, so I'll remove all of those instances.

mellinoe:
I don't really understand why these are going here. Is this just a convenient place to drop them for folks to grab? They aren't being used for anything here (i.e. in the buildtools nuget package), but perhaps I'm missing something?

akoeplinger:
That's a good point. I think a better place to put would be https://github.com/NuGet/NuGet.BuildTasks

rchande:
@mellinoe: Correct, we're adding these files so others can reference them if neccessary. I wasn't involved in the discussion but I think @Pilchie / @weshaggard / @Petermarcu decided this was as good a place as any to put them.

@akoeplinger The Nuget.BuildTasks repo is specifically for the build task used when you have a project.json; these files probably make even less sense over there.

akoeplinger:
@rchande yeah, maybe just wishful thinking on my part given that we're already submoduling that repo in Mono :smile:

mellinoe:
These would probably make most sense in the MSBuild repo, but I don't think it actually goes with the code that they ship from there, either.

> @akoeplinger The Nuget.BuildTasks repo is specifically for the build task used when you have a project.json; these files probably make even less sense over there.

A lot of these targets are for enabling various pieces of NuGet functionality in portable projects, aren't they?

weshaggard:
There isn't a strong reason to put these here beyond the reason there isn't another good repo and that they aren't worth their own repo. We will likely start including some other common targets and tasks for .NET Core here as well, including the targeting pack packages and such.

weshaggard:
@Pilchie @rchande I realize that Portable 5.0 doesn't actually have any reference assemblies as they come from packages but I was inferring from the title there were potentially other versions. Also there have been various talks of producing the nuget packages for the targeting pack references that include these. 

What are the goal for keeping these in-sync? Are these going to be a mirrored version or is the plan to make them the source of truth?

mhutch:
Is there anything blocking merging this?

ericstj:
I think there is still some open questions from @weshaggard to @Pilchie and @rchande.  I tend to agree with @mellinoe that it doesn't feel correct to put these in the buildtools repo.  Buildtools is all about what we need to enable build of our other repos, whereas these aren't even functional as is, they're only being checked in for reference.

Chrisboh:

mhutch:
We still need these 🙂

We ended up merging this branch into our own fork but it would be good to have an official location

PR #644, opened 2016-04-18
Using OfficialBuildId as the seed date if it gets passed in.

joperezr:
Using OfficialBuildId as the seed date if it gets passed in.

cc: @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
@@ -51,14 +51,15 @@ namespace Microsoft.DotNet.Build.Tasks
 
         public override bool Execute()
         {
-            // Calculating GeneratedRevision
+            // If OfficialBuildId is passed in, then use that to calculate the version and revision.
             if (string.IsNullOrEmpty(OfficialBuildId))
             {
                 GeneratedRevision = "0";
             }
             else
             {
-                GeneratedRevision = GetCurrentRevisionForBuildName(OfficialBuildId);
+                bool success = SetVersionAndRevisionFromBuildId(OfficialBuildId);
+                return success;
             }
 
             // Calculating GeneratedVersion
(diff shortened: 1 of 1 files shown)


joperezr:
FYI: @jhendrixMSFT Once this get's merged and consumed by corefx, then OfficialBuildId will be used as the seed date for the buildNumberMajor if passed in.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -93,15 +94,26 @@ public override bool Execute()
...
-                return match.Groups[1].Value;
+                DateTime buildIdDate;
+                if (!DateTime.TryParseExact(match.Groups[1].Value, dateFormat, enUS, DateTimeStyles.AssumeLocal, out buildIdDate))
+                {
+                    Log.LogError("The OfficialBuildId doesn't follow the expected({0}.rr) format: '{1}'", dateFormat, match.Groups[1].Value);
+                    return false;
+                }
+                buildIdDate = buildIdDate.ToUniversalTime();
+                GeneratedVersion = GetCurrentVersionForDate(buildIdDate, ComparisonDate);
+                GeneratedRevision = match.Groups[2].Value;
```
jhendrixMSFT: Don't we require this to always be two digits?  If so then I think we should validate that invariant.

joperezr: We do for corefx, but this code is intended to be used by different repos, so we decided to not  have that constrain on the BuildNumberMinor as some other consumer might want the rev to be only 1 digit, or more than 2 if they produce more than 99 builds a day.

jhendrixMSFT:
LGTM

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -93,15 +94,26 @@ public override bool Execute()
...
-            Regex regex = new Regex(@"\.(\d+)$");
+            Regex regex = new Regex(@"(\d{8})\.(\d+)$");
+            string dateFormat = "yyyyMMdd";
             Match match = regex.Match(buildId);
-            if (match.Success && match.Groups.Count > 1)
+            if (match.Success && match.Groups.Count > 2)
             {
-                return match.Groups[1].Value;
+                DateTime buildIdDate;
+                if (!DateTime.TryParseExact(match.Groups[1].Value, dateFormat, enUS, DateTimeStyles.AssumeLocal, out buildIdDate))
```
weshaggard: Is AssumeLocal correct?

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
weshaggard: Agreed we shouldn't be strict on this and allow for flexibility.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
```diff
@@ -93,15 +94,26 @@ public override bool Execute()
             return true;
         }
 
-        public string GetCurrentRevisionForBuildName(string buildId)
+        public bool SetVersionAndRevisionFromBuildId(string buildId)
         {
-            Regex regex = new Regex(@"\.(\d+)$");
+            Regex regex = new Regex(@"(\d{8})\.(\d+)$");
```
weshaggard: At a future point we should consider making the regex and the dateFormat as a parameter to the task.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
joperezr: I tried AssumeUniversal but for some reason the comparison still required me to convert it to Universal again, so I left AssumeLocal here and if you see [this line](https://github.com/dotnet/buildtools/pull/644/files#diff-67ed61292b05c4dc9a3ee429a0e1413aR110) I am transforming it back to UTC.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
joperezr: That is a good point. It would be very helpful to have it like that for flexibility. I can do this work later.

src/Microsoft.DotNet.Build.Tasks/GenerateCurrentVersion.cs
weshaggard: AssumeLocal here is just telling the parsing to assume the string it is parsing is matching the local timezone which may or may not be correct in all cases but for now I think we are safe to assume that.

PR #646, opened 2016-04-19
Remove package lists from BuildTools

ericstj:
This change removes package lists used for dependency harvesting from
buildtools.  Instead the lists should come from repositories consuming
buildtools.  This allows us to avoid buildtools updates when these lists
change.

I've added some errors to catch the case where a repo hasn't been
configured with a baseline.

Additionally I have added validation that any package in the baseline
must be built with some matching version in the baseline.  In doing this
I've permitted multiple versions of a package to be present in the
baseline and ApplyBaseline will choose the closest sufficient baseline
version.

CoreFx changes
Simple move: https://github.com/ericstj/corefx/commit/967eeae90d8e8c14e8d4dff03aaa9c0034d86109
Update: https://github.com/ericstj/corefx/commit/06bec361ada70f3ac009040a2a9e62e4db1568e4

WCF change:
https://github.com/ericstj/wcf/commit/28a716dd23b079aba5b03323d4f4cf704f0b2ce6

/cc @weshaggard @chcosta @roncain


ericstj:
@dotnet-bot test this please

ericstj:
Heads up on this, I plan to merge ASAP since we need the baseline update tonight.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -553,7 +549,10 @@
         <Exclude Condition="'%(FilePackageDependency.IsReferenceAsset)' != 'true'">Compile</Exclude>
       </FilePackageDependency>
     </ItemGroup>
-      
+    
+    <Error Condition="'@(BaseLinePackage)' == ''" Text="The BaseLinePackage item is not defined: ensure you have imported Microsoft.Private.PackageBaseLine.props from the Microsoft.Private.PackageBaseLine package" />
```
weshaggard: Do these need to be hard errors? That seems a little harsh for some repo's that might use packaging targets from BuildTools but don't care about the baselining. (i.e. I'm thinking CLI where I build the meta-package).

ericstj: I didn't want folks to miss this.  I suppose I could condition them better.

Can you remind me why you wouldn't want the baselining to happen in CLI?  I would imagine that solves some issues for you (eg: maintaining version list).

Issue #596, opened 2016-03-30
Need to fix the test library signing when running against the desktop


tarekgh:
we have manually defining the property 

```Xml
<DelaySign>false</DelaySign>
```

in the csproj file to workaround the signing issue, 
otherwise we'll get error cannot load the test library because it is not signed.

tarekgh:
@weshaggard just to tell I am not going to get into this soon. I just wanted to make sure it is ok to wait a little bit.

tarekgh:
Fixed by the linked PR

PR #647, opened 2016-04-19
Disable delay signing for Desktop tests

tarekgh:
To run the desktop test we need to have the test assembly get signed with the test.snk. Currently we have some workaround in some of the test projects to disable the delay signing. The change here is to centerline this in the tools and avoid duplicating the workaround in every test project

https://github.com/dotnet/buildtools/issues/596

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets
@@ -20,6 +20,15 @@
 
     <DelaySign>true</DelaySign>
     <DelaySign Condition="'$(UseOpenKey)' == 'true'">false</DelaySign>
+    <!--
+      We enable the signing for the test projects only on Windows because Roslyn not supporting full signing on non-Windows platforms.
+      Issue: https://github.com/dotnet/roslyn/issues/8210
+    -->
+    <DelaySign Condition="'$(OSEnvironment)' == 'Windows_NT' And '$(IsTestProject)' == 'true'">false</DelaySign>
+
+    <SignType Condition="'$(SignType)'=='' And '$(UseOpenKey)' == 'true'">real</SignType>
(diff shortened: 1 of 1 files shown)


tarekgh:
@weshaggard could you please have a look?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -43,6 +43,7 @@
   <PropertyGroup Condition="'$(TestWithCore)' != 'true'">
     <XunitExecutable Condition="'$(XunitExecutable)' == ''">xunit.console.exe</XunitExecutable>
     <DebugTestFrameworkFolder>dnx45</DebugTestFrameworkFolder>
+    <DelaySign Condition="'$(OSEnvironment)' == 'Windows_NT'">false</DelaySign>
```
weshaggard: I don't want this to be specific here. I want us to more generally support this in sign.targets

tarekgh:
@weshaggard let me know if this change match your thinking.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets
```diff
@@ -20,6 +20,11 @@
 
     <DelaySign>true</DelaySign>
     <DelaySign Condition="'$(UseOpenKey)' == 'true'">false</DelaySign>
+    <DelaySign Condition="'$(OSEnvironment)' == 'Windows_NT' And '$(IsTestProject)' == 'true'">false</DelaySign>
```
weshaggard: Please add a comment about why Windows_NT is necessary and reference the roslyn issue tracking it.

weshaggard:
Other then adding the comment LGTM

Thanks @tarekgh

tarekgh:
https://github.com/dotnet/roslyn/issues/8210

PR #648, opened 2016-04-19
Updating the version of Build Tools

joperezr:
Updating BuildTools + Adapting to the new PreReleaseResolvePackageAssets that were recently made.

FYI: @ericstj This new BuildTools contains the fix to use OfficialBuildId as the seed date when provided, which will fix the bug we have today where you can't produce a build tools package after 4pm.

cc: @weshaggard @mellinoe


joperezr:
AFAIK with this change [this lines in dir.props](https://github.com/dotnet/buildtools/blob/master/dir.props#L54-L58) shouldn't be used anymore so it should be safe to remove them. Is this correct @weshaggard?

src/Microsoft.DotNet.Build.Tasks.Packaging/test.Desktop/project.json
```diff
@@ -8,9 +8,11 @@
     "xunit.runner.visualstudio": "2.1.0"
   },
   "frameworks": {
-    "net45": { }
```
weshaggard: Why are you not referencing test-runtime in this case?

weshaggard:
>AFAIK with this change these lines in dir.props shouldn't be used anymore so it should be safe to remove them. Is this correct @weshaggard?

Yes I think that is correct.

Can you remind me what made you need to pass in all the RIDs?

joperezr:
Ok, I'll remove from dir.props then. The reason why we added the RIDs here is because of the changes that were recently made to PreReleaseResolveNugetPackageAssets task by @mellinoe, that will cause now that if the runtime ID that your project uses is not present in the project.json, then it throws warnings. Since we want the package to run in all of our supported runtimes, I added them all to the different project.jsons and that removed the warnings.

src/Microsoft.DotNet.Build.Tasks.Packaging/test.Desktop/project.json
joperezr: I believe that this tests do not even run when we do a build of buildtools, which is why it didn't need it. @ericstj is there a reason why Desktop packaging tests do not run on builds?

joperezr: After looking, it seems that even though there is a project for Packaging Desktop Tests, looks like the project is empty and has no tests at all, which would explain why no tests are run on the build. In this case we should either remove the project, or if we might add tests in the future then I guess I could add the test-runtime reference to have the setup ready.

mellinoe:
From what I understand, we are going to need to explicitly list our runtimes regardless, because `--infer-runtimes` is being removed eventually, and at that point it will be required.

weshaggard:
I can understand listing RIDs for projects that have an entry point like an exe or test which need to deploy specific runtime assets but for libraries I wouldn't expect them to have RIDs. @ericstj what is your thoughts?

mellinoe:
That sounds right, I believe we only need to list the runtimes if the project specifies CopyNuGetImplementations=true, which is true for Exe's for example. I wouldn't think that Microsoft.Cci.Extensions needs it, at the very least. I think some of our MSBuild tasks may need it as well, but I wouldn't expect Cci or Cci.Extensions to need it.

joperezr:
Ok, so I basically had to redo the work, but this time I'm only setting the runtimes on the project.jsons that needs them, and setting the right NugetRuntimeIdentifier on each project. @weshaggard @mellinoe PTAL to see if this was more or less what you envisioned. I have validated that the buildtools package still contains the required dlls.

dir.props
```diff
@@ -14,6 +14,11 @@
     <RoslynPackageName>Microsoft.Net.ToolsetCompilers</RoslynPackageName>
   </PropertyGroup>
 
+  <!-- MSBuild will set NuGetRuntimeIdentifier to be 'win' so we pre set it correctly here -->
+  <PropertyGroup>
+    <NuGetRuntimeIdentifier Condition="'$(NuGetRuntimeIdentifier)'==''">None</NuGetRuntimeIdentifier>
```
weshaggard: What happens if it is None? I would really like to not have this here either. If it is necessary can we put it in the targets file that actually uses this property,.

src/ApiCompat/ApiCompat.csproj
```diff
@@ -7,7 +7,6 @@
     <AssemblyName>ApiCompat</AssemblyName>
     <FileAlignment>512</FileAlignment>
     <DefineConstants>$(DefineConstants);COREFX</DefineConstants>
-    <NuGetRuntimeIdentifier>win7-x64</NuGetRuntimeIdentifier>
```
weshaggard: Isn't it needed here?

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
```diff
@@ -3,7 +3,8 @@
   <PropertyGroup>
     <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <NuGetRuntimeIdentifier>win</NuGetRuntimeIdentifier>
+    <NuGetRuntimeIdentifier>win7-x64</NuGetRuntimeIdentifier>
+    <CopyNuGetImplementations>true</CopyNuGetImplementations>
```
weshaggard: Please make sure that we don't end up with more things we don't want in our BuildTools package because of this now.

dir.props
joperezr: If we don't specifically set this, then it would be defaulted to win by MSBuild targets. The property is already being set [in the targets that use it](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets#L61) the problem is that it gets set by MSBuild earlier in the build process. That is why what I did is to set the default in here to be None, and then only override it on the projects that actually need it.

src/ApiCompat/ApiCompat.csproj
joperezr: No. even though this is an .exe, [we only copy the exe into the package](https://github.com/dotnet/buildtools/blob/master/src/nuget/Microsoft.DotNet.BuildTools.nuspec#L33) and not its dependencies, so it should be fine to have the runtime identifier be None for building. I only added this runtime identifier on the projects where we want to deploy the nuget dependencies to the package.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
joperezr: I have already, this basically doesn't make a difference to what we use to have with 'win' as the RID.

dir.props
weshaggard: If specifying "win" is trouble for the non-RID specific projects then we should pick a better default in the packageresolve.targets and not require this to be set in dir.props.

src/ApiCompat/ApiCompat.csproj
weshaggard: OK

dir.props
weshaggard: actually if it doesn't impact any library that doesn't need a RID I'm fine with just forcing the projects that do need a RID to specify it. I much prefer that then setting this here. 

The reason I'm pushing on this is that I want to strip down dir.props pretty far and move most of it to build.common.props and not have anything that isn't necessary in it.

joperezr: the defaulting of 'win' is not comming from packageresolve.targets, it's comming from: `C:\Program Files (x86)\MSBuild\Microsoft\NuGet\Microsoft.NuGet.targets` which we can't really change. Actually packageresolve.targets are simply mimicking what Microsoft.NuGet.targets do.

joperezr:
@weshaggard I have addressed the issue of having the global property in dir.props, let me know if this approach looks better.

weshaggard:
LGTM

PR #650, opened 2016-04-20
Make baseline errors less aggressive & include RID graph

ericstj:
Per previous PR feedback softening the baseline error messages.

Also fixing an old bug where we were missing the runtime.json from buildtools.  I also added a condition to search for it in the repo in order to help us remove some repo-configuration from corefx.

/cc @weshaggard


weshaggard:
:+1:

Issue #517, opened 2016-03-07
Depend on existing project model for project(.lock).json validation


dagood:
Currently we use Json.NET calls which were coded only to the scenarios I've encountered. Using an existing library that creates models from project.json files removes the logic duplication. A project model will also be more resilient to changes in the lockfile and project.json file formats.

For example, we could use `NuGet.ProjectModel` or `Microsoft.DotNet.ProjectModel`.

This needs to apply to the project.json validation and post-restore lockfile validation. Depending on whether these project models are mutable, there may be more difficulty changing auto-upgrade.

MichaelSimons:
Triage: we don't want to do this because we want to limit dependencies.

PR #652, opened 2016-04-20
Add lockfile validation for exact nuget restores

dagood:
Examines lockfiles to make sure all dependencies were found and restored correctly.

When a `project.json` specifies a version of a package that doesn't exist, NuGet tries to find a best match. In our case, we want an error when this happens because it can cause a bunch of problems:

1. The build isn't repeatable. If these fuzzy-matched dependencies are used, the build depends on which packages are available on the feed which may change at any time. (Similar to `*` dependencies.)
1. Restore is slowed. Now that NuGet depends on the feed state, it needs to fetch a list of versions from the feed every restore.
1. NuGet can encounter corrupted files that otherwise would be ignored, like in https://github.com/dotnet/corefx/issues/7902.
1. The wrong packages sometimes break builds.

This would activate the validation in CoreFX: https://github.com/dotnet/corefx/compare/master...dagood:ensure-exact-restore

This doesn't run in sync.cmd, but neither does project.json validation right now. That might be something to change when adding this to CoreFX, but either way it would be caught by CI `RestoreDuringBuild`.

/cc @ericstj @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/ValidateExactRestore.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/ValidateExactRestore.cs
@@ -1,0 +1,84 @@
+﻿using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Newtonsoft.Json.Linq;
+using NuGet.Versioning;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class ValidateExactRestore : Task
+    {
+        [Required]
+        public ITaskItem[] ProjectLockJsons { get; set; }
+
+        private HashSet<string> _errors = new HashSet<string>();
+
+        public override bool Execute()
+        {
+            foreach (var projectLockItem in ProjectLockJsons)
+            {
+                ValidateLockFile(projectLockItem.ItemSpec);
+            }
+            return !Log.HasLoggedErrors;
+        }
+
+        private void ValidateLockFile(string lockFilePath)
+        {
+            var lockfile = VisitProjectDependencies.ReadJsonFile(lockFilePath);
+
+            var lockedDependencyVersions = lockfile["libraries"]
+                .Children<JProperty>()
+                .Select(p => p.Name.Split('/'))
+                .ToLookup(libParts => libParts[0], libParts => NuGetVersion.Parse(libParts[1]));
+
+            // Look at dependencies for all frameworks.
+            var requestedDependencies = lockfile["projectFileDependencyGroups"]
+                .Children<JProperty>()
+                .SelectMany(group => group.Value.Value<JArray>().Values<string>());
+
+            foreach (var dependency in requestedDependencies)
+            {
+                string[] dependencyParts = dependency.Split(' ');
+                string requestedId = dependencyParts[0];
+                if (dependencyParts.Length < 3)
+                {
+                    // Some dependencies have no versions, e.g. 'test-runtime' and 'net46-test-runtime'.
+                    continue;
+                }
+
+                string requestedVersion = dependencyParts[2];
+                NuGetVersion requestedNuGetVersion = NuGetVersion.Parse(requestedVersion);
+
+                IEnumerable<NuGetVersion> restoredVersions = lockedDependencyVersions[requestedId];
+
+                // Check if the requested package is exactly included in the "libraries" section.
+                if (!restoredVersions.Contains(requestedNuGetVersion))
+                {
+                    HandleNonExistentDependency(requestedId, requestedVersion, restoredVersions, lockFilePath);
+                }
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/ValidateExactRestore.cs
```diff
@@ -0,0 +1,84 @@
...
+                .ToLookup(libParts => libParts[0], libParts => NuGetVersion.Parse(libParts[1]));
+
+            // Look at dependencies for all frameworks.
+            var requestedDependencies = lockfile["projectFileDependencyGroups"]
+                .Children<JProperty>()
+                .SelectMany(group => group.Value.Value<JArray>().Values<string>());
+
+            foreach (var dependency in requestedDependencies)
+            {
+                string[] dependencyParts = dependency.Split(' ');
```
ericstj: FYI: here is the source for this string https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs#L66

ericstj:
This looks ok.  We should also file bugs for why https://github.com/dotnet/cli/tree/rel/1.0.0/src/Microsoft.DotNet.ProjectModel doesn't work for us.

ericstj:
Thanks for doing this, it will fix a lot of random issues we've been hitting with the build.

dagood:
Sure, glad it's getting in. :smile:

> We should also file bugs for why https://github.com/dotnet/cli/tree/rel/1.0.0/src/Microsoft.DotNet.ProjectModel doesn't work for us.

In the other issue I didn't mean that it won't work, just that pulling the packages into buildtools added enough work that my priorities worked out differently. Looking at the lockfile model (and the one in `NuGet.ProjectModel`) it actually seems fine. Maybe I could go ahead and do that for this PR--it seems like the changes should be straightforward. Do you think we should go with the CLI model or NuGet one, though? I'm leaning towards NuGet because we're already using the versioning part, but maybe there's more to consider here.

mellinoe:
> When a project.json specifies a version of a package that doesn't exist, NuGet tries to find a best match. In our case, we want an error when this happens because it can cause a bunch of problems:

This seems like something more than just we would want if it can cause a bunch of problems. Is there something we can configure in nuget to catch this, or should there be?

dhusemann:
@mellinoe as a view from the outside watching the sausage being made, I agree with you, Nuget shouldn't be completing the restore or import with Files that are best fit when exact fit is called out. It should error out and offer a recovery point to attempt a redownload.  in CI it should document what file didn't load and fail that build. although the additional checks to verify also sounds like a good idea. now back to watching.

ericstj:
@mellinoe @dhusemann we could use exact match semantics in our project.json `[4.0.1-rc3-12345]`

dagood:
@ericstj Hmm, is that compatible with lifting? I remember trying that a long time ago and it failed somewhere along the way, but if that should be valid maybe we should look into it.

But yeah, right now we use `>=` dependencies so NuGet's in the clear trying to find matches that are `>=` what we specify, but `[something]` evaluates to `package.id >= something <= something` in the lockfile.

ericstj:
> Hmm, is that compatible with lifting

We actually don't lift anymore, that said I still don't like it.  Every time we've tried to use exact version syntax in the past it has caused problems.  P2P references are broken.  Even though we don't use it today, dotnet pack from project.json would get the wrong dependencies in the package.  I'd like to do work to make CoreFx to set an example, and using exact version is not the example we want to set.

dagood:
If it did work for P2P references and dotnet pack (and whatever else) would it be reasonable? Or are these issues caused by design/spec decisions?

I'm guessing the latter, because in the end we want to produce NuGet packages that use `>=` dependencies and it seems like it would be good for the dependencies in the project.json to match that. (But maybe the link between project.json and nupkg isn't as strong as I think.)

ericstj:
>I'm guessing the latter, because in the end we want to produce NuGet packages that use >= dependencies and it seems like it would be good for the dependencies in the project.json to match that. 

Exactly.

Issues caused are by-design.  Folks on the NuGet team discourage use of exact-version syntax.

ericstj:
This LGTM as is.  I think we should file a separate issue to address using ProjectModel where possible.

dagood:
Ok, that makes sense. Do you think it would be reasonable then for NuGet to support something more to the point like `--no-fuzzy-match`?

I'll update the existing issue for using a ProjectModel (https://github.com/dotnet/buildtools/issues/517), get this merged and start a PR to use it. Thanks!

ericstj:
Yeah, I think some parameter on restore is a better option.

weshaggard:
Yes I was thinking the same thing after reading this thread. Should we file an issue on nuget?

dagood:
I think it's worth filing even just to get the NuGet opinion on the problem--I'll do it.

PR #653, opened 2016-04-20
Hides non-compiled entries from VisualStudio

manu-silicon:
Doing so prevents tools such as ReSharper from trying to analyze files that are not valid if we do not met the condition.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
@@ -46,14 +46,26 @@
     </EmbeddedResource>
   </ItemGroup>
   
-  <ItemGroup Condition="Exists('$(StringResourcesPath)') And '$(SkipCommonResourcesIncludes)'=='' AND '$(OmitResources)'!='true'">
-    <Compile Condition="'$(MSBuildProjectExtension)' == '.csproj'" Include="$(CommonPath)/System/SR.cs">
-      <Visible>true</Visible>
-      <Link>Resources/Common/SR.cs</Link>
-    </Compile>
-    <Compile Condition="'$(MSBuildProjectExtension)' == '.vbproj'" Include="$(CommonPath)/System/SR.vb">
-      <Visible>true</Visible>
-      <Link>Resources/Common/SR.vb</Link>
-    </Compile>
-  </ItemGroup>
+  <Choose>
+    <When Condition="Exists('$(StringResourcesPath)') And '$(SkipCommonResourcesIncludes)'=='' AND '$(OmitResources)'!='true'">
+      <Choose>
+        <When Condition="'$(MSBuildProjectExtension)' == '.csproj'">
+          <ItemGroup>
+            <Compile Include="$(CommonPath)/System/SR.cs">
+              <Visible>true</Visible>
+              <Link>Resources/Common/SR.cs</Link>
+            </Compile>
+          </ItemGroup>
+        </When>
+        <When Condition="'$(MSBuildProjectExtension)' == '.vbproj'">
+          <ItemGroup>
+            <Compile Include="$(CommonPath)/System/SR.vb">
+              <Visible>true</Visible>
+              <Link>Resources/Common/SR.vb</Link>
+            </Compile>
+          </ItemGroup>
+        </When>
+      </Choose>
+    </When>
+  </Choose>
 </Project>


jkotas:
@weshaggard Could you please take a look?

weshaggard:
cc @tarekgh 

@manu-silicon do you know why this actually helps or did you just do trial and error to find something that works?

manu-silicon:
@weshaggard What is happening is that in Visual Studio with ReSharper, Resharper perform an analysis of all visible files in the Solution Explorer. If those files do not compile (due to C# compiler errors), then ReSharper will tell you about those errors. One could say it is a ReSharper limitation that it does not look into the `Condition' attribute of the ItemGroup, but it might not be obvious to figure this out since it is mostly part of the msbuild engine processing to figure this out.

However when choosing the ```<Choose><When Condition"...">``` tactic, VisualStudio would know not to display those classes and ReSharper won't see them either. Thus no more error.

It is not a critical issue, but since I cannot get away from not using ReSharper, having those errors appearing all the time is quite distracting and thus the reason for this PR.

weshaggard:
Thanks @manu-silicon. I just wanted to clarify one aspect of what you said, which is after your change they don't show-up at all in solution explorer? I'm pretty sure @tarekgh put Visible=true on them so they could show up and people could F12 to them and so all the usages of SR don't show up with squiggles in VS. I'm less concerned about the files being listed in Solution Explorer but I really would prefer that SR calls in the code don't show up as errors with squiggles, so if you can ensure that then I'm good with the change.

manu-silicon:
This is correct. The files won't show because in the CoreRT project the condition is evaluated to false. If in a project it evaluates to true they will show. So it is my understanding that F12 will still work for project that uses them and no red squiggles will appear. 

Manu

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -46,14 +46,18 @@
...
-    <Compile Condition="'$(MSBuildProjectExtension)' == '.vbproj'" Include="$(CommonPath)/System/SR.vb">
-      <Visible>true</Visible>
-      <Link>Resources/Common/SR.vb</Link>
-    </Compile>
-  </ItemGroup>
+  <Choose>
+    <When Condition="Exists('$(StringResourcesPath)') And '$(SkipCommonResourcesIncludes)'=='' AND '$(OmitResources)'!='true'">
+      <ItemGroup>
+        <Compile Condition="'$(MSBuildProjectExtension)' == '.csproj'" Include="$(CommonPath)/System/SR.cs">
+          <Visible>true</Visible>
```
weshaggard: Given these aren't actually visible any longer we probably can remove the Visible and the Link metadata from here so it doesn't confuse anyone expecting them to be displayed in solution explorer.

manu-silicon: I understand removing Visible, but shouldn't we keep Link?

weshaggard: Link is just metadata that tell VS what folder to put this in when displaying in solution explorer. So if it isn't being displayed any longer then it doesn't really add any value.

manu-silicon: It might get displayed by VS when the condition evaluates to true, so this is why I think we should keep it.

weshaggard: OK I think I miss-understood part of your earlier explanation as I thought where never going to be displayed in VS now. If they are then perhaps we do still need Visible as well.

weshaggard:
@joperezr would you mind testing out this change with the investigation you are doing around the resources?

joperezr:
Will do. Although I won't be able to do this until next week since I'm OOF, is that ok? Is the testing of this blocking this PR to get merged?

weshaggard:
I would like to validate that this doesn't regress the VS scenarios you are working on so yes lets wait until you can verify it next week before we merge this.

@manu-silicon Thanks for the contribution and patience

joperezr:
@manu-silicon I don't see any scenarios regressed by this so I think it's fine to merge. However, if we are already doing this change, could we also move to a `<Choose>` statement the Itemgroup conditioning of when to use `SR.cs` and when to use `SR.vb`? This way, when the `<When>` condition is true, only the file that makes sense (.cs or .vb) is shown in the solution explorer, since today both of them will in this case.
The code would look something like:
```xml
<Choose>
    <When Condition="Exists('$(StringResourcesPath)') And '$(SkipCommonResourcesIncludes)'=='' AND '$(OmitResources)'!='true'">
      <Choose>
        <When Condition="'$(MSBuildProjectExtension)' == '.csproj'">
          <ItemGroup>
            <Compile Include="$(CommonPath)/System/SR.cs">
              <Visible>true</Visible>
              <Link>Resources/Common/SR.cs</Link>
            </Compile>
          </ItemGroup>
        </When>
        <When Condition="'$(MSBuildProjectExtension)' == '.vbproj'">
          <ItemGroup>
            <Compile Include="$(CommonPath)/System/SR.vb">
              <Visible>true</Visible>
              <Link>Resources/Common/SR.vb</Link>
            </Compile>
          </ItemGroup>
        </When>
      </Choose>
    </When>
  </Choose>
```

Let me know what you guys think.

manu-silicon:
@joperezr I've updated the PR according to your suggestion

joperezr:
Perfect, thanks! LGTM.

weshaggard:
LGTM 

Thanks @manu-silicon

PR #655, opened 2016-04-21
Hook .ilproj files into the build

mellinoe:
This is a few small changes to enable building "ilproj" files as a regular participant in the build. This just imports the existing targets file if the file's extension is detected as ".ilproj". This will be used to build the `System.Unsafe` library in corefx (https://github.com/mellinoe/corefx/tree/system.unsafe).

There is also a minor modification to the calling syntax for ilasm so that it will work with the x-plat version of ilasm. "Forward-slash syntax" doesn't work in the x-plat version of ilasm. Currently it seems like the nuget packages for ilasm are broken, though, so I have not been able to incorporate it into the tool runtime yet. I have done manual testing of the scenario.

* Import IL.targets if the project extension is ".ilproj"
* Use hyphen syntax for ilasm flags, the forward-slash syntax does not
  work on non-Windows ilasm.

@weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -60,6 +60,9 @@
   <Import Project="depProj.targets"
           Condition="'$(MSBuildProjectExtension)' == '.depproj'" />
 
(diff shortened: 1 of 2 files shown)


weshaggard:
LGTM

mellinoe:
Thanks Wes.

PR #656, opened 2016-04-22
Fix UpdatePackageDependencyVersion: shouldn't require a version on all dependencies

dagood:
In corefx this fixes references to 'test-runtime' and 'net46-test-runtime', which don't have `version` values.

I only noticed this now because `UpdatePackageDependencyVersion` is the easiest way to fix incorrect package versions when a prerelease upgrade doesn't account for the major, minor, or patch version changing, and I tried out this flow when preparing the corefx lockfile validation PR.

/cc @ericstj @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/UpdatePackageDependencyVersion.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/UpdatePackageDependencyVersion.cs
@@ -22,7 +22,7 @@ namespace Microsoft.DotNet.Build.Tasks
             string dependencyVersion;
             if (package.Value is JObject)
             {
-                dependencyVersion = package.Value["version"].Value<string>();
+                dependencyVersion = package.Value["version"]?.Value<string>();
             }
             else if (package.Value is JValue)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/UpdatePackageDependencyVersion.cs
```diff
@@ -22,7 +22,7 @@ public override bool VisitPackage(JProperty package, string projectJsonPath)
             string dependencyVersion;
             if (package.Value is JObject)
             {
-                dependencyVersion = package.Value["version"].Value<string>();
+                dependencyVersion = package.Value["version"]?.Value<string>();
```
weshaggard: Is this going to cause issues in any other places were someone forgets the version? Should we consider also checking for type=project?

dagood: If you mean something like `"System.Foo": {}`, then maybe we should include that in the project.json validation. It should be very rare that someone actually calls `UpdatePackageDependencyVersion` though, so I think it makes sense for this task to be lenient.

weshaggard: I missed this is part of the update I thought it was part of the validate, so I'm OK with this.

weshaggard:
LGTM

ericstj:
LGTM

Issue #632, opened 2016-04-11
Long sub-paths in Microsoft.DotNet.BuildTools


anurse:
`Microsoft.DotNet.BuildTools` has a very long sub-path in it, which makes it prone to long-path issues on Windows.

I've ended up with a path like this in my dotnet/cli clone: `C:\Users\anurse\Code\dotnet\cli\pkg\packages\Microsoft.DotNet.BuildTools\1.0.25-prerelease-00228-02\lib\portableTargets\packages\Microsoft.Portable.Targets\0.1.1-dev\contentFiles\any\any\Extensions\Microsoft\Portable\v5.0\Microsoft.Portable.VisualBasic.targets` which works out to 260 characters, AKA `MAX_PATH` :(.

I realize I could shorten some of my own paths to manage this, but the sub-path within the package is 160 characters (excluding the package ID and version), which gives users very little space to work with :).

Is there any way to shorten the paths to these items?

/cc @eerhardt @weshaggard

joperezr:
Yes, we will basically move the restore to the top packages dir. I'll take care of this.

PR #657, opened 2016-04-22
Fixing long paths in some packages due to the restore location

joperezr:
Previously we had a nested 'packages' dir inside the BuildTools package, which made paths of the restored packages to be super long. This PR will use the original packages dir instead.

Note: this doesn't require the packages dir to be passed in from the consumer's init-tools, but if it is not passed, we will assume that the packages dir is `{repoRoot}/packages`

cc: @weshaggard 

fixes: #632

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -4,6 +4,10 @@ setlocal
 set PROJECT_DIR=%~1
 set DOTNET_CMD=%~2
 set TOOLRUNTIME_DIR=%~3
+set PACKAGES_DIR=%4
+if [%PACKAGES_DIR%] == [] set PACKAGES_DIR=%TOOLRUNTIME_DIR%
+:: Remove quotes to the packages directory
+set PACKAGES_DIR=%PACKAGES_DIR:"=%
 IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -4,6 +4,10 @@ setlocal
 set PROJECT_DIR=%~1
 set DOTNET_CMD=%~2
 set TOOLRUNTIME_DIR=%~3
+set PACKAGES_DIR=%4
+if [%PACKAGES_DIR%] == [] set PROJECT_DIR=%PROJECT_DIR%\packages\
```
weshaggard: I'm not sure I like making this assumption. Can we instead just put this under the TOOLRUNTIME itself?

joperezr: do  you mean putting it there always? or just as a default in the case where Packages_DIR is not passed in?

weshaggard: I was meaning always, but now that I'm thinking about it perhaps we should just use it as a default. That will hopefully help keep things self-contained under the tool runtime directory.

joperezr: Sounds good. I just pushed a rebased commit to fix the merge conflicts and includded the fix from your feedback.

weshaggard:
LGTM

joperezr:
@AlfredoMS Looks like the Pre-Merge signing tool is having issues. I'm getting the following message:
```
Error communicating with Github, please try again later.
``` 
I'll merge now and will come back to sign whenever the service is back online.

PR #659, opened 2016-04-22
MS.CCI.Extensions needs to support desktop

ericstj:
We use this library in tools that run on desktop.  As such it needs to
be built as a PCL.

/cc @mellinoe @weshaggard


weshaggard:
LGTM

mellinoe:
LGTM

PR #660, opened 2016-04-22
Fix calculation of NETStandard min version

ericstj:
We were not running ExpandProjectReferences during GetPackageIdentity
evaluation if the Version was explicitly set by the project.  The result
was a race condition when calculating NETStandard version.  If the
pkgproj had previously built it would have gotten the correct version,
if not it would have gotten the default: NETStandard1.0.

This was causing https://github.com/dotnet/corefx/issues/7881.

I've cleaned up a bit of the target sequence to make it clearer when
thing run.

/cc @weshaggard @chcosta


weshaggard:
LGTM

chcosta:
LGTM

PR #662, opened 2016-04-22
Rolling back the change where I added 'set -x' since it doesnt work as expected

joperezr:
Looks like `set -x` and `set +x` don't exactly work as using `echo on/off` on Windows, since even if you are redirecting the output to a file, it would still get printed in the standard output. Rolling back to have the echos for now, until we have a better solution for that.

cc: @dagood


dagood:
Thanks, LGTM!

PR #663, opened 2016-04-25
Ensure Desktop assets work with packages.config

ericstj:
Desktop projects by default use packages.config, we shouldn't require
folks to use project.json + RID if we don't actually depend on RID
selection as is the case for most desktop packages.

We should make sure that project types that can use packages.config
are supported by our packages.  I've added empty RID entires to the
DefaultValidateFramework list for all of these frameworks.

I'm also addressing a problem where we might get a different asset when
someone uses a RID over not having a RID: eg the package has lib/net45
asset that is different from its runtimes/win/lib/netstandard1.1 asset.
To catch this case I added a check that tries to see if there is a
better matching asset by framework for any runtime asset.

This caught a number of cases where we were making this mistake for
desktop, so I've made a general rule that any desktop project with a
RID will also put the asset in the RID-less folder.

/cc @weshaggard 

CoreFx change consuming this buildtools update: https://github.com/ericstj/corefx/commit/0c3d7e74d0161d3cf42a074fdf5806029d1f7666

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
@@ -73,10 +73,12 @@
       <XmlDocFile Remove="@(XmlDocFile)" Condition="!Exists('%(XmlDocFile.Identity)')"/>
     </ItemGroup>
 
-    <!-- Desktop libraries that contain types in classic assemblies need to be included in ref
-         so that the types unify with the classic assemblies from the targeting pack. -->
     <PropertyGroup>
-      <PackageAsRefAndLib Condition="'$(PackageAsRefAndLib)' == '' AND '$(IsPartialFacadeAssembly)' == 'true' AND $(PackageTargetFramework.StartsWith('net4', StringComparison.OrdinalIgnoreCase))">true</PackageAsRefAndLib>
+      <IsNetFxLibrary Condition="'$(IsNetFxLibrary)' == '' AND $(PackageTargetFramework.StartsWith('net4', StringComparison.OrdinalIgnoreCase))">true</IsNetFxLibrary>
+      
+      <!-- Desktop libraries that contain types in classic assemblies need to be included in ref
+         so that the types unify with the classic assemblies from the targeting pack. -->
+      <PackageAsRefAndLib Condition="'$(PackageAsRefAndLib)' == '' AND '$(IsPartialFacadeAssembly)' == 'true' AND '$(IsNetFxLibrary)' == 'true'">true</PackageAsRefAndLib>
 
       <!-- A reference asset is any file contributed from a project that is contributing a reference,
            not just the file in the ref folder. -->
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -73,18 +73,20 @@
       <XmlDocFile Remove="@(XmlDocFile)" Condition="!Exists('%(XmlDocFile.Identity)')"/>
     </ItemGroup>
 
-    <!-- Desktop libraries that contain types in classic assemblies need to be included in ref
-         so that the types unify with the classic assemblies from the targeting pack. -->
     <PropertyGroup>
-      <PackageAsRefAndLib Condition="'$(PackageAsRefAndLib)' == '' AND '$(IsPartialFacadeAssembly)' == 'true' AND $(PackageTargetFramework.StartsWith('net4', StringComparison.OrdinalIgnoreCase))">true</PackageAsRefAndLib>
+      <IsDesktopLibrary Condition="'$(IsDesktopLibrary)' == '' AND $(PackageTargetFramework.StartsWith('net4', StringComparison.OrdinalIgnoreCase))">true</IsDesktopLibrary>
```
weshaggard: From a naming prospective I think we should call this IsNetFxLibrary instead of IsDesktopLibrary.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -114,6 +116,12 @@
       <PackageDestination Include="$(PackageTargetPath)">
         <TargetFramework Condition="'$(PackageTargetFramework)' != ''">$(PackageTargetFramework)</TargetFramework>
       </PackageDestination>
+      
+      <!-- RID-specific desktop libraries should also be packaged without a RID to work in packages.config projects -->
+      <PackageDestination Condition="'$(IsDesktopLibrary)' == 'true' AND '$(PackageTargetRuntime)' != '' AND '$(ExludeDesktopLibraryFromLib)' != 'true'" Include="lib/$(PackageTargetFramework)">
```
weshaggard: Do we need the other ExludeDesktopLibraryFromLib property? If so I would rename Desktop similarly.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: Sure

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: I want to have it there as an opt-out in case a desktop library needs to have a RID specific implementation.

ericstj:
Applied rename suggestions

weshaggard:
LGTM

ericstj:
I'm going to hold off on merging until corefx and WCF side are ready to merge.

Issue #621, opened 2016-04-07
"Built By" version string in the non-Windows builds can't get the computer name or username


joperezr:
We should figure out a way of getting this info in non-Windows machines.

cc: @gkhanna79

bartonjs:
```
jbarton@jsb-ubuntu2:~/coreclr$ whoami
jbarton
jbarton@jsb-ubuntu2:~/coreclr$ hostname
jsb-ubuntu2
jbarton@jsb-ubuntu2:~/coreclr$ which whoami
/usr/bin/whoami
jbarton@jsb-ubuntu2:~/coreclr$ which hostname
/bin/hostname
```

joperezr:
awesome, do you know if this two commands are available in all Linux distros?

bartonjs:
They've been on every Unix system I've ever used.  And they're common enough that they're part of BusyBox.  https://busybox.net/downloads/BusyBox.html#commands

PR #664, opened 2016-04-25
Calculate BuiltBy String for non-Windows environments

joperezr:
CC: @weshaggard 

FYI: @gkhanna79 

fixes: #621

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -233,17 +233,20 @@
 
   <Target Name="GenerateBuiltByString" Condition="'$(BuiltByString)'==''">
 
-    <!-- If building on Windows, default BuiltByString to use Computername and username -->
-    <BuiltByString Condition="'$(OSEnvironment)'=='Windows_NT'">%20built by: $(COMPUTERNAME)-$(USERNAME)</BuiltByString>
+    <PropertyGroup>
+      <VersionUserName Condition="'$(VersionUserName)'==''">$(USERNAME)</VersionUserName>
+      <VersionHostName Condition="'$(VersionHostName)'==''">$(COMPUTERNAME)</VersionHostName>
+    </PropertyGroup>
 
-    <Exec Command="whoami" Condition="'$(OSEnvironment)'!='Windows_NT'" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+    <Exec Command="whoami" Condition="'$(OSEnvironment)'!='Windows_NT' AND '$(VersionUserName)'==''" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="VersionUserName" />
     </Exec>
-    <Exec Command="hostname" Condition="'$(OSEnvironment)'!='Windows_NT'" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+    <Exec Command="hostname" Condition="'$(OSEnvironment)'!='Windows_NT' AND '$(VersionHostName)'==''" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="VersionHostName" />
     </Exec>
-    <PropertyGroup Condition="'$(BuiltByString)'==''">
-      <BuiltByString>%20built by: $(VersionUserName)-$(VersionHostName)</BuiltByString>
+
+    <PropertyGroup>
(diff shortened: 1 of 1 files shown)


gkhanna79:
What is the value of this string on VSO machines?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -230,7 +230,18 @@
       <OldBuildVersionFiles Include="$(BuildVersionFilePath)BuildVersions-*.props" Exclude="$(BuildVersionFilePath)%(BuildVersionFileItem.Filename).props" />
     </ItemGroup>
     <Delete Files="@(OldBuildVersionFiles)" TreatErrorsAsWarnings="true"/>
+  </Target>
 
+  <Target Name="GenerateBuiltByString" Condition="'$(OSEnvironment)'!='Windows_NT' AND '$(BuiltByString)'==''">
+    <Exec Command="whoami" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
```
weshaggard: Are these tools guaranteed to be on the machines? If they aren't then I would probably just avoid this completely.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -178,7 +178,7 @@
 
     <ItemGroup>
       <SourceFileLines />
-      <SourceFileLines Include="static char sccsid%5B%5D %5F%5Fattribute%5F%5F%28%28used%29%29 %3D %22%40%28%23%29Version $(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)%22%3B" />
+      <SourceFileLines Include="static char sccsid%5B%5D %5F%5Fattribute%5F%5F%28%28used%29%29 %3D %22%40%28%23%29Version $(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor) $(BuiltByString)%22%3B" />
```
weshaggard: You should be able to remove the space before BuiltByString property as you included it in the string.

joperezr: That is correct, good catch, I missed that.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
joperezr: AFAIK they are present in almost all Unix systems already. In case where they are not present, I am still setting the IgnoreExitCode to false so worst case scenario build will still pass and builtbystring would be something like `built by: -`

joperezr:
> What is the value of this string on VSO machines?

I'm not sure there is a lot of value except on knowing if the binary came out from an official build rather than from a private build, in any case the reason why we are doing this is because this string is always displayed in Windows and we want to have parity between Windows and non-Windows builds and version strings.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -26,7 +26,7 @@
 
   <PropertyGroup>
     <GenerateAssemblyInfo Condition="'$(GenerateAssemblyInfo)'==''">true</GenerateAssemblyInfo>
-    <BuiltByString Condition="'$(BuiltByString)'==''">%20built by: $(COMPUTERNAME)-$(USERNAME)</BuiltByString>
+    <BuiltByString Condition="'$(BuiltByString)'=='' AND '$(OSEnvironment)'=='Windows_NT'">%20built by: $(COMPUTERNAME)-$(USERNAME)</BuiltByString>
```
weshaggard: For simplicity I think I would just move this to inside the target as well, and just always default to those properties for all OS's and for then for non-windows call those tools only if those properties aren't set.

joperezr: Done.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -230,7 +229,22 @@
       <OldBuildVersionFiles Include="$(BuildVersionFilePath)BuildVersions-*.props" Exclude="$(BuildVersionFilePath)%(BuildVersionFileItem.Filename).props" />
     </ItemGroup>
     <Delete Files="@(OldBuildVersionFiles)" TreatErrorsAsWarnings="true"/>
+  </Target>
+
+  <Target Name="GenerateBuiltByString" Condition="'$(BuiltByString)'==''">
+
+    <!-- If building on Windows, default BuiltByString to use Computername and username -->
+    <BuiltByString Condition="'$(OSEnvironment)'=='Windows_NT'">%20built by: $(COMPUTERNAME)-$(USERNAME)</BuiltByString>
```
weshaggard: Almost what I had in mind :)

To enable folks to setup an environment so they don't have to run this whoami/hostname commands we can allow them to set COMUTERMAME and USERNAME env variables. So I would default VersionUserName and VersionHostName to those guys and only execute whoami/hostname if we are not on Windows and if the particular property is empty.

joperezr: Oh I see, sorry I didn't understand the first time. I just pushed a new change that hopefully does what you expected. Let me know otherwise.

weshaggard:
LGTM

PR #666, opened 2016-04-25
Support OfficialBuildId with '-' 

chcosta:
Allows buildId in the form "[build number major]-[revision]".  Azure container names only permit letters (lower-case), numbers, and hyphens (non-consecutive).  This change allows us to specify an OfficialBuildId in an Azure container name conformant format.

<s>Also, minor fix to PublishContent.targets to set the token expiration to 1 day.  We aren't currently using these tokens anywhere (or even retrieving them), but a token expiration of 0 causes a warning in the build.</s>

/cc @weshaggard , @joperezr , @jhendrixMSFT , @maririos

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
@@ -25,9 +25,7 @@
     <CreateAzureContainer
       AccountKey="$(CloudDropAccessToken)"
       AccountName="$(CloudDropAccountName)"
-      ContainerName="$(ContainerName)"
-      WriteOnlyTokenDaysValid="1"
-      ReadOnlyTokenDaysValid="1" />
+      ContainerName="$(ContainerName)" />
     <!-- add relative blob path metadata -->
     <ItemGroup>
       <ForPublishing>
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
@@ -25,7 +25,9 @@
     <CreateAzureContainer
       AccountKey="$(CloudDropAccessToken)"
       AccountName="$(CloudDropAccountName)"
-      ContainerName="$(ContainerName)" />
+      ContainerName="$(ContainerName)"
+      WriteOnlyTokenDaysValid="1"
+      ReadOnlyTokenDaysValid="1" />
     <!-- add relative blob path metadata -->
     <ItemGroup>
       <ForPublishing>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
```diff
@@ -25,7 +25,9 @@
     <CreateAzureContainer
       AccountKey="$(CloudDropAccessToken)"
       AccountName="$(CloudDropAccountName)"
-      ContainerName="$(ContainerName)" />
+      ContainerName="$(ContainerName)"
+      WriteOnlyTokenDaysValid="1"
+      ReadOnlyTokenDaysValid="1" />
```
jhendrixMSFT: Is this causing problems in the build?  If not I would prefer to open an issue and get it properly fixed as this is a regression introduced in the refactoring of these tasks.

chcosta: No, it's not causing a problem in the build.  It's just an annoying warning.  I'll pull this out of my change if you prefer, or leave it and file an issue.  I should know better than to try to cleanup unrelated annoyances in my PR's :)

jhendrixMSFT: It's OK to leave it, just be sure to open an issue and assign it to @karajas.  Before commit 46da56 it was fine to specify 0 as a value, it meant that you didn't want a SAS token.

jhendrixMSFT: If you do leave this in, @karajas please be sure to back it out when you fix the handling of zero.

karajas: Sure I can make that fix.

chcosta: Sounds good.  It wasn't clear to me where the previous refactoring had occurred.  I'll remove this change from my PR since I don't want you to have to deal with coordinating our PR's or tracking down another place to make a change.

jhendrixMSFT:
LGTM

joperezr:
LGTM. After doing this change, we need to make sure to change all build definitions to start using the `-` instead of the `.`

chcosta:
Yes, the change is simple and involves changing the "Build number format" value to "$(date:yyyyMMdd)$(rev:-rr)"

joperezr:
yeah, I suppose you have tested that the `$(rev:-rr)` works as expected? If it doesn't for some reason, maybe you can get away with something like `$(date:yyyyMMdd)-$(rev:rr)`

chcosta:
I tested that it worked, then I hit the issue with our builds failing because they did not support parsing a build id with '-', which led to this PR.

PR #668, opened 2016-04-26
Create symbol packages that have a "symbols." id prefix

dagood:
Creates symbol packages that use a `symbols.` id prefix. This is in addition to the MyGet-convention symbol packages that we already make. The prefixing happens at the `NuGetPack` step, so there is still only a single `nuspec` used to create the library package and these two symbol packages.

To use this, set `PrefixedSymbolPackageOutputPath` to something (varies per repo) and set `PackPrefixedSymbolPackage` to `true`.

We'll want to turn off the MyGet-convention symbol packages at some point. I think this should happen after MyGet supports (to some degree) the prefixed-id symbol packages, so that until then, non-corpnet debuggers can use a working MyGet symbol server.

For example, the kind of structure this can have in CoreFX (all in `bin\packages\Debug`):

* `Microsoft.CSharp.4.0.1-rc3-24025-0.nupkg`
* `symbols\Microsoft.CSharp.4.0.1-rc3-24025-0.symbols.nupkg`
* `prefixedsymbols\symbols.Microsoft.CSharp.4.0.1-rc3-24025-0.symbols.nupkg`

/cc @noahfalk @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -999,15 +999,11 @@
                PrefixedSymbolPackageOutputDirectory="$(PrefixedSymbolPackageOutputPath)"
                AdditionalLibPackageExcludes="@(AdditionalLibPackageExcludes)"
                AdditionalSymbolPackageExcludes="@(AdditionalSymbolPackageExcludes)"
-               Condition="'$(_SkipCreatePackage)' != 'true'"/>
+               Condition="'$(_SkipCreatePackage)' != 'true'">
+      <Output TaskParameter="PackagesCreated" ItemName="OutputPackagePath" />
+    </NugetPack>
 
     <!-- Create markers that record the paths to the generated packages -->
-    <ItemGroup>
-      <OutputPackagePath Include="$(PackageOutputPath)$(Id).$(Version).nupkg" />
-      <OutputPackagePath Include="$(SymbolPackageOutputPath)$(Id).$(Version).symbols.nupkg" />
-      <OutputPackagePath Include="$(PrefixedSymbolPackageOutputPath)symbols.$(Id).$(Version).symbols.nupkg" />
-    </ItemGroup>
-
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -994,11 +995,20 @@
...
                AdditionalSymbolPackageExcludes="@(AdditionalSymbolPackageExcludes)"
                Condition="'$(_SkipCreatePackage)' != 'true'"/>
-    <!-- Create a marker that records the path to the generated package -->
-    <WriteLinesToFile Lines="$(PackageOutputPath)$(Id).$(Version).nupkg;$(SymbolPackageOutputPath)$(Id).$(Version).symbols.nupkg"
+
+    <!-- Create markers that record the paths to the generated packages -->
+    <ItemGroup>
+      <OutputPackagePath Include="$(PackageOutputPath)$(Id).$(Version).nupkg" />
+      <OutputPackagePath Include="$(SymbolPackageOutputPath)$(Id).$(Version).symbols.nupkg" />
+      <OutputPackagePath Include="$(PrefixedSymbolPackageOutputPath)symbols.$(Id).$(Version).symbols.nupkg" />
```
noahfalk: This last one should be conditional based on PackPrefixedSymbolPackage right? Perhaps even better, the NugetPack task could generate this item list for you rather than having the targets file do a parallel construction of the naming conventions. A typical place for bugs is when parallel computations of the same result get out of sync.

dagood: I agree, having an output would be better. The second `OutputPackagePath` is in the same boat, actually: sometimes symbol packages aren't created because there are no symbols to package. I didn't do it during the last PR because it doesn't hurt to list files here that don't exist, but getting rid of the duplication's a good point. I'll add this, thanks!

weshaggard:
Sorry I've not been following the symbols work closely but why are we changing this to a prefix instead of using the suffix?

dagood:
Closing: there was more discussion about the way we want to do this and we're continuing with conventional MyGet/NuGet symbol packages.

PR #669, opened 2016-04-26
Fixing issues in version.h that were causing file description changes.

joperezr:
When we added versioning to native components to coreclr repo, we broke the File Description of the assemblies when they had one defined. This changes will only force the versions to be changed, everything else will stay the same if it is already defined.

Once this is merged and consumed by coreclr, it will fix dotnet/coreclr#4367

cc: @weshaggard 

FYI: @AndyAyersMS @mikedn

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -131,32 +131,37 @@
       Condition="'$(NativeVersionHeaderFile)'!='' and '$(GenerateVersionHeader)'=='true'">
 
     <ItemGroup>
-      <!-- Some of these variables might be defined by macros, so undefine them first to avoid build warnings. -->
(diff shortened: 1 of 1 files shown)


weshaggard:
It isn't entirely clear to me how this fixes the issue in coreclr. Did we have actual projects with AssemblyName=build?

joperezr:
No, the reason why the name was "build" was because the version.h file is only generated once per build, and we do it by calling into the project `build.proj`. The project don't really have an assembly Name defined so it just uses the name of the proj file as the assembly name. For this reason, we need to respect the file description if is already defined, since individual native projects will define it separately.

joperezr:
What we were doing previous to this change, was to override the description that the native projects already defined. (This is an example)[https://github.com/dotnet/coreclr/blob/master/src/tools/crossgen/Native.rc#L5] of how native projects define their own file description names.

weshaggard:
OK so the only think you are force overriding now is the version then. 

LGTM

joperezr:
Exactly, only the version will be overriden now.

PR #672, opened 2016-04-26
The "run" tool - dev workflow

maririos:
We are working on a prototype to have a published contract of inputs and outputs to encapsulate the work we've been doing related to the dev workflow.

What I have right now is the config.json file with the general model of what we want to represent. 

@weshaggard @markwilkie @Chrisboh @ChadNedzlek @MattGal @chcosta @jhendrixMSFT


src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
+﻿{
```
joperezr: Why is there a config.json in here? shouldn't that be repo-specific? Or maybe this is just a sample one?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
+﻿using System;
```
joperezr: We'll probably need the Microsoft headers here.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+using System.IO;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+
+namespace Executor
```
joperezr: Should this be Microsoft.* instead? Also, why the name Executor? It seems to me that Parser fits more the purpose of this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+  "properties": {
+    "OSGroup": {
+      "explanation": "Target OS for result binaries.",
+      "type": "property",
+      "values": "Windows_NT, Linux, OSX, FreeBSD, NetBSD",
```
ChadNedzlek: Shouldn't "values" be an array? What if one of my "values" is supposed to have a comma in it? Seems like we'll get into a weird well of escaping otherwise.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+                    for (int i = 1; i < argsLength && !end; i++)
+                    {
+                        switch (args[i])
+                        {
+                            case "info":
+                                executor.usage();
+                                end = true;
+                                break;
+                            case "help":
+                                Console.WriteLine("Usage: build [command]:");
```
ChadNedzlek: I'm not sure I like the entry point being called "build". What if all I want to do is run tests? It's sort of weird to call "build run-tests"... Since the tool itself is very generic, it seems like it should be called something less "buildy" and more like "execute" or "run" or "do".

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+                    //default right now.
+                    string command = "simple";
+                    bool end = false;
+                    for (int i = 1; i < argsLength && !end; i++)
+                    {
+                        switch (args[i])
+                        {
+                            case "info":
+                                executor.usage();
+                                end = true;
```
ChadNedzlek: Why not just "return 0" instead of trying to track the Boolean here?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+            foreach(KeyValuePair<string,string> prop in propertiesToParse)
+            {
+                var value = (prop.Value.Equals(string.Empty)) ? findPropertyValue(prop.Key): prop.Value;
+                if (value!=null)
+                {
+                    commandProperties += string.Format(" {0}",formatProperty(prop.Key, value, findPropertyType(prop.Key), processName));
+                }
+                else
+                {
+                    Console.WriteLine("Error: The property {0} is not specified in the Json file.", prop.Key);
```
ChadNedzlek: Shouldn't this probably abort the command? This error is probably going to get swallowed up in the impending output from a full build or the like.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+            Console.WriteLine("Usage: executor [config file] [command]");
+            Console.WriteLine("  Config file:      json file that contains the properties and the commands to be run. (Required)");
+            Console.WriteLine("Command: [executor command]|[config command]");
+            Console.WriteLine("  Executor command:   Executor tool commands.");
+            Console.WriteLine("         info       Provides information about the executor tool.");
+            Console.WriteLine("         help       Provides information about the content of the config file. ");
+            Console.WriteLine("  Config command:   Config file specific commands.");
+        }
+                
+        public static void Main(string[] args)
```
ChadNedzlek: So is the goal that I can "set" some properties on the command line, and this thing will only allow me to set the ones specified for that command? I'm not entirely sure how the user executes the command they want in the mode they want.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+  "processes": {
+    "msbuild": {
+      "run": {
+        "windows": "msbuild.exe",
+        "non-windows": "..\\corerun.exe ..\\msbuild.exe"
+      },
+      "types": {
+        "property": "/p:{name}={value}",
+        "target": "/t:{name}",
+        "internal": "/{name}"
```
ChadNedzlek: this is more a "switch" than internal. internal things would have values too, like /v:diagnostic. We might want a fourth king of thing to differentiate switches / arguments with values.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
+﻿{
+  "properties": {
```
ChadNedzlek: Since some of these aren't actually properties, might want to pick a word like "options" or something, so it's not weird to find targets inside this collection

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "complete",
+      "property": {}
+    },
+    "all": {
+      "explanation": "Restores packages, builds product and test binaries, runs the tests and builds product packages.",
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "native, managed",
```
ChadNedzlek: What are "extraCommands"? I don't see them being used in the Exector anywhere, so I'm not sure what they are.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+        "BuildTests": "",
+        "BuildPackages": ""
+      }
+    },
+    "testAgainstPackages": {
+      "explanation": "Generates project.jsons for test projects, restores packages, builds product and then builds tests against the generated project.json files.",
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "",
+      "property": {
```
ChadNedzlek: Is this a list of properties I can optionally set? Properties I have to set? Property values that I can't override? Is there a difference between the three (or are only some supported)?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "default": ""
+    },
+    "BatchGenerateTestProjectJsons": {
+      "explanation": "MsBuild target that generates the project Jsons files to build tests against packages.",
+      "type": "target",
+      "values": "",
+      "default": ""
+    }
+  },
+  "commands": {
```
Chrisboh: Out of all of these is there a concept of which one is default?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+                }
+            }
+            else
+            {
+                Console.WriteLine("Error: Json configuration file should be provided.");
+            }
+        }
+    }
+
+    public class Setup
```
Chrisboh: Can we split public classes into their own files?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+using System.IO;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+
+namespace Executor
+{
+    public class Executor
+    {
+        public Setup openFile(string filePath)
```
Chrisboh: Nit, Capitalize the function names

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            string jsonFile = string.Empty;
+            if (File.Exists(filePath))
+            {
+                jsonFile = File.ReadAllText(filePath);
+                Setup jsonSetup = JsonConvert.DeserializeObject<Setup>(jsonFile);
```
Chrisboh: DeserializeObject will throw a JsonReaderException if the Json is invalid.  We should probably handle that and give a meaningful message to the user.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
Chrisboh: When considering a name we should probably not have a class as the same name as the namespace because when you have to write this, Executor.Executor.OpenFile(foo.txt), it feels weird

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+    "simple": {
+      "explanation": "Only builds managed product binaries. It doesn't restore packages.",
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "",
+      "property": {
```
Chrisboh: Can we rename this to properties as it can be multiple?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: As this PR is more to receive feedback than to actually merge the code, for convenience I added the config.json file here.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: @joperezr we can change the name, but in the last meeting we ended up with execute/executor

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: Oh, that's true.
I was thinking more that the entry point of each repo is build. Then build calls executor so when you do `build help` you would actually be calling `executor help`. That way the documentation should point to build.

ChadNedzlek: Well, that's the thing. There are other things to do in a repo other than "build". Like running tests, or syncing, updating the toolchain, cleaning, or any number of things. I think calling the entry point "build" limits what the normalization will be able to apply to (or lead to really strange command lines where you call "build orderMeAPizza", when it has nothing to do with building).

If I were looking for a way to run the tests in a repository without building (because I already built them), it would ever occur to me to ask the "build" tool how to do this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: Old habits form school. :confused: I'll change it

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: Yes. Right now I'm not handling the errors the way we would want them. I'm just printing. I'll change them.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: I need to implement that. 
The idea for the extraCommands is to put the commands that are part of the command you want to run. For example, when you do `executor all` you would be building the native and the managed libraries, so we will be calling the command `native` and the command `managed` that are already defined in the config.json file

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: As it is right now these are properties that are mandatory in order for the command to work.
If there is no value provided in the command or by the user, it would look for the default value previously defined.
The value should not be overwrite otherwise the command is not going to work.
Do we want more flexibility here?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: Right now I set it to `simple` but we need to talk to decide what the correct default command is.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+
+namespace Executor
+{
+    public class Executor
+    {
+        public Setup openFile(string filePath)
+        {
+            string jsonFile = string.Empty;
```
chcosta: You can move the declaration of jsonFile into the next scope.  ie, delete this line and add the declaration to "string jsonFile = File.ReadAllText(filePath);  In this way, you're not declaring an unused variable if filePath doesn't exist.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+            Console.WriteLine("  Executor command:   Executor tool commands.");
+            Console.WriteLine("         info       Provides information about the executor tool.");
+            Console.WriteLine("         help       Provides information about the content of the config file. ");
+            Console.WriteLine("  Config command:   Config file specific commands.");
+        }
+                
+        public static void Main(string[] args)
+        {
+            int argsLength = args.Length;
+            if (argsLength > 0)
```
chcosta: May want to flip this to "if (args.Length == 0) { do blah and exit }", so that you fail fast and a dev can immediately see what the result is rather than searching to the end of the if block.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
chcosta: Do you want to return an exit code?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+                    bool end = false;
+                    for (int i = 1; i < argsLength && !end; i++)
+                    {
+                        switch (args[i])
+                        {
+                            case "info":
+                                executor.usage();
+                                end = true;
+                                break;
+                            case "help":
```
chcosta: Unix tends to use "--help", you can enforce that convention, or trim leading '-' from args in case devs use that syntax.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
chcosta: I would prefer "repo", but there may be other wonderful suggestions.  I also don't like "build" but maybe I'm not fully understanding the use-case and I'm attributing my own vision to the tool.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+                                Console.WriteLine("Usage: build [command]:");
+                                ++i;
+                                if (i < argsLength)
+                                {
+                                    end = true;
+                                    if (args[i].Equals("properties"))
```
chcosta: I'd expect you'd want some sort of properties traversal hierarchy.  ie, "build help properties" might return something different than "build orderMeAPizza help properties"

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
chcosta: Perhaps have a base class and that instantiates an instance of a Windows help / build command class or Linux help /build command class / etc based on os, rather than passing an os string around and branching logic in various places.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+            string commandProperty = processes[processName].types[type];
+            commandProperty = commandProperty.Replace("{name}", property).Replace("{value}", value);
+            return commandProperty;
+        }
+
+    }
+
+    public class Command
+    {
+        public string explanation { get; set; }
```
chcosta: casing

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
jhendrixMSFT: Can we use some existing command line parsing code and/or library?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: I made the assumption that we wanted only one .cs file instead that a complete project but now that I think more about it we actually never stated that.
I'll make this changes too

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: Do we have that? that would be good!

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: I agree with the two of you that build is not the best word for all the cases we want to cover.
What I see from the pizza example is that we will maintain a build.cmd script that will build everything (by calling the executor) but that doesn't mean that build is the only entry point for the executor.
In that case we need to agree on a word :)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: @ChadNedzlek what I'm thinking is that the properties we set in the command are the mandatory ones in order for the command to succeed. If the developer wants to add more properties to customize the command he could pass them as extra arguments. 
Another idea is that we could have a group of mandatory and a group of optional/common properties per command that the user can modify. We could show that information with the `help` option to give more guideline on how to customize the command.

maririos: @chcosta yes! Thanks! I'll fix it

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "type": "target",
+      "values": "",
+      "default": ""
+    }
+  },
+  "commands": {
+    "simple": {
+      "explanation": "Only builds managed product binaries. It doesn't restore packages.",
+      "processName": "msbuild",
+      "project": "build.proj",
```
weshaggard: Isn't project too specific?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "extraCommands": "",
```
weshaggard: What is extraCommands?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
weshaggard: I'm still not sure I understand the plan for this yet. This allows you to run chain commands together?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
weshaggard: We may still want to think about the name a little more, but regardless we should put it in a Microsoft namespace.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/project.json
```diff
@@ -0,0 +1,26 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.TestHost": "1.0.0-rc2-23923",
```
weshaggard: Why are you using these packages as opposed to making this a standard CLI application and using Microsoft.NETCore.App and dotnet.exe as the host?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+            
+            return completeCommand;
+        }
+
+        public string buildPropertiesForCommand(Dictionary<string, string> propertiesToParse, string processName)
+        {
+            string commandProperties = string.Empty;
+            foreach(KeyValuePair<string,string> prop in propertiesToParse)
+            {
+                var value = (prop.Value.Equals(string.Empty)) ? findPropertyValue(prop.Key): prop.Value;
```
weshaggard: string.IsNullOrEmpty instead of comparing to string.Empty.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
weshaggard: We should probably make the default the first command in the list.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "project": "build.proj",
+      "extraCommands": "native, managed",
+      "property": {}
+    }
+  },
+  "processes": {
+    "msbuild": {
+      "run": {
+        "windows": "msbuild.exe",
+        "non-windows": "..\\corerun.exe ..\\msbuild.exe"
```
weshaggard: I'm not sure if we want to be adding the logic to figure out how to fork these commands in the config file itself. I think we should just call the same thing in all cases and if there is a need to split we should have a wrapper script or something we call instead.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "native, managed",
+      "property": {}
+    }
+  },
+  "processes": {
+    "msbuild": {
+      "run": {
+        "windows": "msbuild.exe",
```
weshaggard: What is the assumption for where to find these tools? We really want to make sure we don't require them to be on the path. Should it be something more like `%ToolPath%\msbuild.exe` (assuming we restore msbuild.exe under the tool runtime.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "",
+      "property": {
+        "RestoreDuringBuild": false,
+        "BuildTests": false,
+        "BuildPackages": false
+      }
+    },
+    "complete": {
```
weshaggard: What is the difference between all and complete?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "processName": "cmake",
+      "project": "",
+      "extraCommands": "",
+      "property": {}
+    },
+    "managed": {
+      "explanation": "Only builds managed product binaries. It doesn't restore packages.",
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "complete",
```
weshaggard: Why would the managed one chain in complete?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "extraCommands": "",
+      "property": {
+        "RestoreDuringBuild": false,
+        "BuildTests": false,
+        "BuildPackages": false
+      }
+    },
+    "complete": {
+      "explanation": "Restores packages, builds product and test binaries, runs the tests and builds product packages",
+      "processName": "msbuild",
```
weshaggard: If we are trying to mimic Process.Start we would have a process name and a arg list. I know you are building the arg list from the properties but we should probably enable loss args here as well so we could pass logging or other things as example.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "project": "build.proj",
```
weshaggard: For the project we need to enable people to be able to pass other projects to these common command so that people can work on individual projects as well.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "extraCommands": "",
+      "property": {
```
weshaggard: What about the other types? What if one of these is a target for example?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "property": {}
+    },
+    "managed": {
+      "explanation": "Only builds managed product binaries. It doesn't restore packages.",
+      "processName": "msbuild",
+      "project": "build.proj",
+      "extraCommands": "complete",
+      "property": {}
+    },
+    "all": {
```
weshaggard: Are you going to add the other workflow commands like sync, test, and perhaps clean (assuming we can do that one here).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            int argsLength = args.Length;
+            if (argsLength > 0)
+            {
+                Executor executor = new Executor();
+                Setup jsonSetup = executor.openFile(args[0]);
+                if (jsonSetup != null)
+                {
+                    //will be passed by executor.cmd/sh? we only need to now if it is in windows or non-windows.
+                    string os = "windows";
```
weshaggard: I'm hopeful we don't need this but if we do end up needing it you should call our RuntimeInformation class to get it. See https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/GetTargetMachineInfo.cs#L40 as an example.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
weshaggard: Yes I think build.cmd should of course remain and at least do something useful, but we can consider other options for the common entry point.

As some food for thought though today when working with msbuild as your build engine you do end up calling `msbuild /t:Test` to do your testing so using build.cmd as the entry point isn't all too much different from that. We should keep in mind that we aren't trying make this so generic that it can run everything it is just supposed to be a small abstraction around our primary build tools like msbuild and cmake for the most part. I don't think we really have a goal for example to have it ever call xunit directly for us but instead use our engine to setup the test layout and create the command and let the build engine call it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
weshaggard: For some other tools in this repo we use https://github.com/dotnet/buildtools/blob/master/src/common/CommandLine.cs

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
weshaggard: Yes please lets try to follow standard .NET C# coding guidelines throughout :)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+                                        jsonSetup.ProvideHelpInstructions(os);
+                                    }
+                                }
+                                if (!end)
+                                {
+                                    jsonSetup.help(os);
+                                    end = true;
+                                }
+                                break;
+                            default:
```
weshaggard: What is the format for passing the properties through to the executor? I don't see you parsing any of the other arguments that might come from the user yet.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
```diff
@@ -0,0 +1,227 @@
...
+        public Dictionary<string, Command> commands { get; set; }
+        public Dictionary<string, Process> processes { get; set; }
+
+        public void ProvideHelpInstructions(string os)
+        {
+            Console.WriteLine("========Commands========");
+
+            foreach (KeyValuePair<string, Command> instr in commands)
+            {
+                Console.WriteLine(string.Format("* {0} - {1}", instr.Key, instr.Value.explanation));
```
weshaggard: string.format is unnecessary, as you can pass a format string directly to Console.WriteLine.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
```diff
@@ -0,0 +1,126 @@
...
+      "extraCommands": "native, managed",
+      "property": {}
+    }
+  },
+  "processes": {
+    "msbuild": {
+      "run": {
+        "windows": "msbuild.exe",
+        "non-windows": "..\\corerun.exe ..\\msbuild.exe"
+      },
```
weshaggard: As an FYI there will be cases where we will likely want to run under corerun (or dotnet) one windows as well.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: It is, but buid.proj is not always the entry point. As you said in another comment, we could change the `project` value for individual projects if desired (is not yet implemented though...)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: all will run the commands: native and managed.
managed runs the command complete -> This is redundant so I'll just renamed complete for managed.
complete is the the equivalent as running build.cmd today.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: This is the list of properties for the command to run.
We can add targets here too. For example the command `TestAgainstPackages` has the target `BatchGenerateTestProjectJsons` as part of the property list.

Maybe property/properties is not the best name. We could change it to options as Chad suggested.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: Yes, I was thinking in adding build-packages, publish-packages, build-test and run-tests.
For sync and clean, I'm thinking on adding to commands the calls to msbuild that we have in them, like sync /p and clean /p.

We still want to maintain sync.cmd and clean.cmd with the other options, right?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/config.json
maririos: Yes.
We should also be able to pass to the executor an array of commands we want to execute (not implemented yet), right?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/executor/executor.cs
maririos: Not implemented yet

maririos:
I updated the PR with the changes we talked about on Friday (I delete the extra command options/code) and a new nomenclature that we discuss with @markwilkie today.
The changes are:

- Address feedback related to formatting and name structure to the config.json file
- Address feedback related to logic and code structure to the Executor.cs class
- Identify how the options under a command are defined (required vs optional)
- Create a project instead of using just one class
- Command line parsing tool instead of doing it manually.
- The last argument would be the commands the user wants to pass. 
- Add ability to change the project where we want to run the command
- Add Microsoft headers and change namespace for Microsoft.*
- Follow standard .NET C# coding guidelines

For now, if you want to give it a try go to the executor folder and do:
`..\..\..\..\Tools\dotnetcli\dotnet.exe restore`
`..\..\..\..\Tools\dotnetcli\dotnet.exe run -- -?`
`..\..\..\..\Tools\dotnetcli\dotnet.exe run -- -default`
`..\..\..\..\Tools\dotnetcli\dotnet.exe run -- -default -osgroup=Linux`
`..\..\..\..\Tools\dotnetcli\dotnet.exe run -- -default ExtraArguments /p:SkipTests=true`
...

cc: @weshaggard @markwilkie @Chrisboh @ChadNedzlek @MattGal @chcosta @jhendrixMSFT

maririos:
I updated the PR with the latest changes which include:

- Executor now actually executes the commands (not just printing) 
- The application is now a portable CLI application 
- The following dev workflow steps have been added as commands: sync /p, sync /ab, clean /p, clean /b, clean /u, build managed, build-tests, build-packages (only works in windows) and publish-packages.

Issues with the dev workflow:
- Build native right now does nothing as the executor only runs one process per command and the way we build native today is by calling msbuild and then calling build-native.cmd in windows and cmake in non-windows.
- build-packages in non-windows involves an extra step to get the RuntimeOS where we are running. 
We would need to make a change for two issues in order to integrate the steps to the executor tool, right?

If you want to try it:
```
<path>dotnet.exe restore
<path>dotnet.exe run -cleanbinariescommand -project:<repoRoot>\build.proj
<path>dotnet.exe run -cleanpackagescommand -project:<repoRoot>\build.proj
<path>dotnet.exe run -sync -project:<repoRoot>\build.proj
<path>dotnet.exe run -default -project:<repoRoot>\build.proj
<path>dotnet.exe run -build-tests
...
```
If you run the executor in the folder where `build.proj` is then `-project:<repoRoot>\build.proj` is not needed.
Also, there are more commands available to use. For more information you could type 
'<path>dotnet.exe run -?`

DotnetCLIVersion.txt
```diff
@@ -1 +1 @@
-1.0.0-beta-002173
\ No newline at end of file
+1.0.0-preview1-002700
```
maririos: Once @joperezr change is in (PR: #706) we don't need this.

maririos:
The executor is almost there! 
I updated the PR so that Buildtools could use the executor and I added sync, clean and build scripts using the executor (not 1 line, but kind of close).

Please take a look and if you want to try it you would need to pull, build, delete the tools and packages directory and then restore the BT package that you just built.

There is still work to do related to how we are showing and asking for help, but according to feedback we could start preparing a PR to merge the changes and then update the build tools package version in the build tools repo.

cc: @weshaggard @markwilkie @Chrisboh @ChadNedzlek @MattGal @chcosta @jhendrixMSFT

joperezr:
IMO it would me better if we keep this PR just for adding the executor to buildtools packagr, and then move all of the commits that actually onboard it to buildtools repo into a different PR. This way, it would be easy to point people to a specific PR if they want to know the required changes to onboard executor in a different repo. 

If the reason you have those changes in here is only as a preview of how it will look then I'm totally fine with that.

sync.cmd
```diff
@@ -0,0 +1,25 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+set _args=%*
+if /I [%1]==[/?] goto Usage
+if [%1]==[]  set _args=-p
+
+echo Running executor.cmd %_args%
+call executor.cmd %_args%
```
weshaggard: How does it know it is supposed to run sync? I would have expected something like `executor.cmd sync %*`

sync.cmd
```diff
@@ -0,0 +1,25 @@
...
+if NOT [%ERRORLEVEL%]==[0] (
+  echo ERROR: An error occurred. There may have been networking problems so please try again in a few minutes.
+  exit /b 1
+)
+
+echo Done Syncing.
+exit /b 0
+
+:Usage
```
weshaggard: Given there is only one option and it is also the default option why do we need Usage?

sync.cmd
```diff
@@ -0,0 +1,25 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+set _args=%*
+if /I [%1]==[/?] goto Usage
+if [%1]==[]  set _args=-p
+
+echo Running executor.cmd %_args%
+call executor.cmd %_args%
+if NOT [%ERRORLEVEL%]==[0] (
```
weshaggard: I think I could live one a little error handling (so not exactly 1 line) but I think we can get this entire script down to just 
```
call executor.cmd sync %*
exit /b %ERRORLEVEL%
```

sync.cmd
maririos: I'll delete it.

sync.cmd
maririos: By default it is passing -p that is the same thing as sync, I'll change it to `executor.cmd sync %*` as it is a little confusing.

In the corefx repo sync has 3 options: 
-p (restore packages)
-t (creates project.json and restore those packages for testing against packages)
-ab (downloads the product packages stored in Azure)

using the executor will look like:
```
executor.cmd sync                                  == executor -p
executor.cmd testAgainstPackages        == executor.cmd -t
executor.cmd downloadAzurePackages == executor.cmd -ab
```
@Chrisboh mentioned that it would be better if we have something more like 
```
executor.cmd sync -p
executor.cmd sync -ab
```
This last part would involve more changes to the executor but it may help to keep things more organized.
Is there a preference?

weshaggard: I've still not had the chance to look through all your changes but I do prefer the later (i.e. different options to `executor.cmd sync` as opposed to different executor commands.

maririos: ok. I'll start working on that change

maririos:
PR updated with latest changes to the executor. To give it a try go to PR maririos/buildtools#1

config.json
```diff
@@ -0,0 +1,153 @@
...
+      "toolName": "msbuild",
+      "lockedSettings": {
+        "RestoreDuringBuild": false,
+      },
+      "settings": {
+        "MsBuildLogging":"default",
+        "MsBuildParameters":"default"
+      }
+    },
+    "sync-p": {
```
weshaggard: Why isn't this just "sync" and the "-p" as a argument? Similar question for the others.

src/Executor/CommandLine.cs
```diff
@@ -0,0 +1,1855 @@
...
+using System;
+using System.Diagnostics;
+using System.Text;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.DotNet.Execute;
+
+using System.Reflection;
+
+namespace Microsoft.Fx.CommandLine
```
weshaggard: Any clean way to see a diff of what you changed in this file beyond pulling down your change and diffing locally.

maririos: I started changing the file with commit `Improve help usage`. If you go to that commit you could differ the files.It has most of the changes.
The next commits are small fixes or typos.

src/Executor/Executor.cs
```diff
@@ -0,0 +1,163 @@
...
+        public Dictionary<string, string> SettingParameters { get; set; }
+        public Dictionary<string, string> CommandParameters { get; set; }
+        
+        public Executor()
+        {
+            SettingParameters = new Dictionary<string, string>();
+            CommandParameters = new Dictionary<string, string>();
+            DevWorkflowCommands = new Dictionary<string, List<string>>();
+            string executorDirectory = Path.GetDirectoryName(typeof(Executor).GetTypeInfo().Assembly.Location);
+            configFilePath = Path.GetFullPath(Path.Combine(executorDirectory, @"..\"));
```
weshaggard: Bad idea to assume a location for this config.json file. I'm OK if it is a default bet we need a way to specify it.

config.json
maririos: Because of the structure of the config.json file. But it is not mandatory so if it is confusing we could change it.

src/Executor/Run.cs
```diff
@@ -0,0 +1,62 @@
...
+                Console.WriteLine("Error in the command: {0}. => {1}", string.Format("{0} {1}", filename, args), e.Message);
+                return 1;
+            }
+        }
+
+        private static void ReadOutputHandler(object sendingProcess,
+            DataReceivedEventArgs outLine)
+        {
+            // Collect the command output.
+            if (!String.IsNullOrEmpty(outLine.Data))
```
ChadNedzlek: Why are we throwing out blank lines?

src/Executor/Run.cs
```diff
@@ -0,0 +1,62 @@
...
+                _process = new System.Diagnostics.Process();
+                _process.StartInfo = psi;
+
+                // Set our event handler to asynchronously read the output.
+                _process.OutputDataReceived += new DataReceivedEventHandler(ReadOutputHandler);
+
+                _process.Start();
+                _process.BeginOutputReadLine();
+
+                _process.WaitForExit();
```
ChadNedzlek: I feel like I've seen some race conditions with output and the exit event.  Sometimes the exit event will happen while there is still pending output. If this process exits at that point, the "end" of the output will get lost.
This also doesn't redirect the Error stream, (because it's a totally different, needing RedirectStandardError and ErrorDataRecieved).

Also, what about input?

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+        public Dictionary<string, Command> Commands { get; set; }
+        public Dictionary<string, Tool> Tools { get; set; }
+        public Dictionary<string, string> SettingParameters { get; set; }
+        public string Os { get; set; }
+        public string ConfigurationFilePath { get; set; }
+
+
+        private string FindSettingValue(string valueToFind)
+        {
+            if (Settings.ContainsKey(valueToFind))
```
ChadNedzlek: This double access is better handled by Settings.TryGetValue(valueToFind, out foundValue)

src/Executor/Run.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace Microsoft.DotNet.Execute
+{
+    public class Run
```
ChadNedzlek: It's sort of odd that this is an instance class. The class doesn't store any meanginful state, and has only one method.  Could it be static, so that you didn't have to say "new Run()", and just have Run.ExecuteProcess?

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+                    Console.WriteLine("Build Failed.");
+                }
+                Console.ResetColor();
+
+                return result;
+            }
+            return 1;
+        }
+
+        public string[] BuildCommand(string commandSelectedByUser, Dictionary<string, string> parameters = null)
```
ChadNedzlek: Magic arrays (where each index has special meaning) are difficult to understand when reading code, since the "members" aren't self-documenting. You could actually return a "ProcessStartInfo" from this method. :-)

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+
+            Command commandToExecute = Commands[commandSelectedByUser];
+            string toolName = GetTool(commandToExecute, Os, ConfigurationFilePath);
+            if (string.IsNullOrEmpty(toolName))
+            {
+                Console.WriteLine("Error: The process {0} is not specified in the Json file.", commandToExecute.ToolName);
+                return null;
+            }
+
+            string commandParameters = string.Empty;
```
ChadNedzlek: This value isn't actually used in this code block, so it doesn't need to be initialized, or even declared at this scope.

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+                if (string.IsNullOrEmpty(commandValues[optSetting.Key]))
+                {
+                    string value = string.IsNullOrEmpty(optSetting.Value) || optSetting.Value.Equals("default") ? FindSettingValue(optSetting.Key) : optSetting.Value;
+                    if (value != null)
+                    {
+                        commandValues[optSetting.Key] = value;
+                    }
+                    else
+                    {
+                        return false;
```
ChadNedzlek: This one doesn't Console.WriteLine anything... if this fails, does the process just silently do nothing?

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+                string value = string.IsNullOrEmpty(reqSetting.Value) || reqSetting.Value.Equals("default") ? FindSettingValue(reqSetting.Key) : reqSetting.Value;
+                if (value != null && (string.IsNullOrEmpty(commandValues[reqSetting.Key]) || reqSetting.Key.Equals("Project")))
+                {
+                    commandValues[reqSetting.Key] = string.IsNullOrEmpty(value) ? "True" : value;
+                }
+                else
+                {
+                    if (!string.IsNullOrEmpty(value) && !value.Equals(commandValues[reqSetting.Key]))
+                    {
+                        Console.WriteLine("Error: The value for setting {0} can't be overwriten.", reqSetting.Key);
```
ChadNedzlek: Errors should probably go to Console.Error.WriteLine

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+            string[] extraA = extraArguments.Split(' ');
+            foreach(string param in extraA)
+            {
+                colonPosition = 0;
+                equalPosition = param.Length;
+                tempParam = string.Empty;
+
+                colonPosition = param.IndexOf(":");
+                equalPosition = param.IndexOf("=");
+                if(colonPosition!=0)
```
ChadNedzlek: The string "a:b=c" is going to get parsed really weirdly by this command, as the "a:b" key with "c" as the value. param.Split(new []{':', '='}, 2) should get the first of either of them, and even do the substringing for you. :-)

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+            if (type.Equals("passThrough"))
+            {
+                commandOption = string.Format(" {0}", toolName.Equals("console") ? "" : value);
+            }
+            else
+            {
+                if (Tools.ContainsKey(toolName) && !string.IsNullOrEmpty(type))
+                {
+                    commandOption = Tools[toolName].ValueTypes[type];
+                    commandOption = commandOption.Replace("{name}", option).Replace("{value}", value);
```
ChadNedzlek: Is there any worry that a value or name might literally contain the string "{name}" or "{value}"? Do we need some sort of escaping mechanism here? Like replace "{{" with "{"?

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+            }
+            return string.Empty;
+        }
+            
+        public string FormatSetting(string option, string value, string type, string toolName)
+        {
+            string commandOption = null;
+            if (type.Equals("passThrough"))
+            {
+                commandOption = string.Format(" {0}", toolName.Equals("console") ? "" : value);
```
ChadNedzlek: So the console command ignores any "passThrough" arguments? That's unexpected (at least to me)

src/Executor/Setup.cs
```diff
@@ -0,0 +1,290 @@
...
+                {
+                    value = lockedSetting.Value.Equals("default") ? FindSettingValue(lockedSetting.Key) : lockedSetting.Value;
+                    sb.Append(string.Format("    {0} ({1})= {2}", lockedSetting.Key, FindSettingType(lockedSetting.Key), value)).AppendLine();
+                    commandParametersToPrint[lockedSetting.Key] = value;
+                }
+
+                sb.AppendLine().Append("  Other Settings (values can be overwritten): ").AppendLine();
+                foreach (KeyValuePair<string, string> optionalSettings in commandToPrint.Settings)
+                {
+                    value = optionalSettings.Value.Equals("default") ? FindSettingValue(optionalSettings.Key) : optionalSettings.Value;
```
ChadNedzlek: What happens if we need a setting value to literally have the value "default", will it get sort of unexpected replaced by something else? Or cause an error? Magic strings tend to lead to unexpected results. Is there some other way we can indicate "this things value is specified elsewhere" other than having one specific (and now unrepresentable) string?  Maybe having it be null or something?

src/Executor/Setup.cs
maririos: Thanks for pointing this out! I was missing a validation, hence no error message.
I'll change it

src/Executor/Setup.cs
maririos: I don't think it is going to be an issue as it is

src/Executor/Setup.cs
maririos: I had it as null before but after talking to Mark it was not clear that it was actually referring to the default value.
I could change it to `{default}`. That is unlikely to happen.

src/Executor/Setup.cs
maririos: I'll create a structure for this. I don't like using ProccessStartInfo as BuildCommand is also called by the help function. It would look weird using a Process data without executing the process.

src/Executor/Run.cs
maririos: I'll change it

src/Executor/Run.cs
maririos: > I feel like I've seen some race conditions with output and the exit event. Sometimes the exit event will happen while there is still pending output. If this process exits at that point, the "end" of the output will get lost.

I'll take a look and as we talked look at Async functions

> This also doesn't redirect the Error stream, (because it's a totally different, needing RedirectStandardError and ErrorDataRecieved).

Msbuild redirects everything to the input one so I didn't use the Error stream, but as you pointed out when we talked we need the Error stream too for other processes different than msbuild.I'll change it.

>Also, what about input?

I consider that this scenario doesn't apply for the executor.

src/Executor/Setup.cs
maririos: True, but do we have parameters like that? I looked at the documentation for msbuild and cmake to make sure I'm covering those scenarios.
I found that this was not working for `foo=bar` that is a possible parameter for cmake so I fixed it.

src/Executor/Executor.cs
maririos: Ok. So by default it would like for the file there, otherwise the file path should be the first parameter sent to the executor.
I updated the documentation to show that

config.json
```diff
@@ -0,0 +1,153 @@
+﻿{
+  "settings": {
+    "OSGroup": {
+      "description": "Target OS for result binaries.",
```
chcosta: "target OS" and "target os group" are not the same.  I'd be explicit and add "group" to the description. For that matter,  it might be nice to link to a description of the osgroup property.  I'm not certain if we have an MD for for this or not though.

config.json
```diff
@@ -0,0 +1,153 @@
...
+      "toolName": "msbuild",
+      "lockedSettings": {
+        "RestoreDuringBuild":  true,
+        "BatchRestorePackages": "default"
+      },
+      "settings": {
+        "MsBuildParameters":"default"
+      }
+    },
+    "clean-b": {
```
chcosta: To me,  there are too many of these commands.  I'd prefer just a clean command and then -b,  -p,  etc options.  I'm pretty sure I'm always going to type clean as step 1

src/Executor/CommandLine.cs
```diff
@@ -0,0 +1,1857 @@
...
+	            CommandLineParser.ParseForConsoleApplication(args, delegate(CommandLineParser parser)
+	            {
+	                parser.DefineOptionalQualifier("Timeout", ref Timeout, "Timeout in milliseconds to wait for each reply.");
+	                parser.DefineOptionalQualifier("Forever", ref Forever, "Ping forever.");
+	                parser.DefineDefaultParameterSet("Ping sends a network request to a host to reply to the message (to check for liveness).");
+	                parser.DefineParameter("Host", ref Host, "The Host to send a ping message to.");
+	            });
+
+	            // Step 3, use the parameters
+	            Console.WriteLine("Got {0} {1} {2} {3}", Host, Timeout, Forever);
```
chcosta: There are 4 placeholders for 3 parameters here.  Also,  this output is only useful if you have the source code sitting next to you so you can see what the values represent.

src/Executor/CommandLine.cs
```diff
@@ -0,0 +1,1857 @@
...
+        /// new paragraphs.</param>
+        /// </summary>
+        public void DefineOptionalQualifier<T>(string name, ref T retVal, string helpText, string defaultValue, List<string> legalValues)
+        {
+            object obj = DefineQualifier(name, typeof(T), retVal, helpText, false, defaultValue, legalValues);
+            if (obj != null)
+                retVal = (T)obj;
+        }
+        /// <summary>
+        /// Like code:DeclareOptionalQualifier except it is an error if this parameter is not on the command line.
```
chcosta: Do you really need a separate method for this?

src/Executor/CommandLine.cs
```diff
@@ -0,0 +1,1857 @@
...
+        /// <param name="name">The name of the qualifer.</param>
+        /// <param name="retVal">The place to put the parsed value</param>
+        /// <param name="helpText">Text to print for this qualifer.  It will be word-wrapped.  Newlines indicate
+        /// new paragraphs.</param>
+        /// </summary>
+        public void DefineQualifier<T>(string name, ref T retVal, string helpText, string defaultValue, List<string> legalValues)
+        {
+            object obj = DefineQualifier(name, typeof(T), retVal, helpText, true, defaultValue, legalValues);
+            if (obj != null)
```
chcosta: Missing {} around if block

src/Executor/CommandLine.cs
```diff
@@ -0,0 +1,1857 @@
...
+                retVal = (T)obj;
+        }
+        /// <summary>
+        /// Specify additional aliases for an qualifier.  This call must come BEFORE the call to
+        /// Define*Qualifier, since the definition needs to know about its aliases to do its job.
+        /// </summary>
+        public void DefineAliases(string officalName, params string[] alaises)
```
chcosta: Typo in parameter name "alaises"

config.json
maririos: Thanks! I changed it. 
I'm not sure if we have documentation about it. I'm basically taking the values from the dir.props file

config.json
maririos: After talking to @weshaggard  I'm changing the config.json to have more the behavior that you are describing.
I'll put some default options so that you can type `clean`

src/Executor/CommandLine.cs
maririos: This is the documentation of the CommandLine class. It has some errors but for now I'm not changing that.

src/Executor/CommandLine.cs
maririos: For now I'm just modifying the methods that we care about on this class.
Maybe later we can do a clean up

maririos:
PR updated with latest feedback. It only contains the changes related to the executor and the config.json file. To use it and see the changes related to the dev workflow scripts, please go to PR maririos/buildtools#1

- The structure of the config.json file changed to have commands and options/alias specific per command. Also, a command is tool dependent I.E. build-managed and build-native are two different commands as one calls msbuild and the other calls a cmd/sh file.

- Location of the config.json file can be provided by the user. If it is not provided by default it is going to look for it at the root of the repo.

- Add logic to how we run a process.

- Fix bugs.

cc: @weshaggard @markwilkie @Chrisboh @ChadNedzlek @MattGal @chcosta @jhendrixMSFT

config.json
```diff
@@ -0,0 +1,232 @@
...
+        "toolName": "msbuild",
+        "settings": {
+          "MsBuildParameters":"default",
+          "MsBuildLogging":"default"
+        }
+      }
+    },
+    "build-managed":{
+      "alias":{
+        "s":{
```
weshaggard: Why is "s" the alias for the case that doesn't restore packages?

config.json
```diff
@@ -0,0 +1,232 @@
...
+        "verbose":{
+          "description": "Passes /flp:v=normal to the msbuild command or the value passed by the user.",
+          "settings":{
+            "MsBuildLogging": "default"
+          }
+        },
+        "os":{
+          "description": "Sets OSGroup=AnyOS or the value passed by the user.",
+          "settings":{
+            "OSGroup": "default"
```
weshaggard: Is default some keyword that indicates take the default value but also allows you to pass set it?

config.json
```diff
@@ -0,0 +1,232 @@
...
+    },
+    "BuildTests": {
+      "description": "Enables/Disables building tests.",
+      "valueType": "property",
+      "values": ["True", "False"],
+      "defaultValue": true
+    },
+    "MsBuildLogging": {
+      "description": "MsBuild logging options.",
+      "valueType": "passThrough",
```
weshaggard: What does passThrough indicate?

config.json
```diff
@@ -0,0 +1,232 @@
...
+      "defaultValues":{
+        "toolName": "msbuild",
+        "settings": {
+          "MsBuildParameters":"default"
+        }
+      }
+    },
+    "build-native":{
+      "alias":{
+        "debug":{
```
weshaggard: probably need a release for the native builds as well.

config.json
maririos: Hmm I was thinking more of a simple build where you only build the product.
I can see how this is confusing as it doesn't say anything :|

config.json
maririos: Yes, it is a keyword. But if you want to overwrite it, you could do that too.

config.json
maririos: I'll add it

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.cmd
```diff
@@ -0,0 +1,8 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+echo msbuild.exe %*
```
weshaggard: At some point I want to make this configurable so we could chose to use the .NET Core msbuild as well.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.cmd
```diff
@@ -0,0 +1,8 @@
+@if "%_echo%" neq "on" echo off
```
weshaggard: Looks like this cmd file can be condensed some more. I don't think we need the setlocal or the if ERRORLEVEL check just exit with ERRORLEVEL

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+working_tree_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+echo $working_tree_root/corerun $working_tree_root/MSBuild.exe $*
+$working_tree_root/corerun $working_tree_root/MSBuild.exe $*
+if [ $? -ne 0 ]; then
```
weshaggard: exit $?

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -27,7 +27,8 @@
     <file src="Microsoft.DotNet.Build.Tasks\PackageFiles\**\*.*" target="lib" />
     <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DotNet.Build.Tasks.dll" target="lib" />
     <file src="Microsoft.DotNet.Build.Tasks\NuProj\**\*.*" target="lib\NuProj" />
-    <file src="BclRewriter\BclRewriter.exe" target="lib" />
+    <file src="Executor\Executor.exe" target="lib" />
+	<file src="BclRewriter\BclRewriter.exe" target="lib" />
```
weshaggard: nit: indention.

config.json
maririos: That the value that you have is not going to be processed or modified. It would print it as it is.

weshaggard:
Looking better. I think we have all the initial knobs we need to accomplish our goals. I would still like to rename executor because I don't really want that to show up in all our repos. I'm thinking something like `run` is probably good enough.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.cmd
maririos: I'll change it

maririos:
PR updated with latest feedback:

- Changing the name of the tool from Executor to Run

- Deleting the option `extraArguments`. Now all the parameters that are passed and are not part of the config file will be passed to the running process. Those extra parameters should be the last thing that we pass to run.exe

- Fix the order on how `-?` works so for example we can do `run.cmd build-managed %* `

config.json
```diff
@@ -46,7 +46,7 @@
       "description": "MsBuild logging options.",
       "valueType": "passThrough",
       "values": [],
-      "defaultValue": "/flp:v=normal"
+      "defaultValue": "/flp:v=diag"
```
weshaggard: I think the default value should be v=normal but the verbose option should be v=diag.

config.json
```diff
@@ -107,6 +107,12 @@
             "BatchRestorePackages": "default"
           }
         },
+        "verbose":{
+          "description": "Passes /flp:v=normal to the msbuild command or the value passed by the user.",
```
weshaggard: This comment doesn't match what you set it to be above.

src/Executor/CommandLine.cs
```diff
@@ -1127,13 +1128,21 @@ private void ParseWordsIntoQualifiers()
...
+                        if(paramSetEncountered)
+                        {
+                            _helpRequestedFor = parameterSet;
+                        }
+                        else
+                        {
+                            _helpRequestedFor = String.Empty;
+                        }
+                        
+                        /*if (i + 1 < _args.Count)
```
weshaggard: Lets delete commented out code.

src/Run/Run.csproj
```diff
@@ -1,29 +1,25 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
-
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <RootNamespace>Executor</RootNamespace>
-    <AssemblyName>Executor</AssemblyName>
+    <AssemblyName>Run</AssemblyName>
```
weshaggard: To be more cross-plat friendly I think this and the wrapper scripts should probably be lowercase.

config.json
```diff
@@ -118,7 +118,7 @@
     },
     "build-managed":{
       "alias":{
-        "s":{
+        "binaries":{
```
weshaggard: Should this just be the default "" alias for this command?

src/Run/CommandLine.cs
```diff
@@ -0,0 +1,1876 @@
...
+using System;
+using System.Diagnostics;
+using System.Text;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.DotNet.Execute;
+
+using System.Reflection;
+
+namespace Microsoft.Fx.CommandLine
```
weshaggard: Any ideas of how much of this code you are actually using? I don't like the fact that it is forked and would prefer that we make changes to the core version of this file in a compatible way, but if that isn't realistic then I would prefer we prune any dead code in your fork given there is no plans to try and merge it back.

config.json
maririos: what do you mean?

src/Run/Run.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <RootNamespace>Executor</RootNamespace>
+    <AssemblyName>Run</AssemblyName>
+    <NuGetRuntimeIdentifier>win7-x64</NuGetRuntimeIdentifier>
```
weshaggard: One thing that just jumped at me is the fact this is x64. That might cause some problems when you try to get this enabled in corefx and we mirror those into TFS.

config.json
weshaggard: Nevermind I think what you have is good for now. I now see the default empty settings are under "defaultValues", and changing those to this would be a break so at least for now we should keep what you have.

src/Run/CommandLine.cs
maririos: I'm using a great part of it. The methods I'm not using is because I don't have positional parameters or I'm not working with enums.
I also though about making the changes in the core version but there are changes that are too specific for this context. 
We could prune it to make it cleaner

src/Run/Run.csproj
maririos: Interesting. I didn't think about it. I copied the value from the other projects that we have in build tools like GenAPI and GenFacades.
Other option is to set it to `any`. Would that help to make it more compatible?

weshaggard: setting it to any isn't likely going to help much, as this doesn't really have any native dependencies. The bigger issue is getting a version of the runtime that this runs on to be either x86 or x64. We should just ensure that when you change corefx you don't break the TFS builds, you can test them manually before merging into corefx.

src/Run/CommandLine.cs
weshaggard: Don't worry about pruning for now but I would like to file a follow-up issue to merge your changes back into the shared copy, as that is the better option. Even if you are using new specific stuff we should be able to add that as an option to the shared copy so we can keep things consistent and shared.

src/Run/Run.csproj
maririos: I'll test it before going to corefx.
Thanks!

src/Run/CommandLine.cs
maririos: Ok. For now I'm going to open an issue in VSO and once this change is merged I'll create an issue in Github.

weshaggard: Don't bother with the VSO issue just go ahead with the github issue in this repo.

maririos:
Is there something else we want to address? or can I merge this change

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+working_tree_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+echo $working_tree_root/corerun $working_tree_root/MSBuild.exe $*
```
weshaggard: We may consider removing the echo here to be consistent with msbuild.cmd.

weshaggard:
LGTM. Please squash and then merge.

PR #673, opened 2016-04-26
Remove --infer-runtimes restore argument

dagood:
project.json files in this repository already have runtimes, so we should be able to remove `--infer-runtimes`. I haven't validated using all the packages in builds, but diffing the contents of `Microsoft.DotNet.BuildTools` before/after looks as similar I could expect from two builds.

Related CoreFX removal: https://github.com/dotnet/corefx/pull/8107

/cc @ericstj @weshaggard


weshaggard:
LGTM

PR #674, opened 2016-04-26
Conditioning import of IL Targets as a temp workaround to allow coreclr onboard the new buildtools

joperezr:
coreclr repo has their own IL.targets which are similar to the ones we have in buildtools except they have additional flags they use to build. For now, We are just conditioning the import of the buildtools il.targets with this PR, and eventually we will merge the coreclr and buildtools version of them to only have one targets file.

cc: @mellinoe 

related: dotnet/coreclr#4598


mellinoe:
LGTM.

PR #675, opened 2016-04-27
Avoid having xunit runner do shadow copy

tarekgh:
When xunit runner does shadow copy for used assemblies it causes a problem when running agaist the desktop. the framework will disallow loading our assemblies which is open source signed.
we pass -noshadow to xunit runner to avoid such issues

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
@@ -172,6 +172,7 @@
 
     <PropertyGroup>
         <OtherRunnerScriptArgs Condition="'$(FilterToTargetGroup)' == 'net46'">$(OtherRunnerScriptArgs) --xunit-test-type=desktop </OtherRunnerScriptArgs>
+        <XunitArgs Condition="'$(FilterToTargetGroup)' == 'net46'"> -noshadow $(XunitArgs)</XunitArgs>
     </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


tarekgh:
@weshaggard @MattGal could you please have a quick look? Thanks.

tarekgh:
@weshaggard @MattGal do you have any comment or you are ok with the change?

AlexGhiondea:
LGTM

tarekgh:
Thanks Alex.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -172,6 +172,7 @@
 
     <PropertyGroup>
         <OtherRunnerScriptArgs Condition="'$(FilterToTargetGroup)' == 'net46'">$(OtherRunnerScriptArgs) --xunit-test-type=desktop </OtherRunnerScriptArgs>
+        <XunitArgs Condition="'$(FilterToTargetGroup)' == 'net46'"> -noshadow $(XunitArgs)</XunitArgs>
```
weshaggard: @MattGal I assume this duplication will go away with your new runner scripts?

tarekgh: @weshaggard yes I talked offline with Matt and he is already working to have unified targets for local and Helix runs. this change just to make sure we'll not have any breaks till Matt's finish his change.

PR #676, opened 2016-04-27
Packaging cleanup

ericstj:
Remove UsePackageTargetRuntimeDefaults since it is no longer used and replaced with OSGroup > PackageTargetRuntime mapping.

Remove System.Numerics.Vectors from generations as part of https://github.com/dotnet/corefx/issues/5900

/cc @chcosta @weshaggard


weshaggard:
LGTM

PR #677, opened 2016-04-27
Fix to UploadToAzure to return false when something fail

maririos:
The task UploadToAzure was always returning true even if there were failures in the execution.
With this change we validate that if an exception happened we are returning false. 

cc: @karajas @chcosta @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
@@ -120,10 +120,17 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                     Log.LogMessage(MessageImportance.Normal, "Received response to check whether Container blobs exist");
                 }
             }
-            await ThreadingTask.WhenAll(Items.Select(item => UploadAsync(ct, item, blobsPresent)));
+            try
+            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -120,10 +120,17 @@ public override bool Execute()
...
-            await ThreadingTask.WhenAll(Items.Select(item => UploadAsync(ct, item, blobsPresent)));
-
-           Log.LogMessage(MessageImportance.High, "Upload to Azure is complete, a total of {0} items were uploaded.", Items.Length);
+            try
+            {
+                await ThreadingTask.WhenAll(Items.Select(item => UploadAsync(ct, item, blobsPresent)));
+            }
+            catch (Exception)
+            {
+                Log.LogError("Failed to upload to Azure");
```
joperezr: You can use LogErrorFromException instead :smile:

jhendrixMSFT:
Thanks for fixing this.  On a somewhat related note could you please update line 90 so that the HashSet ctor takes a StringComparer.OrdinalIgnoreCase for its parameter?  I don't believe blobs can be unique based on their casing.

PR #680, opened 2016-04-29
Add wrapper script so we can call the CloudTest targets directly

chcosta:
Adds a wrapper script which imports dir.props and CloudTest.targets so that we can make a direct call to the CloudTest targets.  This enables us to push to Helix by running "msbuild Tools\CloudTestWrapper.targets /t:CloudBuild /p:blah".  The test build does not utilize build,proj, it's simply a build of tests.builds, which doesn't import the cloud test targets.  In addition, it is useful to have this a discrete step in the build pipeline so that our test build does not claim failure if there is an issue pushing to Helix, and it allows us to more easily rerun that single failed step.

/cc @weshaggard @jhendrixMSFT , @MattGal

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTestWrapper.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTestWrapper.targets
@@ -1,0 +1,4 @@
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <Import Project="$(MSBuildThisFileDirectory)CloudTest.targets" />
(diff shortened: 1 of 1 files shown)


chcosta:
@dotnet-bot test Innerloop Windows Debug

weshaggard:
This should be a `.proj` file if you are planning to call it with msbuild.

weshaggard:
Actually another option is to get this targets file imported indirectly by the tests.builds file so you can call the target directly by building that project. Is there any project in corefx today that imports the cloudtargets? If so you can also just use that project and build just the one target.

chcosta:
Oh yes,  it should be a proj file,  duh.  

The only file I'm certain imports that target is build.proj. In my mind,  this is its own distinct step and it's cleaner to have it as such.  I think that may be my own constructed perception though and I'm completely ok using your solution and calling the target directly via build.proj instead.  Thanks for the suggestion!

weshaggard:
Yes I think it is OK to just call the target via build.proj. That is the entry point for lots of targets.

PR #684, opened 2016-05-02
Only update test dependencies which reference specific repo built packages 

chcosta:
/cc @weshaggard @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
@@ -182,6 +182,7 @@
                                   VersionStructureRegex="$(VersionStructureRegex)"
                                   AdditionalDependencies="@(_InjectProjectReferenceDependency)"
                                   ProjectJson="$(OriginalProjectJson)"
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -182,6 +182,7 @@
                                   VersionStructureRegex="$(VersionStructureRegex)"
                                   AdditionalDependencies="@(_InjectProjectReferenceDependency)"
                                   ProjectJson="$(OriginalProjectJson)"
+                                  IdentityRegex="$(BuildTestsAgainstPackagesIdentityRegex)"
```
chcosta: I have an additional PR for the corefx repo which defines this property (regex).  I'll submit that PR after this PR is reviewed / merged.

jhendrixMSFT:
Could you provide an example regex you'd pass to this task?

chcosta:
I'll be passing in the existing corefx validation pattern - https://github.com/dotnet/corefx/blob/master/dir.props#L89

weshaggard:
I'm currently OK with the approach but I'm starting to wonder how far these regex's will degenerate to such a point where we end up with a list of packages. Lets see how this goes.

weshaggard:
LGTM

PR #689, opened 2016-05-03
Add validation support for .NET 4.6.3

ericstj:
This adds a targeting pack file list and a validation moniker for .NET 4.6.3.

/cc @weshaggard @chcosta


weshaggard:
Should we add a comment in the commit or the xml file that this is pre-release and subject to change.

LGTM

ericstj:
These aren't any public statement of support or anything so I don't see why we should add a comment.  It'll just get stale.  We didn't do that for 4.6.1 or 4.6.2.

Issue #687, opened 2016-05-03
Race in ArchiveTestBuild target


chcosta:
There is a race in the ArchiveTestBuild target of publishtest.target - https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets#L232

The DestinationArchive path doesn't take into account that a test library may be built for different target groups.  See https://github.com/dotnet/corefx/blob/master/src/System.Globalization.Extensions/tests/System.Globalization.Extensions.Tests.builds for an example of where a test library is built for different targets in the CoreFx build.

The result is that you will attempt to create the same zip file with different content..
```
 116:2>ArchiveTestBuild:
         Compressing D:\A\_work\15\s\bin/Windows_NT.AnyCPU.Release/System.Globalization.Extensions.Tests/net46/ into D:\A\_work\15\s\bin/tests/Windows_NT.AnyCPU.Release/archive/tests/System.Globalization.Extensions.Tests.zip...
       (ArchiveTestBuild target) -> 
         D:\A\_work\15\s\Tools\publishtest.targets(232,5): error : An exception has occured while trying to compress 'D:\A\_work\15\s\bin/Windows_NT.AnyCPU.Release/System.Globalization.Extensions.Tests/' into 'D:\A\_work\15\s\bin/tests/Windows_NT.AnyCPU.Release/archive/tests/System.Globalization.Extensions.Tests.zip'. [D:\A\_work\15\s\src\System.Globalization.Extensions\tests\System.Globalization.Extensions.Tests.csproj]
```

This needs to be fixed as it is causing non-deterministic failures in our build pipeline.  Assigning to @jhendrixMSFT 

/cc @weshaggard

weshaggard:
cc @tarekgh @MattGal 
We will need to take this into account when building for the different targets and potentially in the universal runner logic, not sure.

jhendrixMSFT:
Fixed in https://github.com/dotnet/buildtools/commit/ae1465a55c36640d4c0ad6588750c931ebd69f88.

PR #690, opened 2016-05-03
Include TargetGroup in the test archive name as required.

jhendrixMSFT:
Some tests build for multiple target groups; in these cases include the
target group value in the archive name to ensure they're unique.

Fixes issue https://github.com/dotnet/buildtools/issues/687#event-649282773.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -218,18 +218,21 @@
 
   <!-- archive the test binaries along with some supporting files -->
   <Target Name="ArchiveTestBuild" AfterTargets="CreateAssemblyListTxt" Condition="'$(ArchiveTests)' == 'true'">
+    <PropertyGroup Condition="'$(TestProjectName)'==''">
+      <TestProjectName>$(MSBuildProjectName)</TestProjectName>
+    </PropertyGroup>
     <PropertyGroup>
       <TestArchiveDir>$(TestWorkingDir)$(OSPlatformConfig)/archive/tests/</TestArchiveDir>
       <ProjectJson Condition="!Exists('$(ProjectJson)')">$(OriginalProjectJson)</ProjectJson>
-    </PropertyGroup>
-    <PropertyGroup Condition="'$(TestProjectName)'==''">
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -218,18 +218,21 @@
...
+      <TestProjectName>$(MSBuildProjectName)</TestProjectName>
+    </PropertyGroup>
     <PropertyGroup>
       <TestArchiveDir>$(TestWorkingDir)$(OSPlatformConfig)/archive/tests/</TestArchiveDir>
       <ProjectJson Condition="!Exists('$(ProjectJson)')">$(OriginalProjectJson)</ProjectJson>
-    </PropertyGroup>
-    <PropertyGroup Condition="'$(TestProjectName)'==''">
-      <TestProjectName>$(MSBuildProjectName)</TestProjectName>
+      <DestinationArchiveFile>$(TestArchiveDir)$(TestProjectName)</DestinationArchiveFile>
+      <DestinationArchiveFile Condition="'$(TargetGroup)' != ''">$(DestinationArchiveFile).$(TargetGroup)</DestinationArchiveFile>
```
chcosta: I don't know enough about how this is consumed to know if there are any issues with modifying the archive name.  The logic of what you're doing looks fine to me.

chcosta: LGTM

jhendrixMSFT: Hmm yes this might cause a problem with Helix testing as we use file archive name minus the extension to determine the test binary name.  I think we'll need to place these archives in their own sub-directory.

weshaggard: Does it need to be in the file name? Can it be in the path similar to https://github.com/dotnet/corefx/blob/master/dir.props#L494?

jhendrixMSFT: Yep that's what I was thinking.  I'm testing out the change and will update the PR once I have it working.

PR #691, opened 2016-05-03
Fix Crypto members is UWP ApiList

ericstj:
I ran into an issue when helping out on the crypto Algorithms work where
we suddenly started seeing Pinvoke errors after targeting netcore50.

I found that this was missed because we don't let projects that use
TargetGroup='' opt into the more-restrictive UWP checks.

Additionally I found that our data file was wrong for these BCrypt APIs.
I double checked our data file with the one in the WACK and these were
the main differences.

There were two others that I didn't fix:
api-ms-win-core-fibers-l2-1-1.dll!CalloutOnFiberStack
mscoree.dll!MetaDataGetDispenser

There were a lot more that were missing from our list which I also 
didn't fix:
https://gist.github.com/ericstj/9e807d06bcde648963e03646a5f1cde9

/cc @AlexGhiondea @steveharter @bartonjs


bartonjs:
LGTM. (I'm happy that the two API lists now agree on which DLL those BCrypt* functions live in)

ericstj:
@bartonjs are you aware of any historical reason for them to disagree?  I have no context and just asserted an assumption (they should agree) and wrote a tool to validate it.
/cc @pallavit in case she has context.

bartonjs:
@ericstj Nope, I feel like we brought it up with them and they went "oh, our bad, we'll fix it"; but I don't think we refreshed our datafiles since just before releasing .NET Native.

Issue #614, opened 2016-04-05
Make sure that GenerateVersionHeader creates the directory if it doesn't exist


joperezr:
Today if you call GenerateVersionHeader target passing in the location where you want to create the file, the target will error if the folder you want to create it on doesn't exist. This should be handled on the target level and not on the consumer side.

cc: @weshaggard 

related: dotnet/corefx#7453

PR #694, opened 2016-05-04
Create Directory if not exists

joperezr:
cc: @weshaggard 

fixes: #614

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -161,7 +161,14 @@
       <NativeVersionLines Condition="'$(Configuration)'=='Debug'" Include="#define VER_DEBUG                   VS_FF_DEBUG" />
       <NativeVersionLines Condition="'$(Configuration)'!='Debug'" Include="#define VER_DEBUG                   0" />
       <NativeVersionLines Include="#endif" />
+      <NativeVersionFile Include="$(NativeVersionHeaderFile)" />
     </ItemGroup>
+
+    <PropertyGroup>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -161,8 +161,15 @@
       <NativeVersionLines Condition="'$(Configuration)'=='Debug'" Include="#define VER_DEBUG                   VS_FF_DEBUG" />
       <NativeVersionLines Condition="'$(Configuration)'!='Debug'" Include="#define VER_DEBUG                   0" />
       <NativeVersionLines Include="#endif" />
+      <NativeVersionFile Include="$(NativeVersionHeaderFile)" />
```
weshaggard: Is NativeVersionHeaderFile set anywhere but above? If not then please prefer to split the property to a path and file there instead of depending on msbuild batching to split the directory off.

joperezr: Today it is being set when we call msbuild in [corefx build.cmd](https://github.com/dotnet/corefx/blob/master/build.cmd#L78) and [coreclr build.cmd](https://github.com/dotnet/coreclr/blob/master/build.cmd#L297). But we could change that to use the ones in the property group instead.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -116,11 +116,9 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(GenerateNativeVersionInfo)'=='true'">
```
joperezr: After taking a closer look, it seems that this PropertyGroup is never used so I'm cleaning it up and using it for this PR's purposes

weshaggard:
LGTM

PR #695, opened 2016-05-05
Add docker cleanup script

chcosta:
Cleanup script removes all containers and images from a VM using docker.

It would be nice to have more fine grained control over what is removed, but having a cleanup script available from BuildTools puts us in a much better place for having clean build machines in our build lab which aren't running out of disk space issues due to docker images / containers lingering.

It would also be nice if this functionality was a part of clean.sh, but I'm not really certain how it would fit into the flow seamlessly (perhaps checking for 'docker' and then running the script if docker is installed).  We wouldn't want this to be a default as it could be extremely disruptive to have all of a devs images / containers disappear.

/cc @weshaggard , @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/cleanup-docker.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/cleanup-docker.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/perl
+
+#
+# ./cleanup-docker.sh 
+#
+
+# cleanup containers
+my $psList = `docker ps -a`;
+my @psItems = split /\n/, $psList;
+foreach(@psItems) {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/cleanup-docker.sh
```diff
@@ -0,0 +1,33 @@
...
+
+#
+# ./cleanup-docker.sh 
+#
+
+# cleanup containers
+my $psList = `docker ps -a`;
+my @psItems = split /\n/, $psList;
+foreach(@psItems) {
+  if($_ =~ /.*\s+([^\s]+)$/ig) {
```
jhendrixMSFT: Why is the regex match necessary?  Would be good to add a comment explaining.

jhendrixMSFT:
LGTM

Issue #473, opened 2016-02-22
[Investigate] init-tools.cmd command is succeeding even when init-tools.log is showing errors


joperezr:
Investigate why sometimes the command succeeds even though there were errors in the restore. Any error should automatically fail the command itself so that a clear error is thrown for the user.

joperezr:
cc: @weshaggard

PR #697, opened 2016-05-05
Capture exit codes, and set the exit code correctly

joperezr:
turns out we were not caching the exit codes when calling dotnet cli for restore or publish commands, so if they failed we were not returning the failure, which in some cases makes the user think that init tools succeeded even though it didin't.

This, plus a change in the consumer's init-tools script to read the exit code of this scripts, will hopefully make errors more clear.

cc: @weshaggard 

fixes: #473

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -31,9 +31,28 @@ ROBOCOPY "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%" /E
 set TOOLRUNTIME_PROJECTJSON=%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\project.json
 @echo on
 call "%DOTNET_CMD%" restore "%TOOLRUNTIME_PROJECTJSON%" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -31,21 +31,45 @@ ROBOCOPY "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%" /E
 set TOOLRUNTIME_PROJECTJSON=%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\project.json
 @echo on
 call "%DOTNET_CMD%" restore "%TOOLRUNTIME_PROJECTJSON%" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json
+set RESTORE_ERROR_LEVEL=%ERRORLEVEL%
+@echo off
+if not [%RESTORE_ERROR_LEVEL%]==[0] (
+	echo ERROR: An error occured on the restore.
```
weshaggard: Can we be a little clearer on the error? i.e what restore? What do I do to fix it? Is there a log I should look at?

joperezr: Since this is the inner init-tools, this error statement will already be *in the log* so that's why I thought that the user can go and find in the log what happened. In the outer init-tools (the one that lives in each repo) I will catch this exit code, and then redirect people to look in the log for more details. If you still think that I should add more verbosity I can certainly do that.

weshaggard: I just want to make sure that if someone calls build.cmd and sees this error they have some idea of what they need to do to fix it.

joperezr:
Merging this now in order to make the required changes in corefx and coreclr sides. Will address any additional feedback later.

PR #704, opened 2016-05-11
Updating test.targets to make the xunit test assembly configurable

schaabs:
This is needed for stress runs as we run an xunit shim assembly in lieu of the assembly we are actually building.  So we need a way of replacing the assembly argument passed to xunit without overriding the well known property TargetFileName which has other consumers.

@MattGal for further review and context.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -72,7 +72,10 @@
     <XunitOptions Condition="'$(Performance)'!='true'">$(XunitOptions) -notrait Benchmark=true</XunitOptions>
 
     <XunitOptions Condition="'$(XunitMaxThreads)'!=''">$(XunitOptions) -maxthreads $(XunitMaxThreads)</XunitOptions>
-    <XunitArguments>$(TargetFileName) $(XunitOptions)</XunitArguments>
+
```
weshaggard: Please collapse the blank lines.

weshaggard:
LGTM

PR #706, opened 2016-05-11
Move to microsoft net core app

joperezr:
Moving exe projects to use Microsoft.NETCore.App package instead, and libraries to use NETStandard.Library. This is the first step to eliminate the runtime from the Tools directory.

CC: @weshaggard @ericstj


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -3,38 +3,27 @@
...
       "dependencies": {
         "NuProj": "0.10.4-beta-gf7fc34e7d8",
+        "MSBuild": "0.1.0-preview-00022",
         "Microsoft.Build.Framework": "0.1.0-preview-00022",
         "Microsoft.Build.Tasks.Core": "0.1.0-preview-00022",
         "Microsoft.Build.Utilities.Core": "0.1.0-preview-00022",
         "Microsoft.Build.Targets": "0.1.0-preview-00022",
         "Microsoft.Build": "0.1.0-preview-00022",
+        "Microsoft.CodeAnalysis.CSharp": "1.3.0-beta1-20160429-01",
+        "Microsoft.CodeAnalysis.VisualBasic": "1.3.0-beta1-20160429-01",
```
joperezr: I had to add this two because of a dependency conflict between Microsoft.NETCore.App and Microsoft.Net.Compilers.NetCore.

maririos:
@joperezr is there more work to be done for this PR to be merged?

src/ApiCompat/project.json
```diff
@@ -1,28 +1,21 @@
 {
+  "buildOptions": {
```
weshaggard: We sholuldn't need these buildOptions when building via msbuild.

src/Microsoft.Cci.Extensions/Microsoft.Cci.Extensions.csproj
```diff
@@ -14,7 +14,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <ProjectGuid>{2179F9B5-1DBA-4563-9402-A94DE75EA9FA}</ProjectGuid>
     <DefineConstants>$(DefineConstants);COREFX</DefineConstants>
-    <NuGetTargetMoniker>.NETStandard,Version=v1.3</NuGetTargetMoniker>
+    <NuGetTargetMoniker>.NETStandard,Version=v1.5</NuGetTargetMoniker>
```
weshaggard: Is there a reason we needed 1.5? Would 1.3 work?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -3,38 +3,27 @@
     "dnxcore50": {
```
weshaggard: This should be changed to netcoreapp1.0

joperezr:
@maririos 
> is there more work to be done for this PR to be merged?

Given that we have shipped RC2, I was planning on updating the CLI to preview1 latest and Microsoft.NetCore.App to the shipped version. I already started that work, but I'm not done yet, I'll update that PR once I'm done.

src/ApiCompat/project.json
joperezr: True, I'll remove them.

src/Microsoft.Cci.Extensions/Microsoft.Cci.Extensions.csproj
joperezr: It would probably work, but since we are moving the project.json to depend on RC2 NetStandard.Library and we want to be consistent with the tool-runtime/project.json I think it is better to have this match the 1.5

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
joperezr: That would be a breaking change because the publish we are doing is for dnxcore50 which is why I added [this here](https://github.com/dotnet/buildtools/pull/706/files#diff-05eb29e66e87abd2de147e792d51d6b3R26). Ultimately I was going to have  it as you suggest, just thought of staging the changes of first changing project.jsons and then changing init-tools scripts, but I can change it to that if you want

src/Microsoft.Cci.Extensions/Microsoft.Cci.Extensions.csproj
weshaggard: I would prefer we stick with netstandard1.3 so we don't accidently take dependencies on newer APIs as netstandard 1.3 is compatible with net46 as well where as 1.5 is not.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
weshaggard: I'm OK with staging it but I do want to remove all references to dnxcore50 from things like this.

maririos:
Thanks @joperezr

dir.props
```diff
@@ -100,6 +100,7 @@
     <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
     <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/nugetbuild/api/v3/index.json" />
+    <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/roslyn-nightly/api/v3/index.json" />
```
weshaggard: I'm not sure we want to add the roslyn-nightly feed. Can we just upload the dependent package to our buildtools or dotnet-core feed?

weshaggard:
Other than the Roslyn feed the changes look good to me.

dir.props
joperezr: right, actually I forgot about this, this is for the two packages that we needed to upload to our feed. I'll try to do it myself to see if I have permissions

joperezr:
@dotnet-bot test this please

joperezr:
Done, I've uploaded the packages to the buildtools feed so as soon as I get a green CI I'll merge

PR #710, opened 2016-05-12
Change NuGet MinClient version to 2.12

ericstj:
NuGet is adding support for NETStandard to 2.12.  As such we can make
our packages depend on this version as the minimum.  Unfortunately there
is no way to express 2.12 <= version < 3.0 || 3.4 <= version, so I added
a note to the package descriptions that says we require 3.4 when using
NuGet 3.x.

/cc @weshaggard @chcosta


weshaggard:
:+1:

PR #714, opened 2016-05-13
Add UrlEncode to block id for block upload url

karajas:
According to the PUT Block documentation,
blockid - Required. A valid Base64 string value that identifies the block. Prior to encoding, the string must be less than or equal to 64 bytes in size.
For a given blob, the length of the value specified for the blockid parameter must be the same size for each block.
Note that the Base64 string must be URL-encoded.

/cc @schaabs @jhendrixMSFT


schaabs:
I tested out this change and it fixes the upload error I was seeing.

jhendrixMSFT:
@schaabs thanks for testing, LGTM

PR #716, opened 2016-05-13
Adding checksum .rsp for managed libraries

joperezr:
cc: @weshaggard 

related: dotnet/corefx#8479

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -12,4 +12,8 @@
   -->
   <Import Project="$(MSBuildThisFileDirectory)ReferenceAssemblies.props" Condition="'$(ExcludeReferenceAssembliesImport)'!='true'" />
 
+  <PropertyGroup>
+    <CompilerResponseFile Condition="'$(ExcludeChecsumParameterFromCsc)'!='true'">$(MSBuildThisFileDirectory)checksum.rsp</CompilerResponseFile>
+  </PropertyGroup>
+
 </Project>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -13,7 +13,7 @@
   <Import Project="$(MSBuildThisFileDirectory)ReferenceAssemblies.props" Condition="'$(ExcludeReferenceAssembliesImport)'!='true'" />
 
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


joperezr:
The way this will be consumed in corefx, is by adding 
```xml
<PropertyGroup>
  <CompilerResponseFile>$(ToolsDir)checksum.rsp</CompilerResponseFile>
</PropertyGroup>
```
To the root dir.props. That way, checksum.rsp will be used always, and for the specific case of System.Runtime, we will merge the contents of coreAssembly.rsp and checksum.rsp

weshaggard:
I would prefer we just have a property that we enable as opposed to specifying the rsp file.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
```diff
@@ -37,4 +35,25 @@
...
+  <Target Name="MergeCoreAssemblyResponseFile">
+    <ItemGroup>
+      <ResponseFilesToMerge Include="$(CompilerResponseFile);$(MSBuildThisFileDirectory)coreAssembly.rsp" />
+    </ItemGroup>
+    <ReadLinesFromFile File="%(ResponseFilesToMerge.Identity)">
+      <Output TaskParameter="Lines" ItemName="ResponseFileLines" />
+    </ReadLinesFromFile>
+    <WriteLinesToFile File="$(IntermediateOutputPath)/resp.rsp" Lines="@(ResponseFileLines)" Overwrite="true" />
+    <PropertyGroup>
+      <CompilerResponseFile>$(IntermediateOutputPath)/resp.rsp</CompilerResponseFile>
```
weshaggard: Please verify this but I think the compiler allows you to specify multiple rsp files so you don't need to merge them manually.

joperezr: You are right. Looks like this works when you use a semicolon as a separator, I believe that in my testing I used a colon by mistake which led me to think it didn't work.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -12,4 +12,8 @@
   -->
   <Import Project="$(MSBuildThisFileDirectory)ReferenceAssemblies.props" Condition="'$(ExcludeReferenceAssembliesImport)'!='true'" />
 
+  <PropertyGroup>
+    <CompilerResponseFile Condition="'$(ExcludeChecsumParameterFromCsc)'!='true'">$(MSBuildThisFileDirectory)checksum.rsp</CompilerResponseFile>
```
weshaggard: I would change this to.
`'$(CheckSumSHA256)'!='false'`

joperezr: Done.

weshaggard:
LGTM

PR #717, opened 2016-05-13
Fix race condition in dependencylist.txt

chcosta:
We can hit file lock issues if a test library is built targeting multiple TargetGroups because the dependencylist.txt file was not unique per targetgroup.  This change fixes the condition by adding the targetgroup to the dependencylist.txt path name.

/cc @MattGal @jhendrixMSFT


jhendrixMSFT:
LGTM

PR #719, opened 2016-05-13
Fix for incorrect auth signing using encoded url parameters

karajas:
Fix for errors introduced in #714 . There was a failure during upload due to the use of encoded url parameters instead of decoded ones during the construction of the string to sign for authorization.

/cc @schaabs @jhendrixMSFT


jhendrixMSFT:
Will merge pending validation from @schaabs

PR #721, opened 2016-05-16
Change default to not use slash for qualifiers

weshaggard:
When tools need to parse arguments for other platforms (i.e. Linux/OSX)
they have files that begin with "/" and that messes up the parser and
causes an "undefined qualifer" error. The best way to handle this is
to disable the uses of slash for qualifiers by default. People can still
manually enable it if they chose to want to support it but for most
of our tools that are now running x-plat we need to disable this and
only support dash for qualifiers.

cc @stephentoub @ellismg

### DIFF
--- a/src/common/CommandLine.cs
+++ b/src/common/CommandLine.cs
@@ -15,36 +15,36 @@ namespace Microsoft.Fx.CommandLine
     // See code:#Overview to get started.
     /// <summary>
     /// #Overview
-    /// 
+    ///
     /// The code:CommandLineParser is a utility for parsing command lines. Command lines consist of three basic
     /// entities. A command can have any (or none) of the following (separated by whitespace of any size).
-    /// 
+    ///
     ///    * PARAMETERS - this are non-space strings. They are positional (logicaly they are numbered). Strings
     ///        with space can be specified by enclosing in double quotes.
-    ///        
+    ///
     ///    * QUALIFIERS - Qualifiers are name-value pairs. The following syntax is supported.
     ///        * -QUALIFIER
     ///        * -QUALIFIER:VALUE
     ///        * -QUALIFIER=VALUE
     ///        * -QUALIFIER VALUE
-    ///        
+    ///
     ///      The end of a value is delimited by space. Again values with spaces can be encoded by enclosing them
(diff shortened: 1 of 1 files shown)


src/common/CommandLine.cs
```diff
@@ -1746,18 +1746,18 @@ private static string GetEntryAssemblyName()
         private bool _noDashOnParameterSets;
         private bool _noSpaceOnQualifierValues;
         private string[] _parameterSetsWhereQualifiersMustBeFirst;
-        private bool _qualifiersUseOnlyDash;
+        private bool _qualifiersUseOnlyDash = true;
```
weshaggard: This is the only interesting change in this diff. Everything else is just line-ending being removed by sublime text.

weshaggard:
This change should take care of the issue we were seeing with https://github.com/dotnet/corefx/pull/8513

ellismg:
LGTM.  I wonder why Linux was working?

src/common/CommandLine.cs
ellismg: FYI, You can pass ?w=1 to the diff view to have it ignore white-space only changes.

weshaggard:
@ellismg good question. It looks like there are some other interesting things happening at least on the ubuntu runs. http://dotnet-ci.cloudapp.net/job/dotnet_corefx/job/master/job/ubuntu14.04_debug_prtest/1104/consoleFull where we are getting:
`/bin/sh: 3: /tmp/tmp03e02cd8044643e78360a95fa48780c1.exec.cmd: /mnt/j/workspace/dotnet_corefx/master/ubuntu14.04_debug_prtest/Tools/CoreRun.exe: not found`

Even in our master CI runs without my changes we seem to be getting some other interesting errors with the ubuntu runs, which aren't registering as errors. see http://dotnet-ci.cloudapp.net/job/dotnet_corefx/job/master/job/ubuntu14.04_debug/326/consoleFull
`chmod: cannot access ‘/mnt/j/workspace/dotnet_corefx/master/ubuntu14.04_debug/bin/tests/Unix.AnyCPU.Debug/Microsoft.Win32.Primitives.Tests/dnxcore50/corerun’: No such file or directory`

I don't have an ubuntu machine setup right now to take a deeper look but it is something worth looking into further.

PR #722, opened 2016-05-16
Azure retry logic incorrectly using stale DateTime in HTTP header.

jhendrixMSFT:
The lambda to create HttpRequestMessage was incorrectly capturing a local
variable for DataTime instead of reading the current DateTime when
creating a new request; this lead to reusing the old DateTime causing the
value to be stale thus failing retries.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
@@ -83,12 +83,12 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                     blockIds.Add(blockId);
                     string blockUploadUrl = blobUploadUrl + "?comp=block&blockid=" + WebUtility.UrlEncode(blockId);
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -135,11 +135,12 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
             }
 
             string blockListUploadUrl = blobUploadUrl + "?comp=blocklist";
-            DateTime dt1 = DateTime.UtcNow;
+
             using (HttpClient client = new HttpClient())
             {
                 Func<HttpRequestMessage> createRequest = () =>
                 {
+                    DateTime dt1 = DateTime.UtcNow;
```
chcosta: These variable names appear to be really inconsistent ("dt", "dt1", "dateTime").

chcosta:
Minor feedback on variable names.

LGTM

PR #724, opened 2016-05-16
Failure to send HTTP request can lead to NullReferenceException.

jhendrixMSFT:
Ensure that response is not null before dereferencing it, and reset it to
null at the start of each iteration.


chcosta:
LGTM

PR #725, opened 2016-05-16
Add bin/obj dir to the include folders for NativeResourceCompile

joperezr:
#694 changed the location of where _version.h file is dropped by default to be BaseIntermediateOutputPath instead of IntermediateOutputPath, so we also need to update the NativeResourceCompile to also include this folder in the include path.


joperezr:
cc: @weshaggard

weshaggard:
LGTM

PR #726, opened 2016-05-17
Fix for OOM in UploadToAzure when a large number of items are specified

schaabs:
Introduced property MaxClients which is used to throttle the number of UploadClients which will be used concurrently to upload blobs


schaabs:
@karajas: Adding Karthik as an fyi of the fix for OOM issue we discussed

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -57,6 +57,9 @@ public class UploadToAzure : Task, ICancelableTask
         /// </summary>
         public bool Overwrite { get; set; } = false;
 
+
+        public int MaxClients { get; set; } = 8;
```
jhendrixMSFT: Can you add a summary for this property?

PR #731, opened 2016-05-18
Packaging fixes

ericstj:
Please review commits individually.

Fail on error.  Previously we'd miss package breaks during CI because we didn't halt the build.  This change ensures that the build will fail.

Refactor validation report: serialize the object graph rather than manually writing JSON.  Also save off the asset resolution info.

Refactor build sequence & rearrange targets.  I needed this to prepare for the last change, but it is split up to show a cleaner diff.  I've confirmed that this change produces no diff in nuspecs.

Automatically reference RID-graph and lineup.  Previously we required an external reference to these packages so folks building stand-alone/ala'carte apps needed to know that they had to reference them.  This adds the reference automatically.

/cc @weshaggard @mellinoe @chcosta

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNetStandardSupportTable.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNetStandardSupportTable.cs
@@ -68,27 +68,22 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
             SortedSet<Version> knownNetStandardVersions = new SortedSet<Version>();
             List<SupportRow> rows = new List<SupportRow>(Reports.Length);
 
-            foreach (var report in Reports.Select(r => r.GetMetadata("FullPath")))
+            foreach (var reportPath in Reports.Select(r => r.GetMetadata("FullPath")))
             {
                 SupportRow row = new SupportRow();
-                row.Name = Path.GetFileNameWithoutExtension(report);
+                row.Name = Path.GetFileNameWithoutExtension(reportPath);
                 row.SuportedVersions = new SortedSet<NETStandardApiVersion>();
 
-                using (var file = File.OpenText(report))
-                using (var reader = new JsonTextReader(file))
+                var report = ValidationReport.Load(reportPath);
+
+                foreach(var supportedFramework in report.SupportedFrameworks)
                 {
-                    var doc = JObject.Load(reader);
-                    var supportedFrameworks = doc["supportedFrameworks"] as JObject;
+                    var fx = NuGetFramework.Parse(supportedFramework.Key);
 
-                    foreach(var supportedFramework in supportedFrameworks.Properties())
+                    if (fx.Framework == FrameworkConstants.FrameworkIdentifiers.NetStandard)
                     {
-                        var fx = NuGetFramework.Parse(supportedFramework.Name);
-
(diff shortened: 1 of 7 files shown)


chcosta:
@jhendrixMSFT , https://github.com/dotnet/buildtools/pull/731/commits/13c8f912088180712629b64222519aa2ba9b427e may have an impact on how you are building the packages in the build pipeline.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNetStandardSupportTable.cs
```diff
@@ -68,27 +68,22 @@ public override bool Execute()
             SortedSet<Version> knownNetStandardVersions = new SortedSet<Version>();
             List<SupportRow> rows = new List<SupportRow>(Reports.Length);
 
-            foreach (var report in Reports.Select(r => r.GetMetadata("FullPath")))
+            foreach (var reportPath in Reports.Select(r => r.GetMetadata("FullPath")))
```
chcosta: The properties in this task aren't really documented at all.  It would be nice to know what a valid "Reports" item is and what metadata is available in case some change needs to be made in the future.

ericstj: Its just a normal MSBuild item with default metadata, pointing to the package report which is described by that type itself which is not directly serialized/deserialized.  This change actually makes the code much more self-documenting then what was here before.

chcosta: Ok, if there's no custom metadata, then documentation isn't necessary.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -660,8 +658,7 @@
   </Target>
 
   <Target Name="EnsureEmptyPackage"
-          DependsOnTargets="DetermineRuntimeDependencies"
-          BeforeTargets="GenerateNuSpec">
```
chcosta: Is there a growing resistance force against the use of BeforeTargets / AfterTargets?  ie, is there some intrinsic property of DependsOnTargets which makes that preferable, or is it just an easier mental model to consistently utilize the simpler inferred behavior of DependsOnTargets?

ericstj: Simpler mental model to be consistent in one set of targets.
BeforeTargets / AfterTargets is fine, it's just complicated when you're using a mix of both.  Sometimes they are necessary when you can't add the right DependsOnTargets, EG: if the target you are sequencing with is in somefile you don't own and that file didn't have an extensible list.

chcosta:
I like this change - https://github.com/dotnet/buildtools/pull/731/commits/49644614d3f1e0f791a48129675254c8eb0315d2

chcosta:
LGTM, as long as @jhendrixMSFT is ready to consume these changes.

jhendrixMSFT:
I discussed this with @chcosta, looked at some logs from our official builds and we think it's good to go.

ericstj:
Great.  Thanks for the review!

PR #735, opened 2016-05-19
Adding Commit hash for native binaries

joperezr:
cc: @gkhanna79 @markwilkie @jhendrixMSFT 

version.c will look something like this:
```c
static char sccsid[] __attribute__((used)) = "@(#)Version 1.0.24117.0 built by: joperezr-MyMachine. Commit Hash: 1d963d41e5b56623ff054b567ec85a650f581275";

```
I'm open to suggestions on the string format.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -182,12 +182,12 @@
   <Target Name="GenerateVersionSourceFile"
       Inputs="$(MSBuildProjectFile)"
       Outputs="$(NativeVersionSourceFile)"
-      DependsOnTargets="CreateVersionFileDuringBuild"
+      DependsOnTargets="CreateVersionFileDuringBuild;GetLatestCommitHash"
       Condition="'$(NativeVersionSourceFile)'!='' and '$(GenerateVersionSourceFile)'=='true'">
 
     <ItemGroup>
       <SourceFileLines />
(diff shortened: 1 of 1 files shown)


joperezr:
related: dotnet/coreclr#5097

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -204,6 +204,17 @@
...
   </Target>
 
+  <Target Name="GetLatestCommitHash" Condition="'$(LatestCommit)'==''">
+    <Exec Command="git rev-parse HEAD" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="LatestCommit" />
+      <Output TaskParameter="ExitCode" PropertyName="LatestCommitExitCode" />
+    </Exec>
+    <!-- We shouldn't fail the build if we can't retreive the commit hash, so in this case just set it to N/A -->
+    <PropertyGroup Condition="'$(LatestCommitExitCode)'!='0'">
+      <LatestCommit>N/A</LatestCommit>
```
gkhanna79: Are we not going to compute it if LatestCommit is not available as discussed?

joperezr: Yes, if you see the target's condition, it will only execute if you haven't defined `LatestCommit` already, for example in official builds this will be already passed in. This error handling is more for the case were LatestCommit is not defined, and the git rev-parse command fails, which may happen if git is not installed or not in the path for example.

gkhanna79: Got it - thanks. LGTM.

PR #738, opened 2016-05-20
Adding Commit hash to native Windows binaries as well

joperezr:
cc: @weshaggard @gkhanna79

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -124,7 +124,7 @@
   <Target Name="GenerateVersionHeader"
       Inputs="$(MSBuildProjectFile)"
       Outputs="$(NativeVersionHeaderFile)"
-      DependsOnTargets="CreateVersionFileDuringBuild"
+      DependsOnTargets="CreateVersionFileDuringBuild;GetLatestCommitHash"
       Condition="'$(NativeVersionHeaderFile)'!='' and '$(GenerateVersionHeader)'=='true'">
 
     <ItemGroup>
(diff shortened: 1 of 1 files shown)


gkhanna79:
While the change looks fine, what is "Windows" specific to this change (compared to x-plat change)?

joperezr:
In order to add versioning to native binaries, you do it in two different ways depending if it's windows or non-windows. If on Windows, you need to have an RC file that imports a header file (version.h) which contains the values you want to set as version, and version.h is generated by the target that I'm modifying in this PR.

For non-windows, you need to actually generate a source file (version.c) that declares the string with the version, and you compile that. version.c is generated by a different target, which is the one I modifyied on my last PR.

joperezr:
Hey @weshaggard when you have some time, can you PTAL?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -57,7 +57,7 @@
       <AssemblyInfoLines Include="[assembly:AssemblyCopyright(&quot;\x00a9 Microsoft Corporation.  All rights reserved.&quot;)]" />
       <AssemblyInfoLines Include="[assembly:AssemblyVersion(&quot;$(AssemblyVersion)&quot;)]" />
       <AssemblyInfoLines Include="[assembly:AssemblyFileVersion(&quot;$(AssemblyFileVersion)&quot;)]" />
-      <AssemblyInfoLines Include="[assembly:AssemblyInformationalVersion(@&quot;$(AssemblyFileVersion)$(BuiltByString)&quot;)]" />
+      <AssemblyInfoLines Include="[assembly:AssemblyInformationalVersion(@&quot;$(AssemblyFileVersion)$(BuiltByString). Commit Hash%3A $(LatestCommit)&quot;)]" />
```
weshaggard: nit: I would have built the entire Commit string into its own property similar to BuiltByString so it could be empty in cases were we cannot get a commit.

weshaggard:
:+1:

PR #739, opened 2016-05-20
PkgProj test reference

ericstj:
Support P2P reference from test to PkgProj.

This adds support for CSProj referencing PkgProj.  I chose a basic way to plug in the resolve task.  Depending on how we decide to deal with publishing for different TestFrameworks/Runtimes we might want to change how this plugs in.

/cc @weshaggard @karajas @tarekgh @MattGal @mellinoe 

PS: will rebase/squash before comitting.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetApplicableAssetsFromPackages.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetApplicableAssetsFromPackages.cs
@@ -1,0 +1,167 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet.Frameworks;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.DotNet.Build.Tasks.Packaging
+{
+    public class GetApplicableAssetsFromPackages : PackagingTask
+    {
+        private Dictionary<string, List<PackageItem>> _packageToPackageItems;
+        private Dictionary<string, PackageItem> _targetPathToPackageItem;
+        private AggregateNuGetAssetResolver _resolver;
+
+        /// <summary>
+        /// All items that make up packages to resolve from.  
+        /// Must have the following metadata
+        ///     Identity - path to the file in the binary directory
+        ///         TargetPath - path to the file in the package
+        ///         PackageId - ID of the package containing the file
+        /// </summary>
+        [Required]
+        public ITaskItem[] PackageAssets { get; set; }
+
+        /// <summary>
+        /// TargetMoniker to use when resolving assets.  EG: netcoreapp1.0, netstandard1.4
+        /// </summary>
+        [Required]
+        public string TargetMoniker { get; set; }
+
+        /// <summary>
+        /// runtime.json that defines the RID graph
+        /// </summary>
+        [Required]
+        public string RuntimeFile { get; set; }
+        
+        /// <summary>
+        /// If specified will be used when resolving runtime assets, otherwise TargetMoniker will be used.
+        /// </summary>
+        public string RuntimeTargetMoniker { get; set; }
+
+        /// <summary>
+        /// If specified will be used when resolving runtime assets, otherwise no RID will be used.
+        /// </summary>
+        public string TargetRuntime { get; set; }
+
+        [Output]
+        public ITaskItem[] CompileAssets { get; set; }
+
+        [Output]
+        public ITaskItem[] RuntimeAssets { get; set; }
+
+        /// <summary>
+        /// Generates a table in markdown that lists the API version supported by 
+        /// various packages at all levels of NETStandard.
+        /// </summary>
+        /// <returns></returns>
+        public override bool Execute()
+        {
+            if (PackageAssets == null || PackageAssets.Length == 0)
+            {
+                Log.LogError("PackageAssets argument must be specified");
+                return false;
+            }
+
+            if (String.IsNullOrEmpty(TargetMoniker))
+            {
+                Log.LogError("TargetMoniker argument must be specified");
+                return false;
+            }
+
+            NuGetFramework fx = NuGetFramework.Parse(TargetMoniker);
+            NuGetFramework runtimeFx = fx;
+
+            if (!String.IsNullOrEmpty(RuntimeTargetMoniker))
+            {
+                runtimeFx = NuGetFramework.Parse(RuntimeTargetMoniker);
+            }
+
+            string targetString = String.IsNullOrEmpty(TargetRuntime) ? fx.ToString() : $"{fx}/{TargetRuntime}";
+
+            LoadFiles();
+
+            CompileAssets = _resolver.ResolveCompileAssets(fx)
+                                .Where(ca => !NuGetAssetResolver.IsPlaceholder(ca))
+                                .Select(ca => PackageItemAsResolvedAsset(_targetPathToPackageItem[ca]))
+                                .ToArray();
+            RuntimeAssets = _resolver.ResolveRuntimeAssets(runtimeFx, TargetRuntime)
+                                .Where(ra => !NuGetAssetResolver.IsPlaceholder(ra))
+                                .Select(ra => PackageItemAsResolvedAsset(_targetPathToPackageItem[ra]))
+                                .ToArray();
+
+            return !Log.HasLoggedErrors;
+        }
+
+        private void LoadFiles()
+        {
+            _packageToPackageItems = new Dictionary<string, List<PackageItem>>();
+            foreach (var file in PackageAssets)
+            {
+                try
+                {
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetApplicableAssetsFromPackages.cs
```diff
@@ -0,0 +1,167 @@
...
+            if (String.IsNullOrEmpty(TargetMoniker))
+            {
+                Log.LogError("TargetMoniker argument must be specified");
+                return false;
+            }
+
+            NuGetFramework fx = NuGetFramework.Parse(TargetMoniker);
+            NuGetFramework runtimeFx = fx;
+
+            if (!String.IsNullOrEmpty(RuntimeTargetMoniker))
```
tarekgh: do we have any case we pass RuntimeTargetMoniker as empty or null?

tarekgh:
LGTM. Thanks @ericstj for getting this ready.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetApplicableAssetsFromPackages.cs
ericstj: Yeah, suppose this is used by something other than a test.  I'm just trying to come up with the right behavior for the task, then you can decide how you want to drive it with targets.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -3,10 +3,47 @@
...
+  
+  <PropertyGroup>
+    <!-- Resolve from pkgProjs before nuget packages -->
+    <ResolveNugetPackagesDependsOn>$(ResolveNugetPackagesDependsOn);ResolvePkgProjReferences</ResolveNugetPackagesDependsOn>
+    <ResolvePkgProjReferencesDependsOn>ResolveProjectReferences</ResolvePkgProjReferencesDependsOn>
+  </PropertyGroup>
+  
+  <Target Name="ResolvePkgProjReferences" Condition="'@(PackageAsset)' != ''" DependsOnTargets="$(ResolvePkgProjReferencesDependsOn)">
+    <PropertyGroup>
+      <!-- Temporary workaround -->
```
weshaggard: Do we need to work items tracking this workaround and the below hack?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -3,10 +3,47 @@
...
+      <NuGetTargetMoniker Condition="'$(NuGetTargetMoniker)' == '.NETPortable,Version=v4.5,Profile=Profile7'">DNXCore,Version=v5.0</NuGetTargetMoniker>
+    </PropertyGroup>
+    
+    <!-- Hack this for now to use TestNugetRuntimeId.  We probably want to run this in a way to get the compile assets for the
+         target platform and runtime assets for the RID-specific platform.  -->
+    <GetApplicableAssetsFromPackages PackageAssets="@(PackageAsset)" 
+                                     TargetMoniker="$(NuGetTargetMoniker)"
+                                     RuntimeTargetMoniker="$(TestNuGetTargetMoniker)"
+                                     TargetRuntime="$(TestNugetRuntimeId)"
```
weshaggard: @tarekgh were you going to introduce TestRID property to represent this?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -3,10 +3,47 @@
...
   <UsingTask TaskName="ValidatePackageTargetFramework" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
+  <UsingTask TaskName="GetApplicableAssetsFromPackages" AssemblyFile="$(PackagingTaskDir)Microsoft.DotNet.Build.Tasks.Packaging.dll"/>
 
+  <Target Name="UpdatePkgProjProjectReferences" BeforeTargets="BeforeResolveReferences">
+    <!-- Update PkgProj references to call the GetValidationReport target -->
+    <ItemGroup>
+      <ProjectReference Condition="'%(Extension)'=='.pkgproj'">
+        <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+        <Targets>Build;GetPackageAssets</Targets>
+        <OutputItemType>PackageAsset</OutputItemType>
```
weshaggard: Is this item group name too general? Will it conflict with appx packaging for example?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: Same workaround exists in packageresolve.targets.  The hack is less of a hack now, but I do expect @tarekgh / @karajas to tune this as part of their work.  I clean up the comments.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: It doesn't conflict with anything today, but I can make it more specific just to avoid any possibility.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
tarekgh: @weshaggard  I don't think we need to introduce TestRID. I was thinking to use TestNugetRuntimeId and this will still be derived from TestTFM. do we really need TestRID property?

weshaggard: We should chat more about it but I do think we should have TestRID as well as TestTFM for consistency and the reason we want the shorter one is because I expect that people will need to specify it in their builds files. If that isn't the case any longer then we can consider not changing this. My main concern is making a connection between the different Test* properties so it is consistent and clear they both effect what we run on.

tarekgh: @weshaggard yes let's chat offline about that. I always welcome your feedback :-)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -38,6 +38,7 @@
     <ImplicitlyExpandTargetFramework>false</ImplicitlyExpandTargetFramework>
     <!-- Disable RAR complaining about us referencing higher .NET Portable libraries as we aren't a traditional portable library -->
     <ResolveAssemblyReferenceIgnoreTargetFrameworkAttributeVersionMismatch>true</ResolveAssemblyReferenceIgnoreTargetFrameworkAttributeVersionMismatch>
+    <NuGetTargetMoniker Condition="'$(NuGetTargetMoniker)' == ''">DNXCore,Version=v5.0</NuGetTargetMoniker>
```
weshaggard: What do we need to do to shift this to NETCoreApp?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -3,10 +3,44 @@
...
+  <PropertyGroup>
+    <!-- Resolve from pkgProjs before nuget packages -->
+    <ResolveNugetPackagesDependsOn>$(ResolveNugetPackagesDependsOn);ResolvePkgProjReferences</ResolveNugetPackagesDependsOn>
+    <ResolvePkgProjReferencesDependsOn>ResolveProjectReferences</ResolvePkgProjReferencesDependsOn>
+  </PropertyGroup>
+
+  <!-- Resolves applicable files from all files in PkgProjs. Similar to ResolveNuGetPackageAssets, except operates on a file list
+       without requiring actual nupkgs, project.json, or restore -->
+  <Target Name="ResolvePkgProjReferences" Condition="'@(PkgProjAsset)' != ''" DependsOnTargets="$(ResolvePkgProjReferencesDependsOn)">
+    <!-- If TestNuGetTargetMoniker is set, use that for resolving runtime assets. -->
```
weshaggard: Where is TestNugetTargetMoniker being set? Are you expecting that @tarekgh sets that based on TestTFM?

weshaggard:
Overall looks good to me.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
ericstj: All the tests need to update their project.json's and we need to update the TestTargetFramework.  I suspect this will be done with @karajas's work.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
ericstj: Yeah, as I've mentioned multiple times, this is a starting point.  I expect @tarekgh / @karajas to tune this as appropriate.

PR #742, opened 2016-05-23
Relax version structure Regex

chcosta:
Permit additional release values in version structure Regex.

BuildNumberOverrideStructureRegex and VersionStructureRegEx are now nearly identical and these should be collapsed into one regex in an additional cleanup PR (not part of this PR).

/cc @jhendrixMSFT @weshaggard


weshaggard:
LGTM

PR #744, opened 2016-05-23
Add missing dependency of System.Runtime.Serialization.Primitives required for xplat builds

joperezr:
cc: @weshaggard 

I missed this runtime dependency with my last merge, and due to it the xplat build won't work when upgrading buildtools. With this change that adds it, building xplat should work now.


joperezr:
Merging now to unblock dotnet/corefx#8790

PR #747, opened 2016-05-24
Add retry logic when downloading the CLI from Azure blob storage.

jhendrixMSFT:
Downloading from Azure can transiently fail for a number of reasons.  Add
some retry logic when downloading to avoid transient failures.

### DIFF
--- a/init-tools.cmd
+++ b/init-tools.cmd
@@ -47,7 +47,7 @@ set DOTNET_ZIP_NAME=dotnet-dev-win-x64.%DOTNET_VERSION%.zip
 set DOTNET_REMOTE_PATH=https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/%DOTNET_VERSION%/%DOTNET_ZIP_NAME%
 set DOTNET_LOCAL_PATH=%DOTNET_PATH%%DOTNET_ZIP_NAME%
 echo Installing '%DOTNET_REMOTE_PATH%' to '%DOTNET_LOCAL_PATH%' >> %INIT_TOOLS_LOG%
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Here's the code in a non-compressed form.

$retryCount = 0;
$success = $false;
do
{
    try
    {
        (New-Object Net.WebClient).DownloadFile('%DOTNET_REMOTE_PATH%', '%DOTNET_LOCAL_PATH%');
        $success = $true;
    }
    catch
    {
        if ($retryCount -ge 6)
        {
            throw;
        }
        else
        {
            $retryCount++;
            Start-Sleep -Seconds (5 * $retryCount);
        }
    }
}
while ($success -eq $false);

jhendrixMSFT:
Once we're happy with the pattern I intend on porting it to the product repos.

ChadNedzlek:
Whew. That is a mouthful on the line. But given we're writing a CMD file, we don't have a ton of options, I guess. LGTM

weshaggard:
Sigh... I really don't like seeing all this extra logic being included here. Can we look to see if there is any existing API or way to download a file on windows that supports retrying directly.

markwilkie:
There are APIs that Azure publishes, but they don't support core (yet).  Perhaps as an interim step is to create our own mini-tool set that we publish in build tools to interact with Azure.

weshaggard:
@markwilkie unfortunately I don't think that will help much in this case as this is the initial bootstraping code to pull down the CLI from Azure blob which we need in order to do the restoring of build tools. Whatever we use he has to already be on the system.

markwilkie:
@weshaggard, true.  But at least we can relegate that duplication to bootstrapping code.  The right longer term solution is to use a published API.  Until that's ready, we might should think about putting our own placeholder in.

jhendrixMSFT:
@weshaggard @markwilkie If we want to go the tool route then I would suggest we use wget for Windows as it already has retry logic baked in (i.e. let's not reinvent the wheel; we do this on Linux with great success).  Of course we still need to figure out how to get wget on the machine to begin with.

PR #748, opened 2016-05-24
Add support for external package dependencies with fixed versions

chcosta:
This is an interim change while we sort out inter-dependency versioning.

This change allows us to specify package versions for packages not built in the current repo, and then use those package versions when generating a new project.json during the test library build when we are adding a new package dependency into a framework.  ie, Supersede package versions we find in a PackagesDrop with those specified in ExternalPackages ITaskItem.  

I don't expect this to be how we ultimately handle these external packages (though perhaps some form of this will be involved), but this will unblock the corefx repo for Helix failures where it is trying to use the latest build of System.IO.Compression from the CoreFx repo instead of the publically published version which has stable bits.

There is a small related corefx change (the package versions file) which I will send a PR for today.

/cc @weshaggard @dagood @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -41,6 +41,9 @@ namespace Microsoft.DotNet.Build.Tasks
         [Required]
         public string ProjectJson { get; set; }
 
+        // External package dependency versions.
+        public ITaskItem[] ExternalPackages { get; set; }
+        
         // The directory to put the generated project.json in
         [Required]
         public string OutputProjectJson { get; set; }
@@ -91,13 +94,13 @@ namespace Microsoft.DotNet.Build.Tasks
             }
 
             // Update default dependencies section
-            JObject dependencies = GenerateDependencies(projectRoot);
+            JObject dependencies = GenerateDependencies(projectRoot, ExternalPackages);
             projectRoot = UpdateDependenciesProperty(projectRoot, dependencies);
 
             // Update framework dependencies sections
             for (int i = 0; i < Frameworks.Length; i++)
             {
-                dependencies = GenerateDependencies(projectRoot, Frameworks[i]);
+                dependencies = GenerateDependencies(projectRoot, ExternalPackages, Frameworks[i]);
                 projectRoot = UpdateDependenciesProperty(projectRoot, dependencies, Frameworks[i]);
             }
             WriteProject(projectRoot, OutputProjectJson);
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -200,12 +203,16 @@ private JObject GenerateDependencies(JObject projectJsonRoot, string framework =
                 // Don't add a new dependency if one already exists.
                 if (returnDependenciesList.FirstOrDefault(rd => ((JProperty)rd).Name.Equals(name)) == null)
                 {
-                    NuGetVersion dependencyVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
-
-                    string version = string.Join(".", dependencyVersion.Major, dependencyVersion.Minor, dependencyVersion.Patch);
-                    if (!string.IsNullOrWhiteSpace(PackageBuildNumberOverride))
+                    var externalPackageVersion = externalPackageVersions.FirstOrDefault(epv => epv.ItemSpec.Equals(name, StringComparison.OrdinalIgnoreCase))?.GetMetadata("Version");
+                    string version = externalPackageVersion;
```
weshaggard: Why do you need both version and externalPackageVersion variables?

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -203,13 +203,12 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
...
                 {
-                    var externalPackageVersion = externalPackageVersions.FirstOrDefault(epv => epv.ItemSpec.Equals(name, StringComparison.OrdinalIgnoreCase))?.GetMetadata("Version");
-                    string version = externalPackageVersion;
+                    var version = externalPackageVersions.FirstOrDefault(epv => epv.ItemSpec.Equals(name, StringComparison.OrdinalIgnoreCase))?.GetMetadata("Version");
                     if (version == null)
                     {
                         NuGetVersion dependencyVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
                         version = string.Join(".", dependencyVersion.Major, dependencyVersion.Minor, dependencyVersion.Patch);
-                        if (!string.IsNullOrWhiteSpace(PackageBuildNumberOverride) && externalPackageVersion == null)
+                        if (!string.IsNullOrWhiteSpace(PackageBuildNumberOverride) && version == null)
```
weshaggard: Looking at this logic I'm now less convinced the variable was unnecessary. With the current logic I do't see how version could ever be null here but externaPackageVersion could have been null.

PR #749, opened 2016-05-24
Update rids for netcoreapp1.0

ellismg:


ericstj:
These are just here to try and represent the concrete RIDs that we build shared framework for and that folks will use to deploy stand-alone apps, in order to make sure that our packages will restore correctly on those combinations.
If CLI won't ever use that RID I'm OK if you remove it.

PR #750, opened 2016-05-24
Use platform specific CoreRun in API tasks

ellismg:
This should address dotnet/corefx#8815

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
@@ -44,7 +44,7 @@
     </ItemGroup>
 
     <PropertyGroup>
-      <_ApiCompatCmd>"$(ToolsDir)CoreRun.exe" "$(ToolsDir)ApiCompat.exe" "$(ReferenceAssembly)" </_ApiCompatCmd>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
```diff
@@ -44,7 +44,7 @@
     </ItemGroup>
 
     <PropertyGroup>
-      <_ApiCompatCmd>"$(ToolsDir)CoreRun.exe" "$(ToolsDir)ApiCompat.exe" "$(ReferenceAssembly)" </_ApiCompatCmd>
+      <_ApiCompatCmd>$(ToolHostCmd) "$(ToolsDir)ApiCompat.exe" "$(ReferenceAssembly)" </_ApiCompatCmd>
```
weshaggard: There is one more location below with this same problem. Can you please update it as well?

weshaggard:
Thanks for fixing this @ellismg. 

LGTM (after you fix the other location as well).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.targets
ellismg: Thanks, fixed.

PR #751, opened 2016-05-24
Switch test framework to netcoreapp1.0

ericstj:
This switches the default framework and the test framework from
DNXCore50 to netcoreapp1.0.  I've also taken the opportunity to clean
up all non-essential refs to DNXCore50, DNXCore, and DNX.

The only remaining references to these deprecated monikers are:
1. Import statements: needed until all packages are updated.
2. Microsoft.DotNet.Build.Tasks.Packaging.Tests: needed until we can
update the version of BuildTools used by BuildTools to get this change.

/cc @weshaggard @karajas @MattGal @joshfree

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/test/EnsureOOBFrameworkTests.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/test/EnsureOOBFrameworkTests.cs
@@ -152,8 +152,8 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging.Tests
         {
             ITaskItem[] files = new[]
             {
-                CreateItem(@"D:\K2\binaries\x86ret\NETCore\Libraries\dnxcore\System.Xml.XmlSerializer.dll", "lib/DNXCore50", "dnxcore50"),
-                CreateItem(@"D:\K2\binaries\x86ret\NETCore\Libraries\dnxcore\System.Xml.XmlSerializer.dll", "lib/netcore50", "netcore50"),
+                CreateItem(@"D:\K2\binaries\x86ret\NETCore\Libraries\System.Xml.XmlSerializer.dll", "lib/netcoreapp1.0", "netcoreapp1.0"),
+                CreateItem(@"D:\K2\binaries\x86ret\NETCore\Libraries\System.Xml.XmlSerializer.dll", "lib/netcore50", "netcore50"),
                 CreateItem(@"D:\K2\src\NDP\FxCore\src\Packages\_._", "lib/MonoAndroid10", "MonoAndroid10"),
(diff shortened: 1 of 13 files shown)


ericstj:
Right now we need this to have CoreFx pick up new packages.  The number of dependencies on netstandard1.6 has grown such that its no longer tenable to use the hacks we've been using.

CoreFx changes are here: https://github.com/ericstj/corefx/commit/7b696765dd40c4dcee43039a6f8983ab2493d4a4

ericstj:
Ping.  Going to merge this tonight regardless of review. Need to unblock update.

src/Microsoft.DotNet.Build.Tasks.Packaging/test/EnsureOOBFrameworkTests.cs
```diff
@@ -241,7 +241,7 @@ public void OOBFxShouldNotAddWhenAlreadyExists()
                 CreateItem(@"D:\K2\src\Redist\x86\retail\bin\i386\HelpDocs\intellisense\NETCore5\1049\System.Reflection.Emit.xml", "ref/netstandard1.1/ru", ""),
                 CreateItem(@"D:\K2\src\Redist\x86\retail\bin\i386\HelpDocs\intellisense\NETCore5\2052\System.Reflection.Emit.xml", "ref/netstandard1.1/zh-hans", ""),
                 CreateItem(@"D:\K2\src\Redist\x86\retail\bin\i386\HelpDocs\intellisense\NETCore5\3082\System.Reflection.Emit.xml", "ref/netstandard1.1/es", ""),
-                CreateItem(@"D:\K2\binaries\x86ret\NETCore\Libraries\dnxcore\System.Reflection.Emit.dll", "lib/DNXCore50", "DNXCore50"),
+                CreateItem(@"D:\K2\binaries\x86ret\NETCore\Libraries\System.Reflection.Emit.dll", "lib/netcoreapp1.0", "netcoreapp1.0"),
```
bartonjs: Are all the netcoreapp1.0 things going to be netcoreapp1.1 with the next release? If so, do we want a const/static-readonly?  If not... do we anyways? :)

bartonjs:
LGTM

src/Microsoft.DotNet.Build.Tasks.Packaging/test/EnsureOOBFrameworkTests.cs
ericstj: I only changed these so that we weren't testing with a deprecated framework.  They won't change again.We'd add new test cases.
Will netcoreapp1.0 become netcoreapp1.1?  Depends on what changes.  I could imagine supporting corert would warrant such a change

bartonjs: Gotcha.  Thanks for the explanation.

weshaggard:
LGTM

Thanks for pushing this rename through.

PR #754, opened 2016-05-25
Handle case when ExternalPackages is null.

jhendrixMSFT:
The ExternalPackages property is optional, so check for null first.


jhendrixMSFT:
Introduced in https://github.com/dotnet/buildtools/commit/297bef36f7a420349be8352413ef2e9fdb6ca645

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -203,7 +203,10 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
                 // Don't add a new dependency if one already exists.
                 if (returnDependenciesList.FirstOrDefault(rd => ((JProperty)rd).Name.Equals(name)) == null)
                 {
-                    var version = externalPackageVersions.FirstOrDefault(epv => epv.ItemSpec.Equals(name, StringComparison.OrdinalIgnoreCase))?.GetMetadata("Version");
+                    string version = null;
+                    if (externalPackageVersions != null)
```
weshaggard: You should use braces around this single statement.

PR #757, opened 2016-05-26
Add force option when cleaning up docker images.

jhendrixMSFT:


jhendrixMSFT:
/cc @ellismg @chcosta 
I've seen a few cases where the clean-up script fails to delete the image without using "-f".

chcosta:
This is generally because it has dependent images which aren't being deleted.  If we want to have a clean state and ensure we download images everytime,  then adding -f is fine,  but I thought it was probably fine  to not force delete by default and someone could manually force delete it necessary.  I don't think that a few images lingering around is a major problem.  Containers tend to eat up more disk space.

jhendrixMSFT:
@chcosta But will they accumulate over time?

chcosta:
I'm ok with adding this,  when this script was internal,  we had the force option enabled.   I'm only outlining that we control the images on the machines and we don't update them that much, so there are pros and cons to this approach.  Both the pros and the cons are fairly negligible since these machines aren't shared.  I also didn't  want to enforce the force option on anybody else that is potentially using this script (I'm not aware of anybody else using this).

PR #759, opened 2016-05-26
Move to RC3 buildtools and support more distros

ellismg:
We need these changes for bootstrapping CLI (and to get CoreFX building from source on more distros).


ellismg:
@weshaggard Can you please take a look?

weshaggard:
LGTM

PR #760, opened 2016-05-27
Add the ability to filter out RIDs when building tests against packages.

jhendrixMSFT:
We have a situation where new RIDs have been added to the product but we
don't yet have the actual packages; as a result the test build fails when
buliding against packages as it can't resolve these runtime packages.  To
work around this I have added an optional property to filter out RIDs when
generating the test project.json files.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -43,6 +43,11 @@ namespace Microsoft.DotNet.Build.Tasks
 
         // External package dependency versions.
         public ITaskItem[] ExternalPackages { get; set; }
+
+        /// <summary>
+        /// Optional list of RIDs to exclude from the generated project.json.
+        /// </summary>
+        public string[] ExcludedRuntimes { get; set; }
         
         // The directory to put the generated project.json in
         [Required]
@@ -96,6 +101,12 @@ namespace Microsoft.DotNet.Build.Tasks
             // Update default dependencies section
             JObject dependencies = GenerateDependencies(projectRoot, ExternalPackages);
             projectRoot = UpdateDependenciesProperty(projectRoot, dependencies);
+
+            if (ExcludedRuntimes != null)
+            {
+                var excludedRIDs = new HashSet<string>(ExcludedRuntimes, StringComparer.OrdinalIgnoreCase);
+                projectRoot = FilterRuntimes(projectRoot, excludedRIDs);
+            }
 
             // Update framework dependencies sections
             for (int i = 0; i < Frameworks.Length; i++)
@@ -278,6 +289,28 @@ namespace Microsoft.DotNet.Build.Tasks
             var frameworkPathObject = projectJsonRoot.SelectToken(frameworkPath);
             frameworkPathObject["dependencies"] = updatedProperties;
             return projectJsonRoot;
+        }
+
+        private JObject FilterRuntimes(JObject projectRoot, HashSet<string> excludedRIDs)
+        {
+            var runtimes = projectRoot["runtimes"];
+            if (runtimes != null)
+            {
+                var toRemove = new List<JToken>();
+                foreach (JProperty runtime in runtimes)
+                {
+                    if (excludedRIDs.Contains(runtime.Name))
+                        toRemove.Add(runtime);
+                }
+
+                foreach (var token in toRemove)
+                {
+                    Log.LogMessage("Removing RID '{0}' from the list of applicable runtimes.", ((JProperty)token).Name);
+                    token.Remove();
+                }
+            }
+
+            return projectRoot;
         }
 
         private static void WriteProject(JObject projectRoot, string projectJsonPath)
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
@@ -185,7 +185,8 @@
                                   Frameworks="$(GeneratedTargetGroup)"
                                   OutputProjectJson="$(ProjectJson)" 
                                   PackageBuildNumberOverride="$(PackageBuildNumberOverride)"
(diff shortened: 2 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -177,6 +177,11 @@
     <PropertyGroup>
         <PackageBuildNumberOverride Condition="'$(PackageBuildNumberOverride)' == '' and '$(_DerivedPackageNumber)' != ''">$(_DerivedPackageNumber)</PackageBuildNumberOverride>
     </PropertyGroup>
+    <!-- we don't yet have runtime packages for these RIDs so exclude them for now -->
+    <ItemGroup>
+      <ExcludedRuntimes Include="fedora.23-x64" />
```
weshaggard: Lets not include these in the build tools targets. They will be temporary so I would just include them in corefx so we can add and remove them easier.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -280,6 +296,28 @@ private JObject UpdateDependenciesProperty(JObject projectJsonRoot, JObject upda
...
+                foreach (JProperty runtime in runtimes)
+                {
+                    if (excludedRIDs.Contains(runtime.Name))
+                        toRemove.Add(runtime);
+                }
+
+                foreach (var token in toRemove)
+                {
+                    Log.LogMessage("Removing RID '{0}' from the list of applicable runtimes.", ((JProperty)token).Name);
+                    token.Remove();
```
weshaggard: Is there a reason you cannot just do the Remove in the first loop?

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -43,6 +43,11 @@ public class AddDependenciesToProjectJson : Task
 
         // External package dependency versions.
         public ITaskItem[] ExternalPackages { get; set; }
+
+        /// <summary>
+        /// Optional list of RIDs to exclude from the generated project.json.
+        /// </summary>
+        public ITaskItem[] ExcludedRuntimes { get; set; }
```
weshaggard: These could be just `string[]` and then below you can just pass the, directly into the constructor for HashSet.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
jhendrixMSFT: I assumed it would invalidate the collection causing the foreach to fail, I will try it.

weshaggard: Well if I knew the type of runtimes then I might be able to give you more definitive guidance (I hate var :(), but a common pattern is to use a for loop in reverse order to do these type of removals.

jhendrixMSFT: It's a JToken object.  Unfortunately it only exposes IEnumerable, you can't directly index into it.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
jhendrixMSFT: Good idea, will fix.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
jhendrixMSFT: Agreed.  I will see where we can define these in CoreFX (probably the sync project that imports this target).

ellismg: Can you help me understand why they can not be found?  There should be packages for these rids on myget today?  Is this when doing a build against a private feed produced by the new build pipeline?

jhendrixMSFT: @ellismg I didn't realize we had these packages on the feed.  Here's a sample error message.

error : Unable to resolve 'runtime.ubuntu.16.04-x64.runtime.native.System.Net.Security (>= 1.0.1-rc4-24126-06)' for '.NETCoreApp,Version=v1.0 (ubuntu.16.04-x64)

We do include the dotnet.myget.org feed when restoring packages.

weshaggard: The ones on the feed are with different version numbers because they are coming from the TFS builds. So the packages with the new build pipeline aren't self consistent when it comes to these packages yet.

ellismg: So this version: `1.0.1-rc4-24126-06` makes me think it's expecting packages from somewhere else (maybe what was just built from the build pipeline?

We definitely publish packages for builds we do out of TFS (see https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.ubuntu.16.04-x64.runtime.native.System.Net.Security)

jhendrixMSFT: @weshaggard nailed it, the version numbers aren't consistent.

ellismg: @weshaggard Gotcha, makes sense.  I just opened a PR to support building CoreFX on the other distros, so maybe we can remove all this stuff later tomorrow or early next week.  Makes sense to do this now to unblock you.

jhendrixMSFT: One thing that's not clear to me is why the build successfully restores packages for opensuse.13.2, does this RID fall back to something we're already building?

jhendrixMSFT:
@weshaggard I've removed the list of RIDs from the target.  Do you have any other feedback?

weshaggard:
LGTM

PR #762, opened 2016-05-27
Reorganize PublishContent.targets so we can overwrite RelativeBlobPath

wtgodbe:


wtgodbe:
@jhendrixMSFT PTAL

jhendrixMSFT:
LGTM

PR #763, opened 2016-05-28
Enable NugetPack to pass through nuspec properties

cdmihai:
With MSBuild nuget packages, we want to have versioned project to project dependencies. For this, nuspec properties seem like an ideal solution: we add a `$version$` variable in the nuspecs and then compute and inject the actual value at build time.

I also changed packages.targets to allow the user to override the package version and not have it automatically generated.

Usage example of the changes (WIP msbuild branch):
- https://github.com/cdmihai/msbuild/commit/18ea882c144a0a0bc9a8b9fab550eb127675f926#diff-aade4fae49f1d220b531687680bb6718
- https://github.com/cdmihai/msbuild/commit/18ea882c144a0a0bc9a8b9fab550eb127675f926#diff-fcc4a9bafd113347e0ca9e95377d8f73

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/Microsoft.DotNet.Build.Tasks.Packaging.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/Microsoft.DotNet.Build.Tasks.Packaging.csproj
@@ -22,6 +22,7 @@
   <ItemGroup>
     <Compile Include="GetApplicableAssetsFromPackages.cs" />
     <Compile Include="GetMinimumNETStandard.cs" />
+    <Compile Include="NugetPropertyStringProvider.cs" />
     <Compile Include="GetRuntimeJsonValues.cs" />
     <Compile Include="GetRuntimeTargets.cs" />
     <Compile Include="SplitDependenciesBySupport.cs" />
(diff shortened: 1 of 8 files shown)


src/packages.builds
```diff
@@ -12,7 +12,7 @@
     <PackagingTaskDir Condition="'$(BuildToolsTargets45)' == 'true'">$(BinDir)AnyOS.AnyCPU.$(ConfigurationGroup)/Microsoft.DotNet.Build.Tasks.Packaging.Desktop/</PackagingTaskDir>
   </PropertyGroup>
 
-  <Import Project="Microsoft.DotNet.Build.Tasks\PackageFiles\packages.targets" Condition="'$(ImportGetNuGetPackageVersions)' != 'false'" />
+  <Import Project="$(ToolsDir)packages.targets" Condition="'$(ImportGetNuGetPackageVersions)' != 'false'" />
```
ericstj: I don't think you need this after my change here: https://github.com/dotnet/buildtools/blob/5e01895842ec91bc5b7ba4cd6be296415c6707f7/src/packages.builds#L7

jeffkl:
We'd like to get this change in to make progress on a feature of ours.  Can someone please review?

ericstj:
@dotnet-bot test this please.

src/packages.builds
ericstj: Ping: Can you please address this?

ericstj:
This looks good to merge after the comment above is addressed.  I've confirmed that https://github.com/dotnet/buildtools/pull/763/commits/a62764fe7ca8753654cbd048033f04a1bbb73146 is unnecessary.

jeffkl:
I've removed commit a62764f.  Thanks for getting to this.

PR #764, opened 2016-05-29
Always use 'rhel.7-x64' as the RHEL 7 publish RID

ellismg:
I broke building on RHEL with my previous change to simplify the publishing RID, since we don't explicitly list rhel.7.2-x64 in our `tool-runtime.json`.

This change forces us to `rhel.7-x64` as the rid to restore on RHEL. The other supported distros don't bump `$VERSION_ID` for minor releases.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -43,6 +43,13 @@ case $OSName in
             source /etc/os-release
             __PUBLISH_RID=$ID.$VERSION_ID-x64
         fi
+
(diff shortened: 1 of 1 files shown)


ellismg:
/cc @weshaggard @joperezr 

I validated this by building on RHEL (and Ubuntu to validate the changes didn't impact non RHEL distros) with a private build tools with this fix.  I am going to merge it now because the CoreCLR RHEL builds are on the floor because of it.  If you have feedback, I'm happy to address it, but I would like to get builds unblocked over the long weekend.

weshaggard:
FYI @jhendrixMSFT we need this to fix our RHEL build issue.

jhendrixMSFT:
@weshaggard Thanks I'll get the latest version of build tools into CoreFX.

PR #765, opened 2016-05-30
Add RequiresElevation constant

ellismg:
We'll use this in CoreFX to mark tests which require an elevated
context to run (e.g root or Administrator). Previously we conflated
this with OuterLoop.


stephentoub:
LGTM

PR #768, opened 2016-06-01
[WIP] Add Serviceable element to NuSpec

ericstj:
This is a temporary implementation of adding the Serviceable element to
all NuSpecs.  This will be used by the CLI host to determine if it
should write breadcrumbs for our packages.


ericstj:
Closing this as it doesn't work.  NuGet strips the elements.

PR #769, opened 2016-06-01
Add PublishExcludePattern allowing exclusion of packages to publish.

jhendrixMSFT:
We need a way to specify a set of packages to exclude from publishing.


jhendrixMSFT:
@weshaggard @dagood 
Fix for excluding certain packages from publishing.

dagood:
LGTM.

I'm kinda wondering why the items are generated in the buildtools logic rather than passed in by the caller. My guess is that it makes it easy to override the property from the command line without much code in corefx/wcf/etc. Is that right?

jhendrixMSFT:
Closing this as we're no longer going to fix publishing in this way.

PR #771, opened 2016-06-02
Switch to the release Helix API.

jhendrixMSFT:
We shouldn't be using the staging endpoint as it's not production quality
and goes down as changes are made and tested.


jhendrixMSFT:
@weshaggard @ChadNedzlek 
Switch to the production Helix endpoint.

ChadNedzlek:
If anyone was using this before (which I don't think they are), they'll need new access tokens, since we've changed target applications.

jhendrixMSFT:
@dsgouda will this affect you?

dsgouda:
I foresee changes to the rsp files we use, is there something else this change would imply?

jhendrixMSFT:
@dsgouda no you'll just need a new API key.
/cc @schaabs as stress will need to make the same change.

PR #774, opened 2016-06-03
Do not set built by string when unix commands fail

joperezr:
@ellismg @weshaggard 

Looks like we were not handling the case when whoami or hostname commands fail right, and we were setting a wrong built by string. With this, in case of any of those two commands failing, we will just not set a built by string in the version string.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -263,13 +263,15 @@
 
     <Exec Command="whoami" Condition="'$(OSEnvironment)'!='Windows_NT' AND '$(VersionUserName)'==''" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="VersionUserName" />
+      <Output TaskParameter="ExitCode" PropertyName="UserNameExitCode" />
     </Exec>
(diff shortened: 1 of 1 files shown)


joperezr:
Merging now since we need this fix for corefx now.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -263,13 +263,15 @@
...
+      <Output TaskParameter="ExitCode" PropertyName="UserNameExitCode" />
     </Exec>
     <Exec Command="hostname" Condition="'$(OSEnvironment)'!='Windows_NT' AND '$(VersionHostName)'==''" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="VersionHostName" />
+      <Output TaskParameter="ExitCode" PropertyName="HostNameExitCode" />
     </Exec>
 
     <PropertyGroup>
-      <BuiltByString Condition="'$(BuiltByString)'==''">%20built by: $(VersionUserName)-$(VersionHostName)</BuiltByString>
+      <BuiltByString Condition="'$(BuiltByString)'=='' AND '$(UserNameExitCode)'=='0' AND '$(HostNameExitCode)'=='0'">%20built by: $(VersionUserName)-$(VersionHostName)</BuiltByString>
```
weshaggard: This change breaks the windows case, where we don't ever call those, we will no longer get the built-by string. Instead of this checking exit codes why not check to see if the VersionUserName/VersionHostName is not empty.

joperezr: I think you are right, I didn't think of that. We can't check if VersionUserName/VersionHostName is empty, because if the command fails because whoami is not installed, then we will get VersionUserName="error: unkown command whoami" or something like that. One thing I can do though is to set the exit codes to 0 in case they are not defined yet one line above this. What  do you think about that?

PR #775, opened 2016-06-03
updating packageresolve.targets to allow dissabling the local copy of references

schaabs:
-added a property, DisableReferenceCopyLocal, which can be user supplied to dissable copying package references locally

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
@@ -88,7 +88,7 @@
     <ItemGroup>
       <!-- Intersect project-refs with package-refs -->
       <_ReferenceFileNamesToRemove Include="@(Reference)" Condition="'@(_ResolvedProjectReferencePaths->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)'" />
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -88,7 +88,8 @@
     <ItemGroup>
       <!-- Intersect project-refs with package-refs -->
       <_ReferenceFileNamesToRemove Include="@(Reference)" Condition="'@(_ResolvedProjectReferencePaths->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)'" />
-      <_ReferenceCopyLocalPathsFileNamesToRemove Include="@(ReferenceCopyLocalPaths)" Condition="'@(_ResolvedProjectReferencePaths->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)'" />
+      <!-- If local copy is dissabled remove all references, otherwise remove only project refrerences -->
```
weshaggard: nit: typo dissabled

weshaggard:
Looks reasonable although I'm not sure why it should be necessary, I need to dig into this some more because some things seems to be wrong.

PR #776, opened 2016-06-03
Fixing Windows case where exit codes are not defined.

joperezr:
@weshaggard 

Fixing issue introduced here: #774


weshaggard:
LGTM

PR #781, opened 2016-06-07
Update to NuGet 3.5.0-beta2-1437

ericstj:
We need the latest version to unblock pack on Unix.

Remove the workaround for project.json that I added before as it is no longer needed: fixed in latest NuGet.

/cc @weshaggard @joshfree @schellap


weshaggard:
:+1:

PR #789, opened 2016-06-10
Bump package versions

ericstj:
Need to do this to get out of the way of the release branch.

/cc @jhendrixMSFT


jhendrixMSFT:
Why don't the other nuspecs need to be updated?

jhendrixMSFT:
I guess we don't build packages for them.  LGTM

ericstj:
The other two were just added after I'm branching for release/1.0.0 so they won't clash.

PR #790, opened 2016-06-10
Update NuGet to 3.5.0-beta2-1456

ericstj:
This fixes a regression caused by the Serviceable attribute addition.

/cc @weshaggard

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -72,7 +72,7 @@
       <ExpectedVersion>2.1.0</ExpectedVersion>
     </ValidationPattern>
     <ValidationPattern Include="^(?i)(NuGet\..%2A)$">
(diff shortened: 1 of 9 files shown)


weshaggard:
LGTM

I assume this removes the schema change only?

ericstj:
You can see the full set here:
https://github.com/NuGet/NuGet.Client/commit/17e8b0f273cca7fd35d266964007a3314c173aef
 - XPlat only, not a package we pick up
https://github.com/NuGet/NuGet.Client/commit/efb2590f24981729df0746aff860283614c04483
 - Protocol, not a package we pick up
https://github.com/NuGet/NuGet.Client/commit/67c7aaaa1bfe7ba4f26ce678b46a22194b1764b9
 - CoreFx ingestion.  We'll do the same.  General goodness.

PR #791, opened 2016-06-10
Update NuGet to 3.5.0-beta2-1456

ericstj:
This fixes a regression caused by the Serviceable attribute addition.  
Same fix is in release/1.0.0, this ports it to master.


weshaggard:
LGTM

PR #793, opened 2016-06-10
Add some conditional core dump generation for OS X to Unix RunnerTemplate

adityamandaleeka:
This change should enable (limited) core file generation on OS X when running tests. To prevent flooding CI machines with dumps, we will only enable core file generation for a test run if, at the time the test run is started, the "/cores" directory on the machine is empty.

The reason for this change is that CoreFX CI runs have occasionally been hitting mysterious segfaults lately, and having some dumps would be really useful for investigating what could be going on.

@ellismg @ericstj @MattGal PTAL. I've made this change in what seems to be an appropriate and safe place (for instance, we'll definitely know what OS we're on at that point), but my entire experience with this test running infrastructure is limited to the past couple of hours :smile:.


adityamandaleeka:
Sorry, meant to open this against the release branch. Will open another PR.

PR #794, opened 2016-06-10
Add some conditional core dump generation for OS X to Unix RunnerTemplate

adityamandaleeka:
This change should enable (limited) core file generation on OS X when running tests. To prevent flooding CI machines with dumps, we will only enable core file generation for a test run if, at the time the test run is started, the "/cores" directory on the machine is empty.

The reason for this change is that CoreFX CI runs have occasionally been hitting mysterious segfaults lately, and having some dumps would be really useful for investigating what could be going on.

@ellismg @ericstj @MattGal PTAL. I've made this change in what seems to be an appropriate and safe place (for instance, we'll definitely know what OS we're on at that point), but my entire experience with this test running infrastructure is limited to the past couple of hours :smile:.

I opened this PR against release/1.0.0 per @ellismg's advice.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
@@ -27,7 +27,7 @@ fi
 
 echo Executing in $EXECUTION_DIR
 
-# ========================= BEGIN Copying files  =============================== 
+# ========================= BEGIN Copying files  =============================
 if [ ! -d "$EXECUTION_DIR" ]
 then
     mkdir $EXECUTION_DIR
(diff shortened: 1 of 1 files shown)


adityamandaleeka:
cc: @sergiy-k

ellismg:
We don't care about Linux?

ellismg:
One thing we should be aware of is this will also get picked up by Helix (which is probably overall goodness) but we'll need to remember to maybe go clean up dumps on the machines there if we hit something so we can get more cores.

ellismg:
@ericstj We would want to take this buildtools update into the release/1.0.0 branch of CoreFX, is using this branch of buildtools appropriate?

adityamandaleeka:
Linux will be coming soon (and will do a bit more than just generate the dump), but I wanted to do OS X first and get it in the CI since that's where a lot of the recent segfaults have been.

ellismg:
SGTM and LGTM.  I'll talk to @ericstj about logistics and then get this merged and CoreFX updated to a new buildtools with this change.

adityamandaleeka:
Thanks!

ericstj:
@weshaggard set up the mirror and build trigger, so you should just be able to merge and it will mirror and schedule the build as usual.  release/1.0.0 is the 1.0.25 version (master is 1.0.26).
Any idea why the CI build isn't running on this branch?  I suspect we're missing some jenkins config /cc @mmitche

ellismg:
We probabbly need to add the branch here: https://github.com/dotnet/dotnet-ci/blob/master/jobs/data/repolist.txt

ericstj:
I see, I gave that a try, can you review and let me know if it looks good?

PR #797, opened 2016-06-11
Enable core dump generation on Linux in Unix RunnerTemplate

adityamandaleeka:
This change should enable core file generation on Linux when running tests. To prevent flooding CI machines with dumps, we will print the backtrace for each thread from the dump and delete it immediately. The idea here is to see something more useful than "Segmentation fault" when something crashes. It will be nice to have the core files transferred somewhere rather than just deleted, but this is a reasonable starting point (and a vast improvement over the current state of affairs).

The reason for using GDB is that LLDB-3.6 (which is what's on many or most of our CI machines) seems to have a bug on Linux where operations like `bt` and `thread list` don't work in batch mode. If/when we unify all the machines to a working LLDB version with a compatible SOS, we should switch to that instead.

We can consider printing register contents and other information as well once we know that this works.

@ellismg @sergiy-k @ericstj PTAL.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
@@ -13,6 +13,28 @@ else
 fi
 }
 
+function print_info_from_core_file {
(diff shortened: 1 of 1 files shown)


sergiy-k:
LGTM

PR #800, opened 2016-06-13
Using the Shared framework that comes with the cli to run all of the tools that are used to build corefx

joperezr:
With this set of changes, all of the tools produced by buildtools + roslyn (csc & vbc) + msbuild will run on the shared framework that comes with the CLI that we restore in corefx. This will also make restore of tools faster, since now we won't have to restore all of the runtime everytime.

NOTE: Roslyn and MSBuild will only use the shared framework in x-plat builds, since we are still using full framework versions of them in Windows builds.

cc: @weshaggard


src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.runtimeconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "runtimeOptions": {
+    "framework": {
+      "name": "Microsoft.NETCore.App",
+      "version": "1.0.0-rc3-002733"
```
joperezr: I'm still debating on having this runtimeconfig files includded in the package or if they should be created using init-tools instead. The reason is that this version is tied to the framework that comes with the cli, which each repo can have a different one, but also buildtools won't work without them so delegating the creation of the files to a different repo is also not ideal. @weshaggard what do you think?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -3,33 +3,89 @@
...
-        "Microsoft.NETCore.TestHost": "1.0.0-rc3-24123-01",
-        "NETStandard.Library": "1.6.0-rc3-24208-04",
         "System.Diagnostics.TextWriterTraceListener": "4.0.0-rc3-24128-00",
         "System.Runtime.Serialization.Primitives": "4.1.1-rc3-24128-00",
         "Newtonsoft.Json": "7.0.1",
-        "NuGet.Client": "3.5.0-beta2-1456",
-        "NuGet.Versioning": "3.5.0-beta2-1456"
+        "Microsoft.NETCore.Targets": "1.0.1-rc3-24128-00",
+
+        "System.AppContext": "4.1.0-rc3-24128-00",
```
joperezr: The reason why I had to add all of these extra dependencies is because I removed the reference to NETStandard.Library and Microsoft.NETCore.App because they both caused the runtime to be published into the tools dir, and if I didn't manually specify all of these dependencies then the ones restored will be older versions which are brought in as secondary dependencies from msbuild and roslyn.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.runtimeconfig.json
weshaggard: I kind of like generating them as part of init-tools based on the CLI that is pulled down. I think that has a better chance of working, although I do think there is a possibility of things breaking there as well.

weshaggard: I wonder if we shouldn't begin to tie the dotnetcli and buildtools versions together in some way.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
```diff
@@ -8,11 +8,11 @@
          that check the invoked process name against known casing which fail under code coverage.
          This is probably because the code coverage target invocation does not canonicalize the
          process name. -->
-    <ToolHost Condition="'$(OS)' == 'Windows_NT'">CoreRun.exe</ToolHost>
-    <ToolHost Condition="'$(OS)' != 'Windows_NT'">corerun</ToolHost>
+    <ToolHost Condition="'$(OS)' == 'Windows_NT'">dotnet.exe</ToolHost>
+    <ToolHost Condition="'$(OS)' != 'Windows_NT'">dotnet</ToolHost>
```
weshaggard: can we eliminate this and just have it be "dotnet" everywhere.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.runtimeconfig.json
mellinoe: Are these runtimeconfig.json files strictly required by the shared host? That design seems problematic in cases like this, where we just want to try to run on whatever shared framework is available (we don't really care).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
mellinoe: NETStandard.Library shouldn't have a dependency on the runtime, should it?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ApiCompat.runtimeconfig.json
joperezr: it is required as it is the way to tell the CLI that the app must run using the shared framework as opposed to just trying to load the dependencies only from the current folder. However, it looks like if your runtimeconfig references a lower version compared to the one that comes with the cli you are using to run, it looks like this should work as well. Seems like the only scenario where it would break, is if the version of the framework in your runtimeconfig is higher than the shared framework that comes with the cli that you are trying to run your app with, which I guess it kind of makes sense.

mellinoe: Sounds like we are fine with keeping this version as-is then. We can upgrade it to 1.0.0 stable when available.

joperezr: Right, I'll leave them here for now

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
joperezr: It does pull down the runtime, but now I've verified that if I specify `"exclude": "runtime"`then it will not pull down all of the runtime assets. I've updated the PR as such.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets
joperezr: Done.

weshaggard:
LGTM - Thanks for pushing this through. I do want to try and get this move to the RTM stable versions but it doesn't need to block for this PR but we should make sure we track that.

PR #806, opened 2016-06-14
Updates for ProjectReference -> Package dependency conversion.

chcosta:
- Support stable package versions
  - Remove build number override in favor of versions from a package drop or from a versions file
  - Support package versions read from a versions file (not currently being utilized, but this adds support).
  - Eliminate unneccessary regex properties
  - Remove auto-assumed release version and package versions, if you want to specify a specific version, then provide a versions file.
  - Remove ParsePackageNames task which is no longer required.

This will allow us to validate against stable packages in the release branch, if we port this to buildtools/release/1.0.0

/cc @weshaggard @ericstj


chcosta:
@weshaggard , can you take a look at this?  I need these changes for the tick-tock run in Release.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -111,14 +117,75 @@ public override bool Execute()
...
+            if(!File.Exists(versionsFile))
+            {
+                Log.LogError("Specified versions file ({0}) does not exist.", versionsFile);
+            }
+            var lines = File.ReadAllLines(versionsFile);
+            foreach(string line in lines)
+            { 
+                if(!string.IsNullOrWhiteSpace(line))
+                {
+                    string packageVersion = line.Replace(' ', '.');
```
weshaggard: I think this is probably acceptable for now but I think it is fragile and we should make this configurable.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -111,14 +117,75 @@ public override bool Execute()
...
+        }
+
+        private TaskItem CreatePackageItemFromString(string package)
+        {
+            Match m = _packageNameRegex.Match(package);
+            TaskItem packageItem = null;
+            if (m.Success)
+            {
+                packageItem = new TaskItem(m.Groups[0].Value);
+                string name = m.Groups[1].Value;
```
weshaggard: m.Groups["name"].Value

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -1,17 +1,14 @@
...
-  
+
   <ItemGroup>
     <_PackagesDrops Include="$(PackagesDrops)" />
   </ItemGroup>
   <PropertyGroup>
-    <PackageNameRegex Condition="'$(PackageNameRegex)' == ''">(.*)\.(\d+\.\d+\.\d+)-([^-]+-\d+-\d+)</PackageNameRegex>
-    <VersionStructureRegex Condition="'$(VersionStructureRegex)' == ''">(\d+\.\d+\.\d+)-(([^-]+)-?(\d+(-\d+)?)?)</VersionStructureRegex>
-    <BuildNumberOverrideStructureRegex Condition="'$(BuildNumberOverrideStructureRegex)' == ''">([^-]+)?-?(\d+(-\d+)?)?</BuildNumberOverrideStructureRegex>
+    <PackageNameRegex Condition="'$(PackageNameRegex)' == ''">(.*)\.(\d+\.\d+\.\d+)-?([^-]+-\d+-\d+)?</PackageNameRegex>
```
weshaggard: `(?<name>.*)\.(?<version>\d+\.\d+\.\d+)(?<prerelease>-.*)?`

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -171,34 +238,33 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
...
-                                if (m.Success)
-                                {
-                                    NuGetVersion dependencyVersion = nuGetVersion;
-                                    nuGetVersion = NuGetVersion.Parse(string.Join(".", dependencyVersion.Major, dependencyVersion.Minor, dependencyVersion.Patch) + "-" + PackageBuildNumberOverride);
-                                }
+                            string releaseVersion = package.GetMetadata("ReleaseVersion");
+                            // we have package information, so use that.
+                            if (!string.IsNullOrWhiteSpace(releaseVersion))
+                            {
+                                ver = ver + "-" + releaseVersion;
```
weshaggard: remove "-" now as that should be in the prerelease group.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -229,19 +295,26 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
...
                     {
-                        NuGetVersion dependencyVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
-                        version = string.Join(".", dependencyVersion.Major, dependencyVersion.Minor, dependencyVersion.Patch);
-                        if (!string.IsNullOrWhiteSpace(PackageBuildNumberOverride))
+                        version = package.GetMetadata("Version");
+                        string releaseVersion = package.GetMetadata("ReleaseVersion");
+                        if (!string.IsNullOrWhiteSpace(releaseVersion))
                         {
-                            version += "-" + PackageBuildNumberOverride;
+                            version += "-" + releaseVersion;
```
weshaggard: remove "-" now as that should be in the prerelease group.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -229,19 +295,26 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
...
                         }
                     }
+                    else if(!string.IsNullOrWhiteSpace(dependency.GetMetadata("PackageVersion")))
+                    {
+                        version += "-" + dependency.GetMetadata("PackageVersion");
```
weshaggard: remove "-" now as that should be in the prerelease group.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -27,35 +28,34 @@ public class AddDependenciesToProjectJson : Task
...
-        // Regex for which packages to update
+        public string[] PackagesDrops { get; set; }
         [Required]
-        public string IdentityRegex { get; set; }
+        public string PackageNameRegex { get; set; }
 
-        // Permit overriding package versions found in project.json with custom build number version.
-        public string PackageBuildNumberOverride { get; set; }
+        public string[] VersionsFiles { get; set; }
+        public string DownloadFileLocation { get; set; }
```
weshaggard: delete for now.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
ericstj: I think its impossible to reconstruct name and version from the filename since I believe you can have an embedded version and dash in the pre-release portion as well as in the package ID itself.  That's why nuget cracks the zip and looks at the nuspec.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
ericstj: Just split on " " and create an override of CreatePackageItemFromString that takes the ID and version seperately.  Then use NuGet to parse the version rather than trying to roll our own parsing.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -111,14 +117,75 @@ public override bool Execute()
...
+                if (!Directory.Exists(packageDrop))
+                {
+                    Log.LogWarning("PackageDrop does not exist - '{0}'", packageDrop);
+                    continue;
+                }
+                IEnumerable<ITaskItem> packages = Directory.GetFiles(packageDrop).Select(f => new TaskItem(Path.GetFileNameWithoutExtension(f)));
+
+                foreach (ITaskItem package in packages)
+                {
+                    packageNameItems.Add(CreatePackageItemFromString(package.ItemSpec));
```
ericstj: Right way to do this is open the package and read from the NuSpec.  @emgarten I imagine NuGet has a nice API for getting ID and version out of nupkg, right?

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -64,17 +64,28 @@ public override bool Execute()
                 Log.LogError("Cannot find specified project.json - '{0}'", ProjectJson);
                 return false;
             }
-            if (!string.IsNullOrWhiteSpace(PackageBuildNumberOverride))
+
+            List<ITaskItem> packageInformation = new List<ITaskItem>();
```
weshaggard: Any reason to not make this a Dictionary<string, ITaskItem>?

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -111,14 +117,75 @@ public override bool Execute()
...
+        {
+            Match m = _packageNameRegex.Match(package);
+            TaskItem packageItem = null;
+            if (m.Success)
+            {
+                packageItem = new TaskItem(m.Groups[0].Value);
+                string name = m.Groups[1].Value;
+                packageItem.SetMetadata("Name", name);
+                packageItem.SetMetadata("Version", m.Groups[2].Value);
+                packageItem.SetMetadata("ReleaseVersion", m.Groups[3].Value);
```
weshaggard: ReleaseVersion => PrereleaseVersion.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -111,14 +117,75 @@ public override bool Execute()
...
+                if(!string.IsNullOrWhiteSpace(line))
+                {
+                    string packageVersion = line.Replace(' ', '.');
+                    packageNameItems.Add(CreatePackageItemFromString(packageVersion));
+                }
+            }
+            return packageNameItems;
+        }
+
+        private TaskItem CreatePackageItemFromString(string package)
```
weshaggard: Given that you don't look to be handing these back out of the task is there a reason you are using TaskItem here instead of a simple struct? That said I do think it might be worth handing these out at some point so I'm fine with keeping them as a TaskItem.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -171,34 +238,33 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
...
                         NuGetVersion nuGetVersion;
-                        if (externalVersion == null)
+                        if (NuGetVersion.TryParse(property.Value.ToString(), out nuGetVersion))
                         {
-                            if (NuGetVersion.TryParse(property.Value.ToString(), out nuGetVersion))
+                            NuGetVersion dependencyVersion = nuGetVersion;
+                            string ver = null;
+
+                            // a package version was provided, use its version information.
+                            if (package != null)
```
weshaggard: package should always be non-null if you are in this if block to begin with.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -171,34 +238,33 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
...
+                                ver = ver + "-" + releaseVersion;
                             }
-                        }
-                        else
-                        {
-                            nuGetVersion = NuGetVersion.Parse(externalVersion);
+                            nuGetVersion = NuGetVersion.Parse(ver);
                         }
 
                         // Only add the original dependency if it wasn't passed as an AdditionalDependency, ie. AdditionalDependencies may override dependencies in project.json
```
weshaggard: orthogonal but it might be worth creating a dictionary from the AdditionalDependencies list at some point for efficiency. If you stick with Linq there is also an Any() extension method which does exactly what you are trying to do here without needing to check for null.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: I will file an issue to do this the right way.  Looks like NuGet.Repositories may have some package parsing API's.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: I think it's still fragile in the sense that Wes believes the format in the versions files may change, in which case we should have a separate task for actually parsing that file, but your suggestion is slightly better than the custom parsing I was doing.  I'll make an update.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
emgarten: @chcosta take a look at *PackageArchiveReader* and *NuspecReader* from *NuGet.Packaging*

NuGet does some parsing on filenames also, but it is imperfect and should only be used to rule out obvious non-matches. For example packageA.1.2.3 could be either: packageA.1 version: 2.3 or packageA version: 1.2.3. The only way to know for sure is to read the nuspec.

You can see the NuGet code for this here: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol.Core.v3/Utility/LocalFolderUtility.cs#L264

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: if we parse a version via NuGet api's, then the '-' is not part of the prerelease group.  It seems to be clearer if the prerelease group only contains the version that NuGet (or someone reading the code) would understand, and that we should add the '-' as a joining character when appending the release version.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: A dictionary would be nice for lookup purposes, but we have multiple possible input sources (various package drops, or versions files), and the code is not currently written to guarantee a unique key.  I'll file an issue to investigate refactoring this, but I'll leave this as a list for now.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: Good catch, I had cleaned this up elsewhere, but missed this one.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: I also thought there may be the possibility that we pass in package taskitems for additional metadata hints in the future, but at the moment, there is no code path or use-case for that.  I thought the TaskItem struct was general purpose enough that we had more flexibility down the road.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: Yes, a dictionary would be nice, I'll add this to the cleanup issue when I file it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -8,7 +8,8 @@
     <_PackagesDrops Include="$(PackagesDrops)" />
   </ItemGroup>
   <PropertyGroup>
-    <PackageNameRegex Condition="'$(PackageNameRegex)' == ''">(.*)\.(\d+\.\d+\.\d+)-?([^-]+-\d+-\d+)?</PackageNameRegex>
+    <!-- Note: escape msbuild characters with ascii escape codes.  < == %3C, > == %3E-->
+    <PackageNameRegex Condition="'$(PackageNameRegex)' == ''">(?%3Cname%3E.*)\.(?%3Cversion%3E\d+\.\d+\.\d+)-?(?%3Cprerelease%3E.*)?</PackageNameRegex>
```
weshaggard: `(-(?%3Cprerelease%3E.*))?` is more correct for the prerelease.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
weshaggard: Please don't forget to file a clean-up issue for this.

PR #807, opened 2016-06-14
Fix for building test projects in VS and rebuildandtest on command line

karajas:
This is a fix for corefx to allow test projects that now reference the pkgprojs to build in VS.
MS.Common.Targets hardcodes GetTargetPath when resolving projectrefs in the IDE.
Currently VS(+Nuget Extensions 3.5 beta) doesn't understand the p2p reference to the test-runtime, so the nuget restore must be done via msbuild.

This also resolves rebuildandtest on commandline msbuild, by mapping it to the first usage of ProjectReference in a target which happens on both the Build and Rebuild paths.

Changes done with @ericstj 's help.

/cc @weshaggard


weshaggard:
LGTM

Issue #808, opened 2016-06-15
ConditionalTheory silently ignores a test if there are exceptions thrown by the condition properties


CIPop:
The following test theory will silently be ignored (easily missed when there are 104 tests):

```C#
        public static bool IsServerAndDomainAvailable => Crash();
            //Capability.IsDomainAvailable() && Capability.IsNegotiateServerAvailable();
        
        public static bool Crash()
        {
            throw new NotImplementedException();
        }

        public static IEnumerable<object[]> CredentialsData
        {
            get
            {
                yield return new object[] { CredentialCache.DefaultNetworkCredentials };
                yield return new object[] { new NetworkCredential(
                    Configuration.ActiveDirectoryUserName,
                    Configuration.ActiveDirectoryUserPassword,
                    Configuration.ActiveDirectoryName) };
            }
        }

        [OuterLoop]
        [ConditionalTheory(nameof(IsServerAndDomainAvailable))]
        [MemberData(nameof(CredentialsData))]
        public async Task NegotiateStream_KerberosAuthenticationRemote_Success(object credentialObject)
```

Output:
```
 === TEST EXECUTION SUMMARY ===
     System.Net.Security.Tests  Total: 103, Errors: 0, Failed: 0, Skipped: 0, Time: 2.859s
  Finished running tests.  End time=17:47:33.63, Exit code = 0
```

/cc @roncain @davidsh

PR #809, opened 2016-06-15
Fail ConditionalFacts/Theories with throwing conditions

stephentoub:
If the condition member throws an exception, we should treat it as not meeting the condition so that we mark the test(s) as skipped.

Fixes https://github.com/dotnet/buildtools/issues/808
cc: @cipop, @Priya91, @roncain


CIPop:
Thanks @stephentoub ! (Wow that was fast!)

LGTM

bartonjs:
LGTM2

PR #812, opened 2016-06-15
Changed writers to output "throw null;" in empty method bodies

KrzysztofCwalina:
We had a discussion about this offline. throw null results in simpler source and smaller IL.


KrzysztofCwalina:
@weshaggard, could you take a look?

weshaggard:
LGTM

PR #813, opened 2016-06-15
Removing toolruntime/project.json and moving the dependencies to the buildtools package instead.

joperezr:
This PR:

- Removes tool-runtime/project.json
- Moves references to the Buildtools package instead
- Restructures the buildtools package to follow nuget conventions and make it deployable.
- changes in init-tools scripts since now the workflow will change a bit.

cc: @weshaggard @ericstj


joperezr:
FYI: @dagood

joperezr:
fixes: #413 #487

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -1,16 +1,15 @@
-@echo off
+@if "%_echo%" neq "on" echo off
 setlocal
 
 set PROJECT_DIR=%~1
 set DOTNET_CMD=%~2
 set TOOLRUNTIME_DIR=%~3
-set PACKAGES_DIR=%4
+set PROJECT_JSON_FILE=%4
```
weshaggard: I don't want to have to change the arguments we pass to init-tools.cmd as it will break all current usages. I thought we were still going to use the toolruntime\project.json for this publishing and not the actual buildtools package.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -15,30 +15,52 @@
...
+        <dependency id="Microsoft.Build.Targets" version="[0.1.0-preview-00024-160610, )" />
+        <dependency id="Microsoft.Build" version="[0.1.0-preview-00024-160610, )" />
+        <dependency id="Microsoft.Net.Compilers.NetCore" version="[1.3.0-beta1-20160429-01, )" />
+        <dependency id="Microsoft.Net.Compilers.Targets.NetCore" version="[0.1.5-dev, )" />
+        <dependency id="Microsoft.Cci" version="[4.0.0-rc3-24128-00, )" />
+        <dependency id="Microsoft.Composition" version="[1.0.30, )" />
+        <dependency id="System.Diagnostics.TextWriterTraceListener" version="[4.0.0-rc3-24128-00, )" />
+        <dependency id="System.Runtime.Serialization.Primitives" version="[4.1.1-rc3-24128-00, )" />
+        <dependency id="Newtonsoft.Json" version="[7.0.1, )" />
+        <dependency id="NETStandard.Library" version="[1.6.0-rc3-24128-00, )" exclude="runtime" />
```
weshaggard: Why do we need this dependency if we are actually excluding the runtime? I understood why it was needed for the tool-runtime but I don't understand why it is needed here.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: ok, so I'll remove the net45 section of the dependencies and use a new project.json for that.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
joperezr: mm I actually didn't actually verified this, but my theory is that if I don't add this, then back in corefx side when we publish the project.json that contains this package, it will try to restore older version of the framework libraries (since packages like MSBUILD and the compilers depend on them older versions of the framework). That is why if I add this one here, then I can ensure that the libraries will have rc3-24128-00 version.

weshaggard: OK if that is true then I'm fine with this.

joperezr:
Closing this very old PR. I think this should still eventually be done, so whoever wants to pick this up, feel free to take the code I did as a base for your changes.

weshaggard:
cc @crummel @markwilkie

PR #815, opened 2016-06-15
Copy generated core files on Linux to a location that won't be deleted after the run

adityamandaleeka:
Turns out we don't have GDB installed on the CI machines yet, so the strategy of printing stacks from core files and deleting them isn't going to work. Instead, for now, let's copy the core files to a known location that won't be wiped out right after the run.

Like on OS X, we'll only store one core file there at a time in order to avoid filling up the disk.

@ellismg @sergiy-k PTAL


adityamandaleeka:
@dotnet-bot test this please

adityamandaleeka:
@dotnet-bot test  Pre-Merge signing tool

sergiy-k:
LGTM

PR #816, opened 2016-06-15
Filter nupkg assets that overlap with projectrefs

ericstj:
We weren't filtering items that came from pkgproj refs.  Those items are
directly added to the reference item today, so filter that item.

I could have made the pkgproj references behave more like project
references and have them add their output to
_ResolvedProjectReferencePaths, but this would bypass RAR and be less
like an actual package reference.

/cc @karajas @weshaggard


karajas:
LGTM.

PR #817, opened 2016-06-16
Include Unix binaries when testing with local bits

ericstj:
/cc @weshaggard


weshaggard:
LGTM... we so need to clean-up this copying logic :(

ericstj:
Yeah, I hate all the layers of overrides.  This still didn't fix the problem for me so I'm digging deeper.

ericstj:
Here's the real fix: https://github.com/dotnet/buildtools/pull/822

PR #818, opened 2016-06-16
Make Tool Host Overridable for Internal Builds

joperezr:
cc: @weshaggard 

We need to make the Tool Host overridable because internal builds have to run it from a different location.


joperezr:
Merging this now as I need it to unblock internal builds. I'll address any feedback that comes out of this on my next PR.

weshaggard:
LGTM - I would like to understand why this wasn't needed before.

PR #822, opened 2016-06-16
TestWithLocalLibraries should use TargetOS

ericstj:
Previously this was using OSGroup which is the OS the test is building
for not the OS the test is running against.  TargetOS is what the test
is running against.

/cc @weshaggard @karajas

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -92,17 +92,17 @@
       <BaseLocalLibrariesPath Condition="'$(BaseLocalLibrariesPath)'==''">$(BaseOutputPath)</BaseLocalLibrariesPath>
       <AnyOSLocalLibrariesPath>$(BaseLocalLibrariesPath)/AnyOS.$(Platform).$(ConfigurationGroup)</AnyOSLocalLibrariesPath>
       <UnixLocalLibrariesPath>$(BaseLocalLibrariesPath)/Unix.$(Platform).$(ConfigurationGroup)</UnixLocalLibrariesPath>
-      <OSGroupLocalLibrariesPath>$(BaseLocalLibrariesPath)/$(OSGroup).$(Platform).$(ConfigurationGroup)</OSGroupLocalLibrariesPath>
+      <TargetOSLocalLibrariesPath>$(BaseLocalLibrariesPath)/$(TargetOS).$(Platform).$(ConfigurationGroup)</TargetOSLocalLibrariesPath>
     </PropertyGroup>
     <ItemGroup Condition="'$(TestWithLocalLibraries)'=='true'">
       <!-- Replace some of the resolved libraries that came from nuget by exploring the list of files that we are going to copy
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -92,17 +92,17 @@
...
       <!-- Replace some of the resolved libraries that came from nuget by exploring the list of files that we are going to copy
            and replacing them with local copies that were just built -->
       <_ReplacementCandidates Include="@(TestCopyLocal -> '$(AnyOSLocalLibrariesPath)/%(filename)/%(filename).dll')" />
       <_ReplacementCandidates Include="@(TestCopyLocal -> '$(AnyOSLocalLibrariesPath)/%(filename)/%(filename).pdb')" />
-      <_ReplacementCandidates Condition="'$(TargetsUnix)' == 'true'" Include="@(TestCopyLocal -> '$(UnixLocalLibrariesPath)/%(filename)/%(filename).dll')" />
-      <_ReplacementCandidates Condition="'$(TargetsUnix)' == 'true'" Include="@(TestCopyLocal -> '$(UnixLocalLibrariesPath)/%(filename)/%(filename).pdb')" />
-      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(OSGroupLocalLibrariesPath)/%(filename)/%(filename).dll')" />
-      <_ReplacementCandidates Include="@(TestCopyLocal -> '$(OSGroupLocalLibrariesPath)/%(filename)/%(filename).pdb')" />
+      <_ReplacementCandidates Condition="'$(TargetOS)' == 'Linux' OR '$(TargetOS)' == 'OSX'" Include="@(TestCopyLocal -> '$(UnixLocalLibrariesPath)/%(filename)/%(filename).dll')" />
```
bartonjs: What about FreeBSD/NetBSD?  Don't they set TargetsUnix, but won't satisfy this condition now?

weshaggard: Yes is there a reason you cannot still use TargetsUnix here?

weshaggard: Sorry nevermind it is because you are checking the TargetOS and TargetsUnix is only using OSGroup.

weshaggard:
I must admit I don't know if that is enough or not because the file is so complex but I trust you have tested it so I'm good with the change.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
bartonjs: Ugh, well, that's a reason.  Since FreeBSD/NetBSD are probably broken right now anyways (since Linux is) I'd rather see the pull request merged as-is than block on perfection.

bartonjs:
Similar to Wes, I don't know enough to know if this is complete; but aside from the alternate OSes issue I don't see anything wrong :)

ericstj:
I've tested on Jenkins and my Ubuntu VM after I was able to finally get a repro and it worked for me.

PR #826, opened 2016-06-17
Use InvariantCulture for writing constants in MS.CCI.Extensions

akoeplinger:
Before it'd use the current culture when writing the constant value, which results in float/double
using `,` as the decimal separator in cultures where this is the default.

The generated code is invalid C# and doesn't compile. Fixing the problem by using culture-neutral formatting.

Sample code:

```
public class TestLib
{
    public const double PI = 3.141592;
}
```

### DIFF
--- a/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
+++ b/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
@@ -269,7 +269,7 @@ namespace Microsoft.Cci.Writers.CSharp
                 else if (double.IsNaN(val))
                     Write("0.0 / 0.0");
                 else
-                    Write(((double)value).ToString("R"));
+                    Write(((double)value).ToString("R", CultureInfo.InvariantCulture));
             }
             else if (value is float)
             {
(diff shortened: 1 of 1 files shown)


stephentoub:
LGTM

akoeplinger:
```
07:36:06   Restoring all packages...
07:37:43 EXEC : error : Unable to resolve 'Microsoft.NETCore.ConsoleHost (>= 1.0.0-rc2-24027)' for '.NETCoreApp,Version=v1.0'. [D:\j\workspace\innerloop_prtest4ca7949f\build.proj]
07:37:44 EXEC : error : The given key was not present in the dictionary. [D:\j\workspace\innerloop_prtest4ca7949f\build.proj]
07:37:44 EXEC : error : Unable to resolve 'Microsoft.NETCore.ConsoleHost (>= 1.0.0-rc2-24027)' for '.NETCoreApp,Version=v1.0'. [D:\j\workspace\innerloop_prtest4ca7949f\build.proj]
```

I hit that locally as well (had to pull the package from somewhere else). Something up with the feed?

bartonjs:
@dagood FYI (about the package restore failure)

dagood:
@bartonjs Ah, didn't see that this was happening in CI, the internal package storage workaround isn't so useful in this case.

The feed is fine, but recently the ConsoleHost package (all versions) was removed. @weshaggard Could you expain why the purge was needed?

bartonjs:
Test Innerloop Windows Debug please  (package issue should be resolved)

weshaggard:
LGTM

PR #829, opened 2016-06-17
Add file with correlation id to test drop

AlfredoMS:
To be able to track the test progress of tests in Helix we need the correlation id. Currently the only way to read it is during build time, when executing this target. This change uploads a file that contains the correlation id to the same place as the test drop builds are placed making it accessible to retrieve at a later time.

@chcosta @jhendrixMSFT

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
@@ -378,5 +378,24 @@
       EventHubSharedAccessKeyName="$(BuildIsOfficialEventHubSharedAccessKeyName)"
       EventHubSharedAccessKey="$(BuildIsOfficialEventHubSharedAccessKey)"
       EventData="%(TestListFile.OfficialBuildJson)"/>
+    <!-- Upload the Correlation Id to the test drop container for tracking purposes-->
+    <PropertyGroup>
+      <CorrelationFileName>correlation_id-$(ConfigurationGroup)-$(TestNugetRuntimeId).txt</CorrelationFileName>
+    </PropertyGroup>
+    <ItemGroup>
+      <CorrelationFile Include="$(TestArchivesRoot)$(CorrelationFileName)">
+        <RelativeBlobPath>Tracking/$(CorrelationFileName)</RelativeBlobPath>
+      </CorrelationFile>
+    </ItemGroup>
+    <WriteLinesToFile
+        File="@(CorrelationFile)"
+        Lines="$(GeneratedCorrelationId)"
+        Overwrite="true" />
+    <UploadToAzure Condition=" '$(HelixApiAccessKey)' != '' "
+        AccountKey="$(CloudDropAccessToken)"
+        AccountName="$(CloudDropAccountName)"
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -378,5 +378,24 @@
       EventHubSharedAccessKeyName="$(BuildIsOfficialEventHubSharedAccessKeyName)"
       EventHubSharedAccessKey="$(BuildIsOfficialEventHubSharedAccessKey)"
       EventData="%(TestListFile.OfficialBuildJson)"/>
+    <!-- Upload the Correlation Id to the test drop container for tracking purposes-->
+    <PropertyGroup>
+      <CorrelationFileName>correlation_id-$(ConfigurationGroup)-$(TestNugetRuntimeId).txt</CorrelationFileName>
```
chcosta: Do you want to throw an error or skip this logic if TestNugetRuntimeId is not set?

chcosta: Also, is this required everytime SendCompletionEvent is called?  Should this be a separate target or is it intrinsic to SendCompletionEvent (I don't know how SendCompletionEvent is used everywhere)

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -378,5 +378,24 @@
...
+    <ItemGroup>
+      <CorrelationFile Include="$(TestArchivesRoot)$(CorrelationFileName)">
+        <RelativeBlobPath>Tracking/$(CorrelationFileName)</RelativeBlobPath>
+      </CorrelationFile>
+    </ItemGroup>
+    <WriteLinesToFile
+        File="@(CorrelationFile)"
+        Lines="$(GeneratedCorrelationId)"
+        Overwrite="true" />
+    <UploadToAzure Condition=" '$(HelixApiAccessKey)' != '' "
```
chcosta: Have you considered that there may be collisions for this given that multiple builds ( > 10) all specify the same TestNuGetRuntimeId?

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
AlfredoMS: Yes this functionality depends on outputs (correlation id) generated within this Target, so it makes sense to keep it here. It is similar to the steps above the ones I added where it sends some information to an event hub.


Also, if TestNugetRuntimeId is not set, nothing bad would happen. However for our builds we always set it.

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
AlfredoMS: The container name has a suffix that when combined with the TestNugetRuntimeId, makes it unique.

In case the same build configuration is triggered a second time, that is fine, because we only need the latest correlation id.

chcosta:
LGTM

PR #836, opened 2016-06-22
Set coredump_filter to include shared library code.

adityamandaleeka:
This change makes it so the core files generated in CI test runs on Linux will include memory in the shared libraries we load (such as libcoreclr.so). We didn't need it when the plan was just to print stacks when there was a crash in CI, but now that we're actually keeping the dumps sometimes, it's obviously useful for debugging purposes.

@sergiy-k @ellismg PTAL

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
@@ -95,7 +95,15 @@ if [ "$(uname -s)" == "Darwin" ]; then
 elif [ "$(uname -s)" == "Linux" ]; then
   # On Linux, we'll enable core file generation unconditionally, and if a dump
   # is generated, we will print some useful information from it and delete the
-  # dump immediately. 
+  # dump immediately.
+
+  if [ -e /proc/self/coredump_filter ]; then
+      # Include memory in private and shared file-backed mappings in the dump.
+      # This ensures that we can see disassembly from our shared libraries when
+      # inspecting the contents of the dump. See 'man core' for details.
+      echo 0x3F > /proc/self/coredump_filter
+  fi
+
   ulimit -c unlimited
 fi
 # ========================= END Core File Setup ==============================


src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
```diff
@@ -95,7 +95,15 @@ if [ "$(uname -s)" == "Darwin" ]; then
 elif [ "$(uname -s)" == "Linux" ]; then
   # On Linux, we'll enable core file generation unconditionally, and if a dump
   # is generated, we will print some useful information from it and delete the
-  # dump immediately. 
+  # dump immediately.
+
+  if [ -e /proc/self/coredump_filter ]; then
```
weshaggard: Is there a better place to do this? I want to try and keep these test scripts to a minimum and avoid most of these sort of hacks.

weshaggard: This question goes for your earlier core_dump stuff in general https://github.com/adityamandaleeka/buildtools/commit/904d081f00b67ed1b18060b85742a727738f3c99. 

I like the idea of capturing dumps but is there a better place then the runner scripts to do it?

adityamandaleeka: @ellismg recommended putting the core dump logic here. I'm not too familiar with this codebase, but it seems like this is a good place to set up dump generation because it will be enabled only for the shell that this script runs in (which is pretty close to the minimum window we need it for).

Do you have another place in mind that would be more suitable?

ellismg: > I like the idea of capturing dumps but is there a better place then the runner scripts to do it?

The other concern that lead us to this solution was the desire to minimize the number of dumps captured in cases where the entire run-time had problems.  There was fear about filling up the machines with dumps, so the tests happen right before we run tests based on if any other dump is present.

Long term, we would like dumps to be collected anywhere we run tests, I would think.  Do you have a thought about how we would centralize this infrastructure across Jenkins, Helix and local dev boxes?

weshaggard: I'm good with us capturing dumps as part of the runner scripts so it is consistent across the places we run them but I start to worry when we start introducing hacks into them and other things that is unique to just one of the environments.

Also as far as I can tell we are only doing the dumps on unix right now are there plans to add similar dump capturing on windows?

weshaggard: I didn't dig in but we also need to be careful about which tools we start taking a dependency on to capture these dumps. In my ideal world the dump capturing would be handled by the host or a wrapper around the host itself, as opposed to a bunch more code in these scripts.

adityamandaleeka: > Also as far as I can tell we are only doing the dumps on unix right now are there plans to add similar dump capturing on windows?

Dumps on Windows have been enabled via a registry change in the Server 2012 CI image (@mmitche did this recently).

weshaggard: That would still make collecting dumps inconsistent between the various places we run these tests.

adityamandaleeka:
@weshaggard @ellismg If it's okay with you, I'd like to get this change in. There's already stuff in this template for enabling core dumps, and this PR just makes it so those dumps are more useful. Moving all of this logic to a better location is worth looking into, but I don't think it needs to block this PR, especially as we're actively trying to investigate the issue(s) causing segfaults in CoreFX runs.

ellismg:
This sounds good to me.

weshaggard:
@adityamandaleeka I don't have any issues with you merging this but do we have any work items tracking how we want to handle dumps across the various scenarios?.

PR #839, opened 2016-06-23
Allow override of __PUBLISH_RID

mellinoe:
* Also, fall back to Ubuntu 14.04 if an unsupported Ubuntu
  version is encountered.

Similar change as in https://github.com/dotnet/corefx/pull/9646. This lets you override the RID that the tool-runtime project is published for. It also adds a fallback from unsupported Ubuntu versions to 14.04.

@ellismg , @joperezr


ellismg:
My only question is if you feel it's worthwhile to prefix the stuff we pull from the environment with BUILDTOOLS or something.

Your call, LGTM either way.

joperezr:
LGTM

mellinoe:
I don't feel strongly about the name, I think it will be super rare to actually override this, it's just good to have the option. Gonna merge now.

PR #843, opened 2016-06-28
Don't run build on pkgproj references

ericstj:
We'll want to be able to test without building a package.

We're thinking to control package builds for the repo via the
packages.builds file and we don't want a test build to trigger
nupkg build for something we aren't shipping.

/cc @weshaggard @karajas


weshaggard:
LGTM

ericstj:
Running some full build tests on CoreFx then I will merge.

PR #847, opened 2016-06-29
EnsureOOBFramework should promote impl deps

ericstj:
EnsureOOBFramework makes sure that an oob implementation will always
take precedence over a placeholder in the case we have a framework that
transitioned from targeting-pack to package as was the case with
netcore50.

The problem here is that if it promotes only one of the ref or lib we'll
only get the dependencies from that asset, since the other asset's
dependencies will be in the less specific dependency group.

This became a problem when we started merging runtime specific
assets into the same packages as the ref: the fat package work.
Since a netstandard lib that is rid specific will be preferred over the
placeholder that is not rid specific we didn't need to copy the
implementation.   The result was the dependencies would only
be in the netstandard group.

To fix this I added a phase to find the actual implementation and copy
it to a 'netcore50' framework item.  I left a marker in the targetpath
of the copied items so that we wouldn't package the duplicate.

Longer term I'd like to rewrite how we do all the dependency harvesting
so that it is less brittle.  I think the right way to do this would be
to do actuall asset selection for TFM + TFM/RID tuples as appropriate
for all supported app models, get the actual assets that nuget will
select and harvest the dependencies at that point.

Needed for https://github.com/dotnet/corefx/issues/9743

/cc @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
@@ -193,6 +193,7 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
         private List<ManifestFile> GetManifestFiles()
         {
             return (from f in Files.NullAsEmpty()
+                    where !f.GetMetadata(Metadata.FileTarget).StartsWith("$none$", StringComparison.OrdinalIgnoreCase)
                     select new ManifestFile()
                     {
                         Source = f.GetMetadata(Metadata.FileSource),
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/EnsureOOBFramework.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/EnsureOOBFramework.cs
@@ -20,7 +20,7 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
     public class EnsureOOBFramework : PackagingTask
     {
         [Required]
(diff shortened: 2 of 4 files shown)


ericstj:
See https://gist.github.com/ericstj/188a8bae5ecf362fb245c80ee61a987c/revisions for the diff of CoreFx after fixing this.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
```diff
@@ -193,6 +193,7 @@ private Manifest CreateManifest()
         private List<ManifestFile> GetManifestFiles()
         {
             return (from f in Files.NullAsEmpty()
+                    where !f.GetMetadata(Metadata.FileTarget).StartsWith("none", StringComparison.OrdinalIgnoreCase)
```
weshaggard: `<none>`

weshaggard:
:+1:

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
ericstj: Sure, nice that it is an invalid path so no collision with actual usage.

ericstj: Angle brackets don't work since our path APIs reject them.  I used $'s since that'll be less likely to collide with any use case.

ericstj:
@dotnet-bot test this please

Issue #823, opened 2016-06-16
Move UpdatePublishedVersions.ps1 to buildtools


dagood:
`UpdatePublishedVersions.ps1` is in a few places:
* https://github.com/dotnet/corefx/blob/master/UpdatePublishedVersions.ps1
* https://github.com/dotnet/coreclr/blob/master/UpdatePublishedVersions.ps1
* The TFS build auto-publisher
* Soon, WCF

There isn't any big reason to differentiate the file between repos, so it should be centralized.

Also consider centralizing `UpdateDependencies.ps1` (corefx, coreclr, wcf) at the same time.

/cc @weshaggard @jhendrixMSFT @ericstj @eerhardt

weshaggard:
agreed thanks for filing this issue.

Issue #835, opened 2016-06-21
Make UpdateDependencies.ps1 update all in one PR, updating existing PR if possible


dagood:
Instead of opening separate PRs for each dependency group in the versions repo, make a single PR with all upgrades necessary to make dependency state current. This:
* Fixes the merge conflict problem.
* Reduces PR noise.
* Reduces CI strain.
* Simplifies the versions repo subscription.

To help even more, we should update an existing dependency upgrade PR rather than opening a new one if possible. We could have the bot do a force push (to keep number of commits down) or push a new commit (to keep intermediate CI results easily visible). Either way sounds good to me.

(Depends on https://github.com/dotnet/buildtools/issues/823 bringing the script into buildtools.)

Related earlier discussion about this: https://github.com/dotnet/corefx/issues/9156

weshaggard:
I'm all for it.

Issue #732, opened 2016-05-18
Support dependency validation and upgrade using full package versions


dagood:
Now that we have a versions repository that lists every package id and version (https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt) we might want to change from validating and updating dependencies based on prerelease to validating full versions. This solves a few problems:

1. We would no longer need to maintain regexes to match package ids that should follow certain validation rules. (Specifically, this could be a problem if some packages come from different builds.)
2. We wouldn't automatically upgrade a prerelease without upgrading the minor/patch version anymore. For example, if `1.0.0-rc2-01` becomes `1.0.1-rc2-02`, our current upgrading system would upgrade to `1.0.0-rc2-02`, breaking things. We currently make sure this doesn't happen in a post-restore step, but upgrading based on full package version would remove this problem.

This does mean we would have to keep the current version of every dependency checked into a central file in corefx, wcf, and coreclr rather than having a single property for prerelease. It would be great to pair this change up with one that lets us centralize the *actual* dependencies being used, where projects somehow get their versions from a single source.

With the central file, an upgrade can check the latest version on the versions repo and see which versions need to be updated. The central file is updated as well as any project.jsons. Then the dependency validation step just checks project.jsons for consistency with that central file.

/cc @weshaggard @ericstj @jhendrixMSFT @chcosta @eerhardt

weshaggard:
>This does mean we would have to keep the current version of every dependency checked into a central file in corefx, wcf, and coreclr rather than having a single property for prerelease. It would be great to pair this change up with one that lets us centralize the actual dependencies being used, where projects somehow get their versions from a single source.

I don't think we actually need to check that in do we? I was hoping we might just be able to point to the exact package list via it's sha via the github url. Is there any reason that wouldn't work?

dagood:
It would add an extra network dependency during a `sync /p` to validate (maybe just the first `sync /p` if we cache the result), but other than that I like that idea, didn't think of it.

weshaggard:
Yes I think we could pull it down and cache it.

dagood:
For the validation part of this, I'm not sure what to do about our "misc" packages: https://github.com/dotnet/corefx/blob/604ed96/dir.props#L115-L134. Maybe checking to see if all project.jsons have the same version of non-auto-upgraded packages is enough?

weshaggard:
I think it would be reasonable to just point to a package and version list that is checked into the repo for these packages.

PR #848, opened 2016-06-29
Add dependency auto-upgrade scripts to share in dotnet projects

dagood:
`UpdatePublishedVersions` is from CoreFX/CoreCLR/WCF (identical). `UpdateDependencies` is from CoreFX with a small amount of added configurability for WCF and CoreCLR: make the `build.cmd` and PR both opt-in.

This is step 1 in https://github.com/dotnet/buildtools/issues/823. Next is switching to use this script from `Tools\...` in the build pipeline and Maestro build definitions. (Nothing will break in the mean time because this is additive.)

/cc @weshaggard @jhendrixMSFT @ericstj @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/versions/UpdateDependencies.ps1
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/versions/UpdateDependencies.ps1
@@ -1,0 +1,142 @@
+﻿#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+# This script updates the dir.props file with the latest build version,
+# optionally runs the project.json auto-upgrade target, and then optionally
+# creates a GitHub Pull Request for the change.
+
+param(
+    [Parameter(Mandatory=$true)][string]$VersionFileUrl,
+    [Parameter(Mandatory=$true)][string[]]$DirPropsVersionElements,
+    [string]$RepositoryRoot = '.',
+
+    [string]$GitHubUser,
+    [string]$GitHubEmail,
+    [string]$GitHubPassword,
+    [string]$GitHubOriginOwner=$GitHubUser,
+    [string]$GitHubUpstreamOwner='dotnet', 
+    [string]$GitHubUpstreamBranch='master',
+    [string]$GitHubProject,
+    # a semi-colon delimited list of GitHub users to notify on the PR
+    [string]$GitHubPullRequestNotifications='',
+
+    # Run build.cmd to apply the version upgrade to project.json files.
+    [switch]$UpdateInvalidPackageVersions,
+    # Create a pull request based on the GitHub parameters.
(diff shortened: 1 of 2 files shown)


eerhardt:
This may be just my personal preference on programming languages, but I like the C# versions of these scripts in the dotnet/cli repo better than these .ps1 scripts.

UpdateDependencies:  https://github.com/dotnet/cli/tree/rel/1.0.0/build_projects/update-dependencies
UpdatePublishedVesrions:  https://github.com/dotnet/cli/blob/rel/1.0.0/build_projects/shared-build-targets-utils/VersionRepoUpdater.cs

I think it will be easier to maintain and enhance these actions if they were written in C#.  For example, the C# code already uses the _Packages.txt files in `dotnet/versions`, and when we move to _BuildInfo.json files, it will be even easier in C# than in .ps1, IMO.

In order to make sharing across repos possible, we could do the following:

1. Start with the CLI's C# code and make a .csproj/.dll in buildtools with decent APIs.
2. Write a wrapping MSBuild custom task around the APIs, so now the code can be invoked either through MSBuild, or by referencing the .dll and calling the APIs.
3. Write a .targets file in buildtools that invokes the custom task.
4. In each consuming repo (corefx, coreclr, wcf, etc), the script to invoke these:
 1. Pulls down buildtools
 2. Executes `msbuild.exe dependencies.proj /t:UpdateDependencies` (or whatever the .proj should be named). 
 3. The .proj file can have all the parameters/properties in it that the Task consumes (repo root, upstream branch, etc.) and any "Secrets" can be passed on the command line `/p:GitHubPassword=foo`.
5. If other repos (Roslyn, NuGet, MSBuild, etc.) don't want to use the MSBuild .targets file, they can just reference the .dll and invoke the code however they see fit.
6. This .dll is a great place to put the APIs that produce and consume the _BuildInfo.json files that we were discussing in email last week.

Thoughts?

dagood:
@eerhardt I agree with C# and making this a lib + build task, especially with more complexity on the way for https://github.com/dotnet/buildtools/issues/835 (upgrade all in single PR) and https://github.com/dotnet/buildtools/issues/732 (upgrade/validate based on full versions). Currently we do validation of the dependencies during builds which I'm pretty sure CLI doesn't, but that's just something I can add on.

For a bit I was thinking that we could use these scripts as the entry point to the msbuild tooling in the future, but I really don't think that has any benefit over calling `build.cmd /t:UpdateDependencies /p:GitHubUser=Foo;GitHubPassword=Bar...`. (I actually like the removal of that indirection that mostly just adds another language in the mix.) Closing!

weshaggard:
I like @eerhardt idea as well. In general I want to keep as much code as possible out of scripts (cmd/sh/ps1 etc) and in C# or the native build engine we are using to do our work. With @maririos work with the run command we will be able to easily create verbs/commands/alias for these updates as well which will make them more discoverable and self-documenting.

maririos:
It got merged yesterday so it is pretty new. Is not in use yet though, but you can take a look at PR #672

Issue #845, opened 2016-06-29
Include sha in build tools packages


weshaggard:
For servicing and other events we need to include the sha for the commit where we built the build tools package for incase we need to make a targeted fix for the tools in a servicing event. 

cc @markwilkie

PR #849, opened 2016-06-30
Add version info to buildtools packages

chcosta:
Address issue https://github.com/dotnet/buildtools/issues/845

Adds a .version file into the package root.  CLI adds the .versions file into the lib targetpath though it's not clear to me what the benefit there is.

/cc @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -16,4 +16,7 @@
     <CompilerResponseFile Condition="'$(CheckSumSHA256)'!='false'">$(MSBuildThisFileDirectory)checksum.rsp</CompilerResponseFile>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <GetNuGetPackageVersionsDependsOn>CreateVersionFileDuringBuild</GetNuGetPackageVersionsDependsOn>
+  </PropertyGroup>
 </Project>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -215,6 +215,22 @@
     </PropertyGroup>
   </Target>
 
+  <PropertyGroup>
+    <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
+  </PropertyGroup>
+  
+  <Target Name="CreateVersionInfoFile"
+          DependsOnTargets="GetLatestCommitHash">
+    <PropertyGroup>
+      <SyncInfoFile>$(PackagesBasePath)\version.txt</SyncInfoFile>
+    </PropertyGroup>
+    <WriteLinesToFile
+      Condition="'$(LatestCommit)' != ''"
+      File="$(SyncInfoFile)"
+      Lines="$(LatestCommit)"
+      Overwrite="true" />
+  </Target>
+
   <!-- #################################### -->
   <!-- Generate BuildVersion.props -->
   <!-- #################################### -->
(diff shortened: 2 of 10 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/annotatenuspec/AddVersionInfo.targets
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
chcosta: The folder name for these files is a remnant from a previous direction I was going with this.  I'll change the folder name to something more aligned with these files' function.

chcosta: And I'll update the file names

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/annotatenuspec/addversioninfo.cmd
```diff
@@ -0,0 +1,30 @@
+@IF "%_echo%" NEQ "on" ECHO OFF
```
weshaggard: We already have a target that gets the commit hash we should reuse that (https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L207) in fact we should even consider having the versioning.targets file generate this .version file.

chcosta:
Thanks for the hint Wes, I wasn't aware of that target.  

I refactored things, and force pushed my updated branch. 

Unfortunately, I see that the builds are failing, and this makes sense because version.txt won't get produced until we update the version of buildtools which buildtools is producing.  I'll remove the updated nuspecs from this PR, so we can focus on the internals of the change. I'll file an issue so that we can update the nuspec files as soon as buildtools' buildtools version is updated.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -215,6 +215,22 @@
...
   </Target>
 
+  <PropertyGroup>
+    <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
+  </PropertyGroup>
+  
+  <Target Name="CreateVersionInfoFile"
+          DependsOnTargets="GetLatestCommitHash">
+    <PropertyGroup>
+      <SyncInfoFile>$(PackagesBasePath)\version.txt</SyncInfoFile>
```
weshaggard: Did we want to call `version.txt` or `.version`? I personally don't have a preference just curious.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
chcosta: Originally, I called it ".version", because it appeared to me that was what the CLI team was doing.  Then I decided that I didn't want to potentially collide with their file.  In addition, I'm fairly confident that devs file associations are properly setup for opening a text file, so they don't have to extract the file first if they decide they want to open the file in some other program whereas I don't know what devs do or would do with a .version file (it's probably not setup for any file association, which is fine).  It also, corresponded with the type of file we have for the eula agreement files.  All of that made me decide it was preferable to use a .txt file extension vs something else.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -16,4 +16,7 @@
     <CompilerResponseFile Condition="'$(CheckSumSHA256)'!='false'">$(MSBuildThisFileDirectory)checksum.rsp</CompilerResponseFile>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <GetNuGetPackageVersionsDependsOn>CreateVersionFileDuringBuild</GetNuGetPackageVersionsDependsOn>
```
joperezr: Not sure why we wanted to do this

chcosta: I only want a dependency on createversioninfofile if versioning.targets is imported.

PR #850, opened 2016-06-30
Onboarding the run command in build tools

maririos:
Updates the Build Tools package version to 1.0.26-prerelease-00530-02 to consume the Run Tool. 
It modifies and adds the dev workflow scripts sync.cmd, clean.cmd and build.cmd to work running the Run Tool.

cc: @weshaggard  @joperezr @chcosta


run.cmd
```diff
@@ -0,0 +1,30 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+if not defined VisualStudioVersion (
```
weshaggard: At some other point we might consider putting this logic in msbuild.cmd instead of run.cmd, something to consider later.

src/common/test-runtime/project.json
```diff
@@ -0,0 +1,43 @@
+{
+  "dependencies": {
```
weshaggard: Why is this needed?

weshaggard:
@dotnet-bot test this please

run.cmd
```diff
@@ -0,0 +1,30 @@
...
+:: Restore the Tools directory
+call %~dp0init-tools.cmd
+if NOT [%ERRORLEVEL%]==[0] (
+  exit /b 1
+)
+
+set _toolRuntime=%~dp0Tools
+set _dotnet=%_toolRuntime%\dotnetcli\dotnet.exe
+
+echo Running: %_dotnet% %_toolRuntime%\run.exe %*
```
joperezr: Just a small nit here, one way to avoid having to have the command you are running twice every time, you may just toggle echo on/off before and after the command such that if you need to change the command in the future, the change just goes in one place.

clean.cmd
```diff
@@ -0,0 +1,177 @@
+@if "%_echo%" neq "on" echo off
```
joperezr: I'm not sure why this script is not as straight forward as build.cmd. Ideally you would want run.exe to do all of the param parsing. I get how Tools won't be deleted this way, but maybe we should just handle that small special case instead of all of them?

joperezr:

src/common/test-runtime/project.json
joperezr: I believe this is because she is updating to the new buildtools which now has a different way of building dependencies for xunit. This is basically equivalent to what corefx has for it's test projects.

---
In reply to: [69196894](https://github.com/dotnet/buildtools/pull/850#discussion_r69196894) [](ancestors = 69196894)

config.json
```diff
@@ -127,9 +127,7 @@
         "binaries":{
           "description": "Only builds binaries. It doesn't restore packages.",
           "settings":{
-            "RestoreDuringBuild": false,
-            "BuildPackages": false,
-            "BuildTests": false
+            "RestoreDuringBuild": false
```
joperezr: Seems like this change is not particular to the rest of the work you are doing. I wonder if it's better to have it in it's own PR to better explain why do we need this change.

src/common/test-runtime/project.json
weshaggard: I agree with this but I don't see any usage of this in the test projects.

maririos: Without this the project `src\Microsoft.DotNet.Build.Tasks.Packaging\test` fails to restore because it is unable to resolve `test-runtime`

config.json
maririos: I'm making the correspondent changes to build tools in order to use the run tool. This change doesn't change the behavior of build, but cleans the config.json file to use the build options that we have in build tools.
I could create a different PR for it, but I'm not sure if it is worth it.

clean.cmd
maririos: I'll do some clean up.

weshaggard: We definitely should clean this up more but some parts, in particular the tools dir, have to be treated specially. I would like to see this script eventually come to only support one option which would be `all` and for anything else it just calls run with the other arguments.

src/common/test-runtime/project.json
joperezr: right, Packaging tests use to reference the one inside Tools dir, but that one  is outdated which is why I believe she is mimicking what we have in corefx where we have now our own copy of it that we know works with it. Also, this is the reason why she changed global.json

clean.cmd
maririos: Is there a value on doing clean source or clean tools? usually if I delete the tools directory I delete everything else to be consistent with the new versions that I have, and I always clean the whole tree not just what is under src.
I could only leave clean all and the rest leave it to the run tool (cleaning bin, packagesdir and packagesCache)

maririos:
I updated clean.cmd.
For now the only parsing the file does is for `-s` (source) `-t` (tools) `-all`.
I think we may be good with only `-all`.

clean.cmd
weshaggard: I think we should think about this a little more I could be open to a "t" or "s" option as well but I'm not sure it is worth the complexity.

maririos:
I updated clean.cmd again to only have the `-all` option in the script. The other options we could pass to clean are `-b`, `-p`, `-c`. These options are handle by the run command.

clean.cmd
```diff
@@ -0,0 +1,30 @@
...
+
+:Usage
+echo.
+echo Repository cleaning script.
+echo.
+echo Options:
+echo     -b     - Deletes the binary output directory.
+echo     -p     - Deletes the repo-local nuget package directory.
+echo     -c     - Deletes the user-local nuget package cache.
+echo     -all   - Combines all of the above.
```
maririos: @chcosta are we using this option in the build pipeline?
Before the options were passed with `/` and now we do it with `-`. Once I apply this change in Corefx we will need to change how we call that option.

maririos:

joperezr:
:shipit:

clean.cmd
joperezr: Yes, I believe all of the build definitions are using /all today.

joperezr: Actually looks like buildtools build definition do not use it, but corefx and coreclr ones do.

PR #856, opened 2016-07-01
Fix run command overwriting value of setting

maririos:
Let's say that the command `build-managed` in the config.json file is:
```
"build-managed":{
      "alias":{
        "binaries":{
          "description": "Only builds binaries. It doesn't restore packages.",
          "settings":{
            "RestoreDuringBuild": false,
            "OSGroup": "default"
          }
        }
    }
```
where the default value of `OSGroup=AnyOS`
Calling: `build-managed -binaries -OSGroup=Linux` will fail because the value of `OSGroup` is being overwritten.
This behavior es not expected as we want the users to custom the value of the settings.

With this change if a setting has a "default" value, it could be overwritten if the user provides a different value.
Calling: `build-managed -binaries -OSGroup=Linux` will work.

cc: @joperezr

### DIFF
--- a/src/Run/Setup.cs
+++ b/src/Run/Setup.cs
@@ -18,7 +18,7 @@ namespace Microsoft.DotNet.Execute
         private string FindSettingValue(string valueToFind)
         {
             Setting value;
-            if(Settings.TryGetValue(valueToFind, out value))
+            if (Settings.TryGetValue(valueToFind, out value))
             {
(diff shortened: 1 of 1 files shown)


maririos:
@dotnet-bot test Innerloop Windows Debug

joperezr:
@dotnet-bot Innerloop Windows Debug please

joperezr:

PR #858, opened 2016-07-05
Permit completion of builds with no packages

ericstj:
In servicing, some legs may not publish any packages.

/cc @weshaggard @dagood

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
@@ -19,7 +19,7 @@
         <RelativeBlobPath>$([System.String]::Copy('%(RecursiveDir)%(Filename)%(Extension)').Replace('\' ,'/'))</RelativeBlobPath> 
       </ForPublishing>
     </ItemGroup>
-    <Error Condition="'@(ForPublishing)' == ''" Text="No items were found matching pattern '$(PublishPattern)'." />
+    <Warning Condition="'@(ForPublishing)' == ''" Text="No items were found matching pattern '$(PublishPattern)'." />
   </Target>
 
   <!-- publishes items to blob storage in Azure -->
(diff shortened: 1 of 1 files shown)


ericstj:
+@jhendrixMSFT

ericstj:
@chcosta

dagood:
Publishing new 1.0.0 packages would update https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/1.0.0/Latest_Packages.txt with only those packages, which might be confusing. Should that file be a list of latest packages available that were at some point built in that branch, or the packages produced from the latest build? I feel like letting Latest_Packages shrink then updating LKG with the new entries is how it "should" be, but it might not make sense to everyone. /cc @eerhardt

jhendrixMSFT:
LGTM

weshaggard:
>Publishing new 1.0.0 packages would update https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/1.0.0/Latest_Packages.txt with only those packages, which might be confusing. Should that file be a list of latest packages available that were at some point built in that branch, or the packages produced from the latest build? I feel like letting Latest_Packages shrink then updating LKG with the new entries is how it "should" be, but it might not make sense to everyon

Unless I'm miss-understanding something I think it would be simpler for consumers of the version repo if the latest package list always contained the full list of the latest package versions, even if they weren't built and published as part of the latest build. I'm not sure if there is a clean way to do that but  it seems like it would be more logic to me based on the name "latest_packages". If we go with just the subset we should consider calling it "last_published" or something along those lines so it is clearer.

@ericstj as for the changes in this PR the look reasonable, assuming the package list doesn't get messed up in some way.

dagood:
I agree, updating seems more useful for consumers, and if we need to expose only the last build's packages we can choose another name. We will need more logic to handle the update without overwriting the existing list, but with the VersionTools lib centralizing it, that should be reasonable.

chcosta:
This was addressed in the CoreFx branch with https://github.com/dotnet/corefx/commit/8c2b67e16855f4c29f1ff880cd0c09388b9c38b4
, we can probably close this issue.

@dagood, I see that more of the discussion regarding "Latest" occurred in this message.  Yet more reason that we should consider updating Latest to contain the full list of packages.

ericstj:
Ok, will close.

PR #844, opened 2016-06-29
Project reference to package dependency code cleanup

chcosta:
This addresses issues brought up in a previously submitted PR and tracked with issue https://github.com/dotnet/buildtools/issues/830

- order (alphabetically) dependencies in project.json files and add the NuGet.Packaging package
- create a PackageItem class instead of using TaskItem for handling package information
- store package items as a dictionary
- error if package name clash (package information coming from multiple sources with differing versions)
- allow option to allow package name clash and take newest package version (UseNewestPackagesAvailable=true)
- when gathering information from a nupkg, get package information using the NuGet API's
- compare versions with NuGet API's
- fix regex in buildagainstpackages.targets

/cc @weshaggard @ericstj

PR #863, opened 2016-07-07
Add dependency auto-upgrade and versions repo update to buildtools

dagood:
Created a new project for "VersionTools" that so far includes auto-upgrade and build-info update. Started with the files in https://github.com/dotnet/cli/tree/rel/1.0.0/build_projects/update-dependencies and [`Command.cs` and `CommandResult.cs`](https://github.com/dotnet/cli/tree/rel/1.0.0/build_projects/Microsoft.DotNet.Cli.Build.Framework), then refactored and added a few features to work with coreclr/corefx requirements. The library should be sharable as an independent package with a little extra work. (For context, @eerhardt's comment on this approach: https://github.com/dotnet/buildtools/pull/848#issuecomment-229671100)

Added a .targets file and two msbuild tasks to integrate with corefx/coreclr/wcf. Example configuration: https://github.com/dagood/corefx/commit/50ad7f7

Test PR the upgrade task made: https://github.com/dagood/corefx/pull/2 (includes two commits I had for debugging: it committed and pushed the last one).

Once threaded through the builds, this will resolve https://github.com/dotnet/buildtools/issues/823 (share upgrade+update scripts), partially https://github.com/dotnet/buildtools/issues/835 (all in one PR), and the upgrade part of https://github.com/dotnet/buildtools/issues/732 (full package versions).

/cc @weshaggard @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/VersionTools/MsBuildTraceListener.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/VersionTools/MsBuildTraceListener.cs
@@ -1,0 +1,86 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Diagnostics;
+using System.Text;
+
+namespace Microsoft.DotNet.Build.Tasks.VersionTools
+{
+    public class MsBuildTraceListener : TraceListener
+    {
+        private TaskLoggingHelper _log;
+        private TraceEventType _eventType;
+        private StringBuilder _partialLine = new StringBuilder();
+
+        public MsBuildTraceListener(TaskLoggingHelper log, TraceEventType eventType)
+        {
+            _log = log;
+            _eventType = eventType;
+
+            Filter = new TraceEventTypeFilter
+            {
+                ShouldTraceType = eventType
+            };
+        }
+
+        public override void Write(string message)
+        {
+            _partialLine.Append(message);
+        }
+
+        public override void WriteLine(string message)
+        {
+            string fullMessage = _partialLine + message;
+            _partialLine.Clear();
+
+            switch (_eventType)
+            {
+                case TraceEventType.Error:
+                    _log.LogError(fullMessage);
+                    break;
+                case TraceEventType.Warning:
+                    _log.LogWarning(fullMessage);
+                    break;
+                case TraceEventType.Critical:
+                    _log.LogMessage(MessageImportance.High, fullMessage);
+                    break;
+                case TraceEventType.Information:
+                    _log.LogMessage(MessageImportance.Normal, fullMessage);
+                    break;
+                case TraceEventType.Verbose:
+                    _log.LogMessage(MessageImportance.Low, fullMessage);
+                    break;
+            }
+        }
+
+        public override void Flush()
+        {
+            base.Flush();
+            if (_partialLine.Length > 0)
+            {
+                WriteLine(string.Empty);
+            }
+        }
+
+        private class TraceEventTypeFilter : TraceFilter
+        {
+            public TraceEventType ShouldTraceType { get; set; }
+
+            public override bool ShouldTrace(
+                TraceEventCache cache,
+                string source,
+                TraceEventType eventType,
+                int id,
+                string formatOrMessage,
+                object[] args,
+                object data1,
+                object[] data)
+            {
+                return eventType == ShouldTraceType;
+            }
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdatePublishedVersions.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdatePublishedVersions.cs
@@ -1,0 +1,47 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.DotNet.VersionTools.Automation;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks.VersionTools
+{
+    public class UpdatePublishedVersions : Task
+    {
+        [Required]
+        public ITaskItem[] ShippedNuGetPackage { get; set; }
+
+        [Required]
+        public string VersionsRepoPath { get; set; }
+
+        [Required]
+        public string GitHubAuthToken { get; set; }
+        public string GitHubUser { get; set; }
+        public string GitHubEmail { get; set; }
+
+        public string VersionRepo { get; set; }
+        public string VersionRepoOwner { get; set; }
+
+        public override bool Execute()
+        {
+            MsBuildTraceListener[] listeners = Trace.Listeners.AddMsBuildTraceListeners(Log);
+
+            var gitHubAuth = new GitHubAuth(GitHubAuthToken, GitHubUser, GitHubEmail);
+
+            var updater = new VersionRepoUpdater(gitHubAuth, VersionRepoOwner, VersionRepo);
+
+            updater.UpdateBuildInfoAsync(
+                ShippedNuGetPackage.Select(item => item.ItemSpec),
+                VersionsRepoPath)
+                .Wait();
+
+            Trace.Listeners.RemoveMsBuildTraceListeners(listeners);
+
+            return true;
+        }
+    }
+}
--- a/src/Microsoft.DotNet.VersionTools.net45/project.json
+++ b/src/Microsoft.DotNet.VersionTools.net45/project.json
@@ -1,11 +1,9 @@
 {
   "dependencies": {
-    "Microsoft.NETCore.Platforms": "1.0.1",
     "Newtonsoft.Json": "9.0.1",
     "NuGet.Packaging": "3.5.0-rc-1590",
     "NuGet.Versioning": "3.5.0-rc-1590",
-    "System.Collections.Immutable": "1.1.37",
-    "System.Reflection.Metadata": "1.1.0",
+    "Microsoft.NETCore.Platforms": "1.0.1",
     "System.Runtime.InteropServices.RuntimeInformation": "4.0.0"
   },
   "frameworks": {
--- a/src/Microsoft.DotNet.VersionTools/Automation/GitHubHttpClient.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/GitHubHttpClient.cs
@@ -1,0 +1,100 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Diagnostics;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.VersionTools.Automation
+{
+    public class GitHubHttpClient : HttpClient
+    {
+        private GitHubAuth _auth;
+
+        public GitHubHttpClient(GitHubAuth auth)
+        {
+            DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
+            DefaultRequestHeaders.Add("Authorization", $"token {auth.AuthToken}");
+            DefaultRequestHeaders.Add("User-Agent", auth.User);
+            _auth = auth;
+        }
+
+        public async Task PutGitHubFileAsync(
+            string fileUrl,
+            string commitMessage,
+            string newFileContents)
+        {
+            Trace.TraceInformation($"Getting the 'sha' of the current contents of file '{fileUrl}'");
+
+            string currentFile = await GetStringAsync(fileUrl);
+            string currentSha = JObject.Parse(currentFile)["sha"].ToString();
+
+            Trace.TraceInformation($"Got 'sha' value of '{currentSha}'");
+
+            Trace.TraceInformation($"Request to update file '{fileUrl}' contents to:");
+            Trace.TraceInformation(newFileContents);
+
+            string updateFileBody = JsonConvert.SerializeObject(new
+            {
+                message = commitMessage,
+                committer = new
+                {
+                    name = _auth.User,
+                    email = _auth.Email
+                },
+                content = ToBase64(newFileContents),
+                sha = currentSha
+            }, Formatting.Indented);
+
+            var bodyContent = new StringContent(updateFileBody);
+            using (HttpResponseMessage response = await PutAsync(fileUrl, bodyContent))
+            {
+                response.EnsureSuccessStatusCode();
+                Trace.TraceInformation("Updated the file successfully.");
+            }
+        }
+
+        public async Task PostGitHubPullRequestAsync(
+            string title,
+            string description,
+            string originOwner,
+            string originBranch,
+            string upstreamOwner,
+            string upstreamBranch,
+            string project)
+        {
+            string createPrBody = JsonConvert.SerializeObject(new
+            {
+                title = title,
+                body = description,
+                head = $"{originOwner}:{originBranch}",
+                @base = upstreamBranch
+            }, Formatting.Indented);
+
+            string pullUrl = $"https://api.github.com/repos/{upstreamOwner}/{project}/pulls";
+
+            var bodyContent = new StringContent(createPrBody);
+            using (HttpResponseMessage response = await PostAsync(pullUrl, bodyContent))
+            {
+                response.EnsureSuccessStatusCode();
+
+                Trace.TraceInformation($"Created pull request.");
+
+                JObject responseContent = JObject.Parse(await response.Content.ReadAsStringAsync());
+                string htmlUrl = responseContent["html_url"].ToString();
+
+                Trace.TraceInformation($"Pull request page link: {htmlUrl}");
+            }
+        }
+
+        private static string ToBase64(string value)
+        {
+            return Convert.ToBase64String(Encoding.UTF8.GetBytes(value));
+        }
+    }
+}
--- a/src/Microsoft.DotNet.VersionTools/Automation/NupkgNameInfo.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/NupkgNameInfo.cs
@@ -1,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using NuGet.Packaging;
+using NuGet.Packaging.Core;
+
+namespace Microsoft.DotNet.VersionTools.Automation
+{
+    class NupkgNameInfo
+    {
+        public NupkgNameInfo(string path)
+        {
+            using (PackageArchiveReader archiveReader = new PackageArchiveReader(path))
+            {
+                PackageIdentity identity = archiveReader.GetIdentity();
+                Id = identity.Id;
+                Version = identity.Version.ToString();
+                Prerelease = identity.Version.Release;
+            }
+            SymbolPackage = path.EndsWith(".symbols.nupkg");
+        }
+
+        public string Id { get; set; }
+        public string Version { get; set; }
+        public string Prerelease { get; set; }
+        public bool SymbolPackage { get; set; }
+    }
+}
--- a/src/Microsoft.DotNet.VersionTools/Dependencies/FileRegexUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Dependencies/FileRegexUpdater.cs
@@ -1,0 +1,67 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.VersionTools.Dependencies
+{
+    public abstract class FileRegexUpdater : IDependencyUpdater
+    {
+        public string Path { get; set; }
+        public Regex Regex { get; set; }
+        public string VersionGroupName { get; set; }
+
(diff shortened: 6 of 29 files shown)


src/Microsoft.DotNet.VersionTools/project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "dependencies": {
+    "Microsoft.Build.Framework": "0.1.0-preview-00022",
+    "Microsoft.Build.Tasks.Core": "0.1.0-preview-00022",
+    "Microsoft.Build.Utilities.Core": "0.1.0-preview-00022",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.0.2",
+    "NETStandard.Library": "1.6.0",
+    "Newtonsoft.Json": "7.0.1",
```
eerhardt: If you use Newtonsoft.Json version "9.0.1" or higher, you should be able to get rid of the "imports" in your frameworks below.

src/Microsoft.DotNet.VersionTools/project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "dependencies": {
+    "Microsoft.Build.Framework": "0.1.0-preview-00022",
+    "Microsoft.Build.Tasks.Core": "0.1.0-preview-00022",
+    "Microsoft.Build.Utilities.Core": "0.1.0-preview-00022",
```
eerhardt: I don't think this assembly should have any dependencies on MSBuild, unless it is 100% necessary.

src/Microsoft.DotNet.VersionTools/project.json
```diff
@@ -0,0 +1,16 @@
...
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.0.2",
```
eerhardt: Why?  This is just a library, there shouldn't need to be a dependency on CoreCLR.

src/Microsoft.DotNet.VersionTools/project.json
```diff
@@ -0,0 +1,16 @@
...
+    "NETStandard.Library": "1.6.0",
+    "Newtonsoft.Json": "7.0.1",
+    "NuGet.Versioning": "3.5.0-beta2-1456"
```
eerhardt: Not sure how you handle updating versions here, but the CLI preview2 shipped on NuGet 3.5.0-beta2-1484.

src/Microsoft.DotNet.VersionTools.net45/project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "dependencies": {
+    "Newtonsoft.Json": "7.0.1",
+    "NuGet.Versioning": "3.5.0-beta2-1456",
+    "System.Collections.Immutable": "1.1.37",
```
eerhardt: What uses this?

src/Microsoft.DotNet.VersionTools.net45/project.json
```diff
@@ -0,0 +1,16 @@
...
+    "System.Reflection.Metadata": "1.1.0",
+    "Microsoft.NETCore.Platforms": "1.0.1-rc2-24027",
+    "System.Runtime.InteropServices.RuntimeInformation": "4.0.0-rc3-24128-00"
```
eerhardt: Is this necessary?

src/Microsoft.DotNet.VersionTools/Util/Command.cs
```diff
@@ -0,0 +1,348 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.DotNet.VersionTools.Util
+{
+    public class Command
```
eerhardt: I think all these "Utils" should probably be internal. We don't want external callers using these.  (It isn't the job of the "VersionTools" to expose these classes)

src/Microsoft.DotNet.VersionTools/project.json
dagood: Agreed, and it shouldn't be.

eerhardt:
> using Microsoft.Build.Framework;

I assume all .cs files in build-tools should have copyright headers.  Can you ensure they all do?

---
Refers to: src/Microsoft.DotNet.Build.Tasks/VersionTools/MsBuildTraceListener.cs:1 in dc5b318. [](commit_id = dc5b31809a358cfd682cf417f95a37073d73d914, deletion_comment = False)

src/Microsoft.DotNet.VersionTools/project.json
dagood: I should have mentioned: this project.json is copied from the `Microsoft.DotNet.Build.Tasks` one, so these dependencies are mostly from there.

My changes:
* Upgraded `NETStandard.Library` to `1.6.0`
* Removed some `System.*` direct dependencies.
* Added CoreCLR runtime direct dependency. (Will discuss in other comment.)

Because of the way this is shipped (the netstandard and net45 dlls copied to the buildtools package) I wonder if changing this version might break something... I only fiddled with the project.jsons to the point where things worked, so this is something to revisit.

src/Microsoft.DotNet.VersionTools/project.json
dagood: I think this was a misguided attempt to make things work early on, and I didn't go back after I was done to make sure these dependencies are all needed.

src/Microsoft.DotNet.VersionTools.net45/project.json
dagood: This project.json is binary equal to `Microsoft.DotNet.Build.Tasks.net45/project.json`, and I'll need to look at this file more closely to see what's needed. (This particular reference, probably not.)

dagood:
> using Microsoft.Build.Framework;

I'll add them to the files I'm adding here, you're right. I did notice they already don't seem to be consistently applied in Microsoft.DotNet.Build.Tasks, unfortunately.

---
In reply to: [231142273](https://github.com/dotnet/buildtools/pull/863#issuecomment-231142273) [](ancestors = 231142273)

---
Refers to: src/Microsoft.DotNet.Build.Tasks/VersionTools/MsBuildTraceListener.cs:1 in dc5b318. [](commit_id = dc5b31809a358cfd682cf417f95a37073d73d914, deletion_comment = False)

src/Microsoft.DotNet.VersionTools/Upgrade/IDependencyUpgrader.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// A tool that uses buildInfos to perform an upgrade.
+    /// </summary>
+    public interface IDependencyUpgrader
+    {
+        void Upgrade(IEnumerable<BuildInfo> buildInfos);
+
+        /// <summary>
+        /// Which build infos were used during Upgrade. Used for informational purposes.
+        /// </summary>
+        IEnumerable<BuildInfo> BuildInfosUsed { get; }
```
eerhardt: I think this would be better served as the return from Upgrade.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdatePublishedVersions.cs
```diff
@@ -0,0 +1,43 @@
...
+        public string GitHubAuthToken { get; set; }
+        public string GitHubUser { get; set; }
+        public string GitHubEmail { get; set; }
+
+        public string VersionRepo { get; set; }
+        public string VersionRepoOwner { get; set; }
+
+        public override bool Execute()
+        {
+            MsBuildTraceListener[] listeners = Trace.Listeners.AddMsBuildTraceListeners(Log);
```
eerhardt: I think it would be good to have a comment or two either here or in MsBuildTraceListener as to 'why' it is needed.

src/Microsoft.DotNet.Build.Tasks/VersionTools/MsBuildTraceListener.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            Filter = new TraceEventTypeFilter
+            {
+                ShouldTraceType = eventType
+            };
+        }
+
+        public override void Write(string message)
+        {
+            _partialLine.Append(message);
```
eerhardt: Is it OK that a dangling "Write" won't get written to the log?  i.e. if Write is the last call, and not WriteLine.

src/Microsoft.DotNet.VersionTools/Automation/GitHubHttpClient.cs
```diff
@@ -0,0 +1,100 @@
...
+                response.EnsureSuccessStatusCode();
+
+                Trace.TraceInformation($"Created pull request.");
+
+                JObject responseContent = JObject.Parse(await response.Content.ReadAsStringAsync());
+                string htmlUrl = responseContent["html_url"].ToString();
+
+                Trace.TraceInformation($"Pull request page link: {htmlUrl}");
+
+                return response;
```
eerhardt: It is not a good practice to return a disposed object from a method.

src/Microsoft.DotNet.VersionTools/Automation/GitHubHttpClient.cs
```diff
@@ -0,0 +1,100 @@
...
+                sha = currentSha
+            }, Formatting.Indented);
+
+            var bodyContent = new StringContent(updateFileBody);
+            using (HttpResponseMessage response = await PutAsync(fileUrl, bodyContent))
+            {
+                response.EnsureSuccessStatusCode();
+                Trace.TraceInformation("Updated the file successfully.");
+
+                return response;
```
eerhardt: It is not a good practice to return a disposed object from a method.

src/Microsoft.DotNet.VersionTools/Automation/ProjectUpgradePR.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+            if (!hasModifiedFiles)
+            {
+                Trace.TraceWarning("Dependencies are currently up to date");
+                return;
+            }
+
+            string remoteBranchName = $"UpdateDependencies{DateTime.UtcNow.ToString("yyyyMMddhhmmss")}";
+
+            PushNewCommit(commitMessage, remoteBranchName);
```
eerhardt: Can this have a bool option that says not to submit a PullRequest?  In the CLI, I use just "UpdateFiles" locally, and then I manually inspect and maybe build locally before submitting a PR.

I guess maybe the name of the class should change then to not have "Pr" in its name, if the PR is optional.  How about naming it "DependencyUpdater"?

eerhardt: You use the term "Upgrade" in a few spots, so maybe Upgrader.  Although there is nothing in particular saying this has to be an upgrade - it could be used to downgrade, if necessary.

---
In reply to: [69952526](https://github.com/dotnet/buildtools/pull/863#discussion_r69952526) [](ancestors = 69952526)

src/Microsoft.DotNet.Build.Tasks/VersionTools/MsBuildTraceListener.cs
dagood: The way I use `Trace`, `Write` is only ever used by `TraceListener` internals to print a prefix header like "msbuild Information: 0 :". It's possible someone will use `Trace.Write` later, so I'll fix this by adding some kind of `Flush`.

---
In reply to: [69950623](https://github.com/dotnet/buildtools/pull/863#discussion_r69950623) [](ancestors = 69950623)

src/Microsoft.DotNet.VersionTools/Automation/GitHubHttpClient.cs
dagood: Fixed by removing the return--don't know what I was thinking, nothing uses it.

---
In reply to: [69951596](https://github.com/dotnet/buildtools/pull/863#discussion_r69951596) [](ancestors = 69951596)

src/Microsoft.DotNet.VersionTools/Automation/GitHubHttpClient.cs
dagood: Fixed here too by changing to `Task` and removing return.

---
In reply to: [69951690](https://github.com/dotnet/buildtools/pull/863#discussion_r69951690) [](ancestors = 69951690)

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdatePublishedVersions.cs
dagood: Good idea, adding doc to the add/remove methods.

---
In reply to: [69950365](https://github.com/dotnet/buildtools/pull/863#discussion_r69950365) [](ancestors = 69950365)

src/Microsoft.DotNet.VersionTools/Automation/ProjectUpgradePR.cs
dagood: The flag definitely makes sense to me. The naming of these things is free to change more drastically, too. I tried to be consistent with `Update` for changing build-info and `Upgrade` for dependency stuff just to keep them clearly separate, but looking at it now, that doesn't seem necessary. `Update` is more accurate, so I'll name everything that way.

---
In reply to: [69952732](https://github.com/dotnet/buildtools/pull/863#discussion_r69952732) [](ancestors = 69952732,69952526)

dagood: When I went to add it, I realized I don't think a flag is good in a library context. For now I added a new public method for updating that is also used by the public method that does the git and github stuff.

---
In reply to: [69973491](https://github.com/dotnet/buildtools/pull/863#discussion_r69973491) [](ancestors = 69973491,69952732,69952526)

src/Microsoft.DotNet.VersionTools/project.json
dagood: This came from the buildtools project.json, and it's what's available on [https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/NuGet.Versioning](https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/NuGet.Versioning). We can push a new version if an update seems necessary.

---
In reply to: [69943654](https://github.com/dotnet/buildtools/pull/863#discussion_r69943654) [](ancestors = 69943654)

src/Microsoft.DotNet.VersionTools.net45/project.json
dagood: It's used in Command.cs: `OSPlatform` and `RuntimeInformation`.

---
In reply to: [69943776](https://github.com/dotnet/buildtools/pull/863#discussion_r69943776) [](ancestors = 69943776)

dagood:
I updated the project.jsons and tried out UpdateDependencies on win10 and OSX, and they still worked. I upgraded some packages to RTM versions and took some dependencies out.

@ericstj Could you look at these project.jsons to see if the changes should be ok? (A full review would be awesome too. :smile:)

src/Microsoft.DotNet.VersionTools.net45/project.json
eerhardt: Why wouldn't we use the RTM version of it?  Why use the pre-release?

dagood: It looks like you're on iteration 2 (old commit), I changed this to 4.0.0 now. (And lots of other dependencies are RTM now too.)

---
In reply to: [70108692](https://github.com/dotnet/buildtools/pull/863#discussion_r70108692) [](ancestors = 70108692)

eerhardt:
:shipit:

src/Microsoft.DotNet.VersionTools/Automation/NupkgNameInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.DotNet.VersionTools.Automation
+{
+    class NupkgNameInfo
+    {
+        private static Regex s_nugetFileRegex = new Regex(
+            @"^(?<id>.*?)\.(?<version>([0-9]+\.)?[0-9]+\.[0-9]+(-(?<prerelease>[A-z0-9-]+))?)(?<symbols>\.symbols)?\.nupkg$");
+
+        public NupkgNameInfo(string path)
+        {
+            Match match = s_nugetFileRegex.Match(Path.GetFileName(path));
```
ericstj: This isn't safe.  We had a discussion about this here:
https://github.com/dotnet/buildtools/pull/806#discussion_r67536338

And Chris is fixing the way he does this here: 
https://github.com/dotnet/buildtools/pull/844/files#diff-741c73271c193f69634ebe0dba519c22R229

src/Microsoft.DotNet.VersionTools/Util/Command.cs
```diff
@@ -0,0 +1,349 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.DotNet.VersionTools.Util
+{
+    class Command
```
ericstj: Why aren't we just doing this via exec, or making a specific task that wraps the command as a ToolTask.  Do we really need another wrapper for Process?

src/Microsoft.DotNet.VersionTools/PackageInfo.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.VersionTools
+{
+    public class PackageInfo
```
ericstj: There's already a type in NuGet that represents the same: PackageIdentity.

src/Microsoft.DotNet.VersionTools/Dependencies/FileRegexUpdater.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.VersionTools.Dependencies
+{
+    public abstract class FileRegexUpdater : IDependencyUpdater
+    {
+        public string Path { get; set; }
+        public Regex Regex { get; set; }
```
ericstj: Do we really need Regex's still?  Was hoping nuget would have a project model that round-trips by now.  /cc @emgarten

src/Microsoft.DotNet.VersionTools/Automation/NupkgNameInfo.cs
dagood: Nice, I'll change it.

---
In reply to: [71019284](https://github.com/dotnet/buildtools/pull/863#discussion_r71019284) [](ancestors = 71019284)

ericstj:
> Could you look at these project.jsons to see if the changes should be ok?

They look ok, so long as the SLN builds fine, and you get what you expect in the nupkg, and you've tested that.

Added some comments on the code after cursory review.

src/Microsoft.DotNet.VersionTools/Dependencies/FileRegexUpdater.cs
eerhardt: This is to support cases where we have hard-coded version strings in other files, ex. dir.props, or like in the CLI:  https://github.com/dotnet/cli/blob/rel/1.0.0/build_projects/dotnet-cli-build/CliDependencyVersions.cs#L10-L12

dagood: This also replaces elements in xml files, like the dependency validation properties in CoreFX/WCF. `project.json` upgrades are done in `ProjectJsonUpdater`, which is Newtonsoft.Json, so still a problem.

src/Microsoft.DotNet.VersionTools/Util/Command.cs
eerhardt: Yes, let's not make this library dependent on MSBuild.

It appears that there are only a handful of places this is called - just the "git" commands right?  I don't think making the caller invoke the "git" commands is a good approach, that's a lot of benefit of having this logic in a centralized place.

All that the code needs is a way to start a few processes: git commit, push, status.

dagood:
Thanks for the review! I tried it out earlier and the outputs seemed good to me, and worked on Windows and OSX.

I'm adding a new nuspec to the PR that lets us  deliver this lib separately from buildtools, since CLI wants to start using this very soon.

I think we need to talk more about the exec wrapper, but to unblock CLI I want to merge this. Does it sound reasonable to do that, and potentially refactor this later?

src/Microsoft.DotNet.VersionTools/Util/Command.cs
eerhardt: @ericstj - do you have a strong opinion here?  The script needs to kick off a handful of processes (it looks like all 'git' commands).  Having an internal helper makes the code easier to read in those places.

ericstj:
Yeah, that's reasonable.

dagood:
Cool. I squashed and signed, so merging now.

src/Microsoft.DotNet.VersionTools/Dependencies/FileRegexUpdater.cs
ericstj: Well for dir.props MSBuild has a project model as well, but I'm OK keeping this for now.  I have seen folks wrestle with this RegExes as well as see accidental inclusion/exclusion due to bad regexes.  I would hope that we could more more to some strongly typed system that understands the documents and is driven by precise lists of updates.

ericstj:
LGTM

src/Microsoft.DotNet.VersionTools/Dependencies/FileRegexUpdater.cs
dagood: Just to make sure we're talking about the same regexes, the one here would be replacing the content of these elements: https://github.com/dotnet/corefx/blob/2574648/dir.props#L91-L93. So an example would be `<CoreFxExpectedPrerelease>(?<version>.*)<`. A strong model would be better, but at least this is pretty simple.

The package-matching regexes on the other hand are definitely bad, and https://github.com/dotnet/buildtools/issues/732 tracks changing to validate based on the versions repo package lists. (It sounds like you're talking about these with inclusions/exclusions, which I've had my fair share of problems with. :smile:)

src/Microsoft.DotNet.VersionTools/Util/Command.cs
dagood: Yep, only `status`, `commit`, and `push`.

I tried removing the utility classes and putting what I needed in the static `Git` method, and came up with this: https://github.com/dotnet/buildtools/compare/master...dagood:simplify-versiontools-git. @ericstj Maybe it's more reasonable to include this than a full wrapper?

PR #865, opened 2016-07-09
Enable Validation of Xamarin frameworks

ericstj:
This change turns on validation of Xamarin frameworks and updates our
inbox assembly lists to represent the files which must be provided by
the xamarin SDK (either because they already are, or because they
contain types which overlap with types already in their SDK).

/cc @mhutch @migueldeicaza @marek-safar

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/FrameworkLists/MonoAndroid,Version=v1.0/FrameworkList.xml
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/FrameworkLists/MonoAndroid,Version=v1.0/FrameworkList.xml
@@ -26,6 +26,7 @@
 <File AssemblyName="System.Diagnostics.TraceEvent" />
 <File AssemblyName="System.Diagnostics.TraceSource" />
 <File AssemblyName="System.Diagnostics.Tracing" />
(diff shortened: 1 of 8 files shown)


mhutch:
I don't think System.Runtime.InteropServices.WindowsRuntime should be included in the Xamarin framesworks, I don't see any types in it that we could/would support.

mhutch:
I don't think System.Runtime.InteropServices.WindowsRuntime should be included in the Xamarin frameworks, I don't see any types in it that we could/would support.

Apart from that the changes look good to me.

ericstj:
System.Runtime.InteropServices.WindowsRuntime is part of the PCL profiles you already support (EG: net45+win8).  If you don't include those types then folks who are writing PCLs will fail like @onovotny points out here: https://github.com/dotnet/corefx/issues/6684.

mhutch:
Ah, OK, fair enough then.

marek-safar:
We support `System.Runtime.InteropServices.WindowsRuntime` facade already. The only one we haven't added yet is `System.Drawing.Primitives` but we are doing it now

ericstj:
Ok, with this change we'll make sure to defer to your facade for `System.Drawing.Primitives`

PR #867, opened 2016-07-12
Add CPUArch and OSName helpers to Run.exe

Priya91:
cc @maririos @weshaggard @sokket @markwilkie

### DIFF
--- a/src/Run/Constants.cs
+++ b/src/Run/Constants.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.Execute
+{
+    public struct OS
+    {
+        public const string OSX = "OSX";
+        public const string FreeBSD = "FreeBSD";
+        public const string NetBSD = "NetBSD";
+        public const string Linux = "Linux";
+        public const string Unix = "Unix";
+        public const string Windows = "Windows_NT";
+    }
+
+    public struct Arch
+    {
+        public const string X86 = "x86";
+        public const string X64 = "x64";
+        public const string Arm = "arm";
+        public const string Arm64 = "arm64";
+    }
+}
--- a/src/Run/Interop.cs
+++ b/src/Run/Interop.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.DotNet.Execute
+{
+    internal class Interop
+    {
+        public static bool GetUnixVersion(out string result)
+        {
+            result = null;
+            const string OSReleaseFileName = @"/etc/os-release";
+            if (File.Exists(OSReleaseFileName))
+            {
+                string content = File.ReadAllText(OSReleaseFileName);
+                int idIndex = content.IndexOf("ID");
+                int versionIndex = content.IndexOf("VERSION_ID");
+                if (idIndex != -1 && versionIndex != -1)
+                {
+                    string id = content.Substring(idIndex + 3, content.IndexOf(Environment.NewLine, idIndex + 3) - idIndex - 3);
+                    string version = content.Substring(versionIndex + 12, content.IndexOf('"', versionIndex + 12) - versionIndex - 12);
+                    result = $"{id}.{version}";
+                }
+            }
+
+            return result != null;
+        }
+    }
+}
--- a/src/Run/Interop.cs
+++ b/src/Run/Interop.cs
@@ -11,19 +11,36 @@ namespace Microsoft.DotNet.Execute
     {
         public static bool GetUnixVersion(out string result)
         {
+            const string OSId = "ID=";
+            const string OSVersionId = "VERSION_ID=";
             result = null;
             const string OSReleaseFileName = @"/etc/os-release";
             if (File.Exists(OSReleaseFileName))
             {
-                string content = File.ReadAllText(OSReleaseFileName);
-                int idIndex = content.IndexOf("ID");
-                int versionIndex = content.IndexOf("VERSION_ID");
-                if (idIndex != -1 && versionIndex != -1)
+                string[] content = File.ReadAllLines(OSReleaseFileName);
+                string id = null, version = null;
+                foreach (string line in content)
                 {
-                    string id = content.Substring(idIndex + 3, content.IndexOf(Environment.NewLine, idIndex + 3) - idIndex - 3);
-                    string version = content.Substring(versionIndex + 12, content.IndexOf('"', versionIndex + 12) - versionIndex - 12);
-                    result = $"{id}.{version}";
+                    if (line.StartsWith(OSId))
+                    {
+                        id = line.Substring(OSId.Length, line.Length - OSId.Length);
+                    }
+                    else if (line.StartsWith(OSVersionId))
+                    {
+                        int startOfVersion = line.IndexOf('"', OSVersionId.Length) + 1;
+                        int endOfVersion = startOfVersion == 0 ? line.Length : line.IndexOf('"', startOfVersion);
+                        if (startOfVersion == 0)
+                            startOfVersion = OSVersionId.Length;
+
+                        version = line.Substring(startOfVersion, endOfVersion - startOfVersion);
+                    }
+
+                    // Skip parsing rest of the file contents.
+                    if (id != null && version != null)
+                        break;
                 }
+
+                result = $"{id}.{version}";
             }
 
             return result != null;
--- a/src/Run/SettingValueProvider.cs
+++ b/src/Run/SettingValueProvider.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Runtime.InteropServices;
-using System.IO;
 using System;
 using System.Collections.Generic;
 using System.Reflection;
@@ -14,14 +13,14 @@ namespace Microsoft.DotNet.Execute
     {
         private static string s_cpuArch = null;
         private static string s_osName = null;
-        private static string s_osVersion = null;
+        private static string s_osRid = null;
         private static Dictionary<string, string> s_osToDefaultVersionMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
         {
             { OS.OSX, "osx.10.10" },
             { OS.FreeBSD, "osx.10.10" },
             { OS.NetBSD, "osx.10.10" },
             { OS.Linux, "ubuntu.14.04" },
-            { OS.Windows, string.Empty }
+            { OS.Windows, "win7" }
         };
 
         public static string Get(string settingName)
@@ -29,7 +28,7 @@ namespace Microsoft.DotNet.Execute
             PropertyInfo property = typeof(SettingValueProvider).GetTypeInfo().GetProperty(settingName);
             return (property != null) ? (string)property.GetValue(null) : string.Empty;
         }
-        
+
         public static string CPUArch
         {
             get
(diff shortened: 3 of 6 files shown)


src/Run/SettingValueProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+                    {
+                        s_osName = "FreeBSD";
+                    }
+                    else if (RuntimeInformation.IsOSPlatform(OSPlatform.Create("NETBSD")))
+                    {
+                        s_osName = "NetBSD";
+                    }
+                    else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                    {
+                        const string OSReleaseFileName = @"/etc/os-release";
```
maririos: We still need these information as we need to compute the following values:
```
 __TestNugetRuntimeId={s_osName}.{version}-{s_cpuArch} => build.sh
__DOTNET_PKG=dotnet-dev-{[win/osx/ubuntu]}(.version)-x64 =>init-tools.cmd/sh
RuntimeOS =>build-packages.sh. This one has various conditions dependending on the OS
```

maririos:
Once we have all the variables we need, we need to plug it into the logic of the run command tool so that it will use these variables when building the command to execute

src/Run/SettingValueProvider.cs
maririos: We'll need the Number of processors too (build.sh)

src/Run/SettingValueProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+                            int versionIndex = content.IndexOf("VERSION_ID");
+                            if (idIndex != -1 && versionIndex != -1)
+                            {
+                                string id = content.Substring(idIndex + 3, content.IndexOf(Environment.NewLine, idIndex + 3) - idIndex - 3);
+                                string version = content.Substring(versionIndex + 12, content.IndexOf('"', versionIndex + 12) - versionIndex - 12);
+                                s_osName = $"{id}.{version}";
+                            }
+                            else
+                            {
+                                s_osName = "ubuntu.14.04";
```
Priya91: We need the versions to calculate the testnugetruntimeid to restore.. build.sh tries to figure out the version.. I will split this up into 2 properties, one to set the __HostOS and another for __TestNugetRuntimeId

Priya91:
@maririos I was thinking of checking for defaultValue, if it's empty, then use reflection to check if a value can be obtained from SettingValueProvider, based on the Setting name. In which case we will have to match the setting name and the member name in the SettingValueProvider.

Priya91:
@maririos added the reflection and splitting osname and version in the latest commit, working on numproc, will push another commit.

src/Run/Setup.cs
```diff
@@ -14,13 +18,51 @@ public class Setup
...
+            return value;
+        }
 
         private string FindSettingValue(string valueToFind)
         {
             Setting value;
             if (Settings.TryGetValue(valueToFind, out value))
             {
-                return value.DefaultValue;
+                return ParseSettingValue(value.DefaultValue);
```
maririos: In the config.json, the value in Setting for the `default value` would be `${s_cpuArch}`, right?

src/Run/Setup.cs
```diff
@@ -111,7 +153,7 @@ private string BuildParametersForCommand(Dictionary<string, string> commandParam
             {
                 if (!parameters.Key.Equals("toolName") && !string.IsNullOrEmpty(parameters.Value))
                 {
-                    string value = parameters.Value.Equals("default") ? FindSettingValue(parameters.Key) : parameters.Value;
+                    string value = parameters.Value.Equals("default") ? FindSettingValue(parameters.Key) : ParseSettingValue(parameters.Value);
```
maririos: If it is only in the default value under Settings, we don't need to call PaseSettingValue again here

src/Run/Setup.cs
```diff
@@ -14,13 +18,51 @@ public class Setup
         public string Os { get; set; }
         public string ConfigurationFilePath { get; set; }
 
+        private string ParseSettingValue(string inputValue)
+        {
+            string value = string.Empty;
+            int length = inputValue.Length;
+            for (int i = 0; i < length; i++)
+            {
+                if (inputValue[i] == '$')
```
maririos: We could just ask if the first character of the string is `$`. If it is keep processing, otherwise just return the inputValue

maririos:
FYI: @wtgodbe @tyoverby

src/Run/Constants.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Microsoft.DotNet.Execute
+{
+    public struct OS
+    {
+        public const string OSX = "OSX";
+        public const string FreeBSD = "FreeBSD";
+        public const string NetBSD = "NetBSD";
+        public const string Linux = "Linux";
+        public const string Unix = "Unix";
+        public const string Windows = "Windows_NT";
```
sokket: Is this the same for nano server?

src/Run/Setup.cs
Priya91: no, it would be ${CPUArch}.. s_cpuArch is a private static field, we want to use only public properties in the json format..

src/Run/Setup.cs
Priya91: Discussed offline with @maririos, and we decided to not impose rules on where the ${} can be used, but leave it to the config.json author.

src/Run/Setup.cs
Priya91: So the idea of the string format is it can be anything like, ```${OSVersion}-${CPUArch}``` or ```myOS.${OSName}```.. This code will just replace all occurrences of ${} in the string with the value from SettingValueProvider if available.

src/Run/Setup.cs
maririos: yeah, my bad. I meant in the `default value` field, but after our conversation I understand why you don't want that constrain

src/Run/Constants.cs
Priya91: Yeah, currently we don't have a OS Name for specific windows versions in the build system, all windows builds have /p:OSGroup=Windows_NT.

src/Run/SettingValueProvider.cs
```diff
@@ -47,6 +61,44 @@ public static string CPUArch
             }
         }
 
+        public static string UnixVersion
+        {
```
Priya91: Should this be OSVersion instead of UnixVersion?

src/Run/Setup.cs
maririos: It is worth clarifying that `${}` is only valid for Settings value. It is not meant to be used in other parts of the config.json file as we don't want to have a pre-build config.json file step before actually using it.

src/Run/SettingValueProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+                            s_cpuArch = Arch.X64;
+                            break;
+                    }
+                }
+
+                return s_cpuArch;
+            }
+        }
+
+        public static string OSVersion
```
ellismg: This seems like a bad name for this API.  It's trying to give back what exactly?  The RID for the platform?  Something else?  How would this actually be used?

Priya91: This will be used in config.json like,
```
"TestNugetRuntimeId": {
      "valueType": "property",
      "values": ["osx.10.10-x64", "ubuntu.14.04-x64", "ubuntu.16.04-x64", "etc-other-rids"],
      "defaultValue": "${OSVersion}-x64"
     }
```

ellismg: I would call this something like OSRid or something so it was clear what it would be used for.

The default value for windows is string.Empty, which feels like a bad default as well.  Should it be "win7" ?

Priya91: Ok sounds good, yeah, in corefx ```build.sh``` the default value for TestNugetRuntimeId is set as win7-x64.. Will change.

Priya91:
@maririos @ellismg Can you please review the final changes.. Thanks!

src/Run/SettingValueProvider.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                return Environment.ProcessorCount.ToString();
+            }
+        }
+
+        public static string OSRid
+        {
+            get
+            {
+                if (s_osVersion == null)
```
maririos: Just to have consistency, could we rename this variable to `s_osRID` ?

maririos:
LGTM

src/Run/SettingValueProvider.cs
Priya91: yes, haha i missed this :)

src/Run/Interop.cs
```diff
@@ -0,0 +1,32 @@
...
+            const string OSReleaseFileName = @"/etc/os-release";
+            if (File.Exists(OSReleaseFileName))
+            {
+                string content = File.ReadAllText(OSReleaseFileName);
+                int idIndex = content.IndexOf("ID");
+                int versionIndex = content.IndexOf("VERSION_ID");
+                if (idIndex != -1 && versionIndex != -1)
+                {
+                    string id = content.Substring(idIndex + 3, content.IndexOf(Environment.NewLine, idIndex + 3) - idIndex - 3);
+                    string version = content.Substring(versionIndex + 12, content.IndexOf('"', versionIndex + 12) - versionIndex - 12);
```
ellismg: Is this assuming that VERSION_ID is always wrapped with quotes?  I don't believe that's always the case.  Here's the /etc/os-release from Fedora 23

```
[root@4f89a32f501b /]# cat /etc/os-release
NAME=Fedora
VERSION="23 (Twenty Three)"
ID=fedora
VERSION_ID=23
PRETTY_NAME="Fedora 23 (Twenty Three)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:23"
HOME_URL="https://fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=23
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=23
PRIVACY_POLICY_URL=https://fedoraproject.org/wiki/Legal:PrivacyPolicy
```

src/Run/Interop.cs
```diff
@@ -0,0 +1,32 @@
...
+    internal class Interop
+    {
+        public static bool GetUnixVersion(out string result)
+        {
+            result = null;
+            const string OSReleaseFileName = @"/etc/os-release";
+            if (File.Exists(OSReleaseFileName))
+            {
+                string content = File.ReadAllText(OSReleaseFileName);
+                int idIndex = content.IndexOf("ID");
```
ellismg: Might be nice if you change these to be constants, append an = to the strings and then use `<constant_name>.Length` instead of hard-coding the lengths...

src/Run/Interop.cs
Priya91: Yeah, assuming the quotes are always there.. Will handle this case.. I am also assuming ID always comes before Version_ID, otherwise scanning for ID will give location for id in version_id..

src/Run/Interop.cs
Priya91: sure will do.

src/Run/Interop.cs
ellismg: Might be better just to read it line by line, I dunno.

src/Run/SettingValueProvider.cs
```diff
@@ -0,0 +1,128 @@
...
+{
+    internal static class SettingValueProvider
+    {
+        private static string s_cpuArch = null;
+        private static string s_osName = null;
+        private static string s_osRid = null;
+        private static Dictionary<string, string> s_osToDefaultVersionMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+        {
+            { OS.OSX, "osx.10.10" },
+            { OS.FreeBSD, "osx.10.10" },
```
weshaggard: Why does FreeBSD and NetBSD default to osx.10.10?

src/Run/Setup.cs
```diff
@@ -14,13 +18,51 @@ public class Setup
...
+                {
+                    if (inputValue[i + 1] == '{')
```
weshaggard: This will crash for any strings that end in `$` we need to check to be sure `i+1` isn't off the end of the string.

src/Run/SettingValueProvider.cs
```diff
@@ -0,0 +1,128 @@
...
+            { OS.NetBSD, "osx.10.10" },
+            { OS.Linux, "ubuntu.14.04" },
+            { OS.Windows, "win7" }
+        };
+
+        public static string Get(string settingName)
+        {
+            PropertyInfo property = typeof(SettingValueProvider).GetTypeInfo().GetProperty(settingName);
```
weshaggard: Do we have these way to expose these setting values so we know what the options are? Should we make them case-insensitive?

Priya91: No we dont have a way to get all the available settings, yes can make it case insensitive..

src/Run/SettingValueProvider.cs
Priya91: In corefx, build.sh is assinging osx.10.10 as the default values for testnugetruntimeid..

PR #870, opened 2016-07-12
Fix test list filtering for CloudTest

karajas:
/cc @ChadNedzlek @MattGal

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
@@ -1,0 +1,108 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    // Filter the test archives for upload for CloudTest on Helix, based on the OSGroup and TestTFM
+    // This will be used to generate the test list and decide the blobs uploaded to Azure
+    public sealed class FilterForUpload : Task
+    {
+        /// <summary>
+        ///     This is the list of test archives produced with OSPlatformConfig metadata set
+        /// </summary>
+        [Required]
+        public ITaskItem[] UnfilteredUploadItems { get; set; }
+
+        /// <summary>
+        ///     This is the TestTFM folder that we want to select test zips from. This needs to be on the path to the test zip.
+        /// </summary>
+        [Required]
+        public string FilterToTestTFM { get; set; }
+
+        /// <summary>
+        ///     This is the OSGroup folder that we want to select test zips from. This does not require an exact match as some
+        ///     tests
+        ///     may or may not have an exact OSGroup folder match.
+        /// </summary>
+        [Required]
+        public string FilterToOSGroup { get; set; }
+
+        /// <summary>
+        ///     This is the bin folder that should be removed from all paths.
+        /// </summary>
+        [Required]
+        public string BinDir { get; set; }
+
+        [Output]
+        public ITaskItem[] FilteredUploadItems { get; set; }
+
+        public override bool Execute()
+        {
+            List<ITaskItem> filteredBasedOnTestTfm =
+                UnfilteredUploadItems.Where(
+                    item => this.MakeRelativePathTo(item.ItemSpec, BinDir).Contains(FilterToTestTFM)).ToList();
+
+            List<ITaskItem> localFilteredUploadItemsList = new List<ITaskItem>();
+
+            Dictionary<string, List<ITaskItem>> map = new Dictionary<string, List<ITaskItem>>();
+            foreach (var item in filteredBasedOnTestTfm)
+            {
+                string testName = Path.GetFileName(item.ItemSpec);
+                List<ITaskItem> values = null;
+                if (!map.TryGetValue(testName, out values))
+                {
+                    values = new List<ITaskItem>();
+                    map[testName] = values;
+                }
+                values.Add(item);
+            }
+
+            foreach (var item in map)
+            {
+                List<ITaskItem> values = item.Value;
+                ITaskItem selectedItem = null;
+                foreach (var value in values)
+                {
+                    if (value.GetMetadata("OSPlatformConfig").Contains(FilterToOSGroup))
+                    {
+                        selectedItem = value;
+                        break;
+                    }
+                }
+                // Fallback to first value
+                if (selectedItem == null)
+                {
+                    selectedItem = values[0];
+                }
+                localFilteredUploadItemsList.Add(selectedItem);
+            }
+
+            FilteredUploadItems = localFilteredUploadItemsList.ToArray();
+            return true;
+        }
+
+
+        private string MakeRelativePathTo(string targetPath, string basePath)
+        {
+            basePath = Path.GetFullPath(basePath);
+            string fullPath1 = Path.GetFullPath(targetPath);
+            if (fullPath1.StartsWith(basePath, StringComparison.OrdinalIgnoreCase))
+            {
+                return fullPath1.Substring(basePath.Length + 1);
+            }
+            else
+            {
+                return targetPath;
+            }
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
@@ -11,6 +11,7 @@
   <UsingTask TaskName="WriteTestBuildStatsJson" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll"/>
   <UsingTask TaskName="ZipFileCreateFromDirectory" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="ZipFileCreateFromDependencyLists" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="FilterForUpload" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll"/>
 
   <!-- set Helix environment vars based on target platform -->
   <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
@@ -89,14 +90,28 @@
     <Warning Condition="'$(SkipNotifyEvent)' != 'true' and '$(HelixApiAccessKey)' == '' and '$(EventHubInfoMissing)' == '' " Text="EventHubPath and EventHubAccessKeys have been deprecated, use HelixApiAccessKey instead" />
     <Error Condition="'$(SkipNotifyEvent)' != 'true' and '$(HelixApiAccessKey)' == '' and '$(EventHubInfoMissing)' == 'true' " Text="HelixApiAccessKey must be set to start Helix jobs" />
 
-    <!-- gather the test archives for upload -->
-    <!-- This will include every zip in the test archive, to allow CoreCLR tests for perf which don't have a TestTFM directory -->
+    <!-- gather the test archives for upload and filter based on FilterToTestTFM -->
     <ItemGroup>
-      <ForUpload Include="$(TestArchivesRoot)**/*.zip" />
-      <ForUpload Include="$(AnyOSTestArchivesRoot)**/*.zip" />
+      <UnfilteredTestArchives Include="$(TestArchivesRoot)**/*.zip" >
+        <OSPlatformConfig>$(OSPlatformConfig)</OSPlatformConfig>
+      </UnfilteredTestArchives>
+      <UnfilteredTestArchives Include="$(AnyOSTestArchivesRoot)**/*.zip" >
+        <OSPlatformConfig>$(AnyOSPlatformConfig)</OSPlatformConfig>
+      </UnfilteredTestArchives>
       <!-- Only include Unix folders if supported by the current target OS -->
-      <ForUpload Condition="'$(TargetsUnix)' == 'true'" Include="$(UnixTestArchivesRoot)**/*.zip" />
+      <UnfilteredTestArchives Condition="'$(TargetsUnix)' == 'true'" Include="$(UnixTestArchivesRoot)**/*.zip" >
+        <OSPlatformConfig>$(UnixPlatformConfig)</OSPlatformConfig>
+      </UnfilteredTestArchives>
     </ItemGroup>
+    <FilterForUpload UnfilteredUploadItems="@(UnfilteredTestArchives)"
+             FilterToTestTFM="$(FilterToTestTFM)"
+             FilterToOSGroup="$(FilterToOSGroup)" 
+             BinDir="$(BinDir)">
+        <Output TaskParameter="FilteredUploadItems" PropertyName="ForUploadList" />
+    </FilterForUpload>
+    <CreateItem Include="$(ForUploadList)">
+      <Output TaskParameter="Include" ItemName="ForUpload"/>
+    </CreateItem>
 
     <Message Condition="'$(TestTFM)' != ''" Text="Using test archives for TFM: $(TestTFM)" />
     <Message Text="Using OS-Specific test archives from: $(TestArchivesRoot)" />
@@ -194,11 +209,9 @@
           DependsOnTargets="CreateAzureStorage"
           Condition="'$(Performance)' != 'true' or '$(FuncTestsDisabled)' != 'true'">
     <!-- create item group of functional tests -->
-    <ItemGroup>
-      <FunctionalTest Include="$(TestArchivesRoot)**/*.zip" />
-      <FunctionalTest Include="$(AnyOSTestArchivesRoot)**/*.zip" />
-      <FunctionalTest Condition="'$(TargetsUnix)' == 'true'" Include="$(UnixTestArchivesRoot)**/*.zip" />
-    </ItemGroup>
+    <CreateItem Include="$(ForUploadList)">
+      <Output TaskParameter="Include" ItemName="FunctionalTest"/>
+    </CreateItem>
 
     <PropertyGroup>
       <OtherRunnerScriptArgs Condition="'$(FilterToTargetGroup)' == 'net46'">$(OtherRunnerScriptArgs) --xunit-test-type=desktop </OtherRunnerScriptArgs>
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
@@ -23,6 +23,7 @@
     <Compile Include="AzureHelper.cs" />
     <Compile Include="CreateAzureContainer.cs" />
     <Compile Include="DownloadFromAzure.cs" />
+    <Compile Include="FilterForUpload.cs" />
     <Compile Include="GetPerfTestAssemblies.cs" />
     <Compile Include="ListAzureContainers.cs" />
     <Compile Include="SendToEventHub.cs" />


src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Text;
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    public sealed class FilterForUpload : Task
+    {
+        //This task is used to select the appropriate test zips from the different output directories produced by the build,
```
ChadNedzlek: These should be proper XML doc comments. Just type "///" and VS will create the template for you

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,91 @@
...
+            
+            foreach (var filename in filesToSearch)
+            {
+                string selectedTest = string.Empty;
+                using (StreamReader sr = new StreamReader(File.OpenRead(filename)))
+                {
+                    string line;
+                    while ((line = sr.ReadLine()) != null)
+                    {
+                        //Each line is of the format "$(BinDir)/tests/AnyOS.AnyCPU.Debug/Tests/netcoreapp1.0"
```
ChadNedzlek: Is this _literally_ "$(BinDir)" or is the expanded value?  If's the later, we'll have big problems, because my "BinDir" directory might contain all sorts of text you'll match on.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,91 @@
...
+                        //Each test zip is of the format "$(BinDir)/tests/AnyOS.AnyCPU.Debug/archive/tests/netcoreapp1.0/Tests.zip"
```
ChadNedzlek: Is the test name not in the path anywhere? How do we upload multiple tests. If the test name _is_ in there somewhere, you can't use string contains, since the test name might itself contain things like "win"

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,91 @@
...
+                }
+                if (!string.IsNullOrEmpty(selectedTest))
+                {
+                    this._copyFilteredUploadItems.Add(new TaskItem(selectedTest));
+                }
+            }
+            return true;
+        }
+        
+        private int calculateMatchValue(string lines)
```
ChadNedzlek: Casing, and "lines" isn't a good name for this parameter

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,91 @@
...
+                    string line;
+                    while ((line = sr.ReadLine()) != null)
+                    {
+                        //Each line is of the format "$(BinDir)/tests/AnyOS.AnyCPU.Debug/Tests/netcoreapp1.0"
+                        //Each test zip is of the format "$(BinDir)/tests/AnyOS.AnyCPU.Debug/archive/tests/netcoreapp1.0/Tests.zip"
+                        //This is an implicit contract maintained in corefx 
+                        if (calculateMatchValue(line) > calculateMatchValue(selectedTest))
+                        {
+                            int lastIndexOfSlash = line.LastIndexOf('/');
+                            string parentDirectory = line.Substring(0, lastIndexOfSlash);
```
ChadNedzlek: Use Path.DirectoryName() instead of searching for the slashes yourself, especially since you use Path.Combine later.

src/Microsoft.DotNet.Build.Tasks/GenerateTestOutputDirList.cs
```diff
@@ -0,0 +1,60 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Xml.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GenerateTestOutputDirList : Task
```
ChadNedzlek: Nothing seems to use this Task. is it necessary? Or should it not be part of this change

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,91 @@
...
+                                TestArchivesRoot);
+                            //Another contract in corefx is that the test name should be the AssemblyName appended with .Tests.
+                            string testZipName = parentDirectory.Substring(lastIndexOfSlash + 1) + ".zip";
+                            selectedTest = Path.Combine(testArchiveDir, testTFMFolderInArchive, testZipName);
+                        }
+                    }
+                }
+                if (!string.IsNullOrEmpty(selectedTest))
+                {
+                    this._copyFilteredUploadItems.Add(new TaskItem(selectedTest));
```
ChadNedzlek: I think "this" is generally not used in CoreFX/BuildTools unless it's necessary (which it isn't here)

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: This is the expanded BinDir. I used this just for the comment.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,91 @@
...
+        [Required]
+        public string FilterToOSGroup { get; set; }
+        //This is the path under a given Platform/OSGroup/Configuration folder where the Test Archives are placed
+        [Required]
+        public string TestArchivesRoot { get; set; }
+
+        [Output]
+        public ITaskItem[] FilteredUploadItems
+        {
+            get { return this._copyFilteredUploadItems.ToArray(); }
```
ChadNedzlek: This field should be named the same as the property (or, better yet, just us an auto-prop and assign it at the end of Execute)

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: Here the capital Tests is stand in for an actual test name - Tests = Microsoft.CSharp.Tests.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
ChadNedzlek: In which case: fireworks. $(BinDir) could literally be any string under the sun that's a valid path, as we don't control where people clone the repo from. We need to only look specifically for the "AnyOS.AnyCPU.Debug" part and check that (preferably splitting it into it's sensible parts)

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
ChadNedzlek: See comment 3 lines up. We can't substring this then.

src/Microsoft.DotNet.Build.Tasks/GenerateTestOutputDirList.cs
karajas: This is used in tests.targets to generate the list for each file.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: resolved.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: resolved.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: resolved.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: resolved.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: resolved. I remove the value of BinDir from the generated test list during their creation.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: resolved.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// </summary>
+        /// This task is used to select the appropriate test zips from the different output directories produced by the build,
+        /// which are created based on OSGroup and TestTFM.
+        [Required]
+        public string TestOutputDirListsLocation { get; set; }
+
+        /// <summary>
+        /// The test tfm to select.
+        /// </summary>
+        /// This is the TestTFM folder that we want to select test zips from. This needs to be on the path to the test zip.
```
ChadNedzlek: Why are there two kinds of comments here?

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    FilteredUploadItemsList.Add(new TaskItem(selectedTest));
+                }
+            }
+
+            FilteredUploadItems = FilteredUploadItemsList.ToArray();
+            return true;
+        }
+        
+        private int CalculateMatchValue(string testPathToMatch)
```
ChadNedzlek: This still isn't safe, the "Example.Test" might be "CheckUnixWindows.Tests", which is going to do really, really bad things all over the place.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,94 @@
...
+        public string FilterToOSGroup { get; set; }
+
+        [Output]
+        public ITaskItem[] FilteredUploadItems { get; set; }
+
+        private readonly List<ITaskItem> FilteredUploadItemsList = new List<ITaskItem>();
+
+        public override bool Execute()
+        {
+
```
ChadNedzlek: Still some really wonky formatting here. An extra blank line and some strange spaces around List&lt;ITaskItem>

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,94 @@
...
+                    temp.Add(item);
+                }
+            }
+
+            Dictionary<string, List<ITaskItem>> map = new Dictionary<string, List<ITaskItem>>();
+            foreach (var item in temp)
+            {
+                string testName = Path.GetFileName(item.ItemSpec);
+                List<ITaskItem> values = null;
+                if (map.ContainsKey(testName))
```
ChadNedzlek: We've talked about this. Don't do ContainsKey followed by TryGetValue, they literally do the same work twice.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    values = null;
+                    map.TryGetValue(testName, out values);
+                }
+                else
+                {
+                    values = new List<ITaskItem>();
+                }
+                values.Add(item);
+                map[testName] = values;
```
ChadNedzlek: You shouldn't have to set this value except in the else (since it's already in the dictionary)

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// This is the OSGroup folder that we want to select test zips from. This does not require an exact match as some tests 
+        /// may or may not have an exact OSGroup folder match.
+        /// </summary>
+        [Required]
+        public string FilterToOSGroup { get; set; }
+
+        [Output]
+        public ITaskItem[] FilteredUploadItems { get; set; }
+
+        private readonly List<ITaskItem> FilteredUploadItemsList = new List<ITaskItem>();
```
ChadNedzlek: This should be a local variable. And the name should reflect that.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    public sealed class FilterForUpload : Task
+    {
+        /// <summary>
+        /// This task is used to select the appropriate test zips from the different output directories produced by the build,
+        /// which are created based on OSGroup and TestTFM.
+        /// </summary>
+        [Required]
+        public ITaskItem[] UnfilteredUploadItems { get; set; }
```
ChadNedzlek: Something in the comment should indicate that we require the TestTFM and OSPlatformConfig properties to be set.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,87 @@
...
+        public string BinDir { get; set; }
+
+        [Output]
+        public ITaskItem[] FilteredUploadItems { get; set; }
+
+        public override bool Execute()
+        {
+            List<ITaskItem> filteredBasedOnTestTfm =
+                UnfilteredUploadItems.Where(
+                    item => item.ItemSpec.Replace(BinDir, string.Empty).Contains(FilterToTestTFM)).ToList();
```
ChadNedzlek: Usually "Replace" when trying to make files relative is a hack, because whatever you are attempting to make it relative to _might_ appear a second time in the path ("../foo/pizza/pie/../foo" and you're trying to remove "../foo", it will mutilate that path).

karajas: Noted. The path in the ItemSpec is constructed using the BinDir here though.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            List<ITaskItem> localFilteredUploadItemsList = new List<ITaskItem>();
+
+            Dictionary<string, List<ITaskItem>> map = new Dictionary<string, List<ITaskItem>>();
+            foreach (var item in filteredBasedOnTestTfm)
+            {
+                string testName = Path.GetFileName(item.ItemSpec);
+                List<ITaskItem> values = null;
+                map.TryGetValue(testName, out values);
```
ChadNedzlek: if (!map.TryGetValue(testName, out values))

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
ChadNedzlek: Use this instead of "thing.Replace"

``` csharp
string myRandomPath = ...;
string binDir = ...;

binDir = Path.GetFullPath(binDir);
string fullRandomPath = Path.GetFullPath(myRandomPath);
if (fullRandomPath.StartsWith(binDir)) {
  return fullRandomPath.Substring(binDir.Length + 1)
else
  return myRandomPath;
```

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,103 @@
...
+                }
+                localFilteredUploadItemsList.Add(selectedItem);
+            }
+
+            FilteredUploadItems = localFilteredUploadItemsList.ToArray();
+            return true;
+        }
+
+
+        private string makeRelativePathTo(string path1, string path2)
```
ChadNedzlek: makeRelativePathTo -> MakeRelativePathTo
path1 -> targetPath
path2 -> basePath

ChadNedzlek:
One last tiny change, then squash, and then, assuming we got a good job with the correct number of tests in it (about 160ish) it will look good to me.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Build.CloudTestTasks
+{
+    public sealed class FilterForUpload : Task
```
tarekgh: > public sealed class FilterForUpload : Task [](start = 3, length = 43)

could you please add a comment describing what this task for and how it work?

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -194,11 +209,9 @@
...
           Condition="'$(Performance)' != 'true' or '$(FuncTestsDisabled)' != 'true'">
     <!-- create item group of functional tests -->
-    <ItemGroup>
-      <FunctionalTest Include="$(TestArchivesRoot)**/*.zip" />
-      <FunctionalTest Include="$(AnyOSTestArchivesRoot)**/*.zip" />
-      <FunctionalTest Condition="'$(TargetsUnix)' == 'true'" Include="$(UnixTestArchivesRoot)**/*.zip" />
-    </ItemGroup>
+    <CreateItem Include="$(ForUploadList)">
+      <Output TaskParameter="Include" ItemName="FunctionalTest"/>
+    </CreateItem>
```
tarekgh: why we didn't do 

<ItemGroup>
  <FunctionalTest Include="$(ForUploadList)" />
</ItemGroup>

karajas: It's using the same output of the previous one - ForUploadList?

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,103 @@
...
+            FilteredUploadItems = localFilteredUploadItemsList.ToArray();
+            return true;
+        }
+
+
+        private string makeRelativePathTo(string path1, string path2)
+        {
+            path2 = Path.GetFullPath(path2);
+            string fullPath1 = Path.GetFullPath(path1);
+            if (fullPath1.StartsWith(path2))
```
tarekgh: >ullPath1.StartsWith(path2)) [](start = 17, length = 27)

make thins ordinal ignore case to avoid running into weird cases

ChadNedzlek: I agree that's probably ok, since we can't detect the case-sensitivity of the underlying OS.  In general, I think our build tries hard not to have thing meaningfully differ by case, so it's probably safe.

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            foreach (var item in map)
+            {
+                List<ITaskItem> values = item.Value;
+                ITaskItem selectedItem = null;
+                foreach (var value in values)
+                {
+                    if (value.GetMetadata("OSPlatformConfig").Contains(FilterToOSGroup))
+                    {
+                        selectedItem = value;
```
tarekgh: >selectedItem = value; [](start = 24, length = 21)

should we break the loop if we find selectedItem?

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
```diff
@@ -89,14 +90,28 @@
...
+      </UnfilteredTestArchives>
       <!-- Only include Unix folders if supported by the current target OS -->
-      <ForUpload Condition="'$(TargetsUnix)' == 'true'" Include="$(UnixTestArchivesRoot)**/*.zip" />
+      <UnfilteredTestArchives Condition="'$(TargetsUnix)' == 'true'" Include="$(UnixTestArchivesRoot)**/*.zip" >
+        <OSPlatformConfig>$(UnixPlatformConfig)</OSPlatformConfig>
+      </UnfilteredTestArchives>
     </ItemGroup>
+    <FilterForUpload UnfilteredUploadItems="@(UnfilteredTestArchives)"
+             FilterToTestTFM="$(FilterToTestTFM)"
+             FilterToOSGroup="$(FilterToOSGroup)"
```
tarekgh: can we add like <Error ..> with the condition assert FilterToTestTFM/OSGroup are not empty. I assume this should be the case. right?

src/Microsoft.DotNet.Build.CloudTestTasks/FilterForUpload.cs
karajas: resolved

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets
tarekgh: I am ok to keep what you have but you can empty the previous list if you want. 

<FunctionalTest Remove="@(FunctionalTest )"/>


---
In reply to: [70545684](https://github.com/dotnet/buildtools/pull/870#discussion_r70545684) [](ancestors = 70545684)

tarekgh:
:shipit:

ChadNedzlek:
:shipit:

PR #875, opened 2016-07-13
Update Xamarin profiles - WCF contracts + classic

ericstj:
Xamarin has added facades for these WCF contracts which were already
part of PCL profiles that were supported.

Additionally I am including the list of classic assemblies which are
important as we start to add support for classic surface area to .NET
Core.

We need this to unblock our builds today since we're failing validation
for WCF and System.Json.

/cc @mhutch @marek-safar @StephenBonikowsky


ericstj:
These may not be 100% complete yet, and I'll follow up tomorrow using the actual lists from @marek-safar, but I need this to unblock our build.

PR #876, opened 2016-07-13
Update NuGet & enable higher versioned implementation assemblies

ericstj:
Update NuGet to latest so that we can understand netstandard1.7 & netcoreapp1.1.

Add validation and support moniker for netcoreapp1.1.

Enable higher versioned implementation assemblies so that folks can use a netstandard1.X implementation assembly to satisfy multiple netstandard refs (where refs may be higher than 1.X).  This will help us avoid cross-compiling for some netcoreapp1.0/1.1 assemblies.

/cc @joperezr @danmosemsft @karajas


dir.props
```diff
@@ -72,7 +72,7 @@
       <ExpectedVersion>2.1.0</ExpectedVersion>
     </ValidationPattern>
     <ValidationPattern Include="^(?i)(NuGet\..%2A)$">
-      <ExpectedVersion>3.5.0-beta2-1456</ExpectedVersion>
+      <ExpectedVersion>3.5.0-rc-1590</ExpectedVersion>
```
joperezr: >rc-1590< [](start = 29, length = 8)

I suppose this version is the one that knows about netstandard 1.7 and it's mapping to netcoreapp1.1?

joperezr:
LGTM. Are you planning on sending a different PR that adds System.Json to the Framework List? I ask because we need both in order to be able to consume this in dev/api

ericstj:
> Are you planning on sending a different PR that adds System.Json to the Framework List? 

That was done here: https://github.com/dotnet/buildtools/commit/1618fb88c50ec517ba8c334db955f8c0479d3ebc

And it's already been picked up by CoreFx master with https://github.com/dotnet/corefx/commit/8c62c72a8bd1edc7b295924372d1a05ece391517

We had to fast track an update yesterday to unblock the TFS build for WCF and I rolled the System.Json change into that.

joperezr:
oh sorry I missed that! cool thanks Eric

dir.props
ericstj: Yes, it is a version that has this change.  For pre-release builds of NuGet we don't often have a good LKG so I pick latest and then test it and use it if its good.  I've mirrored this to the buildtools feed.

joperezr:
:shipit:

PR #878, opened 2016-07-14
Add Dev Workflow documentation

maririos:
cc: @joperezr @jaredpar @ellismg @Priya91


Documentation/Dev-workflow.md
```diff
@@ -0,0 +1,38 @@
...
+This provides flexibility as with discrete tasks people can iterate on a specific one without needing to run the full workflow, other processes can reuse the tasks as needed and anyone can put together the tasks as wanted to create a customized workflow.
+
+## Process 
+![Dev Workflow process](images/Dev-workflow.jpg)
+
+## Tasks
+
+**Setup**
+
+Set of instructions in order to have an environment ready to build de repo. It is repo specific.
```
jaredpar: "build **the** repo"

Documentation/Dev-workflow.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Clean**
+
+The clean task is responsible of cleaning and leaving the developer enlistment/environment in a state as closest as the repository/code server.
+
+**Sync**
+
+The sync task gets the latest source history and the dependencies the build needs like for example restore the NuGet packages. Sync is the task in charge of eliminating all the network traffic when we build. This way we are able to hit the network only when we are intentional about doing it and then be able to build even in offline mode.
+
+**Build**
```
jaredpar: Is it the responsibility of the "Build" task to run "Sync" as well?  Or must callers ensure "Sync" was run before hand?

Documentation/Dev-workflow.md
```diff
@@ -0,0 +1,38 @@
...
+
+Builds the source code. The order of how we build depends on each repo or the situation.
+
+* Build product binaries: Builds the product binaries without need of hitting the network to restore packages. It doesn't build tests, run tests or builds packages.
+* Build packages: Builds the NuGet packages from the binaries that were built in the Build product binaries step.
+* Build tests: Builds the tests that are in the repository. It could also mean to queue a build in Helix.
+
+**Run Tests**
```
jaredpar: Is it the responsibility of "Test" to run "Build" as well?  Or must callers ensure "Build" was run before hand?

Documentation/Dev-workflow.md
maririos: It is responsibility of the callers. The "build" should failed if "sync" hasn't been run.

As an example, ProjectK and ProjectN have that behavior, and even though Corefx has the logic for that, it is not the default behavior yet.

Documentation/Dev-workflow.md
maririos: Responsibility of the callers. There can be a case when building tests or even binaries is not needed as the dev may use another mechanism for running tests, like Helix.

When we propose the Dev Workflow, we thought about concrete steps that were flexible and without forcing just one way to interact with the build. We propose a workflow to follow, but according to the needs of the dev, the steps can vary in order and occurrence.

Documentation/Dev-workflow.md
maririos: :| thanks! Spanish ...

maririos:
Is there more information that we would like to add? or is it ok if I merge this.

jaredpar:
Do we want to have an issue on `init-tools`?

maririos:
From what I understand, `init-tools` is in charge of downloading the dependencies that some repos need in order to build. In that case it would be part of the `Sync` task. 
Do we want to be more specific including implementation details for every task?

jaredpar:
> Do we want to be more specific including implementation details for every task?

I was hoping this document could become the document for defining what a given task:

- should accomplish
- can assume ran before it
- etc ... 

Mostly describing what it should do vs. how it is done.

maririos:
Oh, I'll work on that then and I'll update the PR. Thanks @jaredpar

maririos:
PR updated with more detail in each Dev Workflow task

maririos:
FYI @markwilkie

jaredpar:
:+1:

PR #883, opened 2016-07-15
Remove support for unknown parameters

maririos:
Before, the Run Command Tool was concatenating all the unknown parameters that were passed by the user. When running the command, it was passing this new string and it was the responsibility of the tool (i.e. msbuild) to work with those unknown parameters.

`build.cmd cross /p:ConfigurationGroup=Debug => msbuild cross /p:ConfigurationGroup=Debug`

With this change, the Run Command Tool will throw an error when there is an unknown parameter.
To pass extra parameters to the command, the user can use `--`, meaning that all the text that is after `--` will be passed and then it would be the responsibility of the tool (i.e. msbuild) to work with them.

```
build.cmd cross /p:ConfigurationGroup=Debug => Error: Parameter not recognized: cross
build.cmd -- cross /p:ConfigurationGroup=Debug => msbuild cross /p:ConfigurationGroup=Debug
````

cc: @joperezr @Priya91 @markwilkie

### DIFF
--- a/src/Run/CommandLine.cs
+++ b/src/Run/CommandLine.cs
@@ -322,7 +322,10 @@ class CommandLine
             }, (parser, ex) =>
             {
                 Console.Error.WriteLine("Error: " + ex.Message);
-                Console.Error.WriteLine("Use -? for help.");
+                
+                Console.ForegroundColor = ConsoleColor.DarkYellow;
+                Console.WriteLine("Our dev workflow has changed! Use -? for help in the new options we have and how to pass parameters now.");
+                Console.ResetColor();
             },
             setupContent, args);
         }
@@ -784,7 +787,7 @@ class CommandLine
             if(!string.IsNullOrEmpty(globalQualifiers))
             {
                 sb.AppendLine().Append('-', maxLineWidth - 1).AppendLine();
-                sb.Append("Global settings to all commands:").AppendLine();
+                sb.Append("Global Settings:").AppendLine();
                 sb.AppendLine();
                 sb.Append(globalQualifiers);
             }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by markwilkie** [comment]
I presume the space between the '--' and 'cross' was a typo?

**Comment by maririos** [comment]
@markwilkie not really. The `--` is the command to tell the tool that it should not process whatever is after it.
Or, is the idea to use `--` before every word/property that is not recognized by the tool?

**Comment by markwilkie** [comment]
I'm referring to the space between. --prop or -- prop?  I'm assuming it's --prop





On Fri, Jul 15, 2016 at 4:30 PM -0700, "Mariana Rios Flores" <notifications@github.com<mailto:notifications@github.com>> wrote:


@markwilkie<https://github.com/markwilkie> not really. The -- is the command to tell the tool that it should not process whatever is after it.
Or, is the idea to use -- before every word/property that is not recognized by the tool?

-

**Comment by maririos** [comment]
> I'm referring to the space between. --prop or -- prop? I'm assuming it's --prop 

No. I'm treating  `--` as a command, so it is independent of the properties you pass.

**Comment by maririos** [comment]
PR updated to support the scenario @markwilkie pointed out.
```
build.cmd cross /p:ConfigurationGroup=Debug => Error: Parameter not recognized: cross
build.cmd --cross /p:ConfigurationGroup=Debug => msbuild cross /p:ConfigurationGroup=Debug
```

**Comment by maririos** [comment]
I also added documentation about the `--` command when asking for help by typing `-?`

**Comment by joperezr** [comment]
I'm not 100% I love this convention, because `--` is actually commonly used to pass in properties to programs (i.e. for git, you can run something like `git fetch --all`). This makes the `--` not super intuitive into what is doing. Why not use another name, perhaps something like `otherParams` or `passThroughParams`

**Comment by maririos** [comment]
After reading more about `--` it is a common convention to use it to indicate that what it is after it should not be processed and it is going to be passed as it is.
The confusion comes when there is no space between the  `--` and the command or property because it seems like the user is passing a command that consists in more than a character, so instead of doing  `clean -a` I pass `clean --all`.

According to this I consider better if we go back to only support `-- {command}`

**Comment by maririos** [comment]
I added to this PR more documentation for the Devs and I changed the behavior of `--` to process after a space.

**Comment by mellinoe** [comment]
> After reading more about -- it is a common convention to use it to indicate that what it is after it should not be processed and it is going to be passed as it is.

This is how I understand "--" to be used. It's how the `dotnet` program treats it, as well as other programs I've seen. I'm a bit confused about calling it a "command", though. Maybe I'm just not understanding the terminology in this case? From what I understand, "--" by itself (without a word after it), should just be a delimiter for "extra args" to the command indicated by the stuff on the left side of "--".

```
<cmd> <specific-args> -- <unprocessed-args>
```

maririos:
@mellinoe You are right. The terminology `command` is not the right one for `--`. It is only a delimiter.
Thanks for the clarification!

maririos:
Is there something else we need here? or can I merge this PR.
Thanks!

markwilkie:
LGTM

PR #893, opened 2016-07-18
Fix clean.cmd when no option is provided

maririos:
When no option is provided clean.cmd should run `clean -b`.

cc: @joperezr @Priya91 @tyoverby

### DIFF
--- a/clean.cmd
+++ b/clean.cmd
@@ -12,7 +12,8 @@ if [%1] == [-all] (
   exit /b !ERRORLEVEL!
 )
 
-call %~dp0run.cmd clean %*
+if [%1]==[] set __args=-b
+call %~dp0run.cmd clean %__args% %*
 exit /b %ERRORLEVEL%
 
 :Usage


joperezr:
I wonder if the default should be --all instead

maririos:
When we created the scripts we thought about it and we decided to have something not that "hard" as getting to a state where everything is clean.
For now, I haven't heard a concern about this default option, but if there is more value on having `--all` as the default one we can totally change it.

joperezr:
mm I guess this is more of a point of view thing. To me, running clean should get me into a clean environment where I know that if I have no pending changes and try to build, I should get a green build(assuming of course that the repo's build is not broken). The reality is that build might still fail just for me, if my packages folder ran into a bad state or if I didn't restore the tools correctly. In any case, as I said earlier this is just my point of view.

maririos:
Until we have the common Public API we've been working on and that @jaredpar is leading, we could leave the default option of clean to be `-b`. If this turns out to be no longer valid, we could change the script :)

TyOverby:
I'd prefer `--all` to be the default as well.  It coincides with `make clean` on unix.

maririos:
For now, I'm going to merge this as it is. 
Thanks!

weshaggard:
I agree we shouldn't switch to "all" as the default as that also deletes things like our package cache which is not something we want to clean all the time.

PR #895, opened 2016-07-20
Add if file "_version.h" exist condition

maririos:
When we are building the native binaries in CoreFx, we have a condition in the script that checks if the file `_verison.h` is present. If it is not present then it would run the `GenerateVersionHeader` target.
```
if not exist "%__binDir%\obj\_version.h" (
    msbuild "%~dp0build.proj" /nologo /t:GenerateVersionHeader /p:GenerateNativeVersionInfo=true
)
```

With this change I'm deleting the condition from the script and adding it to the version.targets file.

cc: @joperezr


crummel:
Build finished.

joperezr:
This is fine as long as we ensure that we will run this target before the native build of corefx. Talked to @maririos offline and looks like that is her plan so that will work. LGTM

PR #899, opened 2016-07-20
Add null check.

Priya91:
cc @maririos

### DIFF
--- a/src/Run/Setup.cs
+++ b/src/Run/Setup.cs
@@ -20,6 +20,10 @@ namespace Microsoft.DotNet.Execute
 
         private string ParseSettingValue(string inputValue)
         {
+            // Can happen for target or internal valueTypes.
+            if (inputValue == null)
+                return inputValue;
+
             string value = string.Empty;
             int length = inputValue.Length;
             for (int i = 0; i < length; i++)


mellinoe:
Hm, seems kind of strange to pass null to a method called "ParseSettingValue". Can we rename the method or ensure null doesn't get passed into it? It looks like there's just two call sites, can we detect null in those places?

maririos:
What is the case here? Could you please provide an example?

Priya91:
> What is the case here? Could you please provide an example?

When a setting of target or internal valueType is used, there is no value for the key, so the Dictionary is storing default(string)=null for those cases.

Priya91:
>Hm, seems kind of strange to pass null to a method called "ParseSettingValue". Can we rename the method or ensure null doesn't get passed into it? 
ParseSettingValue expects a string to parse. Null is in the subset of valid values a string can have.
> It looks like there's just two call sites, can we detect null in those places?
No, there are 3 call sites, and this condition gets hit from ```FindSettingValue``` method. This happens when the ```value.DefaultValue``` is null. This happens because the JsonDeserializer is returning null for empty strings.

Priya91:
Ok after debugging with @maririos seems like i was using ```default``` instead of ```defaultValue``` object, and hence json was storing null for that.. After correcting that, this issue doesn't happen. Closing the PR.

PR #903, opened 2016-07-21
Returning exit code of the DefineParameters function

maririos:
If there was a failure when parsing arguments, the Run Command Tool was returning 0 as the exit code.

With this change the tool returns the correct exit code.

cc: @Priya91 @joperezr

### DIFF
--- a/src/Run/Executor.cs
+++ b/src/Run/Executor.cs
@@ -159,6 +159,7 @@ namespace Microsoft.DotNet.Execute
                 }
                 return jsonSetup.ExecuteCommand(executor.CommandSelectedByUser, paramSelected);
             }
+            //There was an error when parsing the user input, Define Parameters is in charge of printing an error message.
             return 1;
         }
     }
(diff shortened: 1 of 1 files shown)


src/Run/Executor.cs
```diff
@@ -144,25 +144,22 @@ public static int Main(string[] args)
...
-                            paramSelected.Add(param.Key);
-                        }
+                        paramSelected.Add(param.Key);
                     }
-                    return jsonSetup.ExecuteCommand(executor.CommandSelectedByUser, paramSelected);
                 }
+                return jsonSetup.ExecuteCommand(executor.CommandSelectedByUser, paramSelected);
             }
-            return 0;
+            return 1;
```
joperezr: >return 1; [](start = 12, length = 9)

I would add a comment or something here saying that this means that DefineParameters found an error, or I would add a `Console.error.WriteLine("Error: One or more errors were found on the passed-in parameters")`

joperezr:
other than that, this LGTM

src/Run/Executor.cs
maririos: Added!. Thanks Joe

PR #907, opened 2016-07-21
Documentation on compose commands

jaredpar:
Initial documentation describing a method for composing our repos together.

### DIFF
--- a/Documentation/RepoCompose.md
+++ b/Documentation/RepoCompose.md
@@ -1,0 +1,287 @@
+# Composing our repos
+
+In order to have automated builds that compose the output of our ever growing number of repos we need to get more structured data about their relationships.  Today repos provide nothing in the way of programitic access to dependencies.  All of the information is known by the owners and as a result the builds and composition is fully manual. 
+Going forward repos will be given contracts, or APIs, that provide structured data about there dependencies and outputs.  This will give us the necessary data to compose our repos in our growing set of scenarios:
+
+- Linux distro builds from source
+- Official Windows signed build
+- Builds to quickly test and deploy new packages across repos
+
+These contracts will be included in the `run` command specification.  This document in particular will be discussing the commands which allow us to compose the outputs of our repos.
+
+## Commands
+
+This document describes the commands used to compose different repos together.  There is a larger set of commands including `build`, `sync` and `clean` which is described separately. 
+
+### consumes
+
+The `consumes` command returns json output that describes all of the external artifacts consumed by this repo.  This includes NuGet feeds, packages and arbitrary files from the web or file system.  The format of all these items is describe in the Artifact Specification section below. 
+
+The artifacts are grouped into sections: 
+
+- Build Dependencies: artifacts which are referenced by the build output and include NuGet packages, MSI, etc ...  In order to support a composed build these are further divided into floating and static dependencies:
+    - Floating: dependencies where versions can change as a part of a composed build via the `change` command.  This commonly includes CoreFx, CoreClr, etc ...  
+    - Static: dependencies whose version do not change via the `change` command.  This is commonly used for SDK binaries, legacy tools, etc ...  
+- Toolset Dependencies: artifacts used in the production of build output.  This commonly includes NuGet.exe, compilers, etc ... 
+
+These sections are identified respectively by the following JSON sections:
+
+``` json
+"dependencies": {
+    "floating": { 
+        // Build artifacts
+    },
+    "static": { 
+        // Static artifacts
+    },
+    "toolset": { 
+        // Toolset artifacts
+    }
+}
+```
+
+The data in the output is further grouped by operating system.  Builds of the same repo on different operating system can reasonably consume a different set of resources.  The output of `consumes` reflect this and allows per operating system dependencies:
+
+``` json
+{
+    "os-windows": {
+        "dependencies": { } 
+    },
+    "os-linux": {
+        "dependencies": { }
+    }
+}
+```
+
+In the case the `consumes` output doesn't want to take advantage of OS specific dependencies it can specify `"os-all"` as a catch all. 
+
+In addition to artifacts the consume feed can also optionally list any machine prerequitsites needed to build or test the repo:
+
+``` json
+"prereq": { 
+    "Microsoft Visual Studio": "2015",
+    "CMake" : "1.0",
+}
+```
+
+A full sample output for `consumes` is available in the Samples section.
+
+### produces
+
+The `produces` command returns json output which describes the artifacts produced by the the repo.  This includes NuGet packages and file artifacts.  
+
+The output format for artifacts is special for `produces` because it lacks any hard location information.  For example: 
+
+- NuGet artifacts lack feeds
+- File artifacts lack a `"kind"` and supporting values
+
+This is because the `produces` command represents "what" a repo produces, not "where" the repo produces it.  The "where" portion is controled by the `publish` command.  External components will take the output of `produces`, add the location information in and feed it back to `publish`.  
+
+Like `consumes` the `produces` output is also grouped by the operating system:
+
+``` json
+{
+    "os-windows": {
+        "nuget": { },
+        "file": { }
+    },
+    "os-linux": {
+        "nuget": { },
+        "file": { }
+    }
+}
+```
+
+A ful sample output for `produces` is available in the Samples section.
+
+### change
+
+The `change` command is used to alter the floating build dependencies section.  It can establish new versions of NuGet packages, new locations to find file artifacts, different NuGet feeds, etc ...  
+
+This is the command which allow us to use the build output of one repo as the input of a dependent repo.  The first repo can build using a new output version (say beta5) and dependent repos can be changed to accept this new version.  
+
+This command operates by providing json whose format is a subset of the output of `consumes`.  In particular it will provide the `"dependencies.floating"` section.  
+
+``` json
+"dependencies" {
+    "floating": { 
+        "nuget": {
+            "packages" {
+                "MicroBuild.Core": "0.2.0",
+                "Microsoft.NETCore.Platforms": "1.0.1"
+            }
+        }
+    }
+}
+```
+
+A tool responsible for composing repos would ideally:
+
+1. Execute `run.cmd consumes` and capture the output.
+2. Alter the NuGet package versions in the json to have the correct build identifier: beta5, RTM, etc ..
+3. Execute `run.cmd change` and pass in the altered output.
+
+
+### publish
+
+The `publish` command takes a json input that describes the locations artifacts should be published to.  The input to this command is the output of `produces` that is augmented with location information.  
+
+## Artifact Specification
+
+The json describing artifacts is the same between the `consume`, `produces` and `publish` commands.  These items can be used anywhere artifacts are listed above.
+
+### NuGet packages
+
+The description for NuGet artifacts is it two parts:
+
+1. The set of feeds packages are being read from.
+2. The set of packages that are being consumed and their respective versions.
+
+Example:
+
+``` json
+"nuget": {
+    "feeds": [
+        { 
+           "name": "core-clr",
+           "value": "https://dotnet.myget.org/F/dotnet-coreclr/api/v3/index.json" 
+        },
+        {
+            "name": "dotnet-core",
+            "value": "https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"
+        }
+    ],
+    "packages": {
+        "MicroBuild.Core": "0.2.0",
+        "Microsoft.NETCore.Platforms": "1.0.1"
+    }
+}
+```
+
+### File 
+
+Any file which is not a NuGet package should be listed as a file artifact.  These can be downloaded from the web or copied from local places on the hard drive.  Each type of file entry will have a name uniquely identifying the artifact and a kind property specifying the remainder of the properties:
+
+    - uri: a property named `"uri"` will contain an absolute Uri for the artifact.
+    - filesystem: a property named `"location"` will contain an OS specific file path for the artifact.
+
+Example: 
+
+``` json
+"file": {
+    "nuget.exe":  {
+        "kind": "uri",
+        "uri": "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
+    }, 
+    "run.exe": { 
+        "kind": "filesystem",
+        "location": "c:\\tools\\run.exe"
+    }
+}
+```
+
+## Samples
+
+### consumes
+
+``` json
+{
+    "os-all": {
+        "dependencies": {
+            "floating": { 
+                 "nuget": {
+                    "feeds": [
+                        { 
+                           "name": "core-clr",
+                           "value": "https://dotnet.myget.org/F/dotnet-coreclr/api/v3/index.json" 
+                        },
+                        {
+                            "name": "dotnet-core",
+                            "value": "https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"
+                        }
+                    ],
+                    "packages" {
+                        "MicroBuild.Core": "0.2.0",
+                        "Microsoft.NETCore.Platforms": "1.0.1"
+                    }
+                },
+                "file": {
+                    "nuget.exe":  {
+                        "uri": "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
+                    }
+                }
+            },
+            "toolset": { 
+            },
+            "static": { 
+            }
+        },
+        "prereq": { 
+            "Microsoft Visual Studio": "2015",
+            "CMake" : "1.0"
+        } 
+    }
+}
+```
+
+### produces
+
+``` json
+{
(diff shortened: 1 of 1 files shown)


jaredpar:
CC @maririos, @tyoverby

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
+# Composing our repos
+
+In order to have automated builds that compose the output of our ever growing number of repos we need to get more structured data about their relationships.  Today repos provide nothing in the way of programitic access to dependencies.  All of the information is known by the owners and as a result the builds and composition is fully manual.
```
maririos: typo: programmatic

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
+# Composing our repos
+
+In order to have automated builds that compose the output of our ever growing number of repos we need to get more structured data about their relationships.  Today repos provide nothing in the way of programitic access to dependencies.  All of the information is known by the owners and as a result the builds and composition is fully manual. 
+Going forward repos will be given contracts, or APIs, that provide structured data about there dependencies and outputs.  This will give us the necessary data to compose our repos in our growing set of scenarios:
```
maririos: that provide structured data about ~~there~~ their dependencies...

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
...
+
+- Linux distro builds from source
+- Official Windows signed build
+- Builds to quickly test and deploy new packages across repos
+
+These contracts will be included in the `run` command specification.  This document in particular will be discussing the commands which allow us to compose the outputs of our repos.
+
+## Commands
+
+This document describes the commands used to compose different repos together.  There is a larger set of commands including `build`, `sync` and `clean` which is described separately.
```
maririos: Should we add a link here to the documentation of the dev workflow?

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
+# Composing our repos
+
+In order to have automated builds that compose the output of our ever growing number of repos we need to get more structured data about their relationships.  Today repos provide nothing in the way of programitic access to dependencies.  All of the information is known by the owners and as a result the builds and composition is fully manual. 
+Going forward repos will be given contracts, or APIs, that provide structured data about there dependencies and outputs.  This will give us the necessary data to compose our repos in our growing set of scenarios:
+
+- Linux distro builds from source
+- Official Windows signed build
+- Builds to quickly test and deploy new packages across repos
+
+These contracts will be included in the `run` command specification.  This document in particular will be discussing the commands which allow us to compose the outputs of our repos.
```
maririos: Should we add a link here to the documentation of the Run Command Tool?

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
...
+                "Microsoft.NETCore.Platforms": "1.0.1"
+            }
+        }
+    }
+}
+```
+
+A tool responsible for composing repos would ideally:
+
+1. Execute `run.cmd consumes` and capture the output.
```
maririos: Could you please add a graphic of how the commands interact with each other? I think that would help to understand how the flow and dependencies of the commands are.

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
...
+                        }
+                    ],
+                    "packages" {
+                        "MicroBuild.Core": "0.2.0",
+                        "Microsoft.NETCore.Platforms": "1.0.1"
+                    }
+                },
+                "file": {
+                    "nuget.exe":  {
+                        "uri": "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
```
maririos: the `kind` property is missing

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
...
+{
+    "os-all": {
+        "nuget": {
+            "packages" {
+                "MicroBuild.Core": "0.2.0",
+                "Microsoft.NETCore.Platforms": "1.0.1"
+            }
+        },
+        "file": {
+            "nuget.exe": { }
```
maririos: Same here, missing properties

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+}
+```
+
+## Implementation Stages
+
+Fully implementing the commands described in this document requires a decent amount of work.  The expectation is that it will be implemented in stages.  Each stage provides benefit to an existing tool or allows us to write a new tool for a major scenario:
+
+- Stage 1: NuGet build information.  NuGet packages are how the majority of repo artifacts are shared today.  Getting the NuGet section of the above commands implemented allows us to automate the majority of our composed builds.  This is really just a formalization of the commands repos already implement as a part of being inside the Maestro tool.
```
maririos: Is there documentation about the Maestro Tool? I don't know what it is :|

Documentation/RepoCompose.md
jaredpar: I'm a terrible speller 😦

Documentation/RepoCompose.md
jaredpar: Definitely.

Documentation/RepoCompose.md
jaredpar: I will try and get a graphic uploaded separately.  Want to get the documentation settled + ordering then I'll get some graphics for it.

Documentation/RepoCompose.md
jaredpar: fixed

Documentation/RepoCompose.md
jaredpar: This is intentional in the `produces` command.  The documentation for `produces` that the json describes "what" not "where".  

Think of it this way.  The job of Roslyn is to produce the Microsoft.CodeAnalysis nuget package.  Where that package is placed is defined by the arguments to the `publish` command.

Documentation/RepoCompose.md
jaredpar: @eerhardt may have some documentation for it.  The tool itself lives here

https://github.com/dotnet/versions

jaredpar:
Responded to PR feedback and repushed.

Documentation/RepoCompose.md
maririos: >File artifacts lack a "kind" and supporting values

I totally missed that. Thanks for the clarification!

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+    "Microsoft Visual Studio": "2015",
+    "CMake" : "1.0",
+}
+```
+
+A full sample output for `consumes` is available in the Samples section.
+
+### produces
+
+The `produces` command returns json output which describes the artifacts produced by the the repo.  This includes NuGet packages and file artifacts.
```
maririos: >produced by the ~~the~~ repo

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+        "file": { }
+    }
+}
+```
+
+A ful sample output for `produces` is available in the Samples section.
+
+### change
+
+The `change` command is used to alter the floating build dependencies section.  It can establish new versions of NuGet packages, new locations to find file artifacts, different NuGet feeds, etc ...
```
maririos: I know in the `consume` section you mentioned that `change` is changing that file, but could we say that again here to make it more obious? That way I don't have to go back to make sure I understood the flow correctly.

"...to alter the floating build dependencies section of the output file of `consume`."

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+
+This document describes the commands used to compose different repos together.  There is a larger set of commands including `build`, `sync` and `clean` which is [described separately](https://github.com/dotnet/buildtools/blob/master/Documentation/Dev-workflow.md). 
+
+### consumes
+
+The `consumes` command returns json output that describes all of the external artifacts consumed by this repo.  This includes NuGet feeds, packages and arbitrary files from the web or file system.  The format of all these items is describe in the Artifact Specification section below. 
+
+The artifacts are grouped into sections: 
+
+- Build Dependencies: artifacts which are referenced by the build output and include NuGet packages, MSI, etc ...  In order to support a composed build these are further divided into floating and static dependencies:
```
maririos: Does this means that when want to build, we are going to take a look at this file and according to the`where` and `what` we are going to restore/sync?

jaredpar: This information is more relevant to an orchestrator tool.  Essentially something like building up a build graph of the repos.  It would look at the consumes output, build dependencies, and combine that with the data in produces to determine build order.

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+    },
+    "os-linux": {
+        "dependencies": { }
+    }
+}
+```
+
+In the case the `consumes` output doesn't want to take advantage of OS specific dependencies it can specify `"os-all"` as a catch all. 
+
+In addition to artifacts the consume feed can also optionally list any machine prerequitsites needed to build or test the repo:
```
chcosta: I understand this is not a simple problem to solve (implement) for all cases, but do we have any mechanism whereby we can also check for these prereq's or enable to repo to plugin their own prereq checking on a given machine?  I'd assume the answer is no, but I'd be pleasantly surprised if it wasn't or if something was in the works.  This may be a question more targeted at @maririos .

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+        "file": { }
+    },
+    "os-linux": {
+        "nuget": { },
+        "file": { }
+    }
+}
+```
+
+A ful sample output for `produces` is available in the Samples section.
```
chcosta: *full

Documentation/RepoCompose.md
jaredpar: At this point no.  Most of our thinking around prereqs is been in the other direction.  Essentially:

> How can we get prereqs down to the absolute minimal possible set? 

Prereqs are a pain because it adds a lot of complexity to creating build environments.  Every prereq is a new installation step on build machines used for composed builds.  This is both a pain to setup and hard to maintain.  

Think back to the days where building VS components relied on having specific builds of VS installed.  Absolute nightmare to keep build machines up to date with that prereq.  Also no real solution if different repos depended on different builds of VS. 

For the time being we're trying to get all of our prereqs into a known state.  From there see what we can get rid of and over time widdle it down to the minimal possible set.  Once get there we can establish a build environment with only those pre-reqs installed and use it to verify that we don't slip up any.

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+
+The `publish` command takes a json input that describes the locations artifacts should be published to.  The input to this command is the output of `produces` that is augmented with location information.  
+
+## Artifact Specification
+
+The json describing artifacts is the same between the `consume`, `produces` and `publish` commands.  These items can be used anywhere artifacts are listed above.
+
+### NuGet packages
+
+The description for NuGet artifacts is it two parts:
```
chcosta: * "in two parts"

chcosta: Actually, probably something like "The description for NuGet artifacts consists of two parts:"

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+
+1. The set of feeds packages are being read from.
+2. The set of packages that are being consumed and their respective versions.
```
chcosta: Is there no direct correlation between a package and a feed?

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+```
+
+### File 
+
+Any file which is not a NuGet package should be listed as a file artifact.  These can be downloaded from the web or copied from local places on the hard drive.  Each type of file entry will have a name uniquely identifying the artifact and a kind property specifying the remainder of the properties:
+
+    - uri: a property named `"uri"` will contain an absolute Uri for the artifact.
+    - filesystem: a property named `"location"` will contain an OS specific file path for the artifact.
```
chcosta: Is the kind property ("uri" or "filesystem") and its associated property ("uri" or "location"), redundant?  Could you get away with just having a "uri" property or a "filesystem" property and getting rid of the "kind" property?

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+                        }
+                    ],
+                    "packages" {
+                        "MicroBuild.Core": "0.2.0",
+                        "Microsoft.NETCore.Platforms": "1.0.1"
+                    }
+                },
+                "file": {
+                    "nuget.exe":  {
+                        "kind": "url",
```
chcosta: url or uri?

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+                        "uri": "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
+                    }
+                }
+            },
+            "toolset": {
```
chcosta: Are these always required even if they're empty?  I'm ok with that as a requirement, just curious

jaredpar: yes they should be always present.

Documentation/RepoCompose.md
chcosta: > This is really just a formalization of the commands repos already implement as a part of being inside the Maestro tool.  

This sentence is awkward, and I don't really know what it means.  Now that I read it a few times, I'd assume that it's supposed to be "commands **that** repos already implement", and not referring to some "commands repos" I am unaware of.

Documentation/RepoCompose.md
jaredpar: No.  There are tendencies but no requirements for packages to be on a given feed.  The fact that Roslyn today gets CoreFX from nuget.org is an implementation detail.  Roslyn's only real requirement is for Systems.Collection.Immutable at version 1.2.0.  Where that package comes from shouldn't really matter.  

In general the goal of this system is to cleanly separate *what* we depend on from *where* we get it.  This separation is necessary to meet our requirements around composed and offline builds.  

For example consider a local composed build of CoreFX and Roslyn on Linux.  In that scenario Roslyn needs to be able to consume the output of building CoreFX.  The easiest way to do that is to publish the files on disk, and then use that disk location as a NuGet feed.

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+- Stage 4: Change command for all artifacts.  The ability to change a dependency from downloading nuget.exe from nuget.org to a place on the local file system.  This will give us the ability to implement fully offline builds.  
+
+## FAQ
+
+### How would a composer tool match the inputs / outputs of repos?
+
+The goal of this effort is to allow a composer tool to examine an arbitrary set of repos and establish a build order.  It can do so by examining the output of `consumes` and `produces` for each repo and establishing a dependency graph based on the outputs.  Artifacts from `produces` can be linked to floating build dependencies in `consumes`.
+
+But there is no guarantee that at any given point in time that the outputs of one repo will match exactly with the inputs of a depndent repo.  The names are likely to match but not the versions.  For example what is the chance for any given commit that dotnet/corefx is outputing System.Collections.Immutable at the exact version that dotnet/roslyn is cosuming for a given commit?  Probably fairly unlikely and getting them to agree requires coordination which is expensive and intended to be avoided by this very design.
```
chcosta: Remove the preposition ("but")

Documentation/RepoCompose.md
chcosta: Ah, I now see your note in the FAQ section about this, I hadn't gotten there yet.  Makes sense.

chcosta: I hope this doesn't lead to information rot as feeds change and people unfamiliar with the specifics don't know what to do with the feed list, you end up with unused feeds listed.  It's more of a problem / responsibility for the repo owner, but I could see it happening.

Documentation/RepoCompose.md
jaredpar: Doh!  If it's not abundantly clear yet, grammar is not my strong suit :frowning: 

I'm going to submit a follow up PR later today to fix all the issues you found.

Documentation/RepoCompose.md
```diff
@@ -0,0 +1,289 @@
...
+### Your samples have comments in JSON.  That's not legal.
+
+Yes they do.  It's a sample :smile:
+
+## Open Issues
+
+There are a series of open issues we need to track down here.
+
+- What is the full set of operating system identifiers?  The set I have listed above is just a place holder and was given no real thought.  More information is needed here. 
+- How can we relate file names between repos.  It's easy for us to understand that Microsoft.CodeAnalysis.nupkg is the same between repos.  How do we know that a repo which produces core-setup.msi is the input for a repo that consumes core-setup.msi?  Perhaps we have to say that output file identifiers must be unique across repos?  That seems like the simplest approach.
```
chcosta: How can we relate file names between repos**?**

PR #911, opened 2016-07-22
Update bt version to 1.0.26-prerelease-00621-06

maririos:


joperezr:
LGTM. Just FYI, you can't use an earlier version of buildtools since it will brake the official build, but this version already contains my fix.

maririos:
Thanks @joperezr

PR #912, opened 2016-07-22
TarGzFileCreateFromDirectory

piotrpMSFT:
Adds a TarGzFileCreateFromDirectory task to create tar.gz files on Linux & OS X. The task implements the same signature as ZipFileCreateFromDirectory. It requires that `tar` be available on the PATH.


weshaggard:
I don't like the assumption that tar is on the path but I think we can live with that for now until we have a library that support it.

LGTM

piotrpMSFT:
I'm not able to access the Pre-Merge signing tool due to Server 500. Can I ask someone for signoff?

stephentoub:
Done

PR #913, opened 2016-07-23
Enable S.P.CoreLib as core assembly for partial facades

weshaggard:
@stephentoub @mellinoe PTAL

This will enable us to start referencing S.P.CoreLib as opposed to mscorlib for CoreCLR based partial facades.


stephentoub:
LGTM

PR #914, opened 2016-07-23
TarGzFileExtractToDirectory

piotrpMSFT:
Adds TarGzFileExtractToDirectory
Also cleans up unnecessary override in TarGzFileCreateFromDirectory and out-of-date comments in ZipFileExtractToDirectory

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/TarGzFileExtractToDirectory.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/TarGzFileExtractToDirectory.cs
@@ -1,0 +1,112 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.IO;
+using System.IO.Compression;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public sealed class TarGzFileExtractToDirectory : ToolTask
+    {
+        /// <summary>
+        /// The path to the archive to extract.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


piotrpMSFT:
@weshaggard thoughts?

src/Microsoft.DotNet.Build.Tasks/TarGzFileExtractToDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+                if (OverwriteDestination == true)
+                {
+                    Log.LogMessage(MessageImportance.Low, "'{0}' already exists, trying to delete before unzipping...", DestinationDirectory);
+                    Directory.Delete(DestinationDirectory, recursive: true);
+                }
+                else
+                {
+                    Log.LogWarning("'{0}' already exists. Did you forget to set '{1}' to true?", DestinationDirectory, nameof(OverwriteDestination));
+
+                    retVal = false;
```
weshaggard: If you are going to make this fail should this be a LogError instead?

weshaggard:
One comment otherwise LGTM

piotrpMSFT:
@weshaggard I was going to drop by to chat about this. The Zip version of this capability LogWarning's so I maintained parity. However, Zip version does not even return false, so though it spits out a warning it will continue along its way to attempt a decompression. I'm glad to make them match one another but was not familiar with Zip's intention.

My proposal is that we make both LogError and both return 'false'. This is in keeping with my expectation as a user.

Returning false is not likely to be strictly necessary but I'd want to introduce the pattern for folks who choose not to LogError in the future.

Thoughts?

weshaggard:
>My proposal is that we make both LogError and both return 'false'. This is in keeping with my expectation as a user.

That seems reasonable to me.

Chrisboh:

PR #915, opened 2016-07-24
Add S.P.CoreLib to the ignore list for generation validation

weshaggard:
When we start referencing S.P.CoreLib generation validation starts complaining about it. So we need to add it to the ignore list similar to mscorlib. 

cc @stephentoub @ericstj


stephentoub:
LGTM

Issue #891, opened 2016-07-18
In VersionTools, separate Update and SubmitPR cleanly


dagood:
For a consumer of VersionTools to have a flag for submitting a PR, currently they have to do something like in https://github.com/dotnet/cli/pull/3878:
```C#
if (args.Length > 0 && string.Equals("--Update", args[0], StringComparison.OrdinalIgnoreCase))  
{
    dependencyUpdater.Update(updaters, buildInfos);  
}
else
{
    dependencyUpdater.UpdateAndSubmitPullRequestAsync(updaters, buildInfos);
}
```

Either way, the `dependencyUpdater` constructor takes github auth, but it's only used for the PR branch. Maybe it's just my opinion, but the args repeated for each branch seem bad to me too.

This API should be separated into an Updater and a PR Submitter. This would also allow the client to be used for other kinds of PRs.

PR #919, opened 2016-07-26
Make VersionTools UpdateDependencies update existing PR when available

dagood:
Resolves https://github.com/dotnet/buildtools/issues/835, although CoreFX/CoreCLR/WCF need to be changed to use this new shared functionality. Example change: https://github.com/dotnet/corefx/compare/master...dagood:example-versiontools-config. I'll thread it through those repos after this PR.

VersionTools picks a PR to update by:
 1. Searching for `repo:{project.Segments}+head:{headPrefix}+author:{author}+state:open`.
 2. If there are no results, a new PR is made. If there's more than one, the first one is examined for the next steps.
 3. If the PR *also* is confirmed to be based on the correct repository, have the right `head` prefix, and have the same authors (someone may have pushed a new commit to dotnet-bot's repo), then it does a force push to the PR's branch and updates the PR title.
 4. If the search found a PR but it isn't valid, then the upgrade is aborted entirely. Someone's made a manual change, or something's gone wrong.

I've exercised it on my corefx fork and it does what it's supposed to.

I also did some refactoring to resolve https://github.com/dotnet/buildtools/issues/891, so the CLI usage will need to change some.

(For context, the PR that introduced VersionTools: https://github.com/dotnet/buildtools/pull/863)

/cc @eerhardt @ericstj @weshaggard


src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
```diff
@@ -0,0 +1,211 @@
...
+            string queryString = $"repo:{project.Segments}+head:{headPrefix}+author:{author}+state:open";
+            string queryUrl = $"https://api.github.com/search/issues?q={queryString}&sort={sortType}&order=desc";
+
+            using (HttpResponseMessage response = await _httpClient.GetAsync(queryUrl))
+            {
+                response.EnsureSuccessStatusCode();
+
+                JObject responseContent = JObject.Parse(await response.Content.ReadAsStringAsync());
+
+                int totalCount = responseContent["total_count"].Value<int>();
```
eerhardt: Do you think it would make sense to have an object model for this object?  Like we do for GitHubPullRequest.

src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
```diff
@@ -0,0 +1,211 @@
...
+            using (HttpResponseMessage response = await _httpClient.SendAsync(request))
+            {
+                response.EnsureSuccessStatusCode();
+
+                Trace.TraceInformation($"Updated pull request #{number}.");
+                Trace.TraceInformation($"Pull request page: {await GetPullRequestUrlAsync(response)}");
+            }
+        }
+
+        public async Task<GitHubPullRequest> FirstPullRequestWithHeadOrDefaultAsync(
```
eerhardt: (nit): I don't think you need "OrDefault" in the name.

What do you think about the name "FindPullRequestAsync"  or "SearchPullRequestsAsync"?  I don't think we need to say "WithHead" either, since the parameters tell you what is being used for searching.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            _auth = auth;
+
+            Origin = origin;
+            UpstreamBranch = upstreamBranch;
+
+            GitAuthorName = gitAuthorName ?? auth.User;
+        }
+
+        public async Task CreateAsync(
```
eerhardt: (nit) "CreateOrUpdateAsync"?  Since this can and will do both.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            using (var client = new GitHubClient(_auth))
+            {
+                var originBranch = new GitHubBranch(
+                    $"{BranchPrefix}{DateTime.UtcNow.ToString("yyyyMMddhhmmss")}",
+                    Origin);
+
+                GitHubPullRequest pullRequestToUpdate = null;
+
+                if (!AlwaysCreateNewPullRequest)
```
eerhardt: Instead of a property, I think this may make more sense as a parameter to this method - "forceCreate" maybe?

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -0,0 +1,205 @@
...
+
+                        if (CanUpdatePullRequest(client, pullRequestToUpdate))
+                        {
+                            originBranch = new GitHubBranch(
+                                pullRequestToUpdate.Head.Ref,
+                                originBranch.Project);
+                        }
+                        else
+                        {
+                            Trace.TraceWarning($"Can't update pull request #{pullRequestToUpdate.Number}.");
```
eerhardt: If this logic is going to bail - it feels like this should be an error, not a warning.

What do you think about creating a new PR anyway in this scenario instead of returning and doing nothing?  That way we'd know pretty quickly when we got into a situation where we were supposed to update a PR, but couldn't.

eerhardt:
:shipit:

src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
dagood: Adding. I think I just didn't look back at this after I started making model classes for other things. :+1:

---
In reply to: [72436271](https://github.com/dotnet/buildtools/pull/919#discussion_r72436271) [](ancestors = 72436271)

src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
dagood: "OrDefault" was to match Linq naming: returns null rather than throwing if it's not found. Switching away from First gets rid of that odd expectation which is probably better. GitHub API docs call it a Search, so `SearchPullRequestsAsync` sounds good to me.


---
In reply to: [72436710](https://github.com/dotnet/buildtools/pull/919#discussion_r72436710) [](ancestors = 72436710)

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
dagood: Agreed, adding `bool forceCreate = false`.

---
In reply to: [72437273](https://github.com/dotnet/buildtools/pull/919#discussion_r72437273) [](ancestors = 72437273)

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
dagood: Hmm, well this is expected to happen once in a while, but true that if it happens without someone expecting it then that would be a problem. We do need some way of telling when this happens. I'm not sure if opening a new PR is the right way, though, since if this was intentional, we don't want someone else to see a newer green PR and merge it, making it harder to apply whatever fixup is being done to the old PR. Maybe it could just append to the description (I already have the API for that) or add a comment?

---
In reply to: [72437681](https://github.com/dotnet/buildtools/pull/919#discussion_r72437681) [](ancestors = 72437681)

eerhardt: I'd be fine if we did *something* that will get noticed.  A thing to remember about these trace messages is that this process is usually run on a VSO build definition that isn't really monitored closely.  So just writing warnings and errors probably won't get noticed by someone - until they start poking around why we don't get new PRs generated.

If we can somehow make unexpected situations noticeable (like adding a comment to the existing PR saying why @dotnet-bot can't update the PR) that would be ideal IMO.

dagood: Good points. I'll change this to add a comment with the specific reason. (Comment because I realized those send emails, but changes to the PR body don't.)

dagood:
You can see an example of the "can't update" comments here: https://github.com/dagood/corefx/pull/7.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependencies.cs
```diff
@@ -39,26 +42,44 @@ public class UpdateDependencies : Task
...
+                    string suggestedMessage = updateResults.GetSuggestedCommitMessage();
+                    string body = string.Empty;
+                    if (NotifyGitHubUsers != null)
+                    {
+                        body += PullRequestCreator.NotificationString(NotifyGitHubUsers.Select(item => item.ItemSpec));
+                    }
 
-            Trace.Listeners.RemoveMsBuildTraceListeners(listeners);
+                    var prCreator = new PullRequestCreator(gitHubAuth, origin, upstreamBranch, GitHubAuthor);
+                    prCreator.CreateOrUpdateAsync(suggestedMessage, suggestedMessage, body).Wait();
```
eerhardt: I think you forgot to pass in AlwaysCreateNewPullRequest

dagood: Yep, fixed.

eerhardt:
:shipit:

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -0,0 +1,201 @@
...
+        public async Task CreateOrUpdateAsync(
+            string commitMessage,
+            string title,
+            string description,
+            bool forceCreate = false)
+        {
+            using (var client = new GitHubClient(_auth))
+            {
+                var originBranch = new GitHubBranch(
+                    $"{BranchPrefix}{DateTime.UtcNow.ToString("yyyyMMddhhmmss")}",
```
weshaggard: Do we want to actually create a ton of branches based on timestamps? When I was thinking about this workflow I was thinking we would reuse the same branch and simply keep updating it and ensuring there is a PR for it if there are new updates in it. Granted I've not thought this all the way through so there may be issues with such an approach.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -0,0 +1,201 @@
...
+            {
+                return $"Owner of head repo '{pullRequest.Head.User.Login}' is not '{Origin.Owner}'";
+            }
+            if (!pullRequest.Head.Ref.StartsWith(BranchPrefix))
+            {
+                return $"Ref name '{pullRequest.Head.Ref}' does not start with '{BranchPrefix}'";
+            }
+
+            GitCommit commit = client.GetCommitAsync(Origin, pullRequest.Head.Sha).Result;
+            if (commit.Author.Name != GitAuthorName)
```
weshaggard: Why do both the author and committer need to match? What about scenarios where someone needs to manually fix-up a PR? In such case I think it might be OK for the committer to be whoever but the Author be dotnet-bot (or whichever service account we use).

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        private void PushNewCommit(
+            GitHubBranch branch,
+            string commitMessage,
+            bool forcePush)
+        {
+            // Set committer in environment rather than through command start options because net45
+            // doesn't have that feature.
+            Environment.SetEnvironmentVariable("GIT_COMMITTER_NAME", GitAuthorName);
```
weshaggard: What does net45 have to do with this? I would much prefer we just pass these as arguments instead of env variables.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
eerhardt: Maybe some sort of "strategy" and let the caller decide how to name the branch?

I used time stamps originally because I needed a unique branch name.

weshaggard:
Thanks for trying to tackle this problem I love the direction. I've left a few comments but otherwise looks like a good starting point to me. 

cc @stephentoub @jkotas as two folks that deal with these version update PR's a lot.

cc @jaredpar as the configuration you planning to do at https://github.com/dotnet/corefx/compare/master...dagood:example-versiontools-config is coming very close the configuration that he is proposing here https://github.com/dotnet/buildtools/pull/907

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
dagood: A new branch is only created here for a fresh PR, and updates are done by force pushing to the same branch. (`force` cleans up merge conflicts, and avoids polluting the commit tree too much.) If people consistently delete the branch after merging in the github UI there shouldn't be any waste.

That said, always using the same branch would work too and not require people to hit the delete button, but we'd still need to search to find the existing PR, so the code wouldn't be significantly simpler.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
dagood: This is here to help with exactly that scenario, so that the bot doesn't overwrite (with `push force`) the fixup changes someone's making. (This doesn't block the PR from being merged or anything.)

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
dagood: I wasn't able to find arguments for committer, unfortunately. The closest I could find is running `git config ...` instead of using the env vars, but that didn't seem any better to me.

net45 here is about using `Environment.SetEnvironmentVariable` instead of setting it on the `ProcessStartInfo`, but looking back at CLI's command class, this should have been fixed with https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/Command.cs#L169-L173. Not sure why I had to remove that in the earlier PR.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
dagood: Hmm, realized that we do need a different upgrade branch for each target branch in the repo. `master`, `rel/1.0.0`, `dev/api`, etc. need separate branches per-PR. Doable, but we'd need to bring in that info somehow--another parameter passed in by the Maestro build, taken from VSTS environment, or another `git` command are the ways I can think of.

@eerhardt Didn't see your comment when I posted mine earlier (I must not have refreshed), but yeah, that could do it. If we want to go with a `git` command or the environment we could pass the branch name into the strategy.

dagood:
Yeah, this should definitely use the same configuration as the repo API work when possible. So far I don't see overlap, but I think we should make them overlap. :smile:

I have some questions about the repo API stuff when it comes to the versions repo, but maybe that's just not specified yet. @jaredpar Is #907 the right place to ask about that or is there a larger discussion going on?

(One definite thing is that adding a full list of packages simplifies https://github.com/dotnet/buildtools/issues/732 by making the download unnecessary. Validation will only need to check for internal consistency.)

dagood:
Some updates:

* `IUpdateBranchNamingStrategy`, where `SingleBranchNamingStrategy` returns names like `master-UpdateDependencies` and `dev/api-UpdateDependencies`, with no date suffix. Examples: https://github.com/dagood/corefx/pulls. (I only have the config set up in master, so the `dev/api` one is ugly.)
* Moved setting the committer environment variable back inside the utility class brought over from CLI.

weshaggard:
LGTM. Lets try to get this merged and integrated into corefx and coreclr as it will help a lot with the current version update PRs. 

Thanks for doing this work @dagood

markwilkie:
Nice job @dagood

dagood:
Thanks! Rebased my feedback commits and merged. I'll work on getting this into corefx and coreclr.

Issue #869, opened 2016-07-12
apigen.exe generating code that does not compile.


esdrubal:
While generating api of mscorlib v4.6.2, apigen.exe  generates code that does not compile:

Using v4.6.2 reference assemblies

And doing on windows:
```
genapi.exe v4.6.2/mscorlib.dll > out.cs
csc.exe /unsafe /w:0 out.cs
```

Here is  one of the problematic lines:
```[System.Runtime.CompilerServices.IteratorStateMachineAttribute(typeof(System.Collections.Concurrent.ConcurrentDictionary<TKey, TValue>.<GetEnumerator>d__35))]```

ericstj:
Dup of https://github.com/dotnet/buildtools/issues/658

ericstj:
Fixed in https://github.com/dotnet/buildtools/pull/922

Issue #904, opened 2016-07-21
Reference Assembly generation


raffaeler:
I am using the following nuget package to generate the sources for a reference assembly:
https://www.nuget.org/packages/Microsoft.DotNet.BuildTools.GenAPI

The generated code contains the following attribute for each method returning a task:
        [System.Runtime.CompilerServices.AsyncStateMachineAttribute(typeof(Xyz.Abc.MyClass.<GetAuthenticationCookie>d__34))]

The problem is the "AsyncStateMachineAttribute" is not available in all the framework flavors.
Is this attribute really needed?
Can I specify not to write it in the generated source code?

Thank you

ericstj:
I believe this is a duplicate of https://github.com/dotnet/buildtools/issues/658

ericstj:
Fixed in https://github.com/dotnet/buildtools/pull/922

Issue #658, opened 2016-04-22
GenAPI generates custom attributes with references to private types


ericstj:
Bug is here: https://github.com/dotnet/buildtools/blob/b6663b70b5956f352af75d7d9accf89baa13929c/src/Microsoft.Cci.Extensions/Filters/PublicOnlyCciFilter.cs#L66

We need to check the visibility of all the types used in the custom attribute arguments.

An easy repro is to run GenAPI on an assembly with public async methods/events.
GenAPI will generate source like:
`[System.Runtime.CompilerServices.AsyncStateMachineAttribute(typeof(MyType.<MyAsyncMethod>d__13))]` where `MyType.<MyAsyncMethod>d__13` is private (not to mention invalid C# syntax).

To workaround this you can specify -excludeApiList with an exclusion file containing
```
T:System.Runtime.CompilerServices.AsyncStateMachineAttribute
T:System.Runtime.CompilerServices.IteratorStateMachineAttribute
```
There may be other attributes that need exclusion, but this pattern should work as a temporary workaround.

/cc @dotMorten

ratishphilip:
@ericstj 
I am facing the same issue.
I have added the NuGet package to my project. My project contains a public async method.

Could you provide details about how I can specify an -excludeApiList. Where should the exclusion file be placed?

ericstj:
You can put the exclusion file anywhere, for instance in your project root, eg: excludeApis.txt

Then in the project file, you just need to specify some additional parameters to pass to genapi.

For example:
```xml
<PropertyGroup>
  <GenAPIAdditionalParameters>$(GenAPIAdditionalParameters) -excludeApiList "$(MSBuildProjectDirectory)\excludeApis.txt"</GenAPIAdditionalParameters>
</PropertyGroup>
```

ratishphilip:
@ericstj 
Thank you very much. 👍

ratishphilip:
@ericstj 
Can we expect this bug to be fixed in the near future so as to avoid this workaround?

ericstj:
I haven't looked at what a fix would look like here.  I'll mark this one up-for-grabs so anyone can pick it up.  /cc @mellinoe @weshaggard

ericstj:
Fixed in https://github.com/dotnet/buildtools/pull/922

ratishphilip:
@ericstj  Will the updated NuGet package be available soon?

ericstj:
[...]

PR #922, opened 2016-07-28
Attribute, operators, extern modifier fixes

esdrubal:
Fixes 3 issues found while compiling generated code for v4.6.1 mscorlib.dll.

Attributes not visible outside the assembly are now filtered.
Attributes using arguments with typeof expression of a type not visible
outside assembly are also filtered.

When we are generating structs for builtin types such as System.Double, we
cannot use the keyword double while declaring operators.

This fixes bad generated methods such as:
public static decimal operator +(decimal d1, decimal d2) { throw null; }
public static implicit operator decimal (char value) { throw null; }

This changes the generator to not replace a type for its keyword when we
are generating the type itself.

DllImportAttribute was added to methods without modifier extern,
this would cause an error while compiling them.

Fixes:
https://github.com/dotnet/buildtools/issues/869
https://github.com/dotnet/buildtools/issues/904
https://github.com/dotnet/buildtools/issues/658

### DIFF
--- a/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
+++ b/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
@@ -50,6 +50,13 @@ namespace Microsoft.Cci.Writers.CSharp
             WriteEmptyBody();
         }
 
+        private void WriteTypeName(ITypeReference type, ITypeReference containingType, bool isDynamic = false)
+        {
+            var useKeywords = containingType.GetTypeName() != type.GetTypeName();
+
+            WriteTypeName(type, isDynamic: isDynamic, useTypeKeywords: useKeywords);
+        }
+
         private void WriteMethodDefinitionSignature(IMethodDefinition method, string name)
         {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
```diff
@@ -50,6 +50,13 @@ private void WriteDestructor(IMethodDefinition method)
             WriteEmptyBody();
         }
 
+        private void WriteTypeName(ITypeReference type, ITypeReference containingType, bool isDynamic = false)
+        {
+            var useKeywords = containingType.GetTypeName() != type.GetTypeName ();
```
weshaggard: nit: `type.GetTypeName();` you ad an extra space.

weshaggard:
Thanks for submitting these fixes @esdrubal. The changes look good only one minor nit.

esdrubal:
@weshaggard thanks for looking into this, nit fixed.

weshaggard:
@esdrubal Thanks for the contribution.

ericstj:
Awesome, thanks for this fix!

PR #925, opened 2016-07-29
Correct the casing of Microsoft.Net.Compilers.netcore

eerhardt:
This is breaking with new NuGet in the CLI.

@dagood @weshaggard


dagood:
LGTM. This is just NuGet getting more strict? (I tried to reason out how the package-cache-lowercasing change could have caused it and came up empty.)

eerhardt:
It's a combination of the new NuGet casing changes mixed with bugs in the CLI that we are fixing in the latest versions.  However, buildtools requires an older version of the CLI, which we can't fix the bugs in.

weshaggard:
Do we need to lowercase the entire package id?

dagood:
Eric explained a little more offline, and I think it's more about the casing being wrong or inconsistent, not lowercase. (It's `Microsoft.Net.Compilers.netcore` in the feed.)

Being able to use pre-lowercasing NuGet lockfiles with post-change NuGet does seem like it would need the whole thing to be lowercased to me, though... but I'm not sure if that's what CLI needs.

PR #929, opened 2016-08-01
Rename "version" to "versions" when referring to dotnet/versions repository

dagood:
Makes naming consistent with the repository name "dotnet/versions" and parameters in the existing `UpdatePublishedVersions.ps1`. I found this issue when I was removing the powershell script functionality and noticed the msbuild properties' names don't match up.

`VersionsRepoUpdater` is called `VersionRepoUpdater` in the CLI repository, but I think `Versions` makes more sense here too.

Minor change, but I wanted to make it before I worked VersionTools into the repositories, making it hard to change later.

@eerhardt It looks like CLI isn't depending on `VersionsRepoUpdater` yet--is that true, making this free to change?


eerhardt:
:shipit:

PR #931, opened 2016-08-02
Allow adding args to restore calls in init-tools.cmd/sh

dagood:
Allows callers of init-tools.cmd/sh to set `INIT_TOOLS_RESTORE_ARGS`/`__INIT_TOOLS_RESTORE_ARGS` to add arguments to the restore commands. This also refactors the restore calls to reuse the arg strings where appropriate.

~~Makes init-tools.sh "dotnet restore" calls configurable by appending `__BUILDTOOLS_EXTRA_RESTORE_ARGS` to the restore commands called in the package's `init-tools.sh`. Specifically, this lets core-setup pass in `--disable-parallel` to access a workaround.~~

~~Also upgrades the `Microsoft.NETCore.App` runtime dependency for MSBuild, since the fixes to core-setup include upgrading the version of CLI being used and the old version of CLI was apparently the only place that specific version of App was available.~~

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -15,6 +15,8 @@ set MICROBUILD_VERSION=0.2.0
 set PORTABLETARGETS_VERSION=0.1.1-dev
 set ROSLYNCOMPILERS_VERSION=1.3.0-beta1-20160429-01
 set MSBUILD_CONTENT_JSON={"dependencies": { "MicroBuild.Core": "%MICROBUILD_VERSION%", "Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%"},"frameworks": {"netcoreapp1.0": {},"net46": {}}}
+set INIT_TOOLS_RESTORE_ARGS=--source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json %INIT_TOOLS_RESTORE_ARGS%
+set TOOLRUNTIME_RESTORE_ARGS=--source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json %INIT_TOOLS_RESTORE_ARGS%
 
 if not exist "%PROJECT_DIR%" (
   echo ERROR: Cannot find project root path at [%PROJECT_DIR%]. Please pass in the source directory as the 1st parameter.
@@ -30,7 +32,7 @@ ROBOCOPY "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%" /E
 
 set TOOLRUNTIME_PROJECTJSON=%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\project.json
 @echo on
-call "%DOTNET_CMD%" restore "%TOOLRUNTIME_PROJECTJSON%" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json
+call "%DOTNET_CMD%" restore "%TOOLRUNTIME_PROJECTJSON%" %TOOLRUNTIME_RESTORE_ARGS%
 set RESTORE_ERROR_LEVEL=%ERRORLEVEL%
 @echo off
 if not [%RESTORE_ERROR_LEVEL%]==[0] (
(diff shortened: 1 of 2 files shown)


ericstj:
Do the same in init-tools.cmd?

dagood:
It won't be used Windows-side, but might as well. This PR shows some context: https://github.com/dotnet/core-setup/pull/224.

I also added a change to allow the package names to be lowercased when doing the copies--this is needed for the NuGet/CLI upgrade. (The `[Xx]` should keep it compatible with new and old.)

dagood:
I added an upgrade to the MSBuild runtimedependency file because it's also required for core-setup to be fixed, which is the real point of the PR. I'll update the description to reflect that. I think that makes this PR unmergeable for now, since that version has to correspond to the installed shared framework/CLI. (And there is no `1.0.0-rc3-002733` on MyGet!)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/MSBuild.runtimeconfig.json
```diff
@@ -2,7 +2,7 @@
   "runtimeOptions": {
     "framework": {
       "name": "Microsoft.NETCore.App",
-      "version": "1.0.0-rc3-002733"
```
joperezr: We shouldn't do this now since as you say it makes it unmergable. There are bigger changes that will have to happen in order for this to work as expected, for instance, we need to:
-Update all of the runtimeconfig.json in the buildtools package since now all of them will run with RTM shared framework. (this is the easy part)
- Make all of the tools compile against RTM shared framework since that is what they will have at runtime. (This is the hard part, since it is not only updating the dependencies in buildtools, but also the ones for msbuild and for roslyn)
- Update the CLI in all of the repos that consume buildtools, in order for them to be able to consume this change.

cc: @mellinoe

dagood: Agreed, unfortunately this blocks the core-setup fix because it needs to update CLI as part of getting builds working again. (And there's no reason to merge this PR if it doesn't fix core-setup.) Good to know about those extra steps, thanks: my Linux build with this package works, but the Windows one doesn't, and that second point might be the cause. I haven't looked very far into the Windows error yet, though.

ericstj:
LGTM

src/Microsoft.DotNet.Build.Tasks/PackageFiles/MSBuild.runtimeconfig.json
mellinoe: We're going to need to update CLI soon in order to support new distros (among other reasons), so it's something we need to do anyways.

dagood:
I've backed out the `1.0.0-rc3... -> 1.0.0` and case insensitivity changes, so this PR won't conflict with one about upgrading the CLI. (Those commits are still available on https://github.com/dotnet/buildtools/compare/master...dagood:fix-core-setup)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -30,7 +30,7 @@ ROBOCOPY "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%" /E
 
 set TOOLRUNTIME_PROJECTJSON=%BUILDTOOLS_PACKAGE_DIR%\tool-runtime\project.json
 @echo on
-call "%DOTNET_CMD%" restore "%TOOLRUNTIME_PROJECTJSON%" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json
+call "%DOTNET_CMD%" restore "%TOOLRUNTIME_PROJECTJSON%" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json %BUILDTOOLS_EXTRA_RESTORE_ARGS%
```
weshaggard: To make this feel less of a hack can we change this variable to be used for all the arguments with built in extensibility.  I'm thinking something like:

```
set INIT_TOOLS_RESTORE_ARGS=--source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json %INIT_TOOLS_RESTORE_ARGS%

... 
call "%DOTNET_CMD%" restore "%TOOLRUNTIME_PROJECTJSON%" %INIT_TOOLS_RESTORE_ARGS%
```

That way we can share them in this file as well as provide extra ones outside if necessary. Something similar for the sh script as well.

dagood: I really like this idea, I'll do it that way.

dagood:
Updated the PR title/description and pushed suggested changes.

Some testing on OSX led me to keep the `--packages ...` arg in the command itself, because I couldn't find a way to allow spaces in the path while jamming it into a variable. But when I was trying it out, I found that corefx's `init-tools.sh` doesn't support spaces in filenames anyway, even before it gets to the buildtools `init-tools.sh`.

weshaggard:
LGTM

PR #932, opened 2016-08-02
Redist NuPkg assets

ericstj:
See the individual commits for details.  This PR adds support for redistributing assets from the last shipped packages.  It also validates the latest package to ensure it supports all frameworks supported by the last package.

/cc @weshaggard @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
@@ -1,0 +1,105 @@
+﻿using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using NuGet.Versioning;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Build.Tasks.Packaging
+{
+    public class GetLastStablePackage : PackagingTask
+    {
+        /// <summary>
+        /// List of packages to look up.
+        /// The closest StablePackage version that is less than the version of each of these packages will be returned in LastStablePackages.
+        ///   Identity: Package ID
+        ///   Version: Package version.
+        /// </summary>
+        [Required]
+        public ITaskItem[] LatestPackages { get; set; }
+        
+        /// <summary>
+        /// List of previously shipped packages.
+        ///   Identity: Package ID
+        ///   Version: Package version.
+        /// </summary>
+        [Required]
+        public ITaskItem[] StablePackages { get; set; }
+
+        /// <summary>
+        /// Latest version from StablePackages for all packages in LatestPackages.
+        /// If a version isn't found for an item in LatestPackage that will not be included in this set.
+        /// </summary>
+        [Output]
+        public ITaskItem[] LastStablePackages { get; set; }
+
+        public override bool Execute()
+        {
+            if (LatestPackages == null || LatestPackages.Length == 0)
+            {
+                return true;
+            }
+
+            Dictionary<string, Version> latestPackages = new Dictionary<string, Version>();
+            Dictionary<string, Version> lastStablePackages = new Dictionary<string, Version>();
+            
+            foreach(var latestPackage in LatestPackages)
+            {
+                var packageId = latestPackage.ItemSpec;
+
+                var versionString = latestPackage.GetMetadata("Version");
+                NuGetVersion nuGetVersion = null;
+                if (versionString == null || !NuGetVersion.TryParse(versionString, out nuGetVersion))
+                {
+                    Log.LogMessage($"Could not parse version {versionString} for LatestPackage {packageId}, will use latest stable.");
+                }
+
+                latestPackages[packageId] = nuGetVersion?.Version;
+            }
+
+            foreach(var stablePackage in StablePackages)
+            {
+                var packageId = stablePackage.ItemSpec;
+
+                Version latestVersion;
+                if (!latestPackages.TryGetValue(packageId, out latestVersion))
+                {
+                    continue;
+                }
+
+                var versionString = stablePackage.GetMetadata("Version");
+                Version stableVersion;
+                if (versionString == null || !Version.TryParse(versionString, out stableVersion))
+                {
+                    Log.LogError($"Could not parse version {versionString} for StablePackage {packageId}");
+                    continue;
+                }
+
+                // need a version less than current version
+                if (latestVersion != null && stableVersion >= latestVersion)
+                {
+                    continue;
+                }
+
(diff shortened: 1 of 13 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
```diff
@@ -0,0 +1,359 @@
...
+
+        /// <summary>
+        /// Set to partial paths to exclude from file harvesting.
+        /// </summary>
+        public string[] PathsToExclude { get; set; }
+
+        /// <summary>
+        /// Set to partial paths to suppress from both file and support harvesting.
+        /// </summary>
+        public string[] PathsToSuppress { get; set; }
```
ericstj: Do you think it would make more sense if I had a parameter that controlled suppression of the harvested support frameworks rather than making someone realize that in order to suppress support for those they have to suppress the files that resulted in supporting that framework?

ericstj:
https://github.com/ericstj/corefx/commit/cca213b1262c44d73ff1c4917d7f8702a59962b1 has the changes to corefx to pick this up and start using it.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            foreach(var stablePackage in StablePackages)
+            {
+                var packageId = stablePackage.ItemSpec;
+
+                Version latestVersion;
+                if (!latestPackages.TryGetValue(packageId, out latestVersion))
+                {
+                    continue;
```
joperezr: >continue; [](start = 20, length = 9)

I'm not sure I understand when this can happen? In theory if you have a stable package there should be an entry in latestPackages already for it, or not?

ericstj:
https://github.com/ericstj/corefx/commit/ebe98307189415345461d3906812a74f356ff4fd shows how this can simplify library and package projects.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
ericstj: The stable package list has all stable packages for the repo: https://github.com/dotnet/corefx/blob/master/pkg/Microsoft.Private.PackageBaseline/stable.packages.props

We're pulling out just the ones we care about.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Build.Tasks.Packaging
+{
+    public class GetLastStablePackage : PackagingTask
+    {
+        /// <summary>
+        /// List of packages to look up.  Stable version less than this version will be returned.
```
weshaggard: I'm not sure what "Stable version less than this version will be returned." means exactly.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
```diff
@@ -0,0 +1,104 @@
...
+        ///   Identity: Package ID
+        ///   Version: Package version.
+        /// </summary>
+        [Required]
+        public ITaskItem[] LatestPackages { get; set; }
```
weshaggard: Where is the LastestPackages list coming from?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
weshaggard: I do think it might be easier to understand if we suppress based on frameworks but we probably also need a mechanism for path suppression as well. For now lets just stick with what you have and see if what issues we run into, if any, and then adjust.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
```diff
@@ -0,0 +1,364 @@
...
+                    bool hasRuntimeAsset, hasRuntimePlaceHolder;
+                    NuGetAssetResolver.ExamineAssets(Log, "Runtime", package, target, runtimeAssets, out hasRuntimeAsset, out hasRuntimePlaceHolder);
+
+                    isSupported &= hasCompileAsset == hasRuntimeAsset;
+                    isSupported &= hasCompilePlaceHolder == hasRuntimePlaceHolder;
+
+                    if (!isSupported)
+                    {
+                        Log.LogMessage(LogImportance.Low, $"Skipping {fx} because it is not supported on {target}.");
+                        continue;
```
weshaggard: break

weshaggard:
LGTM

markwilkie:
Including @crummel who's looking into making buildtools more consumable by folks not CoreFx

ericstj:
@markwilkie / @crummel these changes only impact folks using pkgproj and only if they opt into this behavior.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
ericstj: We have path exclusion already.  I think framework is better here, based on our discussion I will change it.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
ericstj: I'll clarify the comment.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLastStablePackage.cs
ericstj: It's calculated in packaging.targets based on the pkgproj being built and it's run-time dependencies.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
ericstj: Sure

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
ericstj: Actually now I remember why I made this paths.  We need to both exclude the old files that supported that framework *and* exclude the framework support.  So rather than do that separately I created a single param.  I'm going to leave as is for now.  This should be a rare exception.

PR #939, opened 2016-08-05
Preserve BOM when updating files in VersionTools

### LINKED ISSUE #935: VersionTools: FileRegexPackageUpdater adds UTF-8 BOM
Microsoft.DotNet.VersionTools.Dependencies.FileRegexPackageUpdater adds a UTF-8 BOM to all files it updates with its Update method.  It should instead write the file with the same BOM as it had when reading it.

naamunds:
Fixes #935.

cc: @dagood @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
@@ -37,7 +37,7 @@ namespace Microsoft.DotNet.VersionTools.Dependencies
                 }
                 catch (Exception e)
                 {
-                    Trace.TraceWarning($"Non-fatal exception occurred reading or writing '{projectJsonFile}'. Skipping file. Exception: {e}. ");
+                    Trace.TraceWarning($"Non-fatal exception occurred processing '{projectJsonFile}'. Skipping file. Exception: {e}. ");
                     continue;
                 }
 
--- a/src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.DotNet.VersionTools.Util;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using NuGet.Packaging.Core;
@@ -9,7 +10,6 @@ using NuGet.Versioning;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.IO;
 using System.Linq;
 
 namespace Microsoft.DotNet.VersionTools.Dependencies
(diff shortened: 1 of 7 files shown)


naamunds:
cc: @MichaelSimons

src/Microsoft.DotNet.VersionTools/Util/FileUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+    static class FileUtils
+    {
+        public static bool ReplaceFileContents(string path, Func<string, string> replacement)
+        {
+            string contents;
+            Encoding encoding;
+
+            // Atttempt to preserve the file's encoding, using ASCII encoding (which has no BOM) if the file's
+            // encoding cannot be detected. 
+            using (StreamReader reader = new StreamReader(new FileStream(path, FileMode.Open), Encoding.ASCII, true))
```
eerhardt: Why ASCII?  I would think UTF8 would be a better default.

src/Microsoft.DotNet.VersionTools/Util/FileUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.DotNet.VersionTools.Util
+{
+    static class FileUtils
```
eerhardt: (nit) `internal`

eerhardt:
:shipit: after addressing my comments.

src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
```diff
@@ -29,33 +29,20 @@ public IEnumerable<BuildInfo> Update(IEnumerable<BuildInfo> buildInfos)
...
+
                 try
                 {
-                    projectRoot = ReadProject(projectJsonFile);
+                    FileUtils.ReplaceFileContents(projectJsonFile, contents => ReplaceAllDependencyVersions(contents, projectJsonFile, buildInfos, out buildInfosUsed));
                 }
                 catch (Exception e)
                 {
-                    Trace.TraceWarning($"Non-fatal exception occurred reading '{projectJsonFile}'. Skipping file. Exception: {e}. ");
+                    Trace.TraceWarning($"Non-fatal exception occurred reading or writing '{projectJsonFile}'. Skipping file. Exception: {e}. ");
```
dagood: This is more than just reading and writing, on second thought--searching for the buildinfo and looking at json is in here too. I'm actually thinking that since the Maestro runs aren't typically monitored, we should have some way to put these warnings in the PR description so someone knows about them. (Probably something to file an issue for, though--this is fine by me for now.)

MichaelSimons: +1 on this suggestion - but agree it is a separate issue.

---
In reply to: [73746886](https://github.com/dotnet/buildtools/pull/939#discussion_r73746886) [](ancestors = 73746886)

MichaelSimons:
Looks good, I have no additional comments to the ones already made.

src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
dagood: Created https://github.com/dotnet/buildtools/issues/942 to track it.

dagood:
Looks good to me too, modulo comments.

src/Microsoft.DotNet.VersionTools/Util/FileUtils.cs
naamunds: If the file does not have a BOM, then the StreamReader cannot detect its encoding, so it uses the default encoding.  Thus, we want the default encoding to correspond to no BOM, so that's why ASCII is the default here.

If UTF-8 were the default here, then it would add the UTF-8 BOM when the file originally had no BOM, which is current behavior that this PR is fixing.

src/Microsoft.DotNet.VersionTools/Util/FileUtils.cs
naamunds: I omitted the access modifier to follow the pattern used by the other classes in this folder. @dagood do you know why this pattern was established?  Do you have a preference on whether explicit access modifiers should be used here?

eerhardt: See bullet number 5:  https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

dagood: That was my mistake--dotnet style guides say to use the modifier rather than the implicit value. If you don't mind adding it to the rest of them that would be great. :smile:

src/Microsoft.DotNet.VersionTools/Util/FileUtils.cs
eerhardt: If you create the UTF8Encoding object yourself, you can create one that doesn't write a BOM.

`new UTF8Encoding(encoderShouldEmitUTF8Identifier: false)`

This is better because it handles unicode characters in the file.

src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
naamunds: I don't see anything that could reasonably throw an exception in the code searching for the build info, as it just grabs the Id and Version properties from the LatestPackages PackageIdentity objects, so I think it's fine for the warning message to say that it occurred reading or writing the project.json file.

Agreed about putting the warnings in the PR description as a separate issue.

src/Microsoft.DotNet.VersionTools/Util/FileUtils.cs
naamunds: Sounds good. I've added the `internal` access modifier to all four classes/structs in this folder.

src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
dagood: I'm thinking about https://github.com/naamunds/buildtools/blob/85446ddcbb1f7d15ee4491b5fea3d5266b69aaee/src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs#L87-L118. I think it wouldn't be hard to craft some json that crashes that. (And I wouldn't classify it as reading or writing.)

naamunds: I agree that the JSON-parsing code could throw an exception, but I guess I was thinking that parsing could be considered part of reading, but I can also see using a more narrow definition of file-reading which excludes parsing.  Regardless, I've changed it to "processing".

dagood: That isn't just parsing, it's changing the contents of the in-memory json model. But anyway, fixed. :+1:

eerhardt:
:shipit:

PR #943, opened 2016-08-08
Add mechanism to specify destination dir for supplemental test data.

jhendrixMSFT:
The current mechanism for copying supplemental test data assumes that the
RecusiveDir metadata is defined.  Added new metadata, DestinationDir, for
copying supplemental test data to be preferred over RecursiveDir.


jhendrixMSFT:
@schaabs @chcosta @MattGal This is for copying the test-runtime project.json into the out directory so it can be included in the zip file.

chcosta:
LGTM

Issue #942, opened 2016-08-08
On auto-upgrade, post errors/warnings to PR to make them visible


dagood:
Currently they only show up in the Maestro build, and nobody checks those unless something goes obviously wrong. It would be useful to post errors and warnings into the PRs, giving them high visibility.

This should be doable by making a trace listener in the msbuild task, but we should aim for something that's reusable with only VersionTools.

dagood:
Small thing to add to this: making sure two loaded buildinfos don't share any packages. That's a consistency error in the versions repo, and this is a good way to surface it.

ViktorHofer:
// Auto-generated message

Going forward, the .NET team is using https://github.com/dotnet/arcade to develop the code and issues formerly in this repository. Feel free to reopen/move this issue if it still applies to servicing branches in this repository, or source code which now resides in the Arcade repository.

PR #944, opened 2016-08-08
Add a way to config VersionTools task to upgrade full version for a package id

dagood:
This enables upgrading things like [`CoreClrPackageVersion`](https://github.com/dotnet/coreclr/blob/0bb191d/tests/dir.props#L47), so we can use the VersionTools msbuild task to upgrade coreclr.

/cc @eerhardt @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependencies.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependencies.cs
@@ -99,13 +99,30 @@ namespace Microsoft.DotNet.Build.Tasks.VersionTools
             {
                 foreach (ITaskItem step in XmlUpdateStep)
                 {
-                    yield return new FileRegexReleaseUpdater
+                    string buildInfoName = step.GetMetadata("BuildInfoName");
+                    string packageId = step.GetMetadata("PackageId");
+
+                    FileRegexUpdater updater;
+
+                    if (!string.IsNullOrEmpty(buildInfoName))
                     {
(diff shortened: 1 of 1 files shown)


weshaggard:
Why wouldn't you just use the buildinfo for this as well, even if it is a checked in buildinfo file? Is this just temporary?

dagood:
It does use the buildinfo, check out https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.VersionTools/Dependencies/FileRegexPackageUpdater.cs. It just searches all buildinfos available and finds the specified package no matter where it comes from. Usage is like this:

```
<XmlUpdateStep Include="CoreClr">
	<Path>$(MSBuildThisFileFullPath)</Path>
	<ElementName>CoreClrPackageVersion</ElementName>
	<PackageId>Microsoft.NETCore.Runtime.CoreCLR</PackageId>
</XmlUpdateStep>
```

weshaggard:
What happens if there is a conflict? As in the package is in multiple build infos? I guess that is more of a question about FileRegexPackageUpdater then your actual change. Looking at the code it just takes the first one which may potentially lead to issues at some point. 

This change LGTM

dagood:
Yeah, if a single package is in multiple buildinfos that's a problem for broader things too, like project.json auto-upgrade and the idea of validation. I'd say it would be a consistency issue in the versions repo, and this is a good place to catch those and make them visible. Added as todo with https://github.com/dotnet/buildtools/issues/942.

Thanks, merging.

PR #946, opened 2016-08-09
Start on adding MSBuild guidelines.

JeremyKuhne:
Start on getting some coding guidelines together for our msbuild files.

@weshaggard, @jhendrixMSFT, @markwilkie

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -1,0 +1,14 @@
+# Documents Index
+
+## Intro to .NET Build Tools
+
+## Project Docs
+
+* [Developer Workflow](Dev-workflow.md)
+* [Repo Composition](RepoCompose.md)
+* [The Run Command Tool](RunCommand.md)
+
+## Coding Guidelines
+
+* [MSBuild Coding Style](msbuild-coding-style.md)
+* [MSBuild Best Practices](msbuild-best-practices.md)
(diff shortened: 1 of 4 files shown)


danmosemsft:
One suggestion on conditions: properties in MSBuild conditions are often treated as booleans, even though they aren't constrained to be true or false.

MSBuild does understand booleans, so for example
`'$(DoStuff)' == 'true'`
will evaluate to true if DoStuff is 'on' 'true' or 'yes'. In practice DoStuff is almost always either 'true' 'false' or blank. So the question is, what should blank mean? MSBuild effectively treats it as a 3rd value matching either 'true' nor 'false'. In a particular build process, for a paritcular property, there will inevitably be a convention either implicit or explicit for what blank means.

There is another issue and that is double negatives. In my experience, a property named negatively eg "DoNotDoStuff" quickly becomes hard to rationalize about. The human brain cannot easily parse `'$(DoNotDoStuff)' == 'false'`. 

Combining these two issues I believe it is best for 

- All properties to be "positive". Where negative is normally the case, default the value of the positive property to false.
- All blank values are considered false. This means all conditions against boolean properties compare to 'true' and never to false or blank.

AlfredoMS:
I like this.
@JeremyKuhne will you follow up?

Chrisboh:

PR #947, opened 2016-08-10
Use 2.0.0-beta3 roslyn compilers

chcosta:
**Windows**
BuildTools defines what version of the Roslyn compiler it restores (to tools\Microsoft.NETCore.Compilers) via buildtools:src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd.  BuildTools then copies Microsoft.NETCore.Compilers to tools\net45\roslyn for use on Windows desktop. 

**Linux**
Version of Roslyn used is determined by the BuildTools:src\Microsoft.DotNet.Build.Tasks\PackageFiles\tools-runtime\project.json file which gets restored on the host OS to the Tools directory (tools\csc.exe).  Linux uses the Microsoft.Net.Compilers.netcore package (vs Microsoft.Net.Compilers on Windows).

/cc @weshaggard @wtgodbe


joperezr:
I assume you tested that these compilers run with the framework specified in the runtime.config file?

chcosta:
I have additional changes for consuming this in CoreFx.  No changes to runtime.config were required, but please let me know if alternative testing is required.

weshaggard:
Perhaps we were already in this world but I assume this now means we aren't using the Roslyn library that  are part of the SharedFX which means they are not cross-gen'ed. That will likely have a noticeable impact on our build times. Have you noticed any decrease?

joperezr:
@weshaggard I believe we have always been in this world for csc.exe itself. This is because ToolRuntime folder has always had a csc.exe which is not crossgened, and this means that the core host will use that instead of using the one inside the sharedFx. The Fx dependencies of csc.exe though are in fact crossgened since we pick the ones inside the cli.

chcosta:
In corefx,  on windows,  the default is still too use the shared compiler,  you have to opt into using this Roslyn compiler via "UseSharedCompilation=false".  On Linux,  the default is to use this Rosslyn compiler.  My changes (for this and subsequently corefx) do not affect the current behavior.

weshaggard:
That's a shame. Can we file a follow-up work-item to get all our corefx builds using the same compiler? If we don't we are quickly going to hit issues when people try to use the new features and they will work on some builds but not others.

joperezr:
@chcosta I thought that the shared compiler that we use by default in Windows is also comming from the roslyn package, or is that not the case?

chcosta:
That is not the case, the default compiler used is from MSBuild 

"C:\Program Files (x86)\MSBuild\14.0\bin\csc.exe"

joperezr:
I know that is what the window says, but it might be worth to launch process explorer and take a look at what is actually running, because when you are using the shared compiler server, the output still will say that is running csc.exe even though it is actually running VBCSCompiler.exe. So let's just make sure which one is actually running.

chcosta:
Ok, putting all the pieces together.  Our build claims it is running csc.exe from MSBuild.  I've validated that what it was actually doing was then running vbcscompiler from Roslyn tools.  It wasn't clear to me exactly what vbcscompiler did, my assumption was that it was just responsible for choosing csc and launching csc, but joperezr explained that it is the compiler itself.  Sooo.... I think that joperezr's last comment is correct, and there is no issue here.  I'll close the issue I created to align compiler use across builds with some additional info.

weshaggard:
Even with VBCSCompiler.exe are we sure it is using the compiler from the package? Can you point me at the properties we are setting to make that happen?

chcosta:
I have a subsequent PR for corefx which updates the properties there.  This PR deals with updating the version of the compiler in buildtools.  I can put up the corefx PR as well, I was blocking on this PR for that..

mellinoe:
LGTM. We need to update the compiler version to run our tools on a newer shared framework, so we should get this in soon.

chcosta:
The CoreFx PR is https://github.com/dotnet/corefx/pull/10928

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <RoslynVersion>2.0.0-beta3</RoslynVersion>
```
weshaggard: Can we avoid having to hard code this version and package name here by moving the logic to the init-tools scripts and just copy the props files into the ToolsRuntimePath similar to the portable targets?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -9,7 +9,7 @@
         "Microsoft.Build.Utilities.Core": "0.1.0-preview-00024-160610",
         "Microsoft.Build.Targets": "0.1.0-preview-00024-160610",
         "Microsoft.Build": "0.1.0-preview-00024-160610",
-        "Microsoft.Net.Compilers.netcore": "1.3.0-beta1-20160429-01",
+        "Microsoft.Net.Compilers.netcore": "2.0.0-beta3",
```
weshaggard: I really don't like seeing this version copied multiple places :( Do we have any strategies to try and avoid this duplication?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -19,4 +20,19 @@
   <PropertyGroup>
     <GetNuGetPackageVersionsDependsOn>CreateVersionFileDuringBuild</GetNuGetPackageVersionsDependsOn>
   </PropertyGroup>
+
+  <!-- Common BuildTools properties -->
+  <PropertyGroup>
```
weshaggard: Thanks for starting the migration of a number of these common properties. I think we should clean out a lot more of the other properties as well but that can be done as a separate work item.

cc @crummel @markwilkie

chcosta:
Thanks @weshaggard , I've filed an issue to track removing some more of the duplication related to the Roslyn versions.  If there is no other feedback, I will submit today so that I can focus on consuming this in CoreFx and getting a clean Jenkins run there.

weshaggard:
The rest LGTM. Thanks for migrating more to BuildTools.

PR #948, opened 2016-08-10
Merging uap10.1 change

SedarG:
@weshaggard , Could you please take a look at the PR.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -412,6 +412,7 @@
     <OutOfBoxFramework Include="netcore50">
       <RuntimeId>win10</RuntimeId>
     </OutOfBoxFramework>
+    <OutOfBoxFramework Include="uap10.1"/>
   </ItemGroup>
   <Target Name="EnsureOOBFramework"
           DependsOnTargets="AddPlaceholders;ConvertItems">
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -412,6 +412,7 @@
     <OutOfBoxFramework Include="netcore50">
       <RuntimeId>win10</RuntimeId>
     </OutOfBoxFramework>
+    <OutOfBoxFramework Condition="'$(IsUap101Defined)'=='true'" Include="uap10.1"/>
```
weshaggard: Where do you expect IsUap101Defined to be set?

SedarG: It's going to be defined in packaging.props on corefx repo on relevant branches where uap10.1 is 'supported' (i.e.: dev/api). 
for ref: https://github.com/SedarG/corefx/commit/cf8d4c0e4a1c5758e27cfa41a08a45e85f491df9

weshaggard: Is it not better to add the OutOfBoxFramework item in corefx as the temporary measure instead of introducing this property?

SedarG: Hmm. That's also possible. I don't like splitting OutOfBoxFramework items across multiple files, but I admit that it'll make like much easier.

PR #949, opened 2016-08-10
Fix copying of supplemental test data.

jhendrixMSFT:
The change introduced in commit 04599fa caused the ordering of items
between SupplementalTestData and TestDir/OutDir items to be incorrect when
a group of SupplementalTestData items contained data with and without the
DestinationDir metadata set.  To avoid this mismatch we now coalesce both
groups of SupplementalDataData into an intermediate item group and use
that when copying.


jhendrixMSFT:
Merging to unblock publishing of test assets to VSTS drop storage.

PR #951, opened 2016-08-11
Significantly reduce the amount of time spent in DesignTimeBuilds

hughbe:
Before, renaming a file took 30s+, now takes less than 1s. Delay in opening projects is also almost completely gone

<details>
<summary>Previous Design Time Build Log</summary>
```
115 ms  HarvestStablePackage                       1 calls
118 ms  ResolveAssemblyReferences                  9 calls
124 ms  OpenSourceSign                             3 calls
212 ms  GetLatestCommitHash                        4 calls
1152 ms  RestorePackages                            2 calls
2055 ms  ValidateApiCompatForSrc                    2 calls
3067 ms  FillPartialFacade                          2 calls
5396 ms  CoreCompile                                8 calls
13395 ms  BuildAllProjects                           1 calls
23182 ms  ExpandProjectReferences                    1 calls
```
</details>

<details>
<summary>New Design Time Build Log</summary>
```
2 ms  ValidatePackageVersions                    1 calls
4 ms  GetPackageFiles                            1 calls
6 ms  CoreCompile                                1 calls
7 ms  ResolvePkgProjReferences                   1 calls
9 ms  FilterTargetingPackResolvedNugetPackages   1 calls
16 ms  ResolveAssemblyReferences                  1 calls
78 ms  GetLatestCommitHash                        1 calls
126 ms  HarvestStablePackage                       1 calls
144 ms  ResolveNuGetPackages                       1 calls
1043 ms  ResolveProjectReferences                   1 calls
```
</details>

Fixes dotnet/corefx#10666

The log end summary is above. If you chaps have any ideas about other optimisations, I'm all ears.

/cc @weshaggard @davkean @Pilchie

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -258,13 +258,12 @@
        This helps minimize the number of project evaluations and targets that build in order
        to determine the contents of the package-->
   <Target Name="ExpandProjectReferences"
-          DependsOnTargets="SplitProjectReferences"
-          Condition="'$(DesignTimeBuild)' == 'false'">
+          DependsOnTargets="SplitProjectReferences">
 
     <!-- Only rebuild project references when specified -->
     <MSBuild Targets="Build"
              BuildInParallel="$(BuildInParallel)"
-             Condition="'$(BuildPackageLibraryReferences)' == 'true'"
+             Condition="'$(BuildPackageLibraryReferences)' == 'true' AND '$(DesignTimeBuild)' == 'false'"
              Projects="@(_NonPkgProjProjectReference)"
              Properties="$(ProjectProperties)"
              ContinueOnError="WarnAndContinue"/>
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -258,7 +258,8 @@
        This helps minimize the number of project evaluations and targets that build in order
        to determine the contents of the package-->
   <Target Name="ExpandProjectReferences"
-          DependsOnTargets="SplitProjectReferences">
+          DependsOnTargets="SplitProjectReferences"
+          Condition="'$(DesignTimeBuild)' == 'false'">
 
     <!-- Only rebuild project references when specified -->
     <MSBuild Targets="Build"


davkean:
As a matter of interest, did the doc help you diagnose this?

hughbe:
@davkean yup it was really useful, without it I wouldn't have been able to do it. Thanks!

I'd suggest mentioning on the docs that a high level "Target Performance Summary" can be found at the bottom, that breaks targets down into their relevant time spent. I spent a while trying to go down chains of targets and their dependencies removing them one by one and and seeing their impact (very time consuming) until I saw that summary

davkean:
@ericstj What's the point of the MSBuild call that calls "Build" again?

davkean:
@hughbe Thanks for the feedback on the doc, I've made a change to it to include it.

weshaggard:
@hughbe Thanks for digging into this issue. While I believe your fix might actually help the problem in VS I think it may actually be the wrong way to fix it as it only scopes this to VS/Design time builds. I suspect there is a larger thing we need to look into further as @davkean suggests. @ericstj is on vacation this week but he would be the best person to help figure out the best way to fix this.

hughbe:
@weshaggard thanks for your comments. Is your suggestion that we investigate further why the target takes so long anyway and did that?
The reason why I simply disabled this target was that several other targets are also disabled on design time builds, such as restoring packages, so it seems there is a precedent. 

@davkean is there a way to disable resolving project references in a design time build? Corefx projects have their own complex build system, and although I'm not completely familiar with buildtools, i don't think that target is useful.

Thanmd

weshaggard:
It is possible that disabling this as part of the design time build may be the right thing to do. However, I feel like that is just a bandaid fix for the real problem which is that we probably shouldn't be calling the "Build" target all the time inside or outside of VS context. So there is likely something else we need to figure out. 

In general whenever I see us needing to disable part of our targets only in VS/Design Time red flags go up for me because that naturally means that we will have differences in our builds depending on how we build, which I want to avoid as much as possible because I want our builds to be consistent in all the ways we build.

hughbe:
@weshaggard understood! I'll wait for comments/suggestions by Eric then

Thanks

ericstj:
I think we should probably make this target more closely resolve the `ResolveProjectReferences` which has further conditions on how it treats ProjectReferences when building in the IDE vs the commandline.  I'm ok with this small fix for now but I agree with @weshaggard that we probably have further issues with how we handle project references.

@hughbe did you notice anything broken by excluding this target?  I'd expect that this would cause some design-time builds to fail if they depended on the types that were brought in through the pkgproj reference.

hughbe:
@ericstj thanks for taking a look. 

I've only taken a look at one project/solution: System.Reflection.TypeExtensions (.Tests)

I didn't see any change there but can look deeper at this and other projects soon. 

Are these issues fixed by running the build a single time or are they necesary for each build?

ericstj:
Right now that target does two things:
1. Actually builds the library projects that make up the package.
2. Get's the ouutput of the library projects that make up the package (`GetFilesToPackage` target in libraries).  The latter is used like `GetTargetPath` for a normal project reference.  We look at the files that make up the package to determine which ones should be passed to the compiler / copied to output.

The first part only needs to happen when the library changes (and should only do so today if incremental build isn't broken).  The latter needs to happen every time, or we need to build something to cache the result.

hughbe:
@ericstj I may be getting things confused, but I haven't seen any impact on the IDE experience with this change.

> I'd expect that this would cause some design-time builds to fail if they depended on the types that were brought in through the pkgproj reference.

Perhaps this is because the projects I'm looking at: I'm sure there may be failures but they are don't seem to affect anything. Do all corefx projects depend on types from pkgproj references, or do only certain projects do? If the latter, can you give an example?

ericstj:
All the tests depend on PkgProj references in order to bring in the live surface area of the assembly they are testing & the live implementation (when running the test).

If a test also specifies a package reference in project.json it would bring in the LKG version of the assembly it is testing.  So I would expect design time build would only fail if the test was using API that wasn't yet in the LKG.

If a test was not specifying a package reference in project.json I expect design time build would always fail.

hughbe:
Sorry for the long delay.
I've done some research. This has no effect on whether design time builds fail or not for the following reason:
The `RestorePackages` target is disabled for design time builds. Further, it is my understanding that any build restoring packages will also resolve project references.
This means that if we disable the `ExpandProjectReferences` target, and we haven't restored any packages, the build would fail anyway. However, if we have already restored packages, `ExpandProjectReferences` will always succeed.
As such, I support not running the `ExpandProjectReferences` in Design Time Builds, as it has no effect anyway, I think.
However, as discussed above, there is the question of what is causing `ExpandProjectReferences` to take so long. My diagnosis is that it is calling `BuildAllProjects`, triggering a lengthy compilation and build process.
However, I am not sure if the `BuildAllProjects` target is avoidable or not. This would have to be an area for discussion by those more knowledgeable about the build system.

ericstj:
Typically we never restore packages during the build.  We expect folks to do a full repo restore up front.  The only reason why we restore during a raw-csproj (not *builds) build is to give a workflow for folks updating project.jsons.

> As such, I support not running the ExpandProjectReferences in Design Time Builds, as it has no effect anyway, I think.

It will have an effect.  What this will do is cause the package to not get any of the "live" builds for the package.  Those "live" builds are represented as project references in the package, and by removing them you're not getting any of those project references.  You might as well not even include the pkgproj reference.

I expect the correct minimal fix here is to condition the MSBuild task that is invoking build at https://github.com/hughbe/buildtools/blob/e817994b15447e7e13dfbe29a6cbf1ab054b9185/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L267.  That will cause this target to get the paths to the files, but not actually trigger the Build target.

Do you think you can try that fix instead and see if it gives you the same perf benefits?

hughbe:
> Typically we never restore packages during the build. We expect folks to do a full repo restore up front. The only reason why we restore during a raw-csproj (not *builds) build is to give a workflow for folks updating project.jsons.

How unusual, I've genuinely never restored everything up front - I just tried it and it takes 15/20 minutes! Please keep this feature if there's ever a discussion of removing it.

Turns out my investigations didn't include when all packages were restored (only when the packages were not restore, so the build failed anyway), so you are right. I will investigate your suggestion

ericstj:
> How unusual, I've genuinely never restored everything up front - I just tried it and it takes 15/20 minutes! 

That's what happens when you run build.cmd.  We are working on improving the time.  It's largely due to scalability issues in nuget (which they are also working on fixing).  We've put in some workarounds to improve this and @karajas and I are working on further workarounds.

Restore up front is a requirement for multi-core builds (NuGet doesn't handle concurrent restores of the same projects) and the only supported way to run restore outside VS (per the NuGet team).

Thanks again for looking into this @hughbe.

hughbe:
@ericstj I've updated the PR as you suggested - the time improvement still remains

ericstj:
Awesome.  Thanks for fixing this!

PR #952, opened 2016-08-11
Add a task to help workaround dotnet/corefx#5230

ellismg:
This task generates a response file we can use to pass the correct set of
defines to the complier without having to worry about MSBuild converting
back slashes to forward slashes when running on Unix systems.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -102,6 +102,26 @@
     </PropertyGroup>
   </Target>
 
+  <!--
+    Cross Platform MSBuild has some logic to replace \ with / when invoking commands to fix up path differences between Windows and
+    *NIX. The define command line argument syntax for VB requires that we both surround some items with quotes and escape the quotes with
+    backslashes. However, due to the above MSBuild logic, this causes an invalid command line to be generated when running on *NIX.
+
+    Microsoft/msbuild#422 tracks an actual fix in MSBuild, but for now we work around the issue by using a custom task that
+    transforms the set of defines we are going to use into a response file we can pass along to the Vbc task along with an
+    empty set of defines.
+  -->
+  <UsingTask TaskName="WriteVisualBasicDefineResponseFile"  AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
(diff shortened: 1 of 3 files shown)


ellismg:
/cc @weshaggard @mellinoe

I would like to also back port this fix to the release/1.0.0 branch so I can start to consume that in CoreFX so we can build 1.0.1 of CoreFX on Linux.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -102,6 +102,26 @@
...
+    backslashes. However, due to the above MSBuild logic, this causes an invalid command line to be generated when running on *NIX.
+
+    Microsoft/msbuild#422 tracks an actual fix in MSBuild, but for now we work around the issue by using a custom task that
+    transforms the set of defines we are going to use into a response file we can pass along to the Vbc task along with an
+    empty set of defines.
+  -->
+  <UsingTask TaskName="WriteVisualBasicDefineResponseFile"  AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
+
+  <Target Name="ConvertDefinesToResonseFile" BeforeTargets="CoreCompile" Condition="'$(MSBuildProjectExtension)' == '.vbproj'">
+    <WriteVisualBasicDefineResponseFile DefineConstants="$(FinalDefineConstants)"
```
ellismg: I tried using `WriteLinesToFile` with some transformations so we didn't need this task, but slashes were still getting converted in that case.

weshaggard: Interesting do you know what part of the system is doing the converting in that case? I wouldn't expect there to be any translation happening when writing a raw string to a file.

weshaggard:
Looks reasonable to me if we cannot find any easier workaround. Given that we have only one vb project can we get away with writing the raw string we need directly in the vbproj itself instead of requiring a buildtools update?

ellismg:
Sorry for dropping this on the floor.

> Given that we have only one vb project can we get away with writing the raw string we need directly in the vbproj itself instead of requiring a buildtools update?

My preference is to centralize this stuff in buildtools so normal projects can be oblivious to any *nix specific workarounds.  I understand that we only have one vbproj now, but I'd rather just have everyone that uses buildtools get stuff for free.

PR #953, opened 2016-08-12
Enable dependency verification based on full package version

dagood:
Allows dependency verification using build output lists published to the dotnet/versions repository. It downloads the buildinfo(s), caching results in `Tools/BuildInfoCache/[versions repo commit sha]/[buildinfo name]/buildinfo.json`. From there, it does the read/parse of an auto-upgrade: if there is anything to upgrade, the dependencies are invalid.

This is how I have the corefx dir.props configured to use this PR: https://github.com/dagood/corefx/commit/5bee11c089ff97de50517e62c640b1721eeb58ca.

With that config, verification happens during `sync -p` before the restore. For me, it takes ~2 seconds uncached, 0.3 seconds cached. (It can be disabled with `/p:VerifyPackageVersions=false`.)

An example upgrade PR that shows the `CurrentRef`s being changed: https://github.com/dagood/corefx/pull/19. Only the ones that are actually upgraded are changed. (This allows one buildinfo to be upgraded independently of other ones, for manual fixes.)

The validation task can also auto-fix dependencies using `/p:FixDependencies=true`. This includes the entries in `dir.props`, not just project.jsons. (Upgrade and validation use the same systems.) The idea is to allow manual tweaks without all the automation details of the auto-PR system.

Something to consider is a separate config file outside of the `dir.props`... I think this depends on how we want to implement the repo API. (cc @chcosta) For now, I left config where it was.

Resolves https://github.com/dotnet/buildtools/issues/732

@weshaggard @ericstj @eerhardt, mind taking a look?


src/Microsoft.DotNet.VersionTools/Dependencies/DependencyUpdateTask.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.DotNet.VersionTools.Automation;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.VersionTools.Dependencies
+{
+    /// <summary>
+    /// A task that can be performed to update dependency versions to values in BuildInfos.
+    /// 
+    /// Like System.Threading.Task, a call to Result blocks on the update being performed.
```
eerhardt: Why not use Task<> instead?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
```diff
@@ -29,6 +35,7 @@
     <UpdateDependencies DependencyBuildInfo="@(DependencyBuildInfo)"
                         ProjectJsonFiles="@(ProjectJsonFiles)"
                         XmlUpdateStep="@(XmlUpdateStep)"
+                        BuildInfoCacheDir="$(BuildInfoCacheDir)"
```
eerhardt: What does "UpdateDependencies" do with the cached information?  Shouldn't it always read from @(DependencyBuildInfo)?

src/Microsoft.DotNet.VersionTools/Dependencies/DependencyUpdateTask.cs
dagood: I need `UsedBuildInfos` and `ReadableDescriptionLines` before evaluating, but maybe subclassing would work. I'll see how that would go.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
dagood: The cache is for buildinfo data downloaded from the versions repo--`@(DependencyBuildInfo)` has pointers to where in the versions repo to look. This is what Wes and I talked about here: https://github.com/dotnet/buildtools/issues/732#issuecomment-220122400

eerhardt: But this is "UpdateDependencies", i.e. we are updating to *new* build infos.  Why use the cache?  In case some of them are the same?

Does this make it confusing?

dagood: Oh, you're right, I must not have noticed that this was specifically about UpdateDependencies. The cache is never used here, so I'll remove that input.

src/Microsoft.DotNet.VersionTools/Dependencies/DependencyUpdateTask.cs
dagood: This worked fine, and it looks a lot cleaner. Thanks. :+1:

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -0,0 +1,163 @@
...
+            {
+                yield return new ProjectJsonUpdater(ProjectJsonFiles.Select(item => item.ItemSpec))
+                {
+                    // If SkipStableVersions is specified (true or non-true) on a build info item,
+                    // then make the updater follow that behavior for versions in that build info.
+                    BuildInfoSkipStableVersions = DependencyBuildInfo
+                        .Select(item => new
+                        {
+                            Info = buildInfos.Single(info => info.Name == item.ItemSpec),
+                            Skip = item.GetMetadata("SkipStableVersions")
```
weshaggard: Where are we setting this to true? I didn't see it in your sample PR's.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -0,0 +1,163 @@
...
+                {
+                    return new BuildInfo
+                    {
+                        Name = item.ItemSpec,
+                        LatestPackages = BuildInfo.ReadPackageListAsync(reader).Result
+                    };
+                }
+            }
+
+            string packageId = item.GetMetadata("PackageId");
```
weshaggard: Is there an advantage of using this over just putting the info into a static file? or visa versa?

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependencies.cs
```diff
@@ -42,96 +37,92 @@ public class UpdateDependencies : Task
...
-                    yield return updater;
+                string suggestedMessage = updateResults.GetSuggestedCommitMessage();
+                string body = string.Empty;
+                if (NotifyGitHubUsers != null)
+                {
+                    body += PullRequestCreator.NotificationString(NotifyGitHubUsers.Select(item => item.ItemSpec));
                 }
+
+                var prCreator = new PullRequestCreator(gitHubAuth, origin, upstreamBranch, GitHubAuthor);
+                prCreator.CreateOrUpdateAsync(
```
weshaggard: If a dev runs updatedependencies locally does this mean that it will always create a auto-PR? What is the workflow for just updating my local repo?

weshaggard: Looking further it looks like that would be VerifyDependencies -FixDependencies? If so it might be worth updating the name some so that it is clear which does which. I would have expected to call UpdateDependencies.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
dagood: This is only set to false in corefx, here: https://github.com/dagood/corefx/commit/5bee11c089ff97de50517e62c640b1721eeb58ca#diff-0b192804a6349e8c26d2b027afbd89a2R127. It defaults to true. The idea here is that on `ProjectJsonUpdater` there's a global `SkipStableVersions`, and this `BuildInfoSkipStableVersions` lets you override it (true or false) for individual buildinfos.

Looking at it now, I think I should really just move `SkipStableVersions` somewhere buildinfo-specific. I'll change that.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
dagood: This way lets you set the version via a msbuild property (like `microsoft.xunit.runner.uwp` in the example config PR). There's no benefit to the static file besides conciseness vs. direct msbuild, so that could be taken out.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependencies.cs
dagood: Yep, that's the right command. Agreed the naming isn't great, how about:
* `VerifyDependencies`
* `UpdateDependencies` (only update, no PR)
* `SubmitUpdateDependenciesPullRequest` (new)

This could be designed with `SubmitPullRequest=true` on `UpdateDependencies`, but I'd like to use msbuild `Required` for the github auth args rather than basing it on a bool.

src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
```diff
@@ -133,5 +168,33 @@ private static IEnumerable<JProperty> FindAllDependencyProperties(JObject projec
...
+
+        private class ParsedBuildInfo
+        {
+            public BuildInfo BuildInfo { get; }
+            public IEnumerable<PackageIdentity> Packages { get; }
+
+            public ParsedBuildInfo(BuildInfo buildInfo)
+            {
+                BuildInfo = buildInfo;
+                Packages = buildInfo.LatestPackages
```
eerhardt: Isn't this what `GetLatestPackageIdentities` does?

src/Microsoft.DotNet.VersionTools/Dependencies/IDependencyUpdater.cs
```diff
@@ -14,6 +14,6 @@ public interface IDependencyUpdater
         /// <summary>
         /// Updates based on the given build infos and returns build infos used during update.
         /// </summary>
-        IEnumerable<BuildInfo> Update(IEnumerable<BuildInfo> buildInfos);
+        IEnumerable<DependencyUpdateTask> Update(IEnumerable<BuildInfo> buildInfos);
```
eerhardt: Is it weird to call this "Update" when this method actually doesn't do the "Update"?  You need to start the tasks in order to do the update.

eerhardt:
:shipit:

src/Microsoft.DotNet.VersionTools/Dependencies/ProjectJsonUpdater.cs
dagood: Ah, yep, I think I didn't realize at this point that I'd want to add that method and I didn't remember to come back here once I did. Fixing.

src/Microsoft.DotNet.VersionTools/Dependencies/IDependencyUpdater.cs
dagood: Yeah, I called this `GetUpdateTasks` on `DependencyUpdater` so I'll rename this also to `GetUpdateTasks`. Thanks.

dagood:
Pushed some commits that should address everything, and the new corresponding corefx config is here: https://github.com/dotnet/corefx/commit/28f1fc32afb317755e03cf09950d226f64c685d0

The static dependency part:
```xml
    <TargetingPackDependency Include="Microsoft.TargetingPack.NETFramework.v4.5" />
    <TargetingPackDependency Include="Microsoft.TargetingPack.NETFramework.v4.5.1" />
    <TargetingPackDependency Include="Microsoft.TargetingPack.NETFramework.v4.5.2" />
    <TargetingPackDependency Include="Microsoft.TargetingPack.NETFramework.v4.6" />
    <TargetingPackDependency Include="Microsoft.TargetingPack.NETFramework.v4.6.1" />
    <TargetingPackDependency Include="Microsoft.TargetingPack.NETFramework.v4.6.2" />
    <TargetingPackDependency Include="Microsoft.TargetingPack.Private.WinRT" />
    <StaticDependency Include="@(TargetingPackDependency)">
      <Version>1.0.1</Version>
    </StaticDependency>

    <XUnitDependency Include="xunit"/>
    <XUnitDependency Include="xunit.runner.utility"/>
    <XUnitDependency Include="xunit.runner.console"/>
    <StaticDependency Include="@(XUnitDependency)">
      <Version>2.1.0</Version>
    </StaticDependency>

    <StaticDependency Include="Microsoft.xunit.netcore.extensions;Microsoft.DotNet.BuildTools.TestSuite">
      <Version>1.0.0-prerelease-00704-03</Version>
    </StaticDependency>

    <DependencyBuildInfo Include="@(StaticDependency)">
      <PackageId>%(Identity)</PackageId>
      <UpdateStableVersions>true</UpdateStableVersions>
    </DependencyBuildInfo>
```

Not so bad, actually. And it benefits over the file by reusing the same version string for sets of packages.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
weshaggard: Given that our build is actually based on msbuild for everything here I think I would just remove the static file case and just put the package list in a .props file and use that as the configuration.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
weshaggard: Yes I saw where you were setting it to false in some cases but I still don't understand the logic of where you treat the default as true. Also as a convention we generally make defaults of bools be false as that make for a more logical choice.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependencies.cs
weshaggard: I would call the last one UpdateDependenciesAndSubmitPullRequest but otherwise I'm fine with those names.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
dagood: Cool, that's basically what I did with the new commits I pushed. They aren't in a separate .props file right now, but maybe they should be... and maybe anything that automatically changes should be in this separate file too?

weshaggard:
Other then the one slight rename suggestion the change looks good.

I would also suggest that the package list in your corefx change be in a separate props file.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependencies.cs
dagood: Yeah, that's easier to read, I'll change it.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
weshaggard: I think having all the things that change in that separate file makes sense.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
dagood: I've changed it so there's just `UpdateStableVersions` on `DependencyBuildInfo` that defaults to false. (The default to true was in `ProjectJsonUpdater`, the "global" `public bool SkipStableVersions { get; set; } = true;`.)

Thanks for the note on the convention, that definitely makes sense. Do you think this should apply to msbuild too? I have a `VerifyPackageVersions` which should maybe be `SkipVerifyPackageVersions`, and I think I've broken this at least once in the past.

weshaggard: Yes I like your change to use UpdateStableVersions.

As for the convention I do think it is a good principle to always default values to the default/empty value if at all possible.

weshaggard:
LGTM

PR #957, opened 2016-08-16
Add /nodeReuse:false to MsBuildParameters setting in config.json

maririos:
cc: @weshaggard @joperezr

### DIFF
--- a/config.json
+++ b/config.json
@@ -34,7 +34,7 @@
       "description": "MsBuild building options.",
       "valueType": "passThrough",
       "values": [],
-      "defaultValue": "/nologo /maxcpucount /verbosity:minimal"
+      "defaultValue": "/nologo /maxcpucount /verbosity:minimal /nodeReuse:false"
     },
     "Project": {
       "description": "Project where the commands are going to be applied.",


config.json
```diff
@@ -34,7 +34,7 @@
       "description": "MsBuild building options.",
       "valueType": "passThrough",
       "values": [],
-      "defaultValue": "/nologo /maxcpucount /verbosity:minimal"
+      "defaultValue": "/nologo /maxcpucount /verbosity:minimal /nodeReuse:false"
```
joperezr: >/nodeReuse:false [](start = 63, length = 16)

Does this run in both Windows and non-Windows? If so then this will fail since `nodeReuse` was removed from .NetCore MSBuild

maririos: It only works on Windows, but build tools is only meant to be build on windows so that is why I added it here by default.

joperezr: I see, I thought this was like a default that might be consumed by corefx. In that case, then this LGTM.

maririos: No. Every repo has its own config.json file.

maririos: Maybe we could have a central config.json file in the future, when we implementing the public API for the repos. I'll keep it in mind.

Thanks @joperezr

weshaggard:
LGTM

Issue #958, opened 2016-08-16
Run Command Tool: Passing setting with passThrough property fails 


maririos:
Calling `sync -MsBuildLogging` fails with:
```
Running: C:\mariariBT\Tools\dotnetcli\dotnet.exe C:\mariariBT\Tools\run.exe sync -p -MsBuildLogging
Running: C:\mariariBT\Tools\msbuild.cmd  /p:RestoreDuringBuild=true  true  /nologo /maxcpucount /verbosity:minimal /t:BatchRestorePackages
MSBUILD : error MSB1009: Project file does not exist.
Switch: true
```

It should pass the defaultValue of MsBuildLogging `/flp:v=normal`  and not true.
This applies to all the passThrough settings.

Issue #857, opened 2016-07-01
Merge change of CommandLine.cs from the run folder to the common folder


maririos:
The Run Command Tool (merged with PR #672 ) is using a copy of the class CommandLine.cs.
Instead of having a copy we would like to merge the changes from the class under Run to the CommandLine.cs class under the common folder.

ViktorHofer:
// Auto-generated message

Going forward, the .NET team is using https://github.com/dotnet/arcade to develop the code and issues formerly in this repository. Feel free to reopen/move this issue if it still applies to servicing branches in this repository, or source code which now resides in the Arcade repository.

PR #961, opened 2016-08-17
Fix passthrough setting

maririos:
When a setting was directly passed by the user, the assigned value was always true. Now it assigns the default value of the setting.

Fix issue: #958 

cc: @weshaggard  @joperezr

### DIFF
--- a/src/Run/CommandLine.cs
+++ b/src/Run/CommandLine.cs
@@ -1263,7 +1263,7 @@ class CommandLine
                         throw new CommandLineParserException("Use " + name + ":" + value + " if " + value +
                             " is meant to be value rather than a named parameter");
                     _args[valuePos] = null;*/
-                    value = "true";
+                    value = string.IsNullOrEmpty(defaultValue) ? "true" : defaultValue;
                 }
                 ret = ParseValue(value, type, name);
                 if (type.IsArray)


src/Run/CommandLine.cs
```diff
@@ -1593,7 +1593,7 @@ private StringBuilder GetIntroTextForHelp(int maxLineWidth)
             Wrap(sb, text, 0, String.Empty, maxLineWidth, true);
             text = "For more information about the Run Command Tool: https://github.com/dotnet/buildtools/blob/master/Documentation/RunCommand.md";
             Wrap(sb, text, 0, String.Empty, maxLineWidth, true);
-            sb.AppendLine().AppendLine().Append("Syntax: run [Command] [Action] (global settings)");
+            sb.AppendLine().AppendLine().Append("Sintaxis: run [Command] [Action] (global settings)");
```
joperezr: >Sintaxis [](start = 49, length = 8)

I believe this is an unwanted change.

joperezr:
>                     if (_noSpaceOnQualifierValues || valuePos >= _args.Count || _args[valuePos] == null)

I know it's not related to your PR, but why do we have all of this commented out code? If we need it for something we should tag it with an issue, if not then we should probably clean it up.

---
Refers to: src/Run/CommandLine.cs:1248 in 709aff7. [](commit_id = 709aff7ff0c0cd9dec244f113f1055124666a45a, deletion_comment = False)

joperezr:
other than the 2 small comments, this LGTM.

src/Run/CommandLine.cs
maririos: Hmm ok.

maririos:
Eventually we are going to merge the changes to the Common CommandLine class so I don't want to delete any code yet.
Issue #857 tracks this.

joperezr:
Do you mind adding the link to that issue to the top of the commented lines, so people may have a context of why that commented code is there?

src/Run/CommandLine.cs
maririos: ok

maririos:
It is not the only thing commented or changed. For now I'm going to leave it as it is. This is blocking a fix that we need in CoreFx

maririos:
Thanks Joe!

PR #962, opened 2016-08-17
Restore fast path

ericstj:
This adds a task that can be used to determine if a restore is actually required.

This reduced incremental restore time from 5 minutes down to 1.5 seconds for corefx on my box.

CoreFx requires these changes: https://github.com/ericstj/corefx/commit/6b88b5a772ceab4c3515193fa889b5aade44b124

This potentially could be removed after https://github.com/NuGet/Home/issues/3330 is fixed and we pick it up.

/cc @weshaggard @dagood


src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
```diff
@@ -0,0 +1,171 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#define read
```
dagood: What does this define do?

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
```diff
@@ -0,0 +1,171 @@
...
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public partial class IsRestoreRequired : Task
+    {
+        [Required]
+        public ITaskItem[] ProjectJsons { get; set; }
+
+        [Required]
+        public string PackagesFolder { get; set; }
+
```
weshaggard: nit: extra empty line.

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
```diff
@@ -0,0 +1,171 @@
...
+                    // Each id/version only needs to be checked once
+                    if (AddLock(packagesChecked, package))
+                    {
+                        // Verify the SHA exists for each package, don't validate the content since we assume our packages are immutable
+                        var hashPath = packageResolver.GetHashPath(package.Id, package.Version);
+
+                        if (!File.Exists(hashPath))
+                        {
+                            Log.LogMessage(MessageImportance.Low, $"{projectJsonPath} requires restore because {package} is missing.");
+                            needRestore.Add(project);
```
dagood: Why isn't a lock used here?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -27,7 +28,11 @@
 
     <Error Condition="'$(DnuRestoreCommand)'=='' and Exists('$(ProjectJson)')" Text="RestorePackages target needs a predefined DnuRestoreCommand property set in order to restore $(ProjectJson)" />
 
-    <Exec Condition="Exists('$(ProjectJson)')" Command="$(DnuRestoreCommand) &quot;$(ProjectJson)&quot;" StandardOutputImportance="Low" CustomErrorRegularExpression="^Unable to locate .*" />
+    <IsRestoreRequired ProjectJsons="$(ProjectJson)" PackagesFolder="$(PackagesDir)">
+      <Output TaskParameter="RestoreRequired" PropertyName="RestoreRequired" />
+    </IsRestoreRequired>
+
+    <Exec Condition="Exists('$(ProjectJson)') AND '$(RestoreRequired)' == 'true'" Command="$(DnuRestoreCommand) &quot;$(ProjectJson)&quot;" StandardOutputImportance="Low" CustomErrorRegularExpression="^Unable to locate .*" />
```
weshaggard: orthogonal but what is your thought on potentially writing a task to do restore as well instead of shelling out?

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            ProjectJsons.AsParallel().ForAll(project =>
+            {
+                string projectJsonPath = project.GetMetadata("FullPath");
+                string projectLockJsonPath = ProjectJsonPathUtilities.GetLockFilePath(projectJsonPath);
+
+                if (!File.Exists(projectLockJsonPath))
+                {
+                    Log.LogMessage(MessageImportance.Low, $"{projectJsonPath} requires restore because {projectLockJsonPath} is missing.");
+                    needRestore.Add(project);
```
weshaggard: List is not thread safe so we run the risk of race conditions when we these adds in parallel.

dagood:
Nice. I like how this leaves it open to restore individual project.jsons in the future. (I figure one way to do that would be to call into NuGet in a buildtools task since there won't be the command length limit there.)

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                EnsureRead(jsonReader);
+            }
+        }
+
+        private static void EnsureRead(JsonTextReader reader)
+        {
+            if (!reader.Read())
+            {
+                throw new EndOfStreamException();
```
weshaggard: Are there ever any cases were we expect to hit the end of the stream while trying to read values? If so perhaps we don't want to throw and catch an exception.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
ericstj: I'll clean that up, I was using it to switch between implementations earlier.

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
ericstj: :ok:

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
ericstj: :ok:

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
ericstj: Yeah @dagood pointed out the same.

src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs
ericstj: We never catch it.  We never expect to hit it.  This would mean that the lock file format didn't meet our expectations.  I considered adding more checks to have a better error (eg: catch when the depth changed as we were advancing the reader) but I decided instead to keep it simple.  Ideally the next major nuget update will let us delete this code.
the task

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
ericstj: I did prototype that before.  It's reasonable, especially if we run NuGet in-proc since we can do things like run multiple restores without incurring constant cost start-up time.

PR #963, opened 2016-08-17
Add RepoUtil sources to the build tools repo.

jhendrixMSFT:
We have a tool, RepoUtil, created by the Roslyn team for implementing the
produces and consumes repo APIs.  Since this is meant to be a general
purpose tool to be consumed by various repos I'm moving the sources to the
build tools repo.  I have also added nuget package authoring for this tool
so that it can easily be consumed by other repos.  In the process I have
converted the tool to be a .NET Core app.

### DIFF
--- a/src/RepoUtil/project.json
+++ b/src/RepoUtil/project.json
@@ -1,0 +1,21 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "version": "1.0.0-rc2-3002702",
+      "type": "platform"
+    },
+    "Newtonsoft.Json": "8.0.3",
+    "System.Collections.Immutable": "1.2.0"
+  },
+  "buildOptions": {
+    "emitEntryPoint": true
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": "dnxcore50"
+    }
+  },
+  "runtimes": {
+    "win7-x64": {}
+  }
+}
--- a/src/RepoUtil/repoutil.runtimeconfig.json
+++ b/src/RepoUtil/repoutil.runtimeconfig.json
@@ -1,0 +1,8 @@
+{
+  "runtimeOptions": {
+    "framework": {
+      "name": "Microsoft.NETCore.App",
+      "version": "1.0.0-rc3-002733"
+    }
+  }
+}
--- a/src/nuget/Microsoft.DotNet.BuildTools.RepoUtil.nuspec
+++ b/src/nuget/Microsoft.DotNet.BuildTools.RepoUtil.nuspec
@@ -1,0 +1,28 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.8.1">
+    <id>Microsoft.DotNet.BuildTools.RepoUtil</id>
+    <version>1.0.0-prerelease</version>
+    <title>Repository Produces/Consumes Utility (RepoUtil)</title>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
+    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=518631</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <summary>A tool that can be used to generate and verify internal and external dependency information.</summary>
+    <description>This package provides utilities for implementing the repository produces/consumes API.  You should not need to reference this package from your project.  This tool requires .NET Core 1.0 or later.</description>
+    <copyright>Copyright © Microsoft Corporation</copyright>
+    <tags></tags>
+    <dependencies>
+      <!--
+        Currently we redistribute the dependencies locally since nuget
+        has no support for treating packages as build dependencies.
+      -->
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="RepoUtil\repoutil.exe" target="tools" />
+    <file src="RepoUtil\repoutil.runtimeconfig.json" target="tools" />
+  </files>
+</package>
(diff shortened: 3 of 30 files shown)


jhendrixMSFT:
@jaredpar Here's the PR for getting RepoUtil into build tools.
@ericstj Could you please take a quick look at the nuspec.

ericstj:
NUSPEC LGTM

src/RepoUtil/project.json
```diff
@@ -0,0 +1,21 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "version": "1.0.0-rc2-3002702",
+      "type": "platform"
```
jaredpar: Believe this part is new.  Does this make the tool a .Net Core app?  Will it still run on Desktop CLR?

src/RepoUtil/project.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    "Newtonsoft.Json": "8.0.3",
+    "System.Collections.Immutable": "1.2.0"
+  },
+  "buildOptions": {
+    "emitEntryPoint": true
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": "dnxcore50"
```
jaredpar: Thought that particular TFM was dead.

src/RepoUtil/repoutil.runtimeconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
jaredpar: Why the lower case in repoutil in the file name?

src/RepoUtil/repoutil.runtimeconfig.json
```diff
@@ -0,0 +1,8 @@
...
+  "runtimeOptions": {
+    "framework": {
+      "name": "Microsoft.NETCore.App",
+      "version": "1.0.0-rc3-002733"
```
jaredpar: Why is the version of Microsoft.NetCore.App different here as compared to project.json?

src/nuget/Microsoft.DotNet.BuildTools.RepoUtil.nuspec
```diff
@@ -0,0 +1,28 @@
...
+    <dependencies>
+      <!--
+        Currently we redistribute the dependencies locally since nuget
+        has no support for treating packages as build dependencies.
+      -->
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="RepoUtil\repoutil.exe" target="tools" />
+    <file src="RepoUtil\repoutil.runtimeconfig.json" target="tools" />
```
jaredpar: Don't we need to package Newtonsoft.Json.dll as well?

jaredpar:
Is this project tied into building of the repo source?  Basically if I build the repo will it build this project?  

If so how is that hooked up?  It wasn't obvious from the change.

jhendrixMSFT:
The traversal build mechanism used in this repo will automatically build this new project, take a look at https://github.com/dotnet/buildtools/blob/master/src/dirs.proj.

src/RepoUtil/project.json
weshaggard: Yes can remove the dnxcore50 import.

src/RepoUtil/repoutil.runtimeconfig.json
weshaggard: Yes we should make them match. This should be whatever we plan on running on.

src/RepoUtil/project.json
```diff
@@ -0,0 +1,21 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "version": "1.0.0-rc2-3002702",
+      "type": "platform"
+    },
+    "Newtonsoft.Json": "8.0.3",
+    "System.Collections.Immutable": "1.2.0"
+  },
+  "buildOptions": {
```
weshaggard: Since we aren't building with the cli then you don't even need the buildoptions section. Only the dependencies are necessary.

src/RepoUtil/project.json
weshaggard: Since we aren't using the cli so this doesn't really make it a .NET Core app. The runtime config and the fact we use dotnet.exe to run it makes it a .NET Core app.

We should have another project that builds this for desktop so we can also run it there.

src/nuget/Microsoft.DotNet.BuildTools.RepoUtil.nuspec
```diff
@@ -0,0 +1,28 @@
...
+    <tags></tags>
+    <dependencies>
+      <!--
+        Currently we redistribute the dependencies locally since nuget
+        has no support for treating packages as build dependencies.
+      -->
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="RepoUtil\repoutil.exe" target="tools" />
```
weshaggard: We should have a net46 for the desktop version.

src/nuget/Microsoft.DotNet.BuildTools.RepoUtil.nuspec
weshaggard: We are going to need all the dependencies we need to run this in isolation .

joperezr:
Do you plan to consume this in corefx? If so perhaps you should also add some bootstrapping in init-tools maybe?

src/RepoUtil/repoutil.runtimeconfig.json
jhendrixMSFT: Fixed.

jhendrixMSFT:
@joperezr It's not clear right now how CoreFX is going to consume this, it will depend on @chcosta implementation of produces/consumes (i.e. if his implementation needs it I'll leave it to him to add it).

jhendrixMSFT:
@jaredpar @weshaggard My latest changes should address your feedback, can you please take a look?

jhendrixMSFT:
@weshaggard In particular the changes to the nuspec include the closure of assemblies for the .NET Core versions, is this what is wanted?

weshaggard:
LGTM

Issue #964, opened 2016-08-17
Package-Dependency verification should not open repo files with write access.


AtsushiKan:
The new dependency verification step injected into "sync -p" by this PR (https://github.com/dotnet/buildtools/pull/953) causes hundreds of "UnauthorizedAccessExceptions" on

  .\dependencies.proj
  and every project.json in the tree.

Yes, I "attrib +r" all repo files I'm not actively working on. It's the one aspect of TFS I prefer to Git. It's also convenient for catching things that sneak changes into my files.

[Goes to add "set SkipVerifyPackageVersions=true" to my cmd windows for CoreFx]

Since this process is only checking (autofix is supposed to be off by default), it should not be demanding  write permissions on these files.

@dagood

dagood:
Thanks for filing, I'll check it out. I agree it shouldn't be trying to get write access.

PR #965, opened 2016-08-17
When reading a file to verify dependencies, use FileAccess.Read

dagood:
Fixes https://github.com/dotnet/buildtools/issues/964

I confirmed that a verify using this change works with a read-only `project.json` and `dependenices.props` (and still detects invalid dependencies), and that `UpgradeDependencies` hits errors when trying to upgrade read-only files.

@naamunds @AtsushiKan 
(FYI @eerhardt)


AtsushiKan:
Thanks – LGTM.

naamunds:
:shipit:

dagood:
Thanks, I'll make a PR to upgrade to this in corefx.

PR #971, opened 2016-08-18
Rename RepoUtil package to RepoUtils.

jhendrixMSFT:
This package will likely end up containing multiple utilities for repo
dependency management so changing the name as there are no dependencies on
this package at present.


jhendrixMSFT:
@weshaggard @jaredpar please take a look.

jaredpar:
👍

PR #972, opened 2016-08-18
Add properties to the tools section for run.exe

### LINKED ISSUE #890: Run.exe crashes on a config.json that doesn't have tools linked up correctly
~~I have no idea why this file is considered malformed, but~~ this config.json breaks run.exe

```json
{
    "settings": {},

    "commands": {
        "sync": {
            "alias": {
                "p": {
                    "description": "Restores all NuGet packages for repository",
                    "settings": {}
                }
            },
            "defaultValues": {
                "toolName": "restore",
                "settings": {}
            }
        }
    },

    "tools": {
        "restore": {
            "run": {
                "windows": "restore.cmd",
                "unix": "foo"
            },
            "valueTypes": {}
        }
    }
}
```

error:
```
C:\Users\tyoverby\workspace\roslyn-internal\Open>run sync -?
Tools are already initialized.
Running: C:\Users\tyoverby\workspace\roslyn-internal\Open\Tools\dotnetcli\dotnet.exe C:\Users\tyoverby\workspace\roslyn-internal\Open\Tools\run.exe sync -?
Error: The process restore is not specified in the Json file.
Error: Object reference not set to an instance of an object.    at Microsoft.DotNet.Execute.Setup.GetHelpCommand(String commandName, String alias)
   at Microsoft.Fx.CommandLine.CommandLineParser.GetHelp(Int32 maxLineWidth, String parameterSetName, Boolean displayGlobalQualifiers)
   at Microsoft.Fx.CommandLine.CommandLineParser.<>c.<ParseForConsoleApplication>b__0_0(CommandLineParser parser)
   at Microsoft.Fx.CommandLine.CommandLineParser.Parse(Action`1 parseBody, Action`1 helpHandler, Action`2 errorHandler, Setup setupContent, String[] args)
   at Microsoft.Fx.CommandLine.CommandLineParser.ParseForConsoleApplication(Action`1 parseBody, String[] args, Setup setupContent)
   at Microsoft.DotNet.Execute.Executor.DefineParameters(String[] args, Setup setupInformation)
```

maririos:
With this change we could add specific tools parameters to pass to all the commands. For example, MsBuildParameters like: /nologo /verbosity:minimal /clp:Summary /maxcpucount are always going to be pass when a command uses the tool MsBuild,
That way we could customize them per OS and we only have to declare the parameters in one place.

I don't intend to merge this PR with the config.json file as it woul break the build. It is here for information purposes. Once the PR is approved I'll create a different PR updating the version of BT and adding the changes to config.json

Related to dotnet/corefx#10864.
Also, fix #890 

cc: @weshaggard @TyOverby

### DIFF
--- a/config.json
+++ b/config.json
@@ -30,12 +30,6 @@
       "values": [],
       "defaultValue": "/flp:v=normal"
     },
-    "MsBuildParameters": {
-      "description": "MsBuild building options.",
-      "valueType": "passThrough",
-      "values": [],
-      "defaultValue": "/nologo /maxcpucount /verbosity:minimal /nodeReuse:false"
-    },
     "Project": {
       "description": "Project where the commands are going to be applied.",
       "valueType": "passThrough",
@@ -86,15 +80,14 @@
         "verbose":{
           "description": "Passes /flp:v=diag to the msbuild command or the value passed by the user.",
           "settings":{
-            "MsBuildLogging": "/flp:v=diag"
+            "MsBuildLogging": "/flp:v=diag;LogFile=sync.log"
           }
         }
       },
       "defaultValues":{
         "toolName": "msbuild",
         "settings": {
-          "MsBuildParameters":"default",
-          "MsBuildLogging":"default"
+          "MsBuildLogging":"/flp:v=normal;LogFile=sync.log"
         }
       }
     },
@@ -140,7 +133,7 @@
       "defaultValues":{
         "toolName": "msbuild",
         "settings": {
-          "MsBuildParameters":"default"
+          "MsBuildLogging":"default"
         }
       }
     },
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -168,28 +161,43 @@
...
       }
     },
-    "terminal": {
-      "run": {
-        "windows": "cmd",
-        "unix": "sh"
+    "script": {
+      "osSpecific":{
+        "windows": {
+          "defaultParameters": "",
```
weshaggard: Shouldn't these properties be optional so they can be empty if they are omitted?

src/Run/Setup.cs
```diff
@@ -149,6 +149,12 @@ private CompleteCommand BuildCommand(string commandSelectedByUser, List<string>
         private string BuildParametersForCommand(Dictionary<string, string> commandParameters, string toolName)
         {
             string commandSetting = string.Empty;
+
+            if (Tools.ContainsKey(toolName))
```
weshaggard: I would also check whether or not this "defaultParameters" is null or empty to allow for people to omit it from the config file.

config.json
maririos: Yes. I'll change it

src/Run/Setup.cs
maririos: Will do

maririos:
PR Updated

src/Run/Setup.cs
```diff
@@ -150,10 +150,7 @@ private string BuildParametersForCommand(Dictionary<string, string> commandParam
         {
             string commandSetting = string.Empty;
 
-            if (Tools.ContainsKey(toolName))
-            {
-                commandSetting = Tools[toolName].osSpecific[Os]["defaultParameters"];
-            }
+            Tools[toolName].osSpecific[Os].TryGetValue("defaultParameters", out commandSetting);
```
weshaggard: Is it a valid assumption that Tools will always have the toolName entry? Below it seems like you are still guarding against that.

src/Run/Setup.cs
```diff
@@ -271,14 +268,14 @@ private string GetTool(Command commandToExecute, string os, string configPath, L
             if (Tools.ContainsKey(toolname))
```
weshaggard: Instead of doing a contains here and then various indexing below. I would do a Tools.TryGetValue here and just use the Tool it finds below.

src/Run/Setup.cs
maririos: Yes, the BuildParametersForCommand function is called by the function [BuildCommand](https://github.com/dotnet/buildtools/pull/972/files/3d163b9d9d81bd9d3b7439ac4a0cf8104a57c2cb..2b2df73d2267fbb83824eb5700f0c1338f4ae497#diff-6f4e4ec63e7b18f5f61e1d0ba5c10087R114) which verifies that the toolName is in Tools.

src/Run/Setup.cs
maririos: True, a while ago I did that refactoring but it seems I missed this one. I'll take a look at the code to see where else I have this conditions and I'll change it.

maririos:
PR Updated

weshaggard:
LGTM. Remember to undo the config file before merging :)

PR #975, opened 2016-08-19
Add "DisabledPackages" buildinfo metadata that ignores a package by id

dagood:
This will normally not be used ([it is needed in the dev/api branch](https://github.com/dotnet/corefx/pull/10970#issuecomment-241046344)), but it allows a package to be ignored from a buildinfo and not used for verification and auto-upgrade. Example usage: https://github.com/dotnet/corefx/commit/f61cbf732569f5b3a7c8a5d7dee24abc53c32ac1 (I added the package to a different `DependencyBuildInfo` to ensure it stays at a given version, even though it won't be auto-upgraded.)

/cc @weshaggard @joperezr


weshaggard:
LGTM

PR #976, opened 2016-08-19
Update bt version and config.json file

maririos:
Updating Build Tools version to consume latest changes of run.exe and changing config.json file to reflect this changes.

For more information about the change please take a look at PR  #972 

FYI: @Priya91 @TyOverby @sokket @wtgodbe . The config.json files needs to be updated to reflect the tools section change if the build tools version of the repo is updated. Otherwise run.exe is going to throw an exception.

cc: @weshaggard


config.json
```diff
@@ -168,28 +161,37 @@
...
     },
-    "terminal": {
-      "run": {
-        "windows": "cmd",
-        "unix": "sh"
+    "script": {
+      "osSpecific":{
+        "windows": {
+          "filesExtension": "cmd"
+        },
```
Priya91: why is this fileExtension? cmd.exe is the program that's launched for a script

maririos: The way the tool `script` works is that it attaches the fileExtension to the name of a file that comes from the setting `Project`.
Before we had it like [this](https://github.com/dotnet/buildtools/pull/976/files#diff-06b2d3b23dce96e1619d2b53d6c947ecL190) and the logic to attached the extension was hard coded in the code when the toolname was `terminal`.

With this new value I delete the hard coded things in run.exe and I added a more clear name for the file extension

maririos:
@joperezr

joperezr:
LGTM

PR #978, opened 2016-08-19
During auto-update, update CurrentBranch property

dagood:
Implements the buildtools side of this idea: https://github.com/dotnet/corefx/pull/10970#discussion_r75404356. The CoreFX-side config: https://github.com/dotnet/corefx/commit/1392eaf5b71e893e8cd61089de4949348239d932

An example auto-PR, simulating the point where dev/api has just split off from master, and this is the first auto-PR with dev/api packages available: https://github.com/dagood/corefx/pull/22

It works the same way as the CurrentRef update, but writes the name of the `AutoUpgradeBranch`, which in this case is passed in as `BranchBeingUpgraded`.

@weshaggard @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
@@ -16,6 +16,16 @@ namespace Microsoft.DotNet.Build.Tasks.VersionTools
 {
     public abstract class BaseDependenciesTask : Task
     {
+        internal const string RawUrlMetadataName = "RawUrl";
+        internal const string RawVersionsBaseUrlMetadataName = "RawVersionsBaseUrl";
+        internal const string BuildInfoPathMetadataName = "BuildInfoPath";
+        internal const string CurrentRefMetadataName = "CurrentRef";
+        internal const string CurrentBranchMetadataName = "CurrentBranch";
+        internal const string PackageIdMetadataName = "PackageId";
+        internal const string VersionMetadataName = "Version";
+
+        internal const string AutoUpgradeBranchMetadataName = "Version";
+
         [Required]
         public ITaskItem[] DependencyBuildInfo { get; set; }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependenciesAndSubmitPullRequest.cs
```diff
@@ -121,5 +126,23 @@ private ITaskItem FindDependencyBuildInfo(string name)
...
+                {
+                    Group g = CreateXmlUpdateRegex(elementName, valueGroup)
+                        .Match(contents)
+                        .Groups[valueGroup];
+
+                    return contents
+                        .Remove(g.Index, g.Length)
+                        .Insert(g.Index, newValue);
+                });
+            replaceAction?.Invoke();
```
weshaggard: Why `?` this should never be null.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependenciesAndSubmitPullRequest.cs
```diff
@@ -55,37 +57,40 @@ protected override void TraceListenedExecute()
                 {
                     item.SetMetadata(s_currentRef, masterSha);
                 }
+                string autoUpgradeBranch = item.GetMetadata(s_autoUpgradeBranch);
+                if (!string.IsNullOrEmpty(autoUpgradeBranch))
+                {
+                    item.SetMetadata(s_currentBranch, autoUpgradeBranch);
```
weshaggard: Where is this metadata used?

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -110,11 +110,18 @@ private static BuildInfo CreateBuildInfo(ITaskItem item, string cacheDir)
             string rawVersionsBaseUrl = item.GetMetadata("RawVersionsBaseUrl");
             string buildInfoPath = item.GetMetadata("BuildInfoPath");
             string currentRef = item.GetMetadata("CurrentRef");
+            // Optional
+            string currentBranch = item.GetMetadata("CurrentBranch");
```
weshaggard: Any reason you aren't sharing these string constants?

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependenciesAndSubmitPullRequest.cs
weshaggard: It looks like it is used in BaseDependenciesTask but I didn't see it because I searched for s_currentBranch but you didn't use that constant there (which I left another comment about now).

weshaggard:
couple nits but otherwise LGTM

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateDependenciesAndSubmitPullRequest.cs
dagood: `ReplaceFileContents` only returns an `Action` if there's a change to make, and there might not be a change. For example: if `AutoUpgradeBranch` is `master` but the `CurrentBranch` is already `master`.

@naamunds suggested recently that I rename the `ReplaceFileContents` method, and I think that would help here. `GetReplaceFileContentsTask` is what I'm thinking, paralleling `IDependencyUpdater::GetUpdateTasks`. I'll add a comment too because it's not clear here if not making a replacement is actually an upgrade error.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
dagood: No reason, just hadn't thought to. I'll change this, thanks.

joperezr:
LGTM

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -16,6 +16,16 @@ namespace Microsoft.DotNet.Build.Tasks.VersionTools
 {
     public abstract class BaseDependenciesTask : Task
     {
+        protected static string RawUrlMetadataName = "RawUrl";
```
weshaggard: I would make these internal constants.

dagood: Fixed `static` -> `const`, and just to make sure I have some ok reasoning, `internal` rather than `protected` to make it easier for others in the assembly to reuse but also limit potential for someone to depend on it who shouldn't?

weshaggard: Yes I always find protected constants a little odd which is why I suggested internal. protected would have been OK as well if the users are limited to just derived types.

PR #979, opened 2016-08-22
Add package index

ericstj:
Now that package versions are divorced from assembly version we need a
data file to track the mapping.

In addition to tracking assembly version to package version, this data
file will track stable packages, baseline packages, and module-to-
package mappings.  This will help speed up our builds by caching the
data rather than parsing giant item lists in every msbuild project.

To keep the data file up to date I added validation with an error
message that tells folks to run a command to update the data file.  I
don't update the file automatically because I can't be certain that
packages with stale data have already been created.

CoreFx change which supports this: https://github.com/dotnet/corefx/pull/10971/commits/254ce644dc75132f787053b67177decf9651a86e

/cc @weshaggard @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyBaseLine.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyBaseLine.cs
@@ -24,9 +24,15 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
         public ITaskItem[] OriginalDependencies { get; set; }
 
         /// <summary>
+        /// Permitted package baseline versions.
+        ///   Identity: Package ID
+        ///   Version: Package version.
+        /// </summary>
+        public ITaskItem[] BaseLinePackages { get; set; }
+
+        /// <summary>
         /// Package index files used to define baseline, and assembly to package version mapping.
         /// </summary>
-        [Required]
         public ITaskItem[] PackageIndexes { get; set; }
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


ericstj:
We need to be careful when merging this since it is a breaking change for CoreFx & WCF.

ericstj:
In order to reduce the breaking nature of this change I'm going to restore the old code and keep the implementations side by side for some time.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -7,8 +7,7 @@
   -->
 
   <PropertyGroup>
-    <Version Condition="'$(StableVersion)' != ''">$(StableVersion)</Version>
-
+    <PackageVersion Condition="'$(PackageVersion)' == '' AND '$(Version)' != ''">$(Version)</PackageVersion>
```
joperezr: >'$(Version)' != '' [](start = 61, length = 18)

I suppose this would be the case when an individual pkgproj wants to force a version? If true, why not just set PackageVersion in the pkgproj instead?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -713,19 +710,21 @@
...
     </ApplyBaseLine>
 
-    <!-- if not baselining, flow the dependencies unmodified -->
-    <ItemGroup Condition="'$(BaseLinePackageDependencies)' == 'false'">
-      <_BaseLinedDependencies Include="@(FilePackageDependency)" />
-    </ItemGroup>
 
-    <Error Condition="'@(_BaseLinedDependencies)' != '' AND '@(StablePackage)' == ''" Text="The StablePackage item is not defined: ensure you have imported Microsoft.Private.PackageBaseLine.props from the Microsoft.Private.PackageBaseLine package" />
-    <ApplyPreReleaseSuffix Condition="'@(_BaseLinedDependencies)' != ''" StablePackages="@(StablePackage)" OriginalPackages="@(_BaseLinedDependencies)" PreReleaseSuffix="$(VersionSuffix)" RevStableToPrerelease="$(DependencyRevStableToPrerelease)">
+    <Error Condition="'@(_BaseLinedDependencies)' != '' AND '@(PackageIndex)' == ''"
```
joperezr: >PackageIndex [](start = 63, length = 12)

Where is this ItemGroup loading the contents of the packageIndex file, I couldn't find it in here.

joperezr: Oh, nvm, I just saw your comment bellow about importing the props file. I suppose you kept Microsoft.Private.PackageBaseLine separate so that you can change the baseline without needing a buildtools update, but should we just add the import of the props file to this targets file?

---
In reply to: [75897216](https://github.com/dotnet/buildtools/pull/979#discussion_r75897216) [](ancestors = 75897216)

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: > when an individual pkgproj wants to force a version

Yes

I'm trying to make this change non-breaking.

ericstj:
This should be good-to-go now and is non breaking.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyBaseLine.cs
```diff
@@ -10,29 +10,57 @@
...
+
+        /// <summary>
+        /// Package index files used to define baseline, and assembly to package version mapping.
+        /// </summary>
+        public ITaskItem[] PackageIndexes { get; set; }
+
+        /// <summary>
+        /// Set to true to apply the package baseline
+        /// </summary>
+        public bool Apply { get; set; }
```
weshaggard: Are there cases where this is false?

ericstj: Yes, when folks want to disable baselining and honor only assembly refs.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/VersionUtility.cs
```diff
@@ -41,5 +49,25 @@ public static Version GetAssemblyVersion(string assemblyPath)
...
         }
+
+        public static Version As3PartVersion(Version version)
+        {
+            if (version.Revision == -1)
+            {
+                return version;
+            }
+
+            return new Version(version.Major, version.Minor, version.Build);
```
weshaggard: version.Build could also end up being -1 and Version would fail in those cases.

ericstj: It's not all that interesting for our scenarios, but I'll add safegaurds since this is a utility method.

ericstj: I improved these methods.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -88,13 +86,6 @@
     <PlatformPackageVersion Condition="'$(PlatformPackageVersion)' == ''">1.0.1</PlatformPackageVersion>
   </PropertyGroup>
 
-  <!-- If this package explicitly sets the StableVersion then add it to the stable list -->
```
weshaggard: I think we still have some packages that directly set the StableVersion (perhaps only TFS I don't recall), so is there a way to do that still or do they need to be put into the config file?

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: It's not a good idea to do this since any package which depends on it will get the wrong version in its dependencies.  We aren't using StableVersion anywhere in CoreFx or WCF.  I do see some use in TFS.  I can restore this behavior to support the TFS projects but we should avoid its use.

Issue #973, opened 2016-08-19
Run tool, suppress run output option?


chcosta:
I intend to add an option to the run tool to suppress its output from printing to the console.  MsBuild, for example, has a "/nologo" option which is used to prevent printing its header information to the console when it runs.  I'd like a similar option to run which suppresses "run" output and only displays output of the command I'm actually running.

Suppressing the output from here - https://github.com/dotnet/buildtools/blob/master/src/Run/Setup.cs#L87

Ideally, this would also be configurable from within a specific command in config.json so that you could always say, "if I execute 'run produces', it just gives me the clean output of that command."  

This feature would enable us to more easily pipe run output to other commands.

Please let me know if there is any concern with this or this conflicts with other work / design goals in progress for run.

/cc @maririos @weshaggard

maririos:
So you would only want the output that MsBuild is directly producing?
https://github.com/dotnet/buildtools/blob/master/src/Run/RunProcess.cs#L57

chcosta:
Yes, I don't want these...

https://github.com/dotnet/buildtools/blob/master/src/Run/Setup.cs#L93
https://github.com/dotnet/buildtools/blob/master/src/Run/Setup.cs#L100
https://github.com/dotnet/buildtools/blob/master/src/Run/Setup.cs#L105

Perhaps there are more.  error output or exceptions from the tool itself might be ok as well.

PR #980, opened 2016-08-22
Enable run quiet mode for run tool (suppress run tool output, only display exec output)

chcosta:
https://github.com/dotnet/buildtools/issues/973

Adds a "RunQuiet" option (default false), which, if true, suppresses extra output from the run tool itself.  The default values for RunQuiet are part of the run tool, but a repo may override these values by either adding a setting with a default value in that repo's config.json, adding a "RunQuiet" property to a Command, or specifying the value on the command-line.

/cc @maririos , @weshaggard

### DIFF
--- a/src/Run/Setup.cs
+++ b/src/Run/Setup.cs
@@ -12,8 +12,8 @@ namespace Microsoft.DotNet.Execute
 {
     public class Setup
     {
-        private const string runQuietReservedKeyword = "RunQuiet";
-        private const string runToolSettingValueTypeReservedKeyword = "runToolSetting";
+        private const string RunQuietReservedKeyword = "RunQuiet";
+        private const string RunToolSettingValueTypeReservedKeyword = "runToolSetting";
 
         public Dictionary<string, string> ToolSettings { get; set; }
         public Dictionary<string, Setting> Settings { get; set; }
@@ -47,7 +47,7 @@ namespace Microsoft.DotNet.Execute
 
         private bool IsReservedKeyword(string keyword)
         {
-            if(keyword.Equals(runQuietReservedKeyword))
+            if(keyword.Equals(RunQuietReservedKeyword))
             {
                 return true;
             }
@@ -154,12 +154,12 @@ namespace Microsoft.DotNet.Execute
             int returnCode = ValidateSettings();
 
             ToolSettings = Settings.Where(s => s.Value.ValueType == null  ||
-                                               s.Value.ValueType.Equals(runToolSettingValueTypeReservedKeyword)
+                                               s.Value.ValueType.Equals(RunToolSettingValueTypeReservedKeyword)
                                                ).ToDictionary(s => s.Key, s => string.Empty) ?? new Dictionary<string, string>();
             // A dev may have overriden the default values for a tool setting, but not specified the ValueType
             foreach(var key in ToolSettings.Keys)
             {
-                Settings[key].ValueType = runToolSettingValueTypeReservedKeyword;
+                Settings[key].ValueType = RunToolSettingValueTypeReservedKeyword;
             }
 
             // Parse run tool settings for any settings which do not apply to a Command, this allows us to have run tool settings
@@ -171,17 +171,17 @@ namespace Microsoft.DotNet.Execute
         private void SetRunToolSettingsDefaults()
         {
             // If RunQuiet is already defined in config.json, don't override it
-            if (!Settings.ContainsKey(runQuietReservedKeyword))
+            if (!Settings.ContainsKey(RunQuietReservedKeyword))
             {
                 Setting runQuietSetting = new Setting()
                 {
                     Values = new List<string>() { "True", "False" },
-                    ValueType = runToolSettingValueTypeReservedKeyword,
+                    ValueType = RunToolSettingValueTypeReservedKeyword,
                     Description = "Run tool specific setting.  Set to True to only display output from the executing command.  Default, False",
                     DefaultValue = "false"
                 };
 
-                Settings.Add(runQuietReservedKeyword, runQuietSetting);
+                Settings.Add(RunQuietReservedKeyword, runQuietSetting);
             }
 
         }
@@ -191,7 +191,7 @@ namespace Microsoft.DotNet.Execute
             ParseRunToolSettings(commandSelectedByUser);
             string runQuietValue;
             bool runQuiet = false;
-            if (ToolSettings.TryGetValue(runQuietReservedKeyword, out runQuietValue))
+            if (ToolSettings.TryGetValue(RunQuietReservedKeyword, out runQuietValue))
             {
                 runQuiet = runQuietValue.Equals("true", StringComparison.OrdinalIgnoreCase);
             }
@@ -439,7 +439,7 @@ namespace Microsoft.DotNet.Execute
             {
                 commandOption = string.Format(" {0}", toolName.Equals("console") ? "" : value);
             }
-            else if(type.Equals(runToolSettingValueTypeReservedKeyword)) { /* do nothing */ }
+            else if(type.Equals(RunToolSettingValueTypeReservedKeyword)) { /* do nothing */ }
             else
             {
                 Tool toolFormat;
--- a/src/Run/Setup.cs
+++ b/src/Run/Setup.cs
@@ -177,7 +177,7 @@ namespace Microsoft.DotNet.Execute
                 {
                     Values = new List<string>() { "True", "False" },
                     ValueType = RunToolSettingValueTypeReservedKeyword,
-                    Description = "Run tool specific setting.  Set to True to only display output from the executing command.  Default, False",
+                    Description = "Run tool specific setting.  Set to True to only display output from the executing command.",
                     DefaultValue = "false"
                 };
 
--- a/src/Run/Setup.cs
+++ b/src/Run/Setup.cs
@@ -12,7 +12,8 @@ namespace Microsoft.DotNet.Execute
 {
     public class Setup
     {
-        private static readonly string runQuietReservedKeyword = "RunQuiet";
+        private const string runQuietReservedKeyword = "RunQuiet";
+        private const string runToolSettingValueTypeReservedKeyword = "runToolSetting";
 
         public Dictionary<string, string> ToolSettings { get; set; }
         public Dictionary<string, Setting> Settings { get; set; }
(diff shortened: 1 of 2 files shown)


chcosta:
I don't think this break is related to my changes.  I'll investigate further though...

error CS1061: 'Setup' does not contain a definition for 'prepareValues' and no extension method 'prepareValues' accepting a first argument of type 'Setup' could be found (are you missing a using directive or an assembly reference?) [D:\j\workspace\innerloop_prtest4ca7949f\src\Run\Run.csproj]

chcosta:
This was related to my changes,  I forgot I added this code cleanup in Setup.cs, and then I left out Executor.cs from my changes because  hadn't made any functional changes there.

src/Run/Setup.cs
```diff
@@ -77,40 +121,83 @@ private string FindSettingType(string valueToFind)
             return null;
         }
 
-        public void prepareValues(string os, Dictionary<string, string> parameters, string configFile)
+        public void PrepareValues(string os, Dictionary<string, string> parameters, string configFile)
         {
             SettingParameters = new Dictionary<string, string>(parameters);
             Os = os;
             ConfigurationFilePath = configFile;
+            // Add defaults for run tool settings if they haven't been define din config.json
```
maririos: >define din

typo

src/Run/Setup.cs
```diff
@@ -77,40 +121,83 @@ private string FindSettingType(string valueToFind)
...
+        private void SetRunToolSettingsDefaults()
+        {
+            // If RunQuiet is already defined in config.json, don't override it
+            if (!Settings.ContainsKey(runQuietReservedKeyword))
+            {
+                Setting runQuietSetting = new Setting()
+                {
+                    Values = new List<string>() { "True", "False" },
+                    ValueType = "runToolSetting",
+                    Description = "Run tool specific setting.  Set to True to only display output from the executing command.  Default, False",
```
maririos: When printing for help, the default value is shown so I don't think it is necessary in the Description

src/Run/Setup.cs
```diff
@@ -77,40 +121,83 @@ private string FindSettingType(string valueToFind)
...
-        public void prepareValues(string os, Dictionary<string, string> parameters, string configFile)
+        public void PrepareValues(string os, Dictionary<string, string> parameters, string configFile)
         {
             SettingParameters = new Dictionary<string, string>(parameters);
             Os = os;
             ConfigurationFilePath = configFile;
+            // Add defaults for run tool settings if they haven't been define din config.json
+            SetRunToolSettingsDefaults();
+            ToolSettings = Settings.Where(s => s.Value.ValueType == null ||
+                                               s.Value.ValueType.Equals("runToolSetting")
```
weshaggard: Can you make "runToolSetting" a constant as it seems to be used a few places?

src/Run/Setup.cs
```diff
@@ -4,13 +4,17 @@
...
 using System.IO;
+using System.Linq;
 using System.Collections.Generic;
 using System.Text;
 
 namespace Microsoft.DotNet.Execute
 {
     public class Setup
     {
+        private static readonly string runQuietReservedKeyword = "RunQuiet";
```
weshaggard: Why not a constant string?

src/Run/Setup.cs
```diff
@@ -77,40 +121,83 @@ private string FindSettingType(string valueToFind)
...
         }
 
         public int ExecuteCommand(string commandSelectedByUser, List<string> parametersSelectedByUser)
         {
+            ParseRunToolSettings(commandSelectedByUser);
+            string runQuietValue;
+            bool runQuiet = false;
+            if (ToolSettings.TryGetValue(runQuietReservedKeyword, out runQuietValue))
+            {
+                runQuiet = runQuietValue.Equals("true", StringComparison.OrdinalIgnoreCase) ? true : false;
```
weshaggard: the `? true : false` isn't actually needed here as the value of .Equals will be exactly what you want :)

src/Run/Setup.cs
```diff
@@ -77,40 +121,83 @@ private string FindSettingType(string valueToFind)
...
-                else
+
+                int result = RunProcess.ExecuteProcess(commandToRun.ToolCommand, commandToRun.ParametersCommand);
+                if (!runQuiet)
                 {
-                    Console.ForegroundColor = ConsoleColor.Red;
-                    Console.WriteLine("Build Failed.");
+                    if (result == 0)
+                    {
+                        PrintColorMessage(ConsoleColor.Green, "Build Succeeded.");
```
weshaggard: orthogonal but I just noticed that we say Build Succeeded or Failed here but run isn't always doing a Build. @maririos we should consider updating this to Run command succeeded or failed.

src/Run/Setup.cs
```diff
@@ -77,40 +121,83 @@ private string FindSettingType(string valueToFind)
...
 
-        public void prepareValues(string os, Dictionary<string, string> parameters, string configFile)
+        public void PrepareValues(string os, Dictionary<string, string> parameters, string configFile)
         {
             SettingParameters = new Dictionary<string, string>(parameters);
             Os = os;
             ConfigurationFilePath = configFile;
+            // Add defaults for run tool settings if they haven't been define din config.json
+            SetRunToolSettingsDefaults();
+            ToolSettings = Settings.Where(s => s.Value.ValueType == null ||
```
chcosta: @maririos pointed out that there is no current validation that these ToolSettings are actually properly formed, and someone could add a new setting but leave out the ValueType, so then we would think it's a "run tool setting".  I'll add validation to ensure that anything which is not a run tool setting has proper values set for "ValueType".

src/Run/Setup.cs
chcosta: Nice

src/Run/Setup.cs
chcosta: A few iterations here led to this statement which is obviously redundant.  Thanks for catching it.

src/Run/Setup.cs
maririos: Good point. 
What about `Command Execution Succeeded or Failed` without the Run word? That way it is more clear that it is the result from the executed command and not from executing Run.

src/Run/Setup.cs
```diff
@@ -12,7 +12,8 @@ namespace Microsoft.DotNet.Execute
 {
     public class Setup
     {
-        private static readonly string runQuietReservedKeyword = "RunQuiet";
+        private const string runQuietReservedKeyword = "RunQuiet";
+        private const string runToolSettingValueTypeReservedKeyword = "runToolSetting";
```
weshaggard: These should be PascalCased. See (12) https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md. https://msdn.microsoft.com/en-us/library/x2dbyw72(v=vs.71).aspx

src/Run/Setup.cs
```diff
@@ -4,20 +4,56 @@
...
         public Dictionary<string, Setting> Settings { get; set; }
         public Dictionary<string, Command> Commands { get; set; }
         public Dictionary<string, Tool> Tools { get; set; }
         public Dictionary<string, string> SettingParameters { get; set; }
         public string Os { get; set; }
         public string ConfigurationFilePath { get; set; }
 
+        private int ValidateSettings()
+        {
+            int returnCode = 0;
```
maririos: Just out of curiosity, why are you using as a return type an int and not a bool?

chcosta: I figured this would get fleshed out more in the future and that there would be a possibility for having other returnCodes for warnings vs errors or something else.  Given that doesn't represent the current state of the code, I'm ok with making this a bool if it's confusing.

src/Run/Setup.cs
```diff
@@ -77,40 +142,92 @@ private string FindSettingType(string valueToFind)
...
+        public int PrepareValues(string os, Dictionary<string, string> parameters, string configFile)
         {
             SettingParameters = new Dictionary<string, string>(parameters);
             Os = os;
             ConfigurationFilePath = configFile;
+            // Add defaults for run tool settings if they haven't been defined in config.json
+            SetRunToolSettingsDefaults();
+
+            // Validate Settings before parsing out ToolSettings
+            int returnCode = ValidateSettings();
```
maririos: we could do a check here and if ``returnCode!=0` then return. That way we will avoid doing all the other things

chcosta: My thought was that the benefit of completing the "PrepareValues" code outweighed the perf gain by failing faster here.

src/Run/Setup.cs
```diff
@@ -57,6 +93,35 @@ private string ParseSettingValue(string inputValue)
...
+                    if (Commands.TryGetValue(commandSelectedByUser, out tempCommand))
+                    {
+                        tempCommand.DefaultValues.Settings.TryGetValue(toolSetting, out tempValue);
+                        ToolSettings[toolSetting] = tempValue;
+                    }
+                }
+                // Attempt to get run tool setting value from settings section
+                if (string.IsNullOrEmpty(ToolSettings[toolSetting]))
+                {
+                    Settings.TryGetValue(toolSetting, out tempSetting);
```
maririos: I'm wondering about this behavior. Looking at the value in Settings means that we are always going to use the default value of the Setting that in the majority of cases will always be turned off (with false or whatever).
But, if the user wants that particular setting to be used and it is not part of a command, then the user should specified that he wants it.
For example, right now the `runquiet` option is only in the `produces` command. I want to use it during sync,
so I should do: `sync -runQuiet` right?

All of this to say that maybe we shouldn't look at the value in the `Settings` dictionary, we should look for it in the `SettingParameters` dictionary https://github.com/dotnet/buildtools/pull/980/files#diff-6f4e4ec63e7b18f5f61e1d0ba5c10087L17

What do you think?

maririos: I completely missed this line: https://github.com/dotnet/buildtools/pull/980/files#diff-6f4e4ec63e7b18f5f61e1d0ba5c10087R104

Sorry :(

chcosta: The behavior here is to look for it as a parameter, then part of the command, then settings as a last default value.  I think this behavior meets your desire.

maririos:
LGTM

src/Run/Setup.cs
weshaggard: Yes something like that, another thing that I find useful for failed builds is also the exit code. So on failures it might be useful to say `Command execution failed with exit code {0}`.

maririos: PR #993

PR #984, opened 2016-08-25
Calling target `Test` with BeforeTargets in tests.targets

maririos:
In CoreFx we started hitting the problem that the result of the build was `Build Succeeded` even when we had test failures. It was not obvious until run.exe got implemented.

The problem relies on how the value of the MsBuild property `MSBuildLastTaskResult` was being propagated.
With the AfterTargets property, the result of the subsequent targets, like [RunTestsForProject](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L151) was not being returned to the target `Test` as this one had already finished before the execution of the tests happened.

With this change, the target `CheckTestPlatforms` and all its dependencies are executed and they return the exit code value before `Test` is executed, so the propagation of the value of `MSBuildLastTaskResult` happens correctly.

I tested the change in CoreFx (windows, ubuntu) and Build Tools.

Fix: dotnet/corefx#10430

cc: @weshaggard  @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -298,7 +298,7 @@
   </Target>
 
   <Target Name="CheckTestPlatforms"
-          AfterTargets="Test">
+          BeforeTargets="Test">
     <GetTargetMachineInfo Condition="'$(TargetOS)' == ''">
       <Output TaskParameter="TargetOS" PropertyName="TargetOS" />
     </GetTargetMachineInfo>


joperezr:
we should probably keep MSBuild team in the loop for this, in case they want to add more(or better) documentation around that property.

FYI: @rainersigwald

joperezr:
LGTM

weshaggard:
LGTM

rainersigwald:
Is this a correct understanding of the issue?

* A target `CheckTestPlatforms` was hooked into `Test` with `AfterTargets`
* In another project, there was an `MSBuild` task call specifying the `Test` target
* Following the [Target Build Order](https://msdn.microsoft.com/en-us/library/ee216359.aspx#Anchor_5), `Test` ran, followed by `CheckTestPlatforms`
* `CheckTestPlatforms` failed (as expected)
* Execution returned to the other project
* The return value from the `MSBuild` task was then inspected using `$(MSBuildLastTaskResult)`, and it indicated success (because `Test` succeeded, even though `CheckTestPlatforms` did not

weshaggard:
@rainersigwald Yes that is correct.

PR #985, opened 2016-08-25
Permit the use of TargetGroup in refs

ericstj:
We weren't honoring TargetGroup for subdirectories in reference assembly projects.
/cc @weshaggard


weshaggard:
LGTM

PR #987, opened 2016-08-26
Place the run tool into its own nuget package.

jhendrixMSFT:
To allow repos to consume the run tool without pulling down all of build
tools.  For now continue to place the run tool in the build tools package
until we're ready to make the complete split.


jhendrixMSFT:
CC @weshaggard @jaredpar @maririos

src/nuget/Microsoft.DotNet.BuildTools.Run.nuspec
```diff
@@ -0,0 +1,28 @@
...
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
+    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=518631</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <summary>.NET Core Framework Run Tool</summary>
+    <description>This package provides build-time support for the .NET Core Framework libraries.  You should not need to reference this package from your project.</description>
+    <copyright>Copyright © Microsoft Corporation</copyright>
+    <tags></tags>
+    <dependencies>
+    <!--
```
weshaggard: This comment isn't meaningful for this package (or the others any longer for that matter).

src/nuget/Microsoft.DotNet.BuildTools.Run.nuspec
```diff
@@ -0,0 +1,28 @@
...
+      Do not add any dependencies to this package as we want to install it versionless
+      (i.e. -ExcludeVersion) and we don't want the dependencies to be installed that way
+    -->
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="run\run.exe" target="tools\netcoreapp1.0" />
```
weshaggard: @jaredpar will this be enough in the package? Where will you get the matching shared framework?

jaredpar: I assumed the bootstrapping would pull down the shared framework / runtime with the package.

weshaggard: OK. I'd be interested in having a look at the bootstrapping code you use will use to start consuming the run tool. 

Another challenge we have been having is agreeing upon the versions of the CLI, because the bootstrapping code has to agree with the version in the runtimeconfig.json file in order to run against a shared framework (or we have to do something like buy-pass the muxer and go directly into the shared framework ourselves, in which case we don't need a runtimeconflig.json file).

src/nuget/Microsoft.DotNet.BuildTools.Run.nuspec
jhendrixMSFT: I've removed the comment.

src/nuget/Microsoft.DotNet.BuildTools.Run.nuspec
jhendrixMSFT: For the CLI version can we use a shared CLI-bootstrapper script on an HTTPS endpoint (still a bit queasy about this but how else can we share it) that is downloaded as part of the run bootstrapper?  This way any tool that requires a CLI can use this script and the version is consistent.  Of course it does mean that as the script rolls the version forward all tools that use it will also roll forward.  So the process would look something like the following.

run.bat checked into Foo repo, downloads CLI bootstrapper script and executes
run.bat restores version of run package and executes "real" run.bat script in the package which does the actual work

weshaggard: A potential shared script could work as well but as you mention even that will need to be versioned  in some way.

jhendrixMSFT: Versioned because the CLI version can change to be out-of-sync with runtimeconfig.json or something else?

weshaggard: We somehow have to keep the versions matching in the config (or eliminate the config) and we also need to make sure we have a compatible SharedFX.

jhendrixMSFT: Makes sense.  What is the best practice, i.e. what are we telling customers to do?  To me it feels like bundling the closure of dependencies and eliminating the config simplifies (then we only have to rely on restoring a runtime and host).  Would you agree?

weshaggard: We don't have a great story here. What do you mean by restoring a runtime and host? Do you mean the full shared FX? In either case you still have to worry about compatibility. The most resilient way would be to include everything in the package but then we would end up with a lot of duplication among our tools. I do think a shared framework is the best way we just have to figure out how we want to run on that without depending on a runtimeconfig file.

jhendrixMSFT: If we bundled the closure of FX libs in the package isn't all we'd need to restore is a runtime (coreclr) and a host (dotnet.exe), or does that still leave us with a compat burden (I would hope it to be a smaller one but perhaps I'm wrong)?
How do we bypass the need for a runtime config file?

jhendrixMSFT: After digging around a bit more I see my previous question doesn't make sense so you can ignore that one.

jhendrixMSFT:
@weshaggard I've updated the nuspec to include the dependent assemblies.

weshaggard:
What all does that *.dll reference pull in? I assume it is just the stuff not in the shared framework?

jhendrixMSFT:
Hmm actually it appears to pull in some stuff that's already in the shared framework.  The only delta I see is Newtonsoft.Json.dll and System.Runtime.Serialization.Primitives.dll.  Presumably I would only need to include these two dlls instead of the entire lot.

src/nuget/Microsoft.DotNet.BuildTools.Run.nuspec
```diff
@@ -0,0 +1,25 @@
...
+    <summary>.NET Core Framework Run Tool</summary>
+    <description>This package provides build-time support for the .NET Core Framework libraries.  You should not need to reference this package from your project.</description>
+    <copyright>Copyright © Microsoft Corporation</copyright>
+    <tags></tags>
+    <dependencies></dependencies>
+  </metadata>
+  <files>
+    <file src="run\run.exe" target="tools\netcoreapp1.0" />
+    <file src="run\run.runtimeconfig.json" target="tools\netcoreapp1.0" />
+    <file src="run\Newtonsoft.Json.dll" target="tools\netcoreapp1.0" />
```
weshaggard: While this will work for this case in isolation we will have to figure out how to deal with conflicts among our different tools when we get more of them doing this.

jhendrixMSFT: I was thinking about this as well.  We could put the tools into their own subdirectory under Tools, where the subdir name is that of the nuget package.  Thoughts?

jaredpar: That seems like a resonable approach.

jhendrixMSFT:
@weshaggard Any further comments or is this good to merge?

jhendrixMSFT:
I also updated the nuspec for the RepoUtils package to only include the assemblies that aren't a part of the shared framework.

src/nuget/Microsoft.DotNet.BuildTools.Run.nuspec
weshaggard: That is a fine first step but we may quickly find a fair amount of duplication.

weshaggard:
LGTM - Lets see how this plays out.

PR #988, opened 2016-08-26
Don't run ValidatePackage for noop builds

ericstj:
Packages can filter build by specifying PackagePlatform but we were
ignoring this when scheduling ValidatePackage.

/cc @weshaggard


weshaggard:
LGTM

PR #989, opened 2016-08-26
Fix System.Private package filtering

ericstj:
I broke this filtering when switching to package index.  Previously we'd
only trim System.Private package references if they were "unknown"
packages.  We did this to handle the netcore50 cases where we have
assemblies that don't ship as packages.

Previously we determined "unkown" by checking if the package was in the
BaselinePackage item.  Now we need to check both that item and the index.

This is a targeted fix to restore the thing I broke.
I'd like to refactor the entire dependency harvesting into a smaller
number of tasks but I'll do that in a future commit.

/cc @weshaggard @roncain


ericstj:
@weshaggard signed off in the meat world

weshaggard:
LGTM

PR #990, opened 2016-08-26
Ensure trailing slash for harvest path exclusions

ericstj:
Previously we weren't ensuring a trailing slash in the path, then doing
a startswith comparison.  This made paths like "lib/net" match
"lib/netstandard" which is not what we want.  Fix this by ensuring
a trailing slash on all exclusion paths.

/cc @tarekgh @weshaggard


tarekgh:
LGTM. thanks Eric for your quick turn around.

ericstj:
Thanks for the review.  BuildTools build should be 1.0.26-prerelease-00726-04 once it's done.

weshaggard:
:+1:

PR #991, opened 2016-08-26
Handle fillenames in TargetPath for exclusion dirs

ericstj:
My previous change didn't account for some targetpaths which may include
filenames.  Here we'll identify paths that have filenames and trim the
the filename.

/cc @weshaggard @tarekgh

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
@@ -312,7 +312,7 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
         {
             if (_pathsToExclude == null)
             {
-                _pathsToExclude = PathsToExclude.NullAsEmpty().Select(EnsureTrailingSlash).ToArray();
+                _pathsToExclude = PathsToExclude.NullAsEmpty().Select(EnsureDirectory).ToArray();
             }
 
             return ShouldSuppress(packagePath) ||
@@ -324,22 +324,48 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
         {
             if (_pathsToSuppress == null)
             {
-                _pathsToSuppress = PathsToSuppress.NullAsEmpty().Select(EnsureTrailingSlash).ToArray();
+                _pathsToSuppress = PathsToSuppress.NullAsEmpty().Select(EnsureDirectory).ToArray();
             }
 
             return _pathsToSuppress.Any(p => packagePath.StartsWith(p, StringComparison.OrdinalIgnoreCase));
         }
 
-        private static string EnsureTrailingSlash(string source)
+        private static string EnsureDirectory(string source)
         {
+            string result;
+
             if (source.Length < 1 || source[source.Length - 1] == '\\' || source[source.Length - 1] == '/')
             {
-                return source;
+                // already have a directory
+                result = source;
             }
             else
             {
-                return source + '/';
+                // could be a directory or file
+                var extension = Path.GetExtension(source);
+
+                if (extension != null && extension.Length > 0 && includedExtensions.Contains(extension, StringComparer.OrdinalIgnoreCase))
+                {
+                    // it's a file, find the directory portion
+                    var fileName = Path.GetFileName(source);
+                    if (fileName.Length != source.Length)
+                    {
+                        result = source.Substring(0, source.Length - fileName.Length);
+                    }
+                    else
+                    {
+                        // no directory portion, just return as-is
+                        result = source;
+                    }
+                }
+                else
+                {
+                    // it's a directory, add the slash
+                    result = source + '/';
+                }
             }
+
+            return result;
         }
 
         private static string GetTargetFrameworkFromPackagePath(string path)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
```diff
@@ -324,22 +324,49 @@ private bool ShouldSuppress(string packagePath)
...
             else
             {
-                return source + '/';
+                // could be a directory of file
+                string extension = Path.GetExtension(source);
+
+                if (extension != null && extension.Length > 0 && includedExtensions.Contains(extension, StringComparer.OrdinalIgnoreCase))
+                {
+                    // it's a file, find the directory portion
+                    int lastSlash = source.LastIndexOfAny(s_slashes);
```
ericstj: I don't use Path.GetDirectoryName since that will change / to \ which I will have to change back.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
```diff
@@ -324,22 +324,49 @@ private bool ShouldSuppress(string packagePath)
...
             if (source.Length < 1 || source[source.Length - 1] == '\\' || source[source.Length - 1] == '/')
             {
-                return source;
+                // already have a directory
+                result = source;
             }
             else
             {
-                return source + '/';
+                // could be a directory of file
```
weshaggard: of -> or

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
```diff
@@ -324,22 +324,49 @@ private bool ShouldSuppress(string packagePath)
...
+                // could be a directory or file
+                string extension = Path.GetExtension(source);
+
+                if (extension != null && extension.Length > 0 && includedExtensions.Contains(extension, StringComparer.OrdinalIgnoreCase))
+                {
+                    // it's a file, find the directory portion
+                    int lastSlash = source.LastIndexOfAny(s_slashes);
+                    if (lastSlash != -1)
+                    {
+                        result = source.Substring(0, lastSlash + 1);
```
weshaggard: Isn't this potentially chopping off chars from the end?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
tarekgh: can you use Path.GetDirectoryName just to check if we have file name and get the length of the filename and extension but you'll use the original string to chop off return the results. the reason I am suggesting this is that will work with any files regardless of the file extensions there.

---
In reply to: [76501340](https://github.com/dotnet/buildtools/pull/991#discussion_r76501340) [](ancestors = 76501340)

tarekgh: can you use Path.GetFileName to check if the path ends with the file name and in same time you'll get the file name length which you can use to chop off the original path string? so you'll not need to change / to \ at that time

---
In reply to: [76501340](https://github.com/dotnet/buildtools/pull/991#discussion_r76501340) [](ancestors = 76501340)

tarekgh:
@ericstj I have tried your tools package and worked fine :-) we should be good to go. it was nice also to get this tested to know there is no problem with buildtools on dev/api. 
please let me know when you submit it and which version would it be to consume in dev/apis

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
ericstj: Yep, that's the point.  If those characters represent a filename.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
ericstj: That'd still cause additional unnecessary allocations , but I agree it's more readable.  I'll make the change.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/HarvestPackage.cs
```diff
@@ -324,22 +324,48 @@ private bool ShouldSuppress(string packagePath)
...
+                    // it's a file, find the directory portion
+                    var fileName = Path.GetFileName(source);
+                    if (fileName.Length != source.Length)
+                    {
+                        result = source.Substring(0, source.Length - fileName.Length);
+                    }
+                    else
+                    {
+                        // no directory portion, just return as-is
+                        result = source;
```
weshaggard: What does it mean if there isn't a directory portion? does it mean it is a directory itself? If so shouldn't it get a slash added? If it is a file shouldn't you strip it off like above and just return a slash?

ericstj: It means it's a filename only, like "foo.dll".  We're only in this scope block if the string matched the extension pattern.

tarekgh:
for reference this change will correspond to the build tools version 1.0.26-prerelease-00729-01

PR #992, opened 2016-08-29
Add xunit Traitattribute to conditionally run tests on TFMs.

### LINKED ISSUE #960: Need a way to disable tests for different TestTFM
Now that we are running our tests for various TestTFM's we need a way to disable a test for a given TestTFM.

See https://github.com/dotnet/buildtools/blob/4bc8723ed65c6e6921649d196ed6f14bdfafdd7a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L61 for the list of TestTFMs.

Priya91:
cc @weshaggard @tarekgh @danmosemsft 

fixes #960

### DIFF
--- a/src/xunit.netcore.extensions/Attributes/SkipOnTargetFramework.cs
+++ b/src/xunit.netcore.extensions/Attributes/SkipOnTargetFramework.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    /// <summary>
+    /// Apply this attribute to your test method to specify this is a platform specific test.
+    /// </summary>
+    [TraitDiscoverer("Xunit.NetCore.Extensions.SkipOnTargetFrameworkDiscoverer", "Xunit.NetCore.Extensions")]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = false)]
+    public class SkipOnTargetFramework : Attribute, ITraitAttribute
+    {
+        public SkipOnTargetFramework(TargetFrameworkMoniker platform) { }
+    }
+}
--- a/src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
+++ b/src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
@@ -1,0 +1,25 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    [Flags]
+    public enum TargetFrameworkMoniker
+    {
+        Net45 = 1,
+        Net451 = 2,
+        Net452 = 4,
+        Net46 = 8,
+        Net461 = 16,
+        Net462 = 32,
+        Net463 = 64,
+        Netcore50 = 128,
+        Netcore50aot = 256,
+        Netcoreapp1_0 = 512,
+        Netcoreapp1_1 = 1024
+    }
+}
--- a/src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
+++ b/src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
@@ -10,16 +10,26 @@ namespace Xunit
     [Flags]
     public enum TargetFrameworkMoniker
     {
-        Net45 = 1,
-        Net451 = 2,
-        Net452 = 4,
-        Net46 = 8,
-        Net461 = 16,
-        Net462 = 32,
-        Net463 = 64,
-        Netcore50 = 128,
-        Netcore50aot = 256,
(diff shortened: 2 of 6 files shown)


src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
```diff
@@ -0,0 +1,25 @@
...
+        Net452 = 4,
+        Net46 = 8,
+        Net461 = 16,
+        Net462 = 32,
+        Net463 = 64,
+        Netcore50 = 128,
+        Netcore50aot = 256,
+        Netcoreapp1_0 = 512,
+        Netcoreapp1_1 = 1024
+    }
```
tarekgh: could you write the numbers in hex please? make it easier for looking at what bit is  enabled

tarekgh:
LGTM

karajas:
LGTM. Thanks @Priya91

weshaggard:
@stephentoub could you please have a look to see if this would cover the cases you wanted?

It looks reasonable to me.

src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
```diff
@@ -0,0 +1,25 @@
...
+    [Flags]
+    public enum TargetFrameworkMoniker
+    {
+        Net45 = 1,
+        Net451 = 2,
+        Net452 = 4,
+        Net46 = 8,
+        Net461 = 16,
+        Net462 = 32,
+        Net463 = 64,
```
stephentoub: Could we restructure these values so that there's a value that covers all of desktop, a value that covers all of netcoreapp, etc.?  Otherwise each time a new version comes out we'd likely need to back and update all of the corresponding attributions.
e.g.
```C#
[Flags]
public enum TargetFrameworkMoniker
{
    NetFramework = 0x1,
    NetCoreApp = 0x2,
    NetCore50 = 0x4,
    
    NetCore45 = 0x100 | NetFramework,
    NetCore451 = 0x200 | NetFramework,
    NetCore453 = 0x400 | NetFramework,
    ...
}
```

stephentoub:
> @stephentoub could you please have a look to see if this would cover the cases you wanted?

Left one comment; otherwise LGTM.

src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
Priya91: Is NetCore45 the same as net45 ?

stephentoub: That was just a typo.

src/xunit.netcore.extensions/Attributes/SkipOnTargetFramework.cs
```diff
@@ -0,0 +1,19 @@
...
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    /// <summary>
+    /// Apply this attribute to your test method to specify this is a platform specific test.
+    /// </summary>
+    [TraitDiscoverer("Xunit.NetCore.Extensions.SkipOnTargetFrameworkDiscoverer", "Xunit.NetCore.Extensions")]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = false)]
+    public class SkipOnTargetFramework : Attribute, ITraitAttribute
```
justinvp: This should be named `SkipOnTargetFrameworkAttribute` (with the Attribute suffix).

src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
```diff
@@ -0,0 +1,35 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    [Flags]
+    public enum TargetFrameworkMoniker
```
justinvp: This should be plural because its a Flags enum.

Priya91: Not necessary, PlatformId is just PlatformId.

src/xunit.netcore.extensions/Attributes/SkipOnTargetFramework.cs
Priya91: This is not a compiler error, but to keep up with the naming scheme of other attributes here, will rename.

src/xunit.netcore.extensions/TargetFrameworkMoniker.cs
justinvp: `PlatformID` really should be plural as well, to follow the design guidelines. Then it wouldn't conflict with `System.PlatformID`.

justinvp: https://msdn.microsoft.com/en-us/library/ms229040(v=vs.110).aspx

> ✓ DO use a plural type name for an enumeration with bit fields as values, also called flags enum.

Priya91: PlatformID rename requires work in other repos on buildtools update, feel free to submit a PR or file an issue.

weshaggard: Yes we should probably make this plural.

PR #993, opened 2016-08-29
Fix help for command without alias

### LINKED ISSUE #969: Run, doesn't print description if no alias provided
If you create a command, and it only contains "defaultValues", no "alias" section, no description is displayed from "run -?"

maririos:
- Changes output message after running the command

- Fix #969 

cc: @chcosta @weshaggard 

As an example, for the [produces](https://github.com/dotnet/corefx/blob/master/config.json#L404) command in CoreFx it will print:
```
produces

Usage: run produces [Action] (global settings)

                                        Settings:
                                        Project = src/packages.builds
                                        ProducesTarget =

                                        It will run:
                                        D:\mariariCoreFx\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary
                                        /maxcpucount /nodeReuse:false
                                        /l:BinClashLogger,Tools\net45\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log
                                        src/packages.builds /t:ProducesTarget
```

### DIFF
--- a/src/Run/CommandLine.cs
+++ b/src/Run/CommandLine.cs
@@ -762,8 +762,7 @@ class CommandLine
                     }
                 }
             }
-
-            if (hasQualifiers)
+            else if (hasQualifiers)
             {
                 sb.AppendLine().Append("  Actions:").AppendLine();
 
@@ -779,6 +778,11 @@ class CommandLine
                         Wrap(sb, commandSettingsHelp, QualifierSyntaxWidth + 7, new string(' ', QualifierSyntaxWidth + 7), maxLineWidth, false);
                     }
                 }
+            }
+            else
+            {
+                string commandSettingsHelp = _setupContent.GetHelpCommand(parameterSetParameter.Name);
+                Wrap(sb, commandSettingsHelp, QualifierSyntaxWidth + 7, new string(' ', QualifierSyntaxWidth + 7), maxLineWidth, false);
             }
 
             string globalQualifiers = null;
--- a/src/Run/CommandLine.cs
+++ b/src/Run/CommandLine.cs
@@ -750,7 +750,7 @@ class CommandLine
 
             if (hasParameters)
             {
-                sb.AppendLine().Append("  Parameters:").AppendLine();
+                sb.AppendLine().Append("Parameters:").AppendLine();
                 for (int i = parameterSetBody; i < _parameterDescriptions.Count; i++)
                 {
                     CommandLineParameter parameter = _parameterDescriptions[i];
@@ -764,7 +764,7 @@ class CommandLine
             }
             else if (hasQualifiers)
             {
-                sb.AppendLine().Append("  Actions:").AppendLine();
+                sb.AppendLine().Append("Actions:").AppendLine();
 
                 for (int i = parameterSetBody; i < _parameterDescriptions.Count; i++)
                 {
@@ -775,14 +775,14 @@ class CommandLine
                     {
                         ParameterHelp(parameter, sb, QualifierSyntaxWidth, maxLineWidth);
                         string commandSettingsHelp = _setupContent.GetHelpCommand(parameterSetParameter.Name, parameter.Name);
-                        Wrap(sb, commandSettingsHelp, QualifierSyntaxWidth + 7, new string(' ', QualifierSyntaxWidth + 7), maxLineWidth, false);
+                        Wrap(sb, commandSettingsHelp, QualifierSyntaxWidth, new string(' ', QualifierSyntaxWidth), maxLineWidth, false);
                     }
                 }
             }
             else
             {
                 string commandSettingsHelp = _setupContent.GetHelpCommand(parameterSetParameter.Name);
-                Wrap(sb, commandSettingsHelp, QualifierSyntaxWidth + 7, new string(' ', QualifierSyntaxWidth + 7), maxLineWidth, false);
+                Wrap(sb, commandSettingsHelp, QualifierSyntaxWidth, new string(' ', QualifierSyntaxWidth), maxLineWidth, false);
             }
 
             string globalQualifiers = null;
@@ -1690,7 +1690,7 @@ class CommandLine
         private static void ParameterHelp(CommandLineParameter parameter, StringBuilder sb, int firstColumnWidth, int maxLineWidth)
         {
             // TODO alias information.
-            sb.Append("    ").Append(parameter.Syntax().PadRight(firstColumnWidth)).Append(' ');
+            sb.Append(parameter.Syntax().PadRight(firstColumnWidth));
             string helpText = parameter.HelpText;
             string defValue = string.Empty;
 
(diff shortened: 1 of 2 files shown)


src/Run/Setup.cs
```diff
@@ -411,19 +411,22 @@ private string GetProject(Command commandToExecute, List<string> parametersSelec
         {
             string project = string.Empty;
             bool moreThanOneProject = false;
-            foreach (string param in parametersSelectedByUser)
+            if(parametersSelectedByUser != null)
```
chcosta: You could also do this as a Linq command, something like

```C#
if (parametersSelectedByUser.Count(p => commandToExecute.Alias[p].Settings.TryGetValue("Project", out project)) > 1) 
{
   Console.Error.WriteLine("Error: There can only be one project execution per command.");
   return string.Empty;
}
```

src/Run/CommandLine.cs
```diff
@@ -762,8 +762,7 @@ public string GetHelp(int maxLineWidth, string parameterSetName = null, bool dis
                     }
                 }
             }
-
-            if (hasQualifiers)
+            else if (hasQualifiers)
```
chcosta: Are these mutually exclusive options (parameters vs qualifiers)?  If parameters aren't supported, may want to consider throwing an exception for "hasParameters".  If parameters and qualifiers are both legit options, then maybe change this to 

`if (hasQualifiers)
{
  ...
}
if(!hasQualifiers && !hasParameters)`

src/Run/Setup.cs
maririos: Thanks Chris

src/Run/CommandLine.cs
maririos: Looking at [this](https://github.com/dotnet/buildtools/blob/master/src/Run/CommandLine.cs#L736), they are mutually exclusive options.

Positional parameters are not being used by run.exe so this code is never going to be called. If there is a desire to start using them the design of run.exe should change and it would involve taking a look at the whole CommandLine class.
For now I consider it safe to leave it there as it is.

weshaggard: @maririos please keep in mind that you want to merge this back with the other CommandLine.cs file at some point so lets try not to make the assumption that it is only used by the run tool. As for the actual issue here I didn't dig in so I'm trusting your understanding that the are mutually exclusive.

src/Run/Setup.cs
```diff
@@ -410,20 +410,16 @@ private string GetTool(Command commandToExecute, string os, string configPath, L
         private string GetProject(Command commandToExecute, List<string> parametersSelectedByUser)
         {
             string project = string.Empty;
-            bool moreThanOneProject = false;
-            foreach (string param in parametersSelectedByUser)
+            
+            if(parametersSelectedByUser != null)
             {
-                if (commandToExecute.Alias[param].Settings.TryGetValue("Project", out project))
+                if (parametersSelectedByUser.Count(p => commandToExecute.Alias[p].Settings.TryGetValue("Project", out project)) > 1)
```
weshaggard: Use .Any instead of .Count as that will stop after the first one as opposed to actually counting all of them.

weshaggard:
Will the command really be indented as far as in your comment? If so why?

src/Run/Setup.cs
chcosta: Isn't any more appropriate for "at least one" whereas this is checking for "greater than one"?

weshaggard: Your correct. I incorrectly thought this was checking >= 1. I retract my comment, Count should work for your purposes.

src/Run/CommandLine.cs
maririos: >so lets try not to make the assumption that it is only used by the run tool

Yes. That's why I don't want to change or delete that part of the code.

maririos:
Yes. I tested this changes in CoreFx and pasted the output as it is. It matches the indentation of the other commands and settings because of https://github.com/dotnet/buildtools/pull/993/files#diff-3962b2edd41ed8b45a26e5794b8f2a48R785

weshaggard:
That looks like a lot of wasted screen real-estate can we change that indention to only be a couple spaces?

maririos:
Yeah, we could. 
All together doesn't look that bad though:
```
clean

Usage: run clean [Action] (global settings)

  Actions:
    [-b]                     Deletes the binary output directory.


                               Settings:
                               CleanAllProjects =
                               MsBuildLogging = /flp:v=normal;LogFile=clean.log

                               It will run:
                               c:\mariariBT\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /maxcpucount
                               /nodeReuse:false
                               /l:BinClashLogger,Tools\net45\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log
                               /t:CleanAllProjects /flp:v=normal;LogFile=clean.log

    [-p]                     Deletes the repo-local nuget package directory.


                               Settings:
                               CleanPackages =
                               MsBuildLogging = /flp:v=normal;LogFile=clean.log

                               It will run:
                               c:\mariariBT\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /maxcpucount
                               /nodeReuse:false
                               /l:BinClashLogger,Tools\net45\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log
                               /t:CleanPackages /flp:v=normal;LogFile=clean.log

    [-c]                     Deletes the user-local nuget package cache.


                               Settings:
                               CleanPackagesCache =
                               MsBuildLogging = /flp:v=normal;LogFile=clean.log

                               It will run:
                               c:\mariariBT\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /maxcpucount
                               /nodeReuse:false
                               /l:BinClashLogger,Tools\net45\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log
                               /t:CleanPackagesCache /flp:v=normal;LogFile=clean.log


----------------------------------------------------------------------------------------------------------------------
Global settings to all commands:

    [-?]                     Print this help guide.

    [-OSGroup]               OS Group for result binaries.
                             => Default value: AnyOS
                             => Legal values: [AnyOS, Windows_NT, Linux, OSX, FreeBSD, NetBSD].

    [-RestoreDuringBuild]    Enables/Disables the restoration of packages.
                             => Default value: true
                             => Legal values: [True, False].
```

weshaggard:
still looks like there is a bunch of wasted space and unnecessary line wrapping.

maririos:
I deleted the extra spaces and now the output looks like this:
```
build-managed

Usage: run build-managed [Action] (global settings)

Actions:
[-binaries]    Only builds binaries.  It doesn't restore packages.


               Settings:
               RestoreDuringBuild = false
               MsBuildLogging = /flp:v=normal

               It will run:
               c:\mariariBT\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /maxcpucount /nodeReuse:false
               /l:BinClashLogger,Tools\net45\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log
               /p:RestoreDuringBuild=false /flp:v=normal

[-packages]    Builds the NuGet packages.


               Settings:
               Project = src/packages.builds
               MsBuildLogging = /flp:v=normal

               It will run:
               c:\mariariBT\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /maxcpucount /nodeReuse:false
               /l:BinClashLogger,Tools\net45\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log src/packages.builds
               /flp:v=normal

[-debug]       Sets ConfigurationGroup=Debug or the value passed by the user.


               Settings:
               ConfigurationGroup = Debug
               MsBuildLogging = /flp:v=normal

               It will run:
               c:\mariariBT\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /maxcpucount /nodeReuse:false
               /l:BinClashLogger,Tools\net45\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log
               /p:ConfigurationGroup=Debug /flp:v=normal

----------------------------------------------------------------------------------------------------------------------
Global settings to all commands:

[-?]                     Print this help guide.

[-OSGroup]               OS Group for result binaries.
                         => Default value: AnyOS
                         => Legal values: [AnyOS, Windows_NT, Linux, OSX, FreeBSD, NetBSD].

[-RestoreDuringBuild]    Enables/Disables the restoration of packages.
                         => Default value: true
                         => Legal values: [True, False].

[-Platform]              Sets the value of the platform.
                         => Default value: AnyCPU
                         => Legal values: [AnyCPU, x86, arm, x64, amd64].
```

weshaggard:
Looks better.

change LGTM

PR #994, opened 2016-08-30
Refactor package destination calculation

ericstj:
This change refactors how we calculate the destination path of a library
in a package.

Previously PackageTargetFramework property (single value) or
PackageDestination (many values, precalculated path) must be
specified.

This had a few problems:
1. PackageTargetFramework was often the same information that was in
NuGetTargetMoniker.
2. The only means for specifying more than one destination was too
verbose and required duplicating the target moniker in the path and
TargetFramework metadata.

This fixes both of those issues by allowing folks to specify a list of
PackageTargetFrameworks either in the PackageTargetFramework
property or item.  When using the item each PackageTargetFramework
can optionally specify a TargetRuntime as well.  Additionally, if folks
wish to package using the same target moniker as NuGetTargetMoniker,
we will fallback to using this if nothing else is specified.

In doing this I refactored how we do our asset promotion for ref/lib for
desktop to make it clearer and more consistently named.  This is a minor
breaking change, so I mitigated that by preserving the old property name
PackageAsRefAndLib.

This change also removes the ValidatePackageTargetFramework target.
That target's purpose was to calculate the netstandard version of an
assembly based on references but its no longer needed since we can now
use NuGet.  It's replacement is validation in GetPackageDestination
which ensures that we package things using a compatible TFM to what
we build them against.

/cc @weshaggard @chcosta @joperezr


ericstj:
CoreFx PR to fix current validation issues: https://github.com/dotnet/corefx/pull/11253

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetPackageDestination.cs
```diff
@@ -0,0 +1,154 @@
...
+                    !compat.IsCompatible(packageFx, nuGetFx) && 
+                    !suppressions.Contains(packageFx))
+                {
+                    Log.LogError($"Project is built as {nuGetFx.GetShortFolderName()} but packaged as {packageFx.GetShortFolderName()} which are not compatible; A {nuGetFx.GetShortFolderName()} asset cannot be referenced by a {packageFx.GetShortFolderName()} project.  To suppress this error you can add <SuppressPackageTargetFrameworkCompatibility Include=\"{packageFx.GetShortFolderName()}\" /> to your project file.");
+                }
+
+                string runtime = packageTargetFramework.GetMetadata("TargetRuntime") ?? PackageTargetRuntime;
+
+
+                Add(packageFx, PackageTargetRuntime);
```
chcosta: should this be `Add(packageFx, runtime);`

ericstj: Yeah, had a fix locally but didn't push.  It's actually a bit more complicated.

ericstj:
Here's what we can do to corefx after we have this in:
https://github.com/ericstj/corefx/commit/22416b48c0c0a44553617fcab5d01ba739885b10
https://github.com/ericstj/corefx/commit/d281cf3cee12c0386496e37a66562555f6c5e28c

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetPackageDestination.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                string runtime = PackageTargetRuntime;
+
+                // MSBuild returns an empty string for both non-existent meta-data as well as existent, but empty metadata.
+                // If someone has defined an empty string here, we want to honor that over PackageTargetRuntime.
+                if (packageTargetFramework.MetadataNames.Cast<string>().Any(md => md.Equals("TargetRuntime")))
+                {
+                    runtime = packageTargetFramework.GetMetadata("TargetRuntime");
+                }
+
```
weshaggard: nit: extra line

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetPackageDestination.cs
ericstj: removed.

PR #995, opened 2016-08-30
Condition the use of locally built shims on a new property instead of TestWithLocalLibraries

joperezr:
cc: @ericstj @weshaggard 

This is required in order to remove the use of TestWithLocalLibraries in corefx CI.

related comment: https://github.com/dotnet/corefx/issues/11203#issuecomment-243207773

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -80,8 +80,11 @@
       <Output Condition="'$(TargetOS)'==''" TaskParameter="TargetOS" PropertyName="TargetOS" />
     </GetTargetMachineInfo>
 
+    <PropertyGroup>
+      <TestWithLocalNativeLibraries Condition="'$(TestWithLocalLibraries)'=='true'">true</TestWithLocalNativeLibraries>
+    </PropertyGroup>
     <!-- Test using locally built shim libraries. -->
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -80,8 +80,11 @@
       <Output Condition="'$(TargetOS)'==''" TaskParameter="TargetOS" PropertyName="TargetOS" />
     </GetTargetMachineInfo>
 
+    <PropertyGroup>
+      <UseLocallyBuiltShimLibraries Condition="'$(TestWithLocalLibraries)'=='true'">true</UseLocallyBuiltShimLibraries>
```
ericstj: Consider calling this TestWithLocalNativeLibraries.

ericstj:
LGTM, we should have an issue tracking a fix to TestWithLocalLibraries that uses pkgprojs.

weshaggard:
LGTM 

@Priya91 do we still need to do this at all? I thought you mentioned about using the pkgproj's for the shims now to avoid this.

Priya91:
> @Priya91 do we still need to do this at all? I thought you mentioned about using the pkgproj's for the shims now to avoid this.

Yeah that was initial idea to not have dependency over TestWithLocalLibraries for native shims as well. But i haven't gotten to implementing it.. The only difference between this and using pkgproj will be we copy only the required native shims as opposed to all of the native shims.

PR #1003, opened 2016-09-01
Allow specifying metadata for skip reason.

Priya91:
cc @weshaggard @joperezr


joperezr:
LGTM. is the idea to add the issue # as the reason?

weshaggard:
LGTM as well. Should we do this for the other attributes?

Priya91:
> is the idea to add the issue # as the reason?

Yes issue number or in general reason for different behavior, I find myself commenting a lot while applying this attribute.

Priya91:
> Should we do this for the other attributes?

Haven't had a need to specify reason on applying other attributes as much as this one. We can add, but it will most probably not be used at all..

PR #1006, opened 2016-09-01
Gen api updates

weshaggard:
cc @ericstj @terrajobst 

Adds attribute filtering to genapi, including changes to cci extensions to filter some fake attributes.
Adds support for genapi to output one file per assembly.

### DIFF
--- a/src/GenAPI/Program.cs
+++ b/src/GenAPI/Program.cs
@@ -35,25 +35,33 @@ namespace GenAPI
                 return 1;
             }
 
-            using (TextWriter output = GetOutput())
-            using (IStyleSyntaxWriter syntaxWriter = GetSyntaxWriter(output))
-            {
-                if (!String.IsNullOrEmpty(s_headerFile))
-                {
-                    if (!File.Exists(s_headerFile))
-                    {
-                        Console.WriteLine("ERROR: header file '{0}' does not exist", s_headerFile);
-                        return 1;
-                    }
+            string headerText = GetHeaderText();
+            bool loopPerAssembly = Directory.Exists(s_out);
 
-                    using (TextReader headerText = File.OpenText(s_headerFile))
+            if (loopPerAssembly)
+            {
+                foreach (var assembly in assemblies)
+                {
+                    using (TextWriter output = GetOutput(GetFilename(assembly)))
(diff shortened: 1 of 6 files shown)


terrajobst:
Very cool! Should we change the ICciFilter interface to explicitly handle fake attributes differently? It seems consumers have to downcast in order to get the type name as a string.

terrajobst:
Speaking of which, I assume we want to port these changes to CoreFxTools, too?

weshaggard:
I prefer we make this the source of truth and just merge any remaining changes from there (if any) and then start consuming the package this repo produces. 

> ICciFilter interface to explicitly handle fake attributes differently? It seems consumers have to downcast in order to get the type name as a string.

I thought about it but I'm not sure I wanted to go that route. I would much prefer to make FakeAttribute more like a real one but that took more work then I wanted to put into this right now.

joperezr:
@weshaggard  does this fix #658 #904 ?

ericstj:
+1 @joperezr's comment, while you're in there can you fix #658?  PublicOnlyCciFilter needs to check if the type arguments to an attribute are public types here: https://github.com/dotnet/buildtools/blob/b6663b70b5956f352af75d7d9accf89baa13929c/src/Microsoft.Cci.Extensions/Filters/PublicOnlyCciFilter.cs#L66.

akoeplinger:
#658 was fixed as part of https://github.com/dotnet/buildtools/pull/922 already.

ericstj:
Right you are.  LGTM

terrajobst:
@weshaggard 

> I prefer we make this the source of truth and just merge any remaining changes from there (if any) and then start consuming the package this repo produces.

Sounds good to me. Do you mind if submit a PR with the remaining diff? I think I changed some things in the last three months.

> I thought about it but I'm not sure I wanted to go that route. I would much prefer to make FakeAttribute more like a real one but that took more work then I wanted to put into this right now.

I'd prefer that, too, but in its current version that's not the case, right?

weshaggard:
@terrajobst I've created https://github.com/dotnet/buildtools/issues/1059 to track the merging.

bartonjs:
@weshaggard Aha, I was wondering why the executable you were using had a feature that the one extracted to corefx\Tools didn't have :smile:.

PR #1007, opened 2016-09-02
Add reason message to OuterLoopAttribute

stephentoub:
We're often using comments to indicate why a test is being marked as OuterLoop.  This lets us do that as a string.  Then once all use has been updated to specify a message, we can remove the ctor without the message, and devs will be forced to specify a reason for why a test is marked as OuterLoop.

cc: @weshaggard, @bartonjs, @ianhays


bartonjs:
LGTM.  Let's see how well it holds up in practice :smile:

ianhays:
Sounds like a good idea to me!

weshaggard:
:+1:

PR #1008, opened 2016-09-02
Fixing two conditions that also need to trigger when TestWithLocalNativeLibraries is true

joperezr:
This change is needed for dotnet/corefx#11366 since when we moved to use TestWithLocalNativeLibraries property we actually broke using the native shims, which now is causing non-Windows CI builds to fail for dev/api. 

cc: @ericstj


joperezr:
Merging now since we need this for corefx. I'll address any feedback in a follow-up PR.

ericstj:
LGTM

PR #1009, opened 2016-09-02
Fix default config file path for run tool.

jhendrixMSFT:
Remove a hard-coded assumption on the location of the run tool, using the
current working directory instead.  Removed some duplicate code.

### DIFF
--- a/src/Run/Executor.cs
+++ b/src/Run/Executor.cs
@@ -4,9 +4,9 @@
 
 using Newtonsoft.Json;
 using System;
-using System.Reflection;
 using System.Collections.Generic;
 using System.IO;
+using System.Reflection;
 using Microsoft.Fx.CommandLine;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
CC @weshaggard @maririos

maririos:
One of the ideas of run.exe is that we should be able to run it in individual projects too.
Using `GetCurrentDirectory` will enforce to always pass the location of the config.json file if we are not running from the root directory

jhendrixMSFT:
Using the wrapper script this shouldn't be a problem though as we can either set the working directory or the path to the config file.  Am I missing something?

maririos:
That is true, we could add the call of the config.json file in the wrapper scripts.

maririos:
Could you do it when updating the BT version on the repos?

jhendrixMSFT:
Which repos need to be updated?  Presumably we can update them before updating their version of build tools, case in point I've updated this PR with the necessary change to run.cmd.

src/Run/Executor.cs
```diff
@@ -25,24 +24,15 @@ public Executor()
...
-            if (configFile == null)
-            {
-                string executorDirectory = Path.GetDirectoryName(typeof(Executor).GetTypeInfo().Assembly.Location);
-                configFilePath = Path.GetFullPath(Path.Combine(executorDirectory, @".."));
-            }
-            else
-            {
-                configFilePath = Path.GetDirectoryName(configFile);
-            }
+            configFilePath = Path.GetDirectoryName(configFile);
```
weshaggard: Why wouldn't we actually accept the full config file path as opposed to hardcoding that the file must be called config.json

jhendrixMSFT: Good point, I can make that change.

src/Run/Executor.cs
```diff
@@ -25,24 +24,15 @@ public Executor()
         {
             SettingParameters = new Dictionary<string, string>();
             CommandParameters = new Dictionary<string, Dictionary<string, string>>();
-            string executorDirectory = Path.GetDirectoryName(typeof(Executor).GetTypeInfo().Assembly.Location);
-            configFilePath = Path.GetFullPath(Path.Combine(executorDirectory, @".."));
+            configFilePath = Directory.GetCurrentDirectory();
```
weshaggard: Just to make sure we don't cause a bunch of breaks I would suggest staging this, by keeping the old behavior as well as adding the new behavior. If `..\config.json` exists use it but if not look in the current directory.

jhendrixMSFT: Good idea I've made this change.

jhendrixMSFT:
@weshaggard Any other comments?

weshaggard:
LGTM

PR #1011, opened 2016-09-02
Support repos that don't define package lists

ericstj:
Allow for packaging to work in the absense of a package index, stable
package list, baseline package list, and native module to package list.

@dagood this should fix the issue you were seeing with coreclr.

/cc @weshaggard


dagood:
LGTM, thanks!

PR #1014, opened 2016-09-06
Support ubuntu.16.10 and opensuse.42.1 in tool-runtime project.json

mellinoe:
This allows openSUSE 42.1 and ubuntu 16.10 to publish buildtools without an alternate `__PUBLISH_RID`. The NETStandard.Library version referenced here is a recent build, from after [this](https://github.com/dotnet/corefx/pull/11356) was merged in last week.

@ericstj , @dagood

### DIFF
--- a/src/ApiCompat/project.json
+++ b/src/ApiCompat/project.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "Microsoft.Cci": "4.0.0-rc3-24214-00",
+    "Microsoft.Cci": "4.0.0-rc4-24217-00",
     "Microsoft.Composition": "1.0.30",
     "Microsoft.NETCore.App": {
       "type": "platform",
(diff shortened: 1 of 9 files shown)


mellinoe:
@wesh

mellinoe:
@weshaggard

src/ApiCompat/project.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "Microsoft.Cci": "4.0.0-rc3-24214-00",
+    "Microsoft.Cci": "4.0.0-rc4-24217-00",
```
weshaggard: What's with the RC4 change here?

dagood: It's because `ProjectKTfsCurrentRef` stayed what it was when we RTM'd, but `DependencyBranch` turned to master--so we got what we last built on "master" (ProjectK) when we RTM'd. https://github.com/dotnet/versions/blob/b6fe7bf0d879e7a750b15e90743d5191e3b478de/build-info/dotnet/projectk-tfs/master/Latest_Packages.txt

We should either separate out the build-info branches, or update the CoreCLR and ProjectK SHAs to latest too. For the most part this should have no effect because dependencies are mostly stable at the moment.

PR #1015, opened 2016-09-06
Register ValidPinvokeMappings for netstandard TargetGroups

bartonjs:
The description of the assignment of ValidPinvokeMappings says that it is
OneCoreUwp for UWP and OneCore otherwise; but the implementation is
a much more complicated setup.

Projects may need to build for two different netstandard versions, and the
easiest way for them to do so in corefx is with the TargetGroup property
(e.g. TargetGroup=netstandard1.3); but assigning TargetGroup to anything
which is not a special value in this file makes the ValidPinvokeMappings
variable be unset, which means all p/invokes fail.

This change makes all Windows_NT builds use the OneCore list, unless
the stronger OneCoreUWP list should be chosen instead.

cc: @ericstj @weshaggard @AlexGhiondea

### DIFF
--- a/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
+++ b/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
@@ -17,9 +17,7 @@
        2. For all other windows configuration we use OneCoreAPIs.txt file.
     -->
 
-    <ValidPinvokeMappings Condition="'$(TargetGroup)'==''">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
-    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard13aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
-    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard15aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
+    <ValidPinvokeMappings>$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -17,9 +17,7 @@
        2. For all other windows configuration we use OneCoreAPIs.txt file.
     -->
 
-    <ValidPinvokeMappings Condition="'$(TargetGroup)'==''">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
-    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard13aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
-    <ValidPinvokeMappings Condition="'$(TargetGroup)'=='netstandard15aot'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
+    <ValidPinvokeMappings>$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
```
weshaggard: I'm fine with this change assuming everything builds. It does actually cause this to be included and thus validated on more projects but I suspect that is probably OK.

weshaggard:
LGTM assuming this doesn't break any existing builds.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
bartonjs: What it does right now is launch the P/Invoke scanner, but with no input source it rejects everything as prohibited.

That said, I don't actually know how to verify this for real... just edit the extracted copy in Tools? Or is there a better approach? (I hope "better approach" :smile:)

weshaggard: Editing the extracted copy in the Tools folder is a decent way to verify things like this. If you want a more complete way then see https://github.com/dotnet/buildtools/wiki/Local-testing-of-BuildTools-in-another-repo

bartonjs:
Validated the change by making the same edit to the extracted version in Tools and building System.Security.Cryptography.Cng with a TargetGroup of netstandard1.6.  I have a corefx build from root running now with the local edit, and when it passes I'll hit merge.

bartonjs:
And it has now passed.  Thanks for the review and doc pointers, Wes.

Issue #1013, opened 2016-09-02
Create common pre-release version used for all of the CoreFx packages.


StephenBonikowsky:
Issue identified by @ericstj excerpt from e-mail...

> Just wanted to share that I think the WCF dependency versions (pre-release portion) is going to pretty busted right now.  The reason being that WCF is building with prerelease version X and using the same version X for the CoreFx dependencies in its packages, but that might not match any corefx version that was ever built (suppose CoreFx is version X+2).  The right thing to do would be to make changes to buildtools to have some common pre-release version used for all of the CoreFx packages.  I could fix that in buildtools, but given that it is point-in-time (goes away when we remove the pre-release for RTM) I don’t think we should try to solve it for 1.0.  I’ll fix this for 1.1.

PR #1020, opened 2016-09-07
Add PreReleaseGroup support

ericstj:
This allows for PackageIndexes to specify a PreReleaseGroup.  This can
be specified at the index level or package level in the package index
file.

If PreReleaseGroups are used in the index then the PreReleaseGroup item
should be defined to provide a mapping from group name to the version
suffix to be used for that group.

For example, for the WCF repo:
```xml
<ItemGroup>
  <PreReleaseGroup Include="CoreFx">
    <PreRelease>$(CoreFxExpectedPrerelease)</PreRelease>
  </PreReleaseGroup>
  <PreReleaseGroup Include="WCF">
    <PreRelease>$(VersionSuffix)</PreRelease>
  </PreReleaseGroup>
<ItemGroup>
```

Fixes https://github.com/dotnet/buildtools/issues/1013
/cc @StephenBonikowsky @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyPreReleaseSuffix.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyPreReleaseSuffix.cs
@@ -89,7 +89,7 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
                 if (!IsStable(packageId, packageVersion))
                 {
                     // pre-release, set with suffix
-                    updatedPackage.SetMetadata("Version", packageVersion.ToString() + PreReleaseSuffix);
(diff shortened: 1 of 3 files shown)


StephenBonikowsky:
Thanks for doing this Eric!

ericstj:
Here's the PR for Corefx to add this info to CoreFx repo.  I'll do a similar (smaller) change for WCF.

ericstj:
Ok I changed the implementation here.  Rather than requiring pre-release groups I just stamp the pre-release string directly in the package index file.  This now leads to simpler consumption: no changes required in WCF other than picking up BuildTools and updating the coreFx package versions.  In corefx I just call a task to update the index file before we package it.

CoreFx side: https://github.com/dotnet/corefx/pull/11522/commits/b28ec58d4b30f4d92f2740631b9f4e5535680a2c

weshaggard:
LGTM

PR #1024, opened 2016-09-08
Fix package downgrade in tool-runtime project.json

mellinoe:
@ericstj , @dagood 

I still get hundreds of downgrade warnings, but that seems to be due to bad dependencies in the MSBuild packages (we have the same versions of all those), which were pre-existing.


ericstj:
Confirmed through local restore.

LGTM

dagood:
LGTM, thanks! Agreed that existing "tame" downgrade warnings aren't a problem, it's only `warning NU1012: Dependency conflict. NETStandard.Library 1.6.1-beta-24506-02 expected System.Net.Http >= 4.3.0-beta-24506-02 but received 4.1.0` that showed up through stderr rather than in `init-tools.log`.

mellinoe:
Thanks, merging now.

PR #1026, opened 2016-09-08
Add new CLI and build tools boot-strapping script.

jhendrixMSFT:
New script to be used when boot-strapping build tools for a repo.  A bash
version will be implemented a bit later.

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -1,0 +1,104 @@
+param
+(
+    [string]$RepositoryRoot,
+    [string]$ToolsLocalPath,
+    [switch]$Force = $false
+)
+
+$rootToolVersions = Join-Path $RepositoryRoot ".toolversions"
+$bootstrapComplete = Join-Path $ToolsLocalPath "bootstrap.complete"
+
+# if the force switch is specified delete the semaphore file if it exists
+if ($Force -and (Test-Path $bootstrapComplete))
+{
+    del $bootstrapComplete
+}
+
+# if the semaphore file exists and is identical to the specified version then exit
+if ((Test-Path $bootstrapComplete) -and !(Compare-Object (Get-Content $rootToolVersions) (Get-Content $bootstrapComplete)))
+{
+    exit 0
+}
+
+$initCliScript = "dotnet-install.ps1"
+$initCliLocalPath = Join-Path $ToolsLocalPath $initCliScript
+
+# blow away the tools directory so we can start from a known state
+if (Test-Path $ToolsLocalPath)
+{
+    rd -recurse $ToolsLocalPath
+}
+mkdir $ToolsLocalPath | Out-Null
+
+# download CLI boot-strapper script
+Invoke-WebRequest "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1" -OutFile $initCliLocalPath
+
+# load the version of the CLI
+$rootCliVersion = Join-Path $RepositoryRoot ".cliversion"
+$dotNetCliVersion = Get-Content $rootCliVersion
+
+# now execute the script
+$cliLocalPath = Join-Path $ToolsLocalPath "dotnetcli"
+Invoke-Expression "$initCliLocalPath -Version $dotNetCliVersion -InstallDir $cliLocalPath"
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@weshaggard @ellismg @chcosta 
Here's the new boot-strapping script.  Once this is merged I will post a new PR that consumes this script.

bootstrap/bootstrap.ps1
```diff
@@ -0,0 +1,107 @@
...
+}
+$packagesPath = Join-Path $RepositoryRoot "packages"
+$dotNetExe = Join-Path $cliLocalPath "dotnet.exe"
+$process = Start-Process -Wait -NoNewWindow -FilePath $dotNetExe -ArgumentList "restore $projectJson --packages $packagesPath --source $packageSource" -PassThru
+if ($process.ExitCode -ne 0)
+{
+    exit $process.ExitCode
+}
+
+# download nuget.exe
```
ellismg: What uses `nuget.exe` instead of `dotnet restore` ?

jhendrixMSFT: Restoration of packages uses nuget.exe, see build.proj and dir.props (NugetRestoreCommand).

ellismg: Ahh, I see.  That's different than CoreFX.  Thanks!

weshaggard: Hmmm.. Nothing should be using nuget.exe any more, can you point me to where it is used.

bootstrap/bootstrap.ps1
```diff
@@ -0,0 +1,107 @@
...
+$bootstrapComplete = Join-Path $ToolsLocalPath "bootstrap.complete"
+
+# if the force switch is specified delete the semaphore file if it exists
+if ($Force -and (Test-Path $bootstrapComplete))
+{
+    del $bootstrapComplete
+}
+
+# if the semaphore file exists and is identical to the specified version then exit
+if ((Test-Path $bootstrapComplete) -and !(Compare-Object (Get-Content $rootToolVersions) (Get-Content $bootstrapComplete)))
```
weshaggard: This doesn't take into account just changes to the cli version. I think that is probably OK for now but we need to make sure we can update cli version and the right thing will happen.

bootstrap/bootstrap.ps1
jhendrixMSFT: Yes see https://github.com/dotnet/buildtools/blob/master/build.proj#L46

weshaggard: Is this only the build tools that uses nuget.exe? If so should this instead be a tool listed in the tool versions list instead of downloaded for everyone?

jhendrixMSFT: I will remove this and look at either migrating to "dotnet restore" or have the build tools repo pull this down as a package.

bootstrap/bootstrap.ps1
jhendrixMSFT: Correct, I'll add a task to the board to track this.

bootstrap/bootstrap.ps1
```diff
@@ -0,0 +1,107 @@
...
+
+# blow away the tools directory so we can start from a known state
+if (Test-Path $ToolsLocalPath)
+{
+    rd -recurse $ToolsLocalPath
+}
+mkdir $ToolsLocalPath | Out-Null
+
+# download CLI boot-strapper script
+Invoke-WebRequest "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1" -OutFile $initCliLocalPath
```
weshaggard: @ellismg I suppose we are going to want a way to override this or cache it so we can still build without an internet connection.

ellismg: Yes.

What I've been doing locally is hacking up init-tools.sh to just copy everything from a fixed directory (which I prepped by using dotnet-install.ps1 offline).  If there was some way to sent an environment variable that would do that, it would be great.

jhendrixMSFT: Presumably we don't need this from the start?  If this is the case I will add a task to the board to track it.

ellismg: Correct.  I'm happy to do the work, BTW.

PR #1029, opened 2016-09-09
Add nuget.org feed to boot-strapper.

jhendrixMSFT:
Some repos might require restoring some packages from nuget.org so I have
added this feed to the list of sources.  In the future we can consider
making the list of feeds configurable at the repo level.

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -26,7 +26,7 @@ $initCliLocalPath = Join-Path $ToolsLocalPath $initCliScript
 # blow away the tools directory so we can start from a known state
 if (Test-Path $ToolsLocalPath)
 {
-    rd -recurse $ToolsLocalPath
+    rd -recurse -force $ToolsLocalPath
 }
 mkdir $ToolsLocalPath | Out-Null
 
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@weshaggard @chcosta Required to unblock build tools migration to new boot-strapper.

bootstrap/bootstrap.ps1
```diff
@@ -66,14 +66,16 @@ $pjContent = $pjContent + "}, `"frameworks`": { `"netcoreapp1.0`": { } } }"
 $pjContent | Out-File $projectJson
 
 # now restore the packages
-$packageSource = "https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json"
+$buildToolsSource = "https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json"
+$nugetOrgSource = "https://www.nuget.org/api/v2"
```
weshaggard: lets use the v3 feed. https://api.nuget.org/v3/index.json

weshaggard:
other then using the v3 feed LGTM

bootstrap/bootstrap.ps1
jhendrixMSFT: Agreed will fix.

PR #1030, opened 2016-09-09
Utilize the new boot-strapping script.

jhendrixMSFT:
This change updates the repo to use the new boot-strapping script,
replacing init-tools.cmd along with its supporting files.  New text files
have been added to contain the desired CLI and tool packages.
The boot-strapping happens in run.ps1, if the run tool cannot be found
then the boot-strapper script is downloaded and executed.  It also
downloads nuget.exe which is a unique requirement for build tools.

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -85,15 +85,23 @@ if ($process.ExitCode -ne 0)
 foreach ($tool in $tools)
 {
     $name, $version = $tool.split("=")
-    $destination = Join-Path $ToolsLocalPath $name
-    if (Test-Path (Join-Path $packagesPath "$name\$version\tools\netcoreapp1.0"))
+
+    # at present we have the following conventions when staging package content:
+    #   1.  if a package contains a "tools" directory then recursively copy its contents
+    #       to a directory named the package ID that's under $ToolsLocalPath.
+    #   2.  if a package contains a "libs" directory then recursively copy its contents
+    #       under the $ToolsLocalPath directory.
+    #   3.  if a package contains a file "lib\init-tools.cmd" execute it.
+
+    if (Test-Path (Join-Path $packagesPath "$name\$version\tools"))
     {
+        $destination = Join-Path $ToolsLocalPath $name
         mkdir $destination | Out-Null
-        copy (Join-Path $packagesPath "$name\$version\tools\netcoreapp1.0\*") $destination
+        copy (Join-Path $packagesPath "$name\$version\tools\*") $destination -recurse
     }
     elseif (Test-Path (Join-Path $packagesPath "$name\$version\lib"))
     {
-        copy (Join-Path $packagesPath "$name\$version\lib\*") $ToolsLocalPath
+        copy (Join-Path $packagesPath "$name\$version\lib\*") $ToolsLocalPath -recurse
     }
 
     if (Test-Path (Join-Path $packagesPath "$name\$version\lib\init-tools.cmd"))


jhendrixMSFT:
CC @weshaggard @chcosta @ellismg

run.ps1
```diff
@@ -0,0 +1,34 @@
...
+if ((Test-Path $runExe) -eq 0)
+{
+    if ((Test-Path $toolsLocalPath) -eq 0)
+    {
+        mkdir $toolsLocalPath | Out-Null
+    }
+
+    # download boot-strapper script
+    $bootStrapperPath = Join-Path $toolsLocalPath "bootstrap.ps1"
+    Invoke-WebRequest "https://raw.githubusercontent.com/dotnet/buildtools/master/bootstrap/bootstrap.ps1" -OutFile $bootStrapperPath
```
chcosta: Will we ever want to specify a different branch?

run.ps1
```diff
@@ -0,0 +1,34 @@
...
+    Invoke-Expression "$bootStrapperPath (Get-Location) $toolsLocalPath" | Out-File (Join-Path (Get-Location) "bootstrap.log")
+    if ($LastExitCode -ne 0)
+    {
+        Write-Output "Boot-strapping failed with exit code $LastExitCode"
+        exit $LastExitCode
+    }
+
+    # copy nuget.exe into the packages directory (this is unique to the build tools repo)
+    $packagesPath = Join-Path (Get-Location) "packages"
+    Get-ChildItem (Join-Path $packagesPath "NuGet.CommandLine") -filter "nuget.exe" -recurse | foreach { copy $_.FullName $packagesPath }
```
chcosta: Potentially have a check for a postbootstrap.ps1 file which, if exists, executes and then move this section there.

run.ps1
jhendrixMSFT: It's conceivable that we won't always want to use master, however I would think that in this case run.ps1 would be updated directly (e.g. in foo/rel/1.1.0 they update to consume bootstrap.ps1 from buildtools/rel/1.1.0).

run.ps1
jhendrixMSFT: I thought about this, the only reason I didn't implement it is I thought this requirement is unique to build tools and I don't know if we want to set such a president.  I vote to leave this as-is for now and implement that when the time comes however if folks already know this is going to be a requirement then I can do it now.

chcosta: The only reason I'd vote for it is because then when this file gets ported to a repo, it can just be ported as-is and the risk of the NuGet stuff being left in there goes down.

jhendrixMSFT: That's a good point.  @weshaggard what do you think about a convention of if file $RepoRoot\bootstrap\postbootstrap.ps1 exists we have bootstrap.ps1 execute it.

run.ps1
```diff
@@ -0,0 +1,34 @@
+# set the base tools directory
+$toolsLocalPath = Join-Path (Get-Location) "Tools"
+$runExe = Join-Path $toolsLocalPath "Microsoft.DotNet.BuildTools.Run\run.exe"
+
+# if the tool doesn't exist then download and run the boot-strapper script
+if ((Test-Path $runExe) -eq 0)
```
weshaggard: This is adding another layer of semaphore checking which will block updating tools if .toolversion updates. I think we need to make some more changes to this script to always call bootstrap.ps1, and presumable cache it in some way for subsequent runs.

run.ps1
```diff
@@ -0,0 +1,34 @@
+# set the base tools directory
+$toolsLocalPath = Join-Path (Get-Location) "Tools"
```
weshaggard: I thought we decided to call this ".tools" directory.

run.ps1
weshaggard: probably also need to figure out a way to detect if the bootstrap.ps1 script needs to be downloaded again as well.

run.ps1
weshaggard: We will need to provide a way to actually override not only just the branch but the entire path to make it all local, like @ellismg is going to look into.

run.ps1
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    # copy nuget.exe into the packages directory (this is unique to the build tools repo)
+    $packagesPath = Join-Path (Get-Location) "packages"
+    Get-ChildItem (Join-Path $packagesPath "NuGet.CommandLine") -filter "nuget.exe" -recurse | foreach { copy $_.FullName $packagesPath }
+}
+
+# execute the tool using the dotnet.exe host
+$dotNetExe = Join-Path $toolsLocalPath "dotnetcli\shared\Microsoft.NETCore.App\version\dotnet.exe"
+$process = Start-Process -NoNewWindow -FilePath $dotNetExe -ArgumentList "$runExe $args" -PassThru
```
weshaggard: Any reason you call Start-Process directly? As opposed to just calling the command and returning $lastexitcode? 

Also I think the `-PassThru` parameter is why you were blocking on child processes, as that will cause you to get a set of process objects not just the one.

run.ps1
```diff
@@ -0,0 +1,34 @@
...
+        exit $LastExitCode
+    }
+
+    # copy nuget.exe into the packages directory (this is unique to the build tools repo)
+    $packagesPath = Join-Path (Get-Location) "packages"
+    Get-ChildItem (Join-Path $packagesPath "NuGet.CommandLine") -filter "nuget.exe" -recurse | foreach { copy $_.FullName $packagesPath }
+}
+
+# execute the tool using the dotnet.exe host
+$dotNetExe = Join-Path $toolsLocalPath "dotnetcli\shared\Microsoft.NETCore.App\version\dotnet.exe"
```
weshaggard: Don't you need to set the dotnet path in the targets? Or does BuildTools not use this?

run.ps1
weshaggard: I do like the value of having a copy-paste script that is the same for all the repo's, and so I would be in favor of something like this, but before we even go there I would like to see if there is some pattern we can use in the bootstrap.ps1 or .toolsversion which would allow for this instead of needing to write this custom code?

run.ps1
jhendrixMSFT: I wasn't sure if we had decided on that, will fix.

run.ps1
jhendrixMSFT: How can I call it directly with the path being in a string?
If I used the "-Wait" option then it would block on all child processes.  Doing it this way appears to return a process object for just the process that was created (the docs are unclear on the exact behavior so this is just empirical evidence).

run.ps1
jhendrixMSFT: In dir.props the property DotnetToolCommand is set to the "root" dotnet.exe executable (tools\dotnetcli\dotnet.exe).

run.ps1
jhendrixMSFT: You are right about the .toolversion updates.  Detecting to re-download bootstrap.ps1 is tricky though as I don't think we want to always hit the network (i.e. if I have done a sync/build I should be able to then go offline and rebuild without any network traffic).  I can change this to detect bootstrap.ps1 and always execute it as that will address the case of .toolversions changing.

run.ps1
jhendrixMSFT: Agreed that avoiding the need for custom code is best.  If we're ok with adding complexity to .toolversions we could include source and destination directories.  Something like this.

```
Microsoft.DotNet.BuildTools,1.0.26-prerelease-00802-01,lib,.tools
```

So this would translate to "copy the contents in lib to the .tools directory".  It alleviates the need for extra scripts but does add complexity/duplication and internal knowledge of packages to this file.

PR #1031, opened 2016-09-09
Improve package indexing and validation

ericstj:
This updates indexing to handle cases where we bump up the package
version during pre-release.

/cc @weshaggard @chcosta


weshaggard:
LGTM

PR #1032, opened 2016-09-09
Use INamedTypeReference instead in order to also copy the nested class type forwards from the contract assembly

joperezr:
cc: @weshaggard 

When a contract contains nested class type-forwards they are getting filtered out by GenFacades because their type is `NonGenericNestedType` which doesn't implement `INamespaceTypeReference`. With this change, I'm using `INamedTypeReference` which they do  implement and is actually also implemented by `INamespaceTypeReference` so they will be includded.

### DIFF
--- a/src/GenFacades/Program.cs
+++ b/src/GenFacades/Program.cs
@@ -311,8 +311,8 @@ namespace GenFacades
 
         private static IEnumerable<string> EnumerateDocIdsToForward(IAssembly contractAssembly)
         {
-            // Make note that all type forwards (including nested) implement INamespaceAliasForType, so do
-            // not be tempted to filter using it, instead, we look at the aliased type to them.
+            // Use INamedTypeReference instead of INamespaceTypeReference in order to also include nested
+            // class type-forwards.
             var typeForwardsToForward = contractAssembly.ExportedTypes.Select(alias => alias.AliasedType)
                                                                       .OfType<INamedTypeReference>();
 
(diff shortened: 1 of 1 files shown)


joperezr:
addresses: #11549

src/GenFacades/Program.cs
```diff
@@ -314,7 +314,7 @@ private static IEnumerable<string> EnumerateDocIdsToForward(IAssembly contractAs
             // Make note that all type forwards (including nested) implement INamespaceAliasForType, so do
```
weshaggard: @nguerrera do this comment ring any bells? We hit a bug where the nested type-forwards aren't being copied but the comment seems to indicate that they should.

@joperezr we should also update this comment if we stick with this change.

weshaggard:
Looks reasonable to me.

src/GenFacades/Program.cs
joperezr: What I thought the comment was indicating is the explanation of why to use `alias.AliasedType` instead of just `alias`. Although, I also get that perhaps we were intentionally filtering out nested forwarders for a reason?

nguerrera: IIRC, we used to filter to namespace types only and add nested forwarders with special logic to replicate the csc behavior, which included private nested types of the destination. Then we decided to change it so that we only forward to types that are actually defined in the reference assembly. See https://github.com/dotnet/buildtools/pull/275. 

I did not study things, but I suspect we broke something in that move. This comment does seem out of date. I'd remove it.

PR #1034, opened 2016-09-12
Make the staging of boot-strapped package contents more generic.

jhendrixMSFT:
For packages with content that follows defined conventions copy the
contents into specific locations.  Please see the comment added in this
commit that describes the algorithm in more detail.


jhendrixMSFT:
@weshaggard This is the change we were discussing earlier.

jhendrixMSFT:
@chcosta can you please take a look.

chcosta:
LGTM

PR #1036, opened 2016-09-12
Switch to response files for GenFacades

weshaggard:
cc @ericstj @mellinoe @joperezr

### DIFF
--- a/src/common/CommandLine.cs
+++ b/src/common/CommandLine.cs
@@ -590,11 +590,26 @@ class CommandLine
 #endif
         public CommandLineParser(string commandLine)
         {
+            _args = new List<string>();
             ParseWords(commandLine);
         }
         public CommandLineParser(string[] args)
         {
-            _args = new List<string>(args);
+            _args = new List<string>();
+            _seenExeArg = true;
+
+            for (int i = 0; i < args.Length; i++)
+            {
+                if (args[i][0] == '@')
+                {
+                    // Parse response files
+                    ParseWords(args[i]);
+                }
+                else
+                {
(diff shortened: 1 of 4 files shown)


src/common/CommandLine.cs
```diff
@@ -590,11 +590,18 @@ public bool LastQualifierWins
...
         }
         public CommandLineParser(string[] args)
         {
-            _args = new List<string>(args);
+            _args = new List<string>();
+            _seenExeArg = true;
+
+            for (int i = 0; i < args.Length; i++)
+            {
+                ParseWords(args[i]);
```
ericstj: Isn't this double-parsring the args? Suppose you had nested quotes with something like `"\"Hello\" \"World\""`.  Previously you'd get a single _arg entry as `"Hello" "World"`.  Now you'll get two entries: `Hello` and `World`.

mellinoe:
What's the value? I usually prefer to have the arguments right there in the command, I find it easier to debug, easier to modify, etc.

src/common/CommandLine.cs
weshaggard: Fair maybe I will just parse the args that start with "@".

weshaggard:
>What's the value? I usually prefer to have the arguments right there in the command, I find it easier to debug, easier to modify, etc.

I'm only fixing this because I've hit the command line limit while working on netstandard and there isn't any other way to handle it besides switching to a response file. Also for me personally I've always found it easier to repro issues with a response file because it is a lot easier to run most of things commands by just passing in that one argument.

ericstj:
LGTM

PR #1038, opened 2016-09-13
Update repo to use the new build tools boot-straper script.

jhendrixMSFT:
Use the new script, bootstrap.ps1, to acquire and initialize packages
required to build the repo.
Fixed a bug in bootstrap.ps1 thas was deleting the script after it was
downloaded, and verify that the package that was restored matches the
expected version.


jhendrixMSFT:
@weshaggard @chcosta @ellismg Updated PR to consume the new boot-strapper.

jhendrixMSFT:
@joshfree Here are the changes to boot-strapping build tools I was telling you about.

.gitignore
```diff
@@ -4,6 +4,7 @@ syntax: glob
 
 # Tool Runtime Dir
 /[Tt]ools/
```
ellismg: Should this be removed?

run.ps1
```diff
@@ -0,0 +1,30 @@
+# set the base tools directory
+$toolsLocalPath = Join-Path (Get-Location) ".tools"
```
ellismg: `Get-Location` will give the current working directory, but I think you want $PSScriptRoot so that it gets the folder that "run.ps1" is in, even if I call run from a different folder?

jhendrixMSFT: Yes you're right I'll fix this.

.gitignore
jhendrixMSFT: I thought it might be nice to wait a bit while this transition is in place however I don't feel too strongly about it and can remove it now if we think that's the better choice.

config.json
```diff
@@ -170,12 +170,12 @@
     "msbuild": {
       "osSpecific":{
         "windows": {
-          "defaultParameters": "/nologo /verbosity:minimal /clp:Summary /maxcpucount /nodeReuse:false /l:BinClashLogger,Tools\\net45\\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log",
-          "path": "Tools/msbuild.cmd"
+          "defaultParameters": "/nologo /verbosity:minimal /clp:Summary /maxcpucount /nodeReuse:false /l:BinClashLogger,.tools\\net45\\Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log",
+          "path": ".tools/msbuild.cmd"
```
mellinoe: I'd prefer to not rename this folder; I don't really like "dot-name" folders, especially not for something like this.

jhendrixMSFT: I believe the convention we're moving to is to prefix directories that aren't a part of the repo with a dot to help avoid collisions.

weshaggard: Yes we are trying to use the "." prefix to make these hidden in more contexts. Is there a particular reason you don't like this naming @mellinoe?

run.ps1
```diff
@@ -0,0 +1,30 @@
...
+# now execute it
+Invoke-Expression "$bootStrapperPath (Get-Location) $toolsLocalPath" | Out-File (Join-Path (Get-Location) "bootstrap.log")
+if ($LastExitCode -ne 0)
+{
+    Write-Output "Boot-strapping failed with exit code $LastExitCode"
+    exit $LastExitCode
+}
+
+# execute the tool using the dotnet.exe host
+$dotNetExe = Join-Path $toolsLocalPath "dotnetcli\shared\Microsoft.NETCore.App\version\dotnet.exe"
```
weshaggard: Sorry bringing forward some other comments from the PR you closed.

I still don't understand why you don't need to set this path in the targets? We will want the targets to point at this same dotnet.exe otherwise all the tools will need to agree on the CLI version again which is what we were trying to avoid/workaround here by directly calling dotnet.exe in the version folder.

run.ps1
```diff
@@ -0,0 +1,30 @@
...
+$runExe = Join-Path $toolsLocalPath "Microsoft.DotNet.BuildTools.Run\netcoreapp1.0\run.exe"
+$process = Start-Process -NoNewWindow -FilePath $dotNetExe -ArgumentList "$runExe $args" -PassThru
```
weshaggard: I would still like to avoid using Start-Process and instead just do `& $dotNetExe $runExe $args` then we don't need to worry about what the different semantics of `-Wait` vs `WaitForExit`. I don't think there is any reason we need to get a process object here. We should be able to just call the command and then call `exit $LastExitCode`.

run.ps1
```diff
@@ -0,0 +1,30 @@
...
+    {
+        mkdir $toolsLocalPath | Out-Null
+    }
+
+    # download boot-strapper script
+    Invoke-WebRequest "https://raw.githubusercontent.com/dotnet/buildtools/master/bootstrap/bootstrap.ps1" -OutFile $bootStrapperPath
+}
+
+# now execute it
+Invoke-Expression "$bootStrapperPath (Get-Location) $toolsLocalPath" | Out-File (Join-Path (Get-Location) "bootstrap.log")
```
weshaggard: I think I would prefer using the call operator here `$` as well as with the Start-Process call below. See http://social.technet.microsoft.com/wiki/contents/articles/7703.powershell-running-executables.aspx. The call operator is a little more limited and thus also a little safer.

run.ps1
```diff
@@ -0,0 +1,30 @@
...
+if ($LastExitCode -ne 0)
```
weshaggard: If I understand how lastexitcode works I believe it only has the exitcode of the last win32 program that was executed so it will not have the return code of the script you are calling directly. Although it is worth testing as I'm not 100% sure.

run.ps1
jhendrixMSFT: The problem here is that we cannot use the dotnet.exe that's under the version directory to run "dotnet restore" as it can't find the restore command (it's implemented in some other dll).  We can either hack up the CLI/FX installation to resolve this or we bifurcate the implementation to have "dotnet restore" use the "top-level" dotnet.exe and use the version-specific dotnet.exe purely as a host (I don't like the idea of hacking up the installation).  Perhaps there are other options here too.

run.ps1
jhendrixMSFT: Great point I'll make this change.

run.ps1
jhendrixMSFT: Thanks for the link I hadn't seen this.  I'll make the switch to the call operator here and in bootstrap.ps1.

run.ps1
jhendrixMSFT: The documents are pretty vague on this but from my testing $LastExitCode does contain the return value from the executed script.

run.ps1
weshaggard: Unfortunately we will likely need to separate these and define a DotnetHost which points to the version for all the tools except for the dotnet sdk tools (i.e. restore).

run.ps1
weshaggard: I would just test it by directly calling exit 1 from your script to ensure it hits your error path. If that works then we can just do this.

run.ps1
jhendrixMSFT: It looks like this has been moved to toolruntime.targets.  Obviously if we just change it to the new path it will prevent consumption of build tools until they migrate to the new boot-strapper.  Presumably this change should be implemented in such a way to support current and new behavior, would you agree?

jhendrixMSFT: There's also some logic in Build.Common.props that calculates this path.

config.json
jhendrixMSFT: I discussed this with @weshaggard offline and we're going to revert this part of the change for now (it will be addressed in a separate PR).

jhendrixMSFT:
@weshaggard Here are the changes we discussed yesterday, can you please take a look.

weshaggard:
LGTM.

Can you please also file the issues we talked about and link them to this? In particular the one which will require a change in dir.props which points to dotnet.exe in different places depending on if it is a build tools tool (i.e. genfacade/apicompat) or part of the dotnet sdk (i.e. restore).

jhendrixMSFT:
Also see https://github.com/dotnet/buildtools/issues/1046.

weshaggard:
See also https://github.com/dotnet/buildtools/issues/1045 which I think is more important then #1046.

PR #1040, opened 2016-09-14
Handle packages with placeholders only in ref

ericstj:
Some packages have placeholders in ref but no reference assembly
(eg: System.Private assemblies)

Make sure we index these to get the correct assembly version.

I noticed this issue when reviewing the dev-api merge.

/cc @weshaggard @joperezr


joperezr:
LGTM.

ericstj:
Merging.  This will require an index update when consuming the new buildtools.

weshaggard:
👍

PR #1042, opened 2016-09-15
Produces API implementation

chcosta:
https://github.com/chcosta/buildtools/blob/producesapi/Documentation/BuildToolsRepoApiProducesImplementation.md

These changes generate a produces.json file based on the contents of a drop folder.  Produces.json contents are pushed to the dotnet/versions repo if an authorization token is provided.  

This change does not include displaying the latest produces results ('run produces' scenario).

@weshaggard @dagood


src/Microsoft.DotNet.VersionTools/project.json
```diff
@@ -1,5 +1,8 @@
-{
+﻿{
   "dependencies": {
+    "Microsoft.Build.Framework": "0.1.0-preview-00022",
+    "Microsoft.Build.Tasks.Core": "0.1.0-preview-00022",
+    "Microsoft.Build.Utilities.Core": "0.1.0-preview-00022",
```
dagood: I'll have to come back to do a full review, but this dependency is the first thing that sticks out. CLI (/cc @eerhardt) takes a dependency on this package and doesn't want the msbuild dependencies or `dnxcore50` import. I worked with this by making the tasks in the buildtools Tasks project, which references VersionTools. Is there a reason you can't do the same here?

chcosta: The only reason I have it here, is because I wanted to access the "UpdateGitHubFileAsync" method from VersionsRepoUpdater.cs.  I can look at moving this to the Tasks project though.

eerhardt: +1 on this.  IMO - it is good to split the actual implementation out from MSBuild, so we can write tools outside of MSBuild that do these operations.

For example, if we wanted to write a tool that consumed one of these "produces.json" files and made a chart/graph/website out of it.  If we had an object model for the "produces.json" structure split away from MSBuild, this would be easy.  And then have MSBuild Tasks that use this object model as necessary.

Chrisboh:

PR #1047, opened 2016-09-15
Prevent run.ps1 from parsing arguments.

jhendrixMSFT:
The arguments passed to run.ps1 are intended for the run.exe tool, so
prefix the args to run.ps1 with a double dash so that powershell doesn't
attempt to parse them (without this it will "eat" any double dashes passed
to run.ps1).

### DIFF
--- a/run.cmd
+++ b/run.cmd
@@ -12,5 +12,5 @@ if not defined VisualStudioVersion (
 )
 
 :Run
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Commit https://github.com/dotnet/buildtools/commit/cae331bdbd73c17668f05d1e66e7a59cb6374f12 broke the official build, this change will fix it.

PR #1051, opened 2016-09-19
Make package harvesting replace refs

ericstj:
Previously we'd always prefer live assets over harvested assets from the
previous shipped package.  That put the burden on the developer to
know when to include references.

This change lets us keep references in the package and only use them
when they are significant.  This should help reduce the concept count
required by folks working in corefx.

/cc @weshaggard @joperezr @stephentoub


ericstj:
Required changes to pick this up in corefx:
https://github.com/ericstj/corefx/commit/cb9568136d2bd44db2c5804c232f362b7a50afe9
https://github.com/ericstj/corefx/commit/295f226adff67940b2549d02d10421163c23a5fc
https://github.com/ericstj/corefx/commit/2459090fb1aa3833c1189449eba42e8d239e398c

PR #1053, opened 2016-09-21
Remove Configuration and set ConfigurationGroup on ProjectReferences

alexperovich:
This change fixes a bug with building inside VS where the Configuration property would flow through ProjectReferences and break the build because a project is build with mismatched configurations. e.g. mismatched NuGetTargetMoniker and TargetGroup.

@weshaggard @tarekgh @JeremyKuhne @ericstj @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -105,6 +105,10 @@
   <Target Name="ConvertCommonMetadataToAdditionalProperties" BeforeTargets="AssignProjectConfiguration">
     <!-- list each append as a seperate item to force re-evaluation of AdditionalProperties metadata -->
     <ItemGroup>
+      <ProjectReference>
+        <UndefineProperties>%(ProjectReference.UndefineProperties);Configuration</UndefineProperties>
+      </ProjectReference>
+
       <!-- Set some basic configuration properties and pass them along to ProjectReference's -->
       <ProjectReference>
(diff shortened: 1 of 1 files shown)


tarekgh:
LGTM. thanks @alexperovich

JeremyKuhne:
:shipit:

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -105,6 +105,10 @@
   <Target Name="ConvertCommonMetadataToAdditionalProperties" BeforeTargets="AssignProjectConfiguration">
     <!-- list each append as a seperate item to force re-evaluation of AdditionalProperties metadata -->
     <ItemGroup>
+      <ProjectReference>
```
weshaggard: Do we actually need to undefine Configuration? I thought all we were missing is flowing the ConfigurationGroup.

alexperovich: This is required to make VS builds succeed. Without this System.Runtime.Tests fails with E:\corefx\Tools\Packaging.targets(589,5): error : Packages that are constrained by runtime should not have runtime dependencies.  They will be ignored by nuget

ericstj: This is happening because configuration gets the OSGroup and/our TargetGroup embedded in it when the SLN uses the meta-configuration and that will leak into project references.  We need to make sure the only thing project references see is the normal configuration string (Debug/Release) and any other relevant properties will be passed through their specific Property names (OSGroup/TargetGroup) rather than leaking in through Configuration.

As such the fix is to either redefine Configuration to the simple string or not pass Configuration and pass the simple string as ConfigurationGroup.

weshaggard:
LGTM - Thanks for updating this @alexperovich

PR #1054, opened 2016-09-21
Remove EnsureOOBFramework

ericstj:
This "magic" functionality is getting in the way, especially now that we're harvesting and trying to remove depprojs.

CoreFx changes required: https://github.com/ericstj/corefx/commit/87d339420f925d408132901c9446a4b51ef839e1

/cc @weshaggard


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PromoteDependencies.cs
```diff
@@ -38,42 +34,45 @@ public override bool Execute()
...
             {
-                foreach (var implementationFx in implementationFxs)
+                foreach (var refFx in refFxs)
                 {
-                    NuGetFramework fx = NuGetFramework.Parse(implementationFx);
-                    if (referenceSets.ContainsKey(fx))
+                    // find best lib (if any)
+                    var nearestLibFx = FrameworkUtilities.GetNearest(refFx, libFxs);
+
+                    if (nearestLibFx != null)
```
weshaggard: Shouldn't you also check if they are different?

ericstj: Will do.

PR #1057, opened 2016-09-22
Remove DotnetCLIVersion.txt, no longer used

dagood:
Bootstrap init flow uses ".cliversion"; remove the old file.

DotnetCLIVersion.txt still being in the source tree is misleading and makes it harder to notice the new file.

@weshaggard @chcosta


chcosta:
It looks like there are a couple of places where DotNetCliVersion.txt will need to be converted to use .cliversion for repo's that consume buildtools (ie in CloudTestTasks and PerfTesting.targets), but none of those instances should affect building the BuildTools repo itself.

dagood:
Ah, interesting... will that make converting product repos to use bootstrap.ps1 a "synchronized"/breaking buildtools change, then?

chcosta:
Yes.

@weshaggard , in general is the intention to make these breaking changes in BuildTools and force repo's to conform when updating their BuildTools version, or to support both init-tools and bootstrapping behaviors in BuildTools so that repo's can consume other changes while transitioning to use the bootstrapping?

weshaggard:
>@weshaggard , in general is the intention to make these breaking changes in BuildTools and force repo's to conform when updating their BuildTools version, or to support both init-tools and bootstrapping behaviors in BuildTools so that repo's can consume other changes while transitioning to use the bootstrapping?

We should try to make them opt-in and phase them out over time. In other words I would suggest we update targets to support both for a while. Can you point me at what is actually reading the DotNetCliVersion.txt file directly? It was never intended to be consumed by anything but the bootstrapping scripts.

Chrisboh:
@dagood can you merge or close this?

PR #1061, opened 2016-09-25
GenAPI: Ignore obsolete-as-error ctors when generating base call

akoeplinger:
The first constructor in the base class is sometimes one that is marked as [Obsolete("message", true)] meaning any usage of that ctor results in a compilation error. Example: System.ComponentModel.AsyncCompletedEventArgs

We now skip those constructors when generating the base call so classes that inherit from those types compile out of the box.

### DIFF
--- a/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
+++ b/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
@@ -262,7 +262,7 @@ namespace Microsoft.Cci.Writers.CSharp
             if (baseType == null)
                 return;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
```diff
@@ -264,13 +264,13 @@ private void WriteBaseConstructorCall(ITypeDefinition type)
 
             var ctors = baseType.Methods.Where(m => m.IsConstructor && _filter.Include(m));
```
weshaggard: I think I would just filter out the ctors here instead of doing it in both places below. We already do other filtering here.

akoeplinger: Fixed.

PR #1068, opened 2016-09-27
Don't write generate project.json if contents match

chcosta:
This change doesn't write a new generated project.json for test projects if the contents match the exisiting project.json.  Doing this allows us to make use of the "IsRestoreRequired" task in CoreFx and improves perf for restoring generated test packages when no changes have been made.  On my machine, restore time goes from ~11 minutes to ~10 seconds.

/cc @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -391,9 +391,13 @@ namespace Microsoft.DotNet.Build.Tasks
 
         private static void WriteProject(JObject projectRoot, string projectJsonPath)
         {
-            string projectJson = JsonConvert.SerializeObject(projectRoot, Formatting.Indented);
-            Directory.CreateDirectory(Path.GetDirectoryName(projectJsonPath));
-            File.WriteAllText(projectJsonPath, projectJson + Environment.NewLine); 
+            string projectJson = JsonConvert.SerializeObject(projectRoot, Formatting.Indented) + Environment.NewLine;
+
+            if (!File.Exists(projectJsonPath) || !projectJson.Equals(File.ReadAllText(projectJsonPath)))
+            {
+                Directory.CreateDirectory(Path.GetDirectoryName(projectJsonPath));
+                File.WriteAllText(projectJsonPath, projectJson);
+            }
         }
 
         /* JProperties are encapsulated with "['" and "']" to assist with matching Paths which


ericstj:
You've already read the project once, why are you reading it again? Just keep track of whether or not you've made changes.

ericstj:
I see, you're just squashing it without comparing in UpdateDependenciesProperty so you don't know if it's changed.  I guess I don't care too much but it seems like this could be better.

chcosta:
The second read is reading the OutputProjectJson, not the input ProjectJson.  So the comparison is, what we've generated, vs what was generated in a previous run, if anything.  I don't think we can make these assumptions about what occurred in a previous generation in UpdateDependenciesProperty or anywhere else.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -391,9 +391,13 @@ private JObject FilterRuntimes(JObject projectRoot, HashSet<string> excludedRIDs
 
         private static void WriteProject(JObject projectRoot, string projectJsonPath)
         {
-            string projectJson = JsonConvert.SerializeObject(projectRoot, Formatting.Indented);
-            Directory.CreateDirectory(Path.GetDirectoryName(projectJsonPath));
-            File.WriteAllText(projectJsonPath, projectJson + Environment.NewLine); 
+            string projectJson = JsonConvert.SerializeObject(projectRoot, Formatting.Indented) + Environment.NewLine;
+
+            if (!File.Exists(projectJsonPath) || !projectJson.Equals(File.ReadAllText(projectJsonPath)))
```
weshaggard: Is doing a straight string compare of the files the right way to determine this?  I guess if you expect to be the only one writing this file it could work but it does seem like it might be a large string compare and if we are doing this hundreds of times that might cause some memory issues.

weshaggard: Can we not do the typical msbuild approach and only update it if the inputs actually have changed and have a newer timestamp then the project.json file you are checking?

chcosta: Determining the correct set of inputs might be a bit of a chore since the inputs depend on

- the original project.json
- an optional versions file
- an optional package drop with varying packages

I think it's simpler, and more reliable, to make this determination in code.

I wouldn't imagine project.json files are that large and likely to lead to memory issues.  I opted for this vs an alternative line by line comparison because it seemed it would be more performant (in my mind).  Obviously, if memory consumption is an issue then perf goes out the window.  I'll have to rely on your experience here if you believe that string comparing project.json files is likely to cause problems.

weshaggard: I can live with that for now.

PR #1069, opened 2016-09-27
Repo API buildtools implementation proposal doc

dagood:
An initial proposal for implementing the Repo API Compose commands in buildtools, at a high level.

We don't need solutions for some of the finer points in the doc yet: the first implementation goal is to make the auto-PR system produce updates on par with what we have now, but using the Repo API.

@chcosta @weshaggard @ellismg @jaredpar 
/cc @markwilkie @dleeapho

### DIFF
--- a/Documentation/RepoCompose.md
+++ b/Documentation/RepoCompose.md
@@ -56,7 +56,7 @@ The data in the output is further grouped by operating system.  Builds of the sa
 
 In the case the `consumes` output doesn't want to take advantage of OS specific dependencies it can specify `"os-all"` as a catch all. 
 
-In addition to artifacts the consume feed can also optionally list any machine prerequitsites needed to build or test the repo:
+In addition to artifacts the consume feed can also optionally list any machine prerequisites needed to build or test the repo:
 
 ``` json
 "prereq": { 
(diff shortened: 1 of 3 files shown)


chcosta:
These changes align with my understanding of the process.

chcosta:
I think the "dependencies.props" file / requirement is throwing me off, but I'll spend some more time investigating to understand its purpose vs just having all of the info in consumes.

dagood:
For `consumes`/`change`/`verify`, `dependencies.props` is for non project.json stuff, specifically properties in msbuild, e.g. the version of package `x` in category `y` should be the value of property `z` in file `w`. We could put that in in consumes.json and have some leeway for extensibility in the Repo API spec, but I don't think that's useful info for anyone using the Repo API. I see a dividing line here between the internal repo config vs. what the repo depends on outside its source tree.

Separately, for auto-upgrade we also need a place to customize which build-infos are used. Adding URI pointers into `consumes` kind of makes sense because the build-infos are an external dependency, but only in the auto-update "build"/process. I'm not sure where to draw the line between build and meta-build.

Documentation/RepoCompose.md
```diff
@@ -118,14 +118,14 @@ This command operates by providing json whose format is a subset of the output o
 
 A tool responsible for composing repos would ideally:
 
-1. Execute `run.cmd consumes` and capture the output.
-2. Alter the NuGet package versions in the json to have the correct build identifier: beta5, RTM, etc ..
-3. Execute `run.cmd change` and pass in the altered output.
+1. Execute `consumes` and capture the output.
```
weshaggard: I'm not sure capturing standard output is the best way to transfer this information. Can we just instead have some sort of parameter to pass the output?

Documentation/buildtools-repo-compose.md
```diff
@@ -0,0 +1,83 @@
...
+ * `change` takes a modified consumes.json and overwrites the checked-in consumes.json file. Then, it changes the project contents so they match the new source of truth.
+
+ * `verify` reads consumes.json and ensures its contents are self-consistent and match the repository contents. This allows devs to check if their project.json or other modifications are valid.
+   * This is not a documented Repo API Compose command, but [RepoUtil][RepoUtil] offers it and the existing buildtools target `/t:VerifyDependencies` is an analogue.
+
+The `produces` command scans artifacts of a completed build for files that the repository produced and intends to publish. More detailed documentation will be available in BuildToolsRepoApiProducesImplementation.md [~~link~~](BuildToolsRepoApiProducesImplementation.md).
+
+The Repo API Compose command document [describes the workflow to update a version](RepoCompose.md#change), but to go into more detail, the corresponding flow in the Buildtools implementation is:
+
+ 1. `run consumes > temp.json`
```
weshaggard: Seems to me like this round-trip is just useless and instead we should just have a convention of a checked in consumes.json file that people can directly get and/or edit it.

Documentation/RepoCompose.md
dagood: What kind of parameter are you imagining, what would it do? I think stdout is good because it's standard and doesn't require temporary file creation if we apply tools over the repo API to coordinate between repos.

Documentation/buildtools-repo-compose.md
weshaggard: They could run the command to generate the file then perhaps. I'm always very uneasy about blindly capturing standard output from complex tool chain, as it seems like extra text could easily just crept in and break people capturing it.

Chrisboh:

PR #1070, opened 2016-09-28
Update package indexing and Validation

ericstj:
This improves the indexing to also index implementation versions,
potentially from runtime-specific packages.  Previously we only indexed
references which missed cases where impl assemblies compiled against
other impl assemblies and the ref didn't change.  Such is the case for
partial facades.

I've also added some validation to catch cases where we add new shim
packages.

This change requires CoreFx change: https://github.com/ericstj/corefx/commit/6965cf0b5a0ce9073f79b8bbeaa8a488f41b4202

/cc @weshaggard


ericstj:
merging this because corefx changes for versioning are blocked on it.

PR #1075, opened 2016-09-29
Allow overriding the AssemblyOriginatorKeyFile property

akoeplinger:
This is useful if a project wants to specify a key other than the MSFT.snk, ECMA.snk or Open.snk.


weshaggard:
LGTM

PR #1077, opened 2016-09-30
Add source index supporting targets

alexperovich:
This change adds support for building a source index by using all the implementation assemblies as references passed to the compiler. This enables the source indexer to connect references between source projects correctly. Specifically this allows references to things that are type forwarded to reference the actual source where the implementation lives.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
@@ -108,8 +108,20 @@
       <_ReferenceCopyLocalPathsFromPackage Remove="@(_ReferenceCopyLocalPathsFileNamesToRemove)" />
 
       <!-- add the filtered resolved package output -->
-      <Reference Include="@(_ReferenceFromPackage)" />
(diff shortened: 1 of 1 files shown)


alexperovich:
@ericstj

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -108,8 +108,20 @@
       <_ReferenceCopyLocalPathsFromPackage Remove="@(_ReferenceCopyLocalPathsFileNamesToRemove)" />
 
       <!-- add the filtered resolved package output -->
-      <Reference Include="@(_ReferenceFromPackage)" />
+      <Reference Condition="'$(SourceIndex)' != 'true'" Include="@(_ReferenceFromPackage)" />
```
ericstj: I'd prefer it if you did this as a separate target.  Looks to me like it it is all just modifications that could be done in a Target that runs AfterTargets=ResolveNuGetPackages.  For that matter this doesn't even need to go into buildtools.

alexperovich: I'll move it into corefx then.

PR #1081, opened 2016-10-01
Move test publishing from post-build to pre-test-run

weshaggard:
Currently test publishing happens during PrepareForRun which is a
post step of build. However we need TestTFM as a parameter in test
publishing which we don't know until test runtime because a given
test build can have a set of TestTFM's that apply to it. To address
this we are moving test publishing as a pre-step for test running.
To do this we are overriding the Test target and adding a set of
DependsOnTargets to allow for ordering them correctly.

cc @joperezr @tarekgh 

fyi @stephentoub this should fix the issue we saw with CodeDom tests.


tarekgh:
LGTM. thanks Wes for cleaning up this.

PR #1082, opened 2016-10-01
Redirect git rev-parse HEAD error output

jkotas:
Redirect git rev-parse HEAD error output to avoid bogus messages about missing git tool when building inside VS and not having git on the path. Any errors are ignored anyway.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -205,7 +205,7 @@
   </Target>
 
   <Target Name="GetLatestCommitHash" Condition="'$(LatestCommit)'==''">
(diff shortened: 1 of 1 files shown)


jkotas:
@joperezr PTLA

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -205,7 +205,7 @@
   </Target>
 
   <Target Name="GetLatestCommitHash" Condition="'$(LatestCommit)'==''">
-    <Exec Command="git rev-parse HEAD" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+    <Exec Command="git rev-parse HEAD 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
```
weshaggard: Is that still going to get the output in the success case?

weshaggard:
I've been wanting to try and figure out a way to have this work better incrementally. Currently it always runs on every build which causes us to do it more often then actually needed.

PR #1084, opened 2016-10-02
Remove Xunit.PlatformID enum

### LINKED ISSUE #999: Xunit.PlatformID should be plural
This is very nit-picky, but it's a pet peeve of mine: It'd be nice if the .NET Core code followed the framework design guidelines, even in test infrastructure code, so it feels familiar to .NET developers.

[`Xunit.PlatformID`](https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/PlatformID.cs) is a flags enum, and according the [design guidelines](https://msdn.microsoft.com/en-us/library/ms229040(v=vs.110).aspx), it should be plural:

> ✓ DO use a singular type name for an enumeration unless its values are bit fields.
> ✓ DO use a plural type name for an enumeration with bit fields as values, also called flags enum.

Thus, I'd like to rename `PlatformID` to `PlatformIDs` (plural) to follow the .NET conventions we're used to.

Renaming has an additional benefit: the plural name would no longer conflict with [`System.PlatformID`](https://github.com/dotnet/corefx/blob/c4538cc3e6d3c28318e50ce81d5c2137c91b0d44/src/System.Runtime.Extensions/ref/System.Runtime.Extensions.cs#L685-L694), which is being added back in the dev/api branch in CoreFX. Renaming the test enum means we'll no longer need to  [fully qualify it with the namespace](https://github.com/dotnet/corefx/blob/722c529bb3b4096564c33303568e8e99a350100b/src/System.Globalization/tests/CultureInfo/CultureInfoCurrentCulture.cs#L103) anymore, e.g. instead of `[PlatformSpecific(Xunit.PlatformID.AnyUnix)]` it could simply be `[PlatformSpecific(PlatformIDs.AnyUnix)]`.

I know this will cause some churn, and for a seemingly minor benefit, but I'd be happy to do the work to make it happen, opening a PR in this repo to do the rename, along with PRs in the repos that need to be updated as a result.

cc: @weshaggard @stephentoub @Priya91

justinvp:
Now that https://github.com/dotnet/corefx/pull/12284 has been merged, we can delete `PlatformID` (replaced by `TestPlatforms`).

I did a search on GitHub and while there are some repos besides corefx using `Microsoft.xunit.netcore.extensions` (e.g. wcf, corefxlab), none that I could find are using `PlatformID`. (There may be private repos that need to be updated at some point that I can't see).

The following open corefx PRs are still using `PlatformID` in some way. I've left comments in each about the rename:
 - https://github.com/dotnet/corefx/pull/11489
 - https://github.com/dotnet/corefx/pull/12238
 - https://github.com/dotnet/corefx/pull/12240
 - https://github.com/dotnet/corefx/pull/12249

Fixes #999


justinvp:
cc: @weshaggard

weshaggard:
@justinvp there are other repo's that are using this right now as well. At least the WCF repo is so I would prefer we give this remove a little more time.

weshaggard:
Sorry I didn't read you full description. Thanks for looking into other usages and if WCF isn't using this then I'm good with going ahead and removing it.

PR #1096, opened 2016-10-06
Fix debugging from VS

ericstj:
VS will only run the build target, so we need to make sure the test is
deployed as part of build.

/cc @weshaggard


ericstj:
Closing in favor of https://github.com/dotnet/buildtools/pull/1095

PR #1102, opened 2016-10-06
Move version.txt to obj directory

chcosta:
Creating a version.txt file in the PackagesBasePath folder was not optimal because PackageBasePath contains the $(OSGroup) which gets filtered for a build.  The result is that you could have multiple PackageBasePaths, which meant generating redundant version.txt files.  We really just want to generate one version.txt file and use that for all of the produced packages.  In some repo's, like buildtools, this doesn't matter as much as in other repo's, like CoreFx, where there are multiple OSGroups for a build.  

This change moves the version.txt file to get generated in the obj directory, and creates a traversal build dependency so that the target gets called once before a traversal build.

@joperezr @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -217,7 +217,6 @@
 
   <PropertyGroup>
     <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
-    <BeforeTraversalBuildDependsOn>CreateVersionInfoFile;$(BeforeTraversalBuildDependsOn)</BeforeTraversalBuildDependsOn>
   </PropertyGroup>
     
   <PropertyGroup>
@@ -227,6 +226,7 @@
   
   <Target Name="CreateVersionInfoFile"
           DependsOnTargets="GetLatestCommitHash"
+          BeforeTargets="BuildAllProjects"
           Inputs="$(LatestCommit)"
           Outputs="$(SyncInfoFile)">
     <MakeDir Condition="!Exists('$(SyncInfoDirectory)')"
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -217,12 +217,12 @@
 
   <PropertyGroup>
     <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
-    <TraversalBuildDependsOn>CreateVersionInfoFile;$(TraversalBuildDependsOn)</TraversalBuildDependsOn>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -217,13 +217,20 @@
 
   <PropertyGroup>
     <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
+    <TraversalBuildDependsOn>CreateVersionInfoFile;$(TraversalBuildDependsOn)</TraversalBuildDependsOn>
+  </PropertyGroup>
+    
+  <PropertyGroup>
+    <SyncInfoDirectory>$(ObjDir)</SyncInfoDirectory>
```
weshaggard: Add conditions to allow these to be overridden if needed.

weshaggard:
other then the one override comment looks good to me.

chcosta:
@weshaggard , updated as per your comment.  I also changed the "TraversalBuildDependsOn" property to "BeforeTraversalBuildDependsOn" because, by convention, versioning.targets gets imported before dir.traversal.targets and I want to be able to call this target before "BuildAllProjects", https://github.com/dotnet/corefx/blob/master/dir.traversal.targets#L295.  The alternative would be to require versioning.targets to be imported after dir.traversal.targets and I didn't like that.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -217,13 +217,20 @@
 
   <PropertyGroup>
     <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
+    <BeforeTraversalBuildDependsOn>CreateVersionInfoFile;$(BeforeTraversalBuildDependsOn)</BeforeTraversalBuildDependsOn>
```
chcosta: Your assessment is correct.  I like your suggestion.  I'll prepare an update

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -218,12 +218,19 @@
   <PropertyGroup>
     <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
   </PropertyGroup>
+    
+  <PropertyGroup>
+    <SyncInfoDirectory Condition="'$(SyncInfoDirectory)' == ''">$(ObjDir)</SyncInfoDirectory>
+    <SyncInfoFile Condition="'$(SyncInfoFile)' == ''">$(SyncInfoDirectory)version.txt</SyncInfoFile>
```
chcosta: I considered this, but for official builds, where we are shipping packages from, this will not be an issue.

joperezr: agreed, but it will be an issue for local builds, although I guess people might not care if the binaries that they build might have an inaccurate version

PR #1105, opened 2016-10-07
Port UseNewestPackagesAvailable option to buildtools-release/1.0.0

chcosta:
In CoreFx release branch builds, we only build / publish packages which we are servicing.  We are unable to build tests with project references converted to packages, because the process does not know what package versions to use (traditionally it can determine this from a package drop which contains all of the packages we built).  The fix for determining package versions, was to check in a versions file which the process can derive information from - https://github.com/dotnet/corefx/commit/252011a470b5586338694219642c411b0d0c3a7d.  However, when a version file is used in conjunction with a package drop (the built packages we are servicing), there can be a conflict between package version numbers.  To address that conflict, we have an option to provide the hint "UseNewestPackagesAvailable".  The change which added this option was not ported to buildtools release/1.0.0.  This change cherry-picks the [commit](https://github.com/dotnet/buildtools/commit/5060b4ab9b0f5dc1a2264401a440617646e15bb6) to bring it in to release/1.0.0 so that we can produce servicing builds using the build pipeline infrastructure going forward.

@weshaggard 

/cc @gkhanna79 @markwilkie

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.net45/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/project.json
@@ -1,10 +1,11 @@
 {
   "dependencies": {
+    "Microsoft.NETCore.Platforms": "1.0.1-rc2-24027",
     "Newtonsoft.Json": "7.0.1",
+    "NuGet.Packaging": "3.5.0-beta2-1456",
     "NuGet.Versioning": "3.5.0-beta2-1456",
     "System.Collections.Immutable": "1.1.37",
     "System.Reflection.Metadata": "1.1.0",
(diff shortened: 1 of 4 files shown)


weshaggard:
While this will help eliminate the conflicts. It does still require us to update the version numbers in package list overtime, as with each servicing release we end up deleting them from the package list we are building and at that point the versions file will need to be updated otherwise we will start preferring the older version with the next servicing event.

chcosta:
Yes,  to address that issue I'm  making a change to support uri based version files.  That's a separate pr though and being tracked with an internal issue.  I hope to have that pr available on Monday.

PR #1107, opened 2016-10-07
Parameterize bootstrapper for CLI install path and shared framework version.

crummel:
This change adds two optional parameters to the bootstrapper script:
1. Selecting the directory to install the CLI (by default, this remains in <toolsDir>\dotnetcli).  This is needed for e.g., CLI, which wants its bootstrapping CLI in the .stage0 folder.
2. Selecting the shared framework version to use.  This is to set up buildtools' dotnetcli\shared\Microsoft.NETCore.App\version symlink correctly.  We could attempt to automatically detect this but the bootstrapper never sees the install package, it's all handled by dotnet-install - the options I see are either modifying dotnet-install to check the version and pass it back, or requiring a Microsoft.NETCore.App.version similar to the other version files in each consuming repo.

The fixes the buildtools portion of #918.

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -1,7 +1,10 @@
 param
 (
-    [string]$RepositoryRoot,
-    [string]$ToolsLocalPath,
+    [Parameter(Mandatory=$false)][string]$RepositoryRoot = $PSScriptRoot,
+    [Parameter(Mandatory=$false)][string]$ToolsLocalPath = (Join-Path $RepositoryRoot "Tools"),
+    [Parameter(Mandatory=$false)][string]$CliLocalPath = (Join-Path $ToolsLocalPath "dotnetcli"),
+    [Parameter(Mandatory=$false)][string]$SharedFrameworkSymlinkPath = (Join-Path $ToolsLocalPath "dotnetcli\shared\Microsoft.NETCore.App\version"),
+    [Parameter(Mandatory=$false)][string]$SharedFrameworkVersion = "<auto>",
     [switch]$Force = $false
 )
 
@@ -42,8 +45,8 @@ $rootCliVersion = Join-Path $RepositoryRoot ".cliversion"
 $dotNetCliVersion = Get-Content $rootCliVersion
 
 # now execute the script
-$cliLocalPath = Join-Path $ToolsLocalPath "dotnetcli"
-Invoke-Expression "$initCliLocalPath -Version $dotNetCliVersion -InstallDir $cliLocalPath"
+Write-Host "$initCliLocalPath -Version $dotNetCliVersion -InstallDir $CliLocalPath"
+Invoke-Expression "$initCliLocalPath -Version $dotNetCliVersion -InstallDir $CliLocalPath"
 if ($LastExitCode -ne 0)
 {
     Write-Output "The .NET CLI installation failed with exit code $LastExitCode"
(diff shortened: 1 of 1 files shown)


crummel:
cc @weshaggard @chcosta 
Another option I just thought of - the consuming repo can be responsible for setting up the version symlink.  I will take a look at that and see if it's cleaner.

bootstrap/bootstrap.ps1
```diff
@@ -52,9 +56,12 @@ if ($LastExitCode -ne 0)
 
 # create a junction to the shared FX version directory. this is
 # so we have a stable path to dotnet.exe regardless of version.
-$junctionTarget = Join-Path $toolsLocalPath "dotnetcli\shared\Microsoft.NETCore.App\1.0.0"
```
weshaggard: I would prefer this stay in here instead of duplicating this in each repo. I also don't like the hardcoding of the 1.0.0 shared framework here or in the individual repos. I like your other suggestion of finding the latest version that installed with the cli and linking to that.

bootstrap/bootstrap.ps1
```diff
@@ -2,6 +2,7 @@ param
 (
     [string]$RepositoryRoot,
     [string]$ToolsLocalPath,
+    [string]$CliLocalPath,
```
weshaggard: This needs to be made optional. We should probably make all these optional and given them reasonable defaults.

weshaggard: While we are here it might be worth adding an optional parameter to allow overriding the "version" link directory path as well.

bootstrap/bootstrap.ps1
crummel: Updated, I've change it so the new behavior is that dotnet-install will optionally write out a Microsoft.NETCore.App.version file to a supplied path (in build_tools in CLI so it's already in .gitignore).  bootstrap will then grab the version out of that file and use that version for the symlink.

bootstrap/bootstrap.ps1
crummel: Made all parameters explicitly optional and added CliSymlinkPath parameter.

bootstrap/bootstrap.ps1
```diff
@@ -42,8 +43,11 @@ $rootCliVersion = Join-Path $RepositoryRoot ".cliversion"
 $dotNetCliVersion = Get-Content $rootCliVersion
 
 # now execute the script
-$cliLocalPath = Join-Path $ToolsLocalPath "dotnetcli"
-Invoke-Expression "$initCliLocalPath -Version $dotNetCliVersion -InstallDir $cliLocalPath"
+if ($CliLocalPath -eq "")
+{
+    $CliLocalPath = Join-Path $ToolsLocalPath "dotnetcli"
```
crummel: Good point, fixed.

bootstrap/bootstrap.ps1
```diff
@@ -1,12 +1,15 @@
 param
 (
-    [string]$RepositoryRoot,
-    [string]$ToolsLocalPath,
+    [Parameter(Mandatory=$false)][string]$RepositoryRoot = $PSScriptRoot,
+    [Parameter(Mandatory=$false)][string]$ToolsLocalPath = (Join-Path $RepositoryRoot "Tools"),
+    [Parameter(Mandatory=$false)][string]$CliLocalPath = (Join-Path $ToolsLocalPath "dotnetcli"),
+    [Parameter(Mandatory=$false)][string]$CliSymlinkPath = (Join-Path $ToolsLocalPath "dotnetcli\shared\Microsoft.NETCore.App\version"),
```
weshaggard: I would call this SharedFrameworkSymlinkPath

crummel: Changed as suggested.

bootstrap/bootstrap.ps1
crummel: Obsoleted by new auto-selection of shared runtime version.

crummel:
Rebased and squashed, will merge after the build passes.

PR #1112, opened 2016-10-10
Enable reading versions file from URI

chcosta:
This change is necessary for our release branches where we want to be able to point to the latest versions of the release branch packages.

This change is in conjunction with https://github.com/dotnet/buildtools/pull/1105 to address servicing requirements.

@weshaggard @dagood

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -91,9 +91,9 @@ namespace Microsoft.DotNet.Build.Tasks
             // Retrieve package information from a versions file
             if (VersionsFiles != null)
             {
-                foreach (var versionsFile in VersionsFiles)
+                foreach (var versionsUri in VersionsFiles.Select(v => new Uri(v)))
                 {
-                    AddPackageItemsToDictionary(ref packageInformation, GatherPackageInformationFromVersionsFile(versionsFile, comparer));
+                    AddPackageItemsToDictionary(ref packageInformation, GatherPackageInformationFromVersionsFile(versionsUri, comparer));
                 }
             }
 
@@ -195,28 +195,40 @@ namespace Microsoft.DotNet.Build.Tasks
             }
         }
 
-        private static async System.Threading.Tasks.Task<string> GetFileAsync(string uri)
+        private static async System.Threading.Tasks.Task<Stream> GetStream(Uri uri)
         {
-            using (var stream = (File.Exists(uri)) ? File.OpenRead(uri) : await new HttpClient().GetStreamAsync(uri))
+            if(uri.Scheme == "file")
+            {
+                return new FileStream(uri.LocalPath, FileMode.Open, FileAccess.Read, FileShare.Read);
+            }
+            else
+            {
+                return await new HttpClient().GetStreamAsync(uri);
+            }
+        }
+
+        private static async System.Threading.Tasks.Task<string> GetFileAsync(Uri uri)
+        {
+            using (var stream = GetStream(uri).Result)
             using (StreamReader reader = new StreamReader(stream))
             {
                 return await reader.ReadToEndAsync();
             }
         }
 
-        // A versions file is of the form https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/corefx/master/Latest_Packages.txt
-        private Dictionary<string, PackageItem> GatherPackageInformationFromVersionsFile(string versionsFile, VersionComparer comparer = null)
+        // A versions file is of the form https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/corefx/release/1.0.0/LKG_Packages.txt
+        private Dictionary<string, PackageItem> GatherPackageInformationFromVersionsFile(Uri uri, VersionComparer comparer = null)
         {
             Dictionary<string, PackageItem> packageItems = new Dictionary<string, PackageItem>();
 
             string contents = string.Empty;
             try
             {
-                contents = GetFileAsync(versionsFile).Result;
+                contents = GetFileAsync(uri).Result;
             }
             catch (AggregateException ae)
             {
-                Log.LogError("Error: Unable to open '{0}', either the file does not exist locally or there is a network issue accessing that URI.", versionsFile);
+                Log.LogError("Error: Unable to open '{0}', either the file does not exist locally or there is a network issue accessing that URI.", uri.ToString());
                 throw ae;
             }
             var lines = contents.Split(new string[] { "\r\n", "\n" }, StringSplitOptions.RemoveEmptyEntries);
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -198,15 +195,34 @@ private void AddPackageItemsToDictionary(ref Dictionary<string, PackageItem> pac
...
+            string contents = string.Empty;
+            using (var stream = (File.Exists(uri)) ? File.OpenRead(uri) : await new HttpClient().GetStreamAsync(uri))
+            using (StreamReader reader = new StreamReader(stream))
+            {
+                contents = await reader.ReadToEndAsync();
+            }
+            return contents;
+        }
+
+        // A versions file is of the form https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/corefx/master/Latest_Packages.txt
```
dagood: Maybe this is a question for a corresponding corefx PR, but should we pin this to a commit SHA for repeatable builds?

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -198,15 +195,34 @@ private void AddPackageItemsToDictionary(ref Dictionary<string, PackageItem> pac
...
-        // A versions file is of the form https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt
+        private static async System.Threading.Tasks.Task<string> GetFileAsync(string uri)
+        {
+            string contents = string.Empty;
+            using (var stream = (File.Exists(uri)) ? File.OpenRead(uri) : await new HttpClient().GetStreamAsync(uri))
+            using (StreamReader reader = new StreamReader(stream))
+            {
+                contents = await reader.ReadToEndAsync();
+            }
+            return contents;
```
dagood: Could you use `return await reader.ReadToEndAsync();` inside the using?

chcosta: Ack, fixed

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
dagood: Your hope is correct, I think I just don't understand how these package versions are used.

dagood: I see... and one thing I think I didn't have right was whether you were going to link [`Latest_Packages.txt`](https://github.com/dotnet/versions/blob/64d27bdd68ba6f9f1d62c7eaf11395090eed8f04/build-info/dotnet/corefx/release/1.0.0/Latest_Packages.txt) or [`LKG_Packages.txt`](https://github.com/dotnet/versions/blob/64d27bdd68ba6f9f1d62c7eaf11395090eed8f04/build-info/dotnet/corefx/release/1.0.0/LKG_Packages.txt). I was thinking Latest due to the comment and PR description, but LKG makes sense and makes me not concerned about repeatability. (Now I remember LKG is what you added in https://github.com/dotnet/corefx/pull/12382.)

dagood: It's manual at the moment, something to do on release. The idea is that we have [build pointers](https://github.com/dotnet/versions#build-pointers) for build qualities, but we haven't automated it yet. (LKG would probably be for a green test build.)

A while back there was some discussion about if it makes more sense for `Latest_Packages` to be a list where packages aren't "removed" when the next build publishes, and have a separate list for only the last build's packages. That hasn't happened--but is that kind of Latest what would be useful here?

chcosta: "That" kind of Latest is exactly what we need here.  Unless the URI we point to contains the old packages plus whatever is newly published, we will be in a state where there is a gap between LKG and what we published in the interim builds between then and the current servicing release.  How much effort is it to produce that kind of Latest file?  We really need that to make this a workable solution.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -198,15 +195,32 @@ private void AddPackageItemsToDictionary(ref Dictionary<string, PackageItem> pac
             }
         }
 
-        // A versions file is of the form https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt
+        private static async System.Threading.Tasks.Task<string> GetFileAsync(string uri)
+        {
+            using (var stream = (File.Exists(uri)) ? File.OpenRead(uri) : await new HttpClient().GetStreamAsync(uri))
```
weshaggard: I'm not sure File.Exists is the best way to do this check. See https://referencesource.microsoft.com/#System.Xml/System/Xml/XmlDownloadManager.cs,27 as an possible example to convert to a Uri and if it isn't a file then use HttpClient.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
dagood: It should just be modification to https://github.com/dotnet/buildtools/blob/d468095793/src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs, although it involves downloading the existing file which isn't currently done there. I'll write an issue to nail down what the end result should be.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -198,15 +195,44 @@ private void AddPackageItemsToDictionary(ref Dictionary<string, PackageItem> pac
...
+            }
+            else
+            {
+                return await new HttpClient().GetStreamAsync(uri);
+            }
+        }
+
+        private static async System.Threading.Tasks.Task<string> GetFileAsync(Uri uri)
+        {
+            using (var stream = GetStream(uri).Result)
```
weshaggard: `var stream = await GetStream(uri)`. In general you should avoid calling .Result when you are in an async method.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -198,15 +195,44 @@ private void AddPackageItemsToDictionary(ref Dictionary<string, PackageItem> pac
...
+            string contents = string.Empty;
+            try
+            {
+                contents = GetFileAsync(uri).Result;
+            }
+            catch (AggregateException ae)
             {
-                Log.LogError("Specified versions file ({0}) does not exist.", versionsFile);
+                Log.LogError("Error: Unable to open '{0}', either the file does not exist locally or there is a network issue accessing that URI.", uri.ToString());
+                throw ae;
```
weshaggard: If you are going to re-throw you should just do `throw;` so the original exception, including its original callstack, is rethrown.

Issue #1001, opened 2016-08-31
When creating auto-update PR, log GitHub error responses


dagood:
GitHub sends [nice summaries of errors](https://developer.github.com/v3/#client-errors) for `422` responses. They should be logged to make debugging auto-PRs easier.

dagood:
We should also give detailed results for `503`s, because they're typically API rate limits.

PR #1113, opened 2016-10-10
Trace GitHub rate limit info, print error contents

dagood:
When there is an error response code in `GitHubClient`, include the response contents in the exception so the error can be diagnosed. Also trace rate limit information with every request for informational purposes.

An example rate limit error with these changes:
> Response code does not indicate success: 403 (Forbidden) with content: {"message":"API rate limit exceeded for 198.51.125.2. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}

Fixes https://github.com/dotnet/buildtools/issues/1001.

@chcosta @weshaggard 
/cc @gkhanna79


chcosta:
LGTM

Issue #1111, opened 2016-10-10
Recent CLI doesn't work with init_tools.sh


janvorli:
While trying to help someone from the community to build coreclr on Linux with kernel version >= 4.6, which requires using recent CLI, I have discovered that init_tools.sh don't work when I modify the DotnetCLIVersion.txt to point to a recent version, e.g. 1.0.0-preview3-003786. There are two issues here. First one is a folder casing issue:
`/home/janvorli/git/coreclr/init-tools.sh: line 137: /home/janvorli/git/coreclr/packages/Microsoft.DotNet.BuildTools/1.0.26-prerelease-00809-01/lib/init-tools.sh: No such file or directory`
But `ls /home/janvorli/git/coreclr/packages` shows that the Microsoft.DotNet.BuildTools is named microsoft.dotnet.buildtools now.
So I've modified init_tools.sh by modifying this path to lowercase. But after that, the second issue occurs:

```
/home/janvorli/git/coreclr/Tools/Microsoft.Common.CurrentVersion.targets(1097,5): warning MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.0" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend. [/home/janvorli/git/coreclr/src/mscorlib/System.Private.CoreLib.csproj]
/home/janvorli/git/coreclr/Tools/Microsoft.Common.CurrentVersion.targets(1097,5): warning MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.0" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend. [/home/janvorli/git/coreclr/src/mscorlib/ref/mscorlib.csproj]
[...]

PR #1114, opened 2016-10-10
In init-tools.sh, lenient casing for restored packages

dagood:
Uses `find` to allow compatibility with CLI versions that have the NuGet change that lowercases package ids in the packages directory.

This should fix the buildtools side of https://github.com/dotnet/buildtools/issues/1111.

Tested on OSX and openSUSE 13.2 with https://github.com/dotnet/coreclr/compare/master...dagood:init-tools-insensitive.

@janvorli @chcosta @mellinoe @ericstj
/cc @weshaggard @gkhanna79

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -98,8 +98,10 @@ if [ "$?" != "0" ]; then
     echo "ERROR: An error ocurred when running: '$__DOTNET_CMD restore \"${__PORTABLETARGETS_PROJECTJSON}\"'. Please check above for more details."
     exit 1
 fi
-cp -R "${__PACKAGES_DIR}/Microsoft.Portable.Targets/${__PORTABLETARGETS_VERSION}/contentFiles/any/any/." "$__TOOLRUNTIME_DIR/."
-cp -R "${__PACKAGES_DIR}/MicroBuild.Core/${__MICROBUILD_VERSION}/build/." "$__TOOLRUNTIME_DIR/."
+find $__PACKAGES_DIR -type d \( \
+    -iwholename "${__PACKAGES_DIR}/Microsoft.Portable.Targets/${__PORTABLETARGETS_VERSION}/contentFiles/any/any" -or \
+    -iwholename "${__PACKAGES_DIR}/MicroBuild.Core/${__MICROBUILD_VERSION}/build" \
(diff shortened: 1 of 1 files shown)


chcosta:
I don't understand how the package name got lower-cased.

dagood:
It's a breaking NuGet change that we're exposed to through `dotnet restore`: https://github.com/NuGet/Home/issues/2522

dagood:
Ping--this is needed to restore buildtools with a CLI version past the NuGet breaking change.

chcosta:
I'm not super familiar with 'find'.  Does this change the behavior?  ie, is this going to hide not finding those packages if they don't exist?

dagood:
Good point, if something happens that lets the restore pass but the packages aren't there, this won't report an error. Also, if multiple copies of the same package are in the packages dir with different directory casing it would also copy each one's contents. Those are the only differences I'm aware of.

Are we currently relying on `cp -R` on the literal paths to return an error? The only place I've seen that error help so far is when dealing with the NuGet capitalization breaking change, which this PR makes non-breaking as far as init-tools is concerned.

chcosta:
I'd expect us to not see an issue with the multiple packages because it would be differently cased packages, but with the same version; and that seems unlikely.  Also, I guess it's not too concerning if the directories aren't present because i'd expect that to manifest as an issue during the previous "package restore" step.

LGTM

dagood:
Agreed, unlikely, but with some potential cases. Filed https://github.com/dotnet/buildtools/issues/1162 to track maybe fixing that, although I wonder if we'll be able to replace that copy with something better at some point.

Merging, thanks.

dagood:
This is breaking some xplat builds when tools are initialized:

[chcosta/dotnetcore:fedora23_prereqs corefx build](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=417523)
> /root/corefx/packages/Microsoft.DotNet.BuildTools/1.0.27-prerelease-01002-04/lib/init-tools.sh: line 101: find: command not found

[chcosta/dotnetcore:alpine_prereqs corefx build](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=417523)
> 2016-11-07T14:03:47.4123920Z find: unrecognized: -iwholename

It looks like `find` isn't as stable across distros as I'd hoped, or even available--we need an alternative to find the name of a folder in a case insensitive way. I'm not familiar enough with the ecosystem to know how to check this without just trying various things.

@ellismg @chcosta @mellinoe Any ideas?

In the meantime, @ericstj plans to make the init-tools copy and msbuild import conditional.

Issue #1117, opened 2016-10-11
Make a build-info file available that maintains previous builds' packages


dagood:
Right now `VersionsRepoUpdater` overwrites `Latest_Packages.txt` with the list of packages created during the current build. Instead, the `Latest_Packages.txt` should contain the latest versions of all packages ever published from that build, unless packages are manually removed. For example, a servicing release would only update the serviced packages in `Latest_Packages.txt` to their new versions.

There should also be a `Last_Build_Packages.txt` that's simply overwritten with the build's package list.

This matches the name "`Latest`" better, and would improve scenarios like https://github.com/dotnet/buildtools/pull/1112#discussion_r82700580 and https://github.com/dotnet/buildtools/pull/858.

We should consider using the [github tree API](https://developer.github.com/v3/git/trees/#create-a-tree) so these files can be updated in the same commit rather than two (or three, when `Latest.txt` is updated).

@chcosta @weshaggard

weshaggard:
Sounds good. I do like the idea of having the separate "LastBuild" list as well.

dagood:
This is done, and happens in corefx master.

chcosta:
How can we get this working for CoreFx release/1.0.0 and release/1.1.0?

dagood:
Getting this in 1.0.0 is what I mentioned here: https://github.com/dotnet/buildtools/pull/1118#issuecomment-253318066, but I don't think the needed changes are considered above the bar at the moment. /cc @markwilkie 

For 1.1.0, we'd need https://github.com/dotnet/buildtools/pull/1118 in the buildtools version that corefx/1.1.0 uses, but 1.1.0 uses VersionTools so in the CoreFX repo I believe it would just be a buildtools update.

chcosta:
Based on conversations with a few people, I think that the bar has shifted slightly. Given that there is no urgency around releases at the moment, and that we could opportunistically get these types of changes in.

PR #1118, opened 2016-10-11
Keep old packages in file when updating Latest_Packages

dagood:
Adds a new build-info file `Last_Build_Packages.txt` that stores only the new packages. When creating `Last_Build_Packages` is enabled, the update to `Latest_Packages` carries the existing `Latest_Packages` entries over into the new version.

Also includes changes to publish all build-info files in a single commit.

Fixes (once uptaken) https://github.com/dotnet/buildtools/issues/1117.

Example commit, using some odd packages I had lying around: https://github.com/dagood/versions/commit/e34d05821c5c713ce8fa1a4f0d95bc7428fc893a.

@chcosta @weshaggard @karajas
/cc @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
@@ -5,14 +5,19 @@
 using Microsoft.DotNet.VersionTools.Automation.GitHubApi;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Net.Http;
 using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.VersionTools.Automation
 {
     public class VersionsRepoUpdater
     {
+        private const int MaxTries = 10;
+        private const int RetryMillisecondsDelay = 5000;
+
         private GitHubAuth _gitHubAuth;
         private GitHubProject _project;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
```diff
@@ -68,25 +71,71 @@ public VersionsRepoUpdater(GitHubAuth gitHubAuth, GitHubProject project)
...
-                string latestFilePath = $"{versionsRepoPath}/Latest.txt";
+                if (updateLatestPackageList)
+                {
+                    string path = $"{versionsRepoPath}/Latest_Packages.txt";
+
+                    var allPackages = new Dictionary<string, string>(packageDictionary);
+
+                    if (updateLastBuildPackageList)
+                    {
+                        Dictionary<string, string> existingPackages = await GetPackagesAsync(client, path);
```
weshaggard: Why wouldn't we just use the same list we are going to push for the LastBuildPackageList above instead of going to get the file from github again?

weshaggard: Sorry I miss-read this as I now realize you are getting the Lastest_Packages.txt file instead. I would consider changing the name of path to latestPackagesPath.

src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
```diff
@@ -96,18 +145,33 @@ public VersionsRepoUpdater(GitHubAuth gitHubAuth, GitHubProject project)
                     message += $" for {prereleaseVersion}";
                 }
 
-                await UpdateGitHubFileAsync(latestFilePath, prereleaseVersion, message);
+                GitTree tree = await client.PostTreeAsync(_project, masterSha, objects.ToArray());
+                GitCommit commit = await client.PostCommitAsync(_project, message, tree.Sha, new[] { masterSha });
+
+                // Only fast-forward. Don't overwrite other changes: throw exception instead.
+                await client.PatchReferenceAsync(_project, masterRef, commit.Sha, force: false);
```
weshaggard: Isn't this going to potentially have race conditions? I suspect the files being written too are different for each official build so perhaps there shouldn't be any conflicts. At any rate we might need to consider some retry logic here if we start seeing failures.

src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
dagood: Will do, I can see reusing the same `path` name in different scopes being confusing.

chcosta:
@dagood, how would this work for a branch like corefx/release/1.0.0.  Do we need to seed https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/1.0.0/Latest_Packages.txt with the packages from https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/1.0.0/LKG_Packages.txt in order to get the full, current package set?

dagood:
Yeah, it needs to be initially seeded/patched.

It looks like I'll also need to port this over to buildtools/release/1.0.0 to get this change in corefx/release/1.0.0 builds--I'll give that a shot.

dagood:
Never mind, corefx/release/1.0.0 still uses [UpdatePublishedVersions.ps1](https://github.com/dotnet/corefx/blob/b0552ed99962c248de2c9f5d987e341cb817da76/UpdatePublishedVersions.ps1) (no VersionTools at all) so we would need a big port and a build def change, or a unique implementation.

Do we need this for release/1.0.0 servicing or is it ok to only have it going forwards?

markwilkie:
We should minimize the churn in our release branches as much as we can.

chcosta:
@dagood, is there a way to gather this information from our builds without doing a major port / build def change to the release branch?  The primary reason I was pushing for this feature was to address a gap which will grow with each servicing build out of release/1.0.0 and release/1.1.0.  

Though, it just now occurs to me, that we may be able to address that gap by providing multiple inputs and letting the build tests against packages logic combine the Latest and LKG files.  The logic to support multiple version files should already be supported in the release branches, I still need to port the change to support URI's instead of just local files though.  If that works, we could get away without this in the release branches.

dagood:
Latest + LKG does sound to me like it'll work.

The "unique implementation" I have in mind is reimplementing the logic here in powershell in `UpdateDependencies.ps1`. Doable if need be--it's small in scope and isolated, but completely new build infra code in the servicing branch. I don't know how that compares with the work to point to two URIs.

chcosta:
Sorry, it occurs to me that this won't work in a sustainable fashion.  In my head I was thinking of the model which this PR covered, which is creating a "Latest" and including the previous "Latest" packages.  Given that model, then LKG + Latest addresses the full current state of the branch.  In release/1.0.0, Latest represents only the newest packages published, so LKG + Latest is only LKG + most recent packages, and that creates a possible delta of packages which aren't in that list.  Does that make sense?  We need some other solution.

dagood:
I might not understand the requirements--we can freely change the LKG file, so in what situation would there be missed packages? If we need to have access to *all* released packages (not just the *latest* released packages) this PR wouldn't solve that either. It would be possible to maintain another file `All_Released_Packages.txt` programmatically, though.

chcosta:
I see, manually changing the LKG file would be the least intrusive.  If there's a doc on producing a servicing release out of 1.0.0, we should add that step.

PR #1123, opened 2016-10-12
Allow BuildAgainstPackages to work with global test project.json 

karajas:
Check for use of common test project.json

Change AddDependenciesToProjectJson to allow for compile only assets in framework specific dependencies of common project.json

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -275,13 +275,13 @@ namespace Microsoft.DotNet.Build.Tasks
            } 
         }
 
-        private JToken GetFrameworkDependenciesSection(JObject projectJsonRoot, string framework = null)
+        private JObject GetFrameworkDependenciesSection(JObject projectJsonRoot, string framework = null)
         {
             if(string.IsNullOrWhiteSpace(framework))
             {
-                return projectJsonRoot["dependencies"];
+                return (JObject) projectJsonRoot["dependencies"];
             }
-            return projectJsonRoot.SelectToken("frameworks." + NewtonsoftEscapeJProperty(framework) + ".dependencies");
+            return (JObject) projectJsonRoot["frameworks"][framework]["dependencies"];
         }
 
         // Generate the combines dependencies from the projectjson jObject and from AdditionalDependencies
@@ -311,7 +311,30 @@ namespace Microsoft.DotNet.Build.Tasks
                             JProperty addProperty;
                             if (nuGetVersion != null)
                             {
-                                addProperty = new JProperty(property.Name, nuGetVersion.ToString());
+                                //If the dependency is of the form 
+                                // "Dependency" : {
+                                // "version": "3.1.4-beta-23456-00",
+                                // "include": "compile"
+                                //  }
+                                if (property.Value.Type==JTokenType.Object)
+                                {
+                                    JObject jo = property.Value as JObject;
+
+                                    if (jo["version"] != null)
+                                    {
+                                        jo.Remove("version");
+                                        jo.Add("version", nuGetVersion.ToString());
+                                    }
+                                    else
+                                    {
+                                        Log.LogMessage("Ignoring Dependency '{0}', property does not have a version", property.Name);
+                                    }
+                                    addProperty = new JProperty(property.Name, jo);
+                                }
+                                else
+                                {
+                                    addProperty = new JProperty(property.Name, nuGetVersion.ToString());
+                                }
                             }
                             else
                             {
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -323,7 +323,7 @@ namespace Microsoft.DotNet.Build.Tasks
                                     if (jo["version"] != null)
                                     {
                                         jo.Remove("version");
-                                        jo.Add("version", nuGetVersion.ToString());
+                                        jo["version"] = nuGetVersion.ToString();
                                     }
                                     else
                                     {
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
@@ -39,12 +39,12 @@
   </PropertyGroup>
   
   <PropertyGroup>
-      <GenerateTestProjectJsonDependsOn Condition="'$(SkipGenerateTestProjectJson)'=='false'">
-        ExtractProjectReferences;
-        GatherProjectReferenceForProjectJson;
-        DetermineProjectJsonPath;
-        AddDependenciesToProjectJson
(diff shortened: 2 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -10,6 +10,7 @@
 using NuGet.Packaging;
 using NuGet.Packaging.Core;
 using NuGet.Versioning;
+using System.Diagnostics;
```
karajas: will remove.

karajas:
@chcosta

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -38,7 +38,7 @@
     </ResolveNuGetPackagesDependsOn>
   </PropertyGroup>
 
-  <Target Name="GenerateTestProjectJson" DependsOnTargets="ExtractProjectReferences;GatherProjectReferenceForProjectJson;DetermineProjectJsonPath;AddDependenciesToProjectJson" />
+  <Target Name="GenerateTestProjectJson" DependsOnTargets="ExtractProjectReferences;GatherProjectReferenceForProjectJson;DetermineProjectJsonPath;AddDependenciesToProjectJson" Condition="'$(UsesCommonTestProjectJson)'!='true'"/>
```
weshaggard: Where does UsesCommonTestProjectJson get set?

weshaggard: By the way regardless of the condition value all the DependsOnTargets will still get executed, is that going to cause issues?

karajas: I initially had the check on the depends on targets but this seems to have the same behavior.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -38,11 +38,11 @@
     </ResolveNuGetPackagesDependsOn>
   </PropertyGroup>
 
-  <Target Name="GenerateTestProjectJson" DependsOnTargets="ExtractProjectReferences;GatherProjectReferenceForProjectJson;DetermineProjectJsonPath;AddDependenciesToProjectJson" />
+  <Target Name="GenerateTestProjectJson" DependsOnTargets="ExtractProjectReferences;GatherProjectReferenceForProjectJson;DetermineProjectJsonPath;AddDependenciesToProjectJson"/>
 
   <Target Name="GenerateAllTestProjectJsons" DependsOnTargets="GenerateTestProjectJson" />
 
-  <Target Name="DetermineProjectJsonPath">
+  <Target Name="DetermineProjectJsonPath" Condition="'$(UsesCommonTestProjectJson)'!='true'">
```
weshaggard: I definitely don't like seeing this condition duplicated all over the place. You should use something along the lines of 

```
  <PropertyGroup>
    <GenerateTestProjectJsonDependsOn Condition="'$(SkipGenerateTestProjectJson)'!='true'">
      ExtractProjectReferences;
      GatherProjectReferenceForProjectJson;
      DetermineProjectJsonPath;
      AddDependenciesToProjectJson
    </GenerateTestProjectJsonDependsOn>
  </PropertyGroup>
 
  <Target Name="GenerateTestProjectJson" DependsOnTargets="$(GenerateTestProjectJsonDependsOn)" />
```

karajas: Made the associated change. Since the propertygroup listed above is global, I had to create another target to evaluate in the given context.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -37,8 +37,21 @@
       DetermineProjectJsonPath;
     </ResolveNuGetPackagesDependsOn>
   </PropertyGroup>
-
-  <Target Name="GenerateTestProjectJson" DependsOnTargets="ExtractProjectReferences;GatherProjectReferenceForProjectJson;DetermineProjectJsonPath;AddDependenciesToProjectJson" />
+  
+  <!--This target is required to evaluate this property uniquely in the context of each project -->
+  <Target Name="EvaluateGenerateTestProjectJsonDependsOn" >
+    <PropertyGroup>
+      <GenerateTestProjectJsonDependsOn Condition="'$(SkipGenerateTestProjectJson)'=='false'">
```
weshaggard: This shouldn't be in a wrapper target. Also I would change this condition to `'$(SkipGenerateTestProjectJson)' == 'true'`

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -38,8 +38,6 @@
     </ResolveNuGetPackagesDependsOn>
   </PropertyGroup>
   
-  <!--This target is required to evaluate this property uniquely in the context of each project -->
-  <Target Name="EvaluateGenerateTestProjectJsonDependsOn" >
     <PropertyGroup>
```
weshaggard: nit: indention should align.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -275,13 +275,13 @@ private static JObject ReadProject(string projectJsonPath)
...
-        private JToken GetFrameworkDependenciesSection(JObject projectJsonRoot, string framework = null)
+        private JObject GetFrameworkDependenciesSection(JObject projectJsonRoot, string framework = null)
         {
             if(string.IsNullOrWhiteSpace(framework))
             {
-                return projectJsonRoot["dependencies"];
+                return (JObject) projectJsonRoot["dependencies"];
             }
-            return projectJsonRoot.SelectToken("frameworks." + NewtonsoftEscapeJProperty(framework) + ".dependencies");
+            return (JObject) projectJsonRoot["frameworks"][framework]["dependencies"];
```
karajas: The other two occurrences look like they want to determine the path to the entry, and so use the NewtonsoftEscapeJProperty method to construct it and refer to it as a JProperty. Here I wanted to retrieve the JObject that is referenced at this path.

chcosta:
@karajas, can you clarify for me why adddependenciestoprojectjson is still needed when using a common project.json?

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: I don't see anything about the other two instances which necessitates they be treated as JProperties.

chcosta:
Are there test projects with project references to things not in the common project.json?

karajas:
Yup there are still test project.json's that are local to the test project.
Also adddependenciestoprojectjson is still used to update dependencies on a project.json of a locally build project, the VersionTools repo update script had some edge cases that were more easily resolve with AddDependencies.

PR #1125, opened 2016-10-13
Add Linux bootstrapper script.

crummel:
I attempted to keep this very close to the PowerShell bootstrapper (some of it is also cribbed from CLI's init-tools.sh).  Only tested on Ubuntu so far.

### DIFF
--- a/bootstrap/bootstrap.sh
+++ b/bootstrap/bootstrap.sh
@@ -1,0 +1,257 @@
+#!/usr/bin/env bash
+
+# Stop script on NZEC
+set -e
+# Stop script if unbound variable found (use ${var:-} if intentional)
+set -u
+# By default cmd1 | cmd2 returns exit code of cmd2 regardless of cmd1 success
+# This is causing it to fail
+set -o pipefail
+
+# Use in the the functions: eval $invocation
+invocation='say_verbose "Calling: ${FUNCNAME[0]}"'
+
+# standard output may be used as a return value in the functions
+# we need a way to write text on the screen in the functions so that
+# it won't interfere with the return value.
+# Exposing stream 3 as a pipe to standard output of the script itself
+exec 3>&1
+
+say_err() {
+    printf "%b\n" "bootstrap: Error: $1" >&2
+}
+
+say() {
+    # using stream 3 (defined in the beginning) to not interfere with stdout of functions
+    # which may be used as return value
+    printf "%b\n" "bootstrap: $1" >&3
+}
+
+say_verbose() {
+    if [ "$verbose" = true ]; then
+        say "$1"
+    fi
+}
+
+machine_has() {
+    eval $invocation
+    
+    hash "$1" > /dev/null 2>&1
+    return $?
+}
+
+check_min_reqs() {
+    if ! machine_has "curl"; then
+        say_err "curl is required to download dotnet. Install curl to proceed."
+        return 1
+    fi
+    
+    return 0
+}
+
+# args:
+# remote_path - $1
+# [out_path] - $2 - stdout if not provided
+download() {
+    eval $invocation
+    
+    local remote_path=$1
+    local out_path=${2:-}
+
+    local failed=false
+    if [ -z "$out_path" ]; then
+        curl --retry 10 -sSL --create-dirs $remote_path || failed=true
+    else
+        curl --retry 10 -sSL --create-dirs -o $out_path $remote_path || failed=true
+    fi
+    
+    if [ "$failed" = true ]; then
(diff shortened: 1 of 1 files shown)


crummel:
cc @weshaggard @chcosta for review.

bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,268 @@
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# See LICENSE file in the project root for full license information.
+
+# Note: This script should be compatible with the dash shell used in Ubuntu. So avoid bashisms! See https://wiki.ubuntu.com/DashAsBinSh for more info
```
mellinoe: Don't we need a shebang?

bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,268 @@
...
+
+# standard output may be used as a return value in the functions
+# we need a way to write text on the screen in the functions so that
+# it won't interfere with the return value.
+# Exposing stream 3 as a pipe to standard output of the script itself
+exec 3>&1
+
+# Setup some colors to use. These need to work in fairly limited shells, like the Ubuntu Docker container where there are only 8 colors.
+# See if stdout is a terminal
+if [ -t 1 ]; then
```
mellinoe: I think we should avoid doing anything with colors in this script, I think. `tput` isn't even installed in a lot of docker images that I've seen, and it seems a bit silly to install it just for this purpose. We should try to have as few dependencies in here as technically possible.

bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,268 @@
...
+say_verbose() {
+    if [ "$verbose" = true ]; then
+        say "$1"
+    fi
+}
+
+machine_has() {
+    eval $invocation
+    
+    which "$1" > /dev/null 2>&1
```
mellinoe: Let's avoid using `which` to validate prereqs; it's not installed by default on a lot of systems. It gets awkward because checks for other programs will fail when what you are really missing is `which` itself.

http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

https://github.com/dotnet/coreclr/issues/6904

bootstrap/bootstrap.sh
mellinoe: Actually, could you clarify this comment/note? Most of our other scripts are written for bash, and I think most folks use bash as their shell, too. Unless I'm misunderstanding how shebangs work (definitely possible 😄 ).

bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,268 @@
...
+# remote_path - $1
+# [out_path] - $2 - stdout if not provided
+download() {
+    eval $invocation
+    
+    local remote_path=$1
+    local out_path=${2:-}
+
+    local failed=false
+    if [ -z "$out_path" ]; then
```
mellinoe: Do we want to do any retries here? Just looking at init-tools in coreclr, and it specifies some other stuff, too. Is any of this interesting?

```
curl --retry 10 -sSL --create-dirs -o $__DOTNET_PATH/dotnet.tar $__CLIDownloadURL
```

bootstrap/bootstrap.sh
crummel: Removed this and added a shebang, this was just copied from the existing scripts.

bootstrap/bootstrap.sh
crummel: Removed colors.

bootstrap/bootstrap.sh
crummel: -S (show errors) and --fail (fail silently) conflict.  I switched this so we don't fail silently, I think it's better to show the error.  Also added retries.

crummel:
@mellinoe I've added the option for a local CLI, let me know if that looks good.  Thanks!

mellinoe:
The changes look good to me. Do we need to copy the `forcedCliLocalPath` into the "Tools" directory, or wherever the install target is? I'm assuming the build targets will expect it in a certain place, would we need to override that path as well?

crummel:
I think the symlinking on line 200 of $toolsLocalPath/dotnetcli/shared/Microsoft.NETCore.App/version to $cliInstallPath/shared/Microsoft.NETCore.App/$version should take care of that, since buildtools just expects that first one to exist.  Is there anything else it depends on?

mellinoe:
Yeah, that's a good point. That should work. Thanks!

crummel:
Rebased and squashed, will merge after the build passes.

Issue #1019, opened 2016-09-07
Support generating package reports for private/runtime packages


weshaggard:
Today we are using the reports to generate flat layouts for indexing our builds but the runtime and private packages don't have reports so we are special casing them by hard coding packages (see https://github.com/dotnet/corefx/blob/dev/api/layout/dir.proj#L19). That is very fragile and keeps breaking. Instead we should produce a package report for all our packages.

cc @danmosemsft

PR #1126, opened 2016-10-13
Always create package report

ericstj:
I've pulled out package report creation so that it doesn't rely on Validation.  I've also refactored validation be driven off the report instead of producing the report.  While doing this I noticed some dead code and cleaned that up.

Fixes https://github.com/dotnet/buildtools/issues/1019

/cc @weshaggard


src/Microsoft.DotNet.Build.Tasks.Packaging/src/Generations.cs
```diff
@@ -16,203 +16,6 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
 {
     public class Generations
     {
-        /// <summary>
-        /// A cache of path to generation definition file > Generations objects to avoid
```
weshaggard: Looks like this is unrelated to package reports.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetApplicableAssetsFromPackageReports.cs
```diff
@@ -79,13 +85,13 @@ public override bool Execute()
 
             CompileAssets = compileAssets.ToArray();
             RuntimeAssets = runtimeAssets.ToArray();
+            BuildProjects = buildProjects.Distinct().Select(bp => bp.ToItem()).ToArray();
```
weshaggard: What do you expect to do with the BuildProjects? Wouldn't we also need to know the set of global properties needed if we were going to rebuild them?

weshaggard: Nevermind I see below that you actually do capture those properties.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -1028,9 +1029,22 @@
                     Condition="'%(PackageFile.IsSourceCodeFile)'!='true'" />
     </ItemGroup>
   </Target>
+  
+  <Target Name="GetPackageReport"
```
weshaggard: What do you expect to trigger this target?

weshaggard: Whatever we do trigger it with an we do it for all packages and address https://github.com/dotnet/buildtools/issues/1019

src/Microsoft.DotNet.Build.Tasks.Packaging/src/Generations.cs
ericstj: Yeah, I kept it in a separate commit.  It was just some cleanup.  If you prefer I can merge it in a separate PR.

weshaggard: You can keep it with this I was just trying to figure out the connection.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetApplicableAssetsFromPackageReports.cs
ericstj: This is all the task work that enabled me to build from the referencing project instead of from within the pkgproj, but I didn't include the targets change because it required some messy UndefineProperties.  You can see that here, I'm not submitting it at the moment: https://github.com/ericstj/buildtools/commit/9bd54529a8ef4b0c73d552aeae8598ff474996a9

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: It's triggered by ValidatePackage.  It can also be triggered on its own.  EG: by adding to dir.traversal.targets and calling it on packages.builds.
I'll also add it to BuildDependsOn to ensure it runs even if SkipValidatePackage is set.

weshaggard: I don't see it chained on ValidatePackage.  In any case I would prefer it always run for pkgproj builds.

ericstj: It's right here: https://github.com/dotnet/buildtools/pull/1126/files#diff-03c502c8604ae1b4ae2efeb1f3d0c45eR1047
Also, now here: https://github.com/dotnet/buildtools/pull/1126/files#diff-03c502c8604ae1b4ae2efeb1f3d0c45eR131

PR #1129, opened 2016-10-13
Fix a couple minor issues found when adding the bootstrap in CLI.

crummel:
- Symlink argument were switched.
- Typo in clearing the local tools directory.


crummel:
cc @mellinoe

mellinoe:
LGTM

PR #1130, opened 2016-10-14
Record native assets in package report

ericstj:
Added support to record native assets in the package report.

/cc @weshaggard


weshaggard:
LGTM

PR #1131, opened 2016-10-14
Fix default parameter values for value types

### LINKED ISSUE #1127: GenAPI doesn't handle value types with default parameters set
Seems like it always sets the value to null.

```C#
public void DoSomethingAsync(CancellationToken = default(CancellationToken))
{
   // ...
}
```

Will generate:
```C#
public void DoSomethingAsync(CancellationToken = null)
{
   // ...
}
```

/cc @weshaggard

davidfowl:
Fixes #1127 

/cc @ericstj @weshaggard


ericstj:
LGTM, thanks for fixing.

weshaggard:
Thanks @davidfowl for fixing this.

PR #1133, opened 2016-10-14
Add TargetGroup to the dependency list file to avoid clashes

joperezr:
cc: @weshaggard @MattGal 
FYI: @karajas @tarekgh 

Adding `TargetGroup` to the dependency list file name, which will help avoid race conditions of two builds trying to write to the same file, which has caused some intermittent build brakes.


tarekgh:
LGTM

weshaggard:
LGTM this should get fixed all up once we fix the larger issue.

PR #1134, opened 2016-10-14
Partial package support

ericstj:
Packages can be built without building all binaries by specifying
AllowPartialPackages=true.

To build a partial package without rebuilding any of the projects use
the following
msbuild <pkgproj> /p:AllowPartialPackages=true;BuildPackageLibraryReferences=false

/cc @ellismg


ericstj:
Ping: @ellismg please have a look.

ellismg:
LGTM

PR #1135, opened 2016-10-14
Resolve ordering issue with IsTestProject

karajas:
Fix condition for SkipGenerateTestProjectJson to match new default

/cc @weshaggard


karajas:
Thanks.

PR #1136, opened 2016-10-14
Generate layout from package reports

ericstj:
This produces flat-file layout directories from corefx.

These look like
ref{tfm}\file.dll
runtime{tfm}{rid}\file.dll

/cc @weshaggard @Petermarcu @ellismg

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLayoutFiles.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLayoutFiles.cs
@@ -1,0 +1,139 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using NuGet.Frameworks;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks.Packaging
+{
+    public class GetLayoutFiles : PackagingTask
+    {
+        /// <summary>
+        /// Package report files
+        /// </summary>
+        [Required]
+        public string[] PackageReports { get; set; }
+
+        /// <summary>
+        /// Destination directory for layout
+        /// </summary>
+        [Required]
+        public string DestinationDirectory { get; set; }
+
+        /// <summary>
+        /// Optional set of frameworks to restrict the layout
+        ///   Identity: Framework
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -1203,4 +1205,29 @@
                         StablePackages="@(StablePackage)"
                         ModuleToPackages="@(ModuleToPackage)" />
   </Target>
+
+  <Target Name="CreateLayout"
+          Condition="'$(IsRuntimePackage)' != 'true'"
```
weshaggard: Why wouldn't we do this for runtime packages as well?

ericstj: The identity package report includes all the runtime package info.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GetLayoutFiles.cs
```diff
@@ -0,0 +1,139 @@
...
+                        }
+
+                        if (rid != null && rids.Count > 0 && !rids.Contains(rid))
+                        {
+                            Log.LogMessage(LogImportance.Low, $"Skipping {fx}/{rid} since it is not in {nameof(Frameworks)}");
+                            continue;
+                        }
+                    }
+
+                    if (!packageReport.SupportedFrameworks.ContainsKey(fx.ToString()))
```
ericstj: I list all targets even those that might not be supported since that can help folks debug what we consider a package as not supported.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -1203,4 +1205,29 @@
                         StablePackages="@(StablePackage)"
                         ModuleToPackages="@(ModuleToPackage)" />
   </Target>
+
+  <Target Name="CreateLayout"
```
ericstj: It'll work either way.   That's why I made it take an array.  I just put it here as a shortcut.  I'm fine with this target not being in packaging.targets and only being in your layout project.

weshaggard: I'm fine with this for now, as you say we can pivot as needed. One nice thing about doing it as part of the individual build is that it will always update if you just rebuild the one library package project. However it might actually be interesting to it from the library build itself.

ericstj:
Removing WIP tag.  This is ready to merge.

PR #1138, opened 2016-10-14
Changing TestNugetTargetMoniker to be a Property instead of an Item

joperezr:
cc: @weshaggard @ericstj @karajas 

Now that each TestPath is unique to a build (since TargetGroup and TestTFM are part of it) we can get rid of the need of having TestNugetTargetMoniker being an item, since we don't need its metadata anymore.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -12,6 +12,13 @@
     <TestsSuccessfulSemaphore>tests.passed</TestsSuccessfulSemaphore>
   </PropertyGroup>
 
+  <!-- In case that TestPath is not yet set, default it here -->
+  <PropertyGroup>
+    <TestTargetOutputRelPath Condition="'$(TestTargetOutputRelPath)'=='' And '$(TargetGroup)'!='' And '$(TestTFM)'!=''">$(TargetGroup).$(TestTFM)/</TestTargetOutputRelPath>
+    <TestTargetOutputRelPath Condition="'$(TestTargetOutputRelPath)'=='' And '$(TargetGroup)'=='' And '$(TestTFM)'!=''">default.$(TestTFM)/</TestTargetOutputRelPath>
+    <TestPath Condition="'$(TestPath)'==''">$(TestWorkingDir)$(OSPlatformConfig)/$(MSBuildProjectName)/$(TestTargetOutputRelPath)</TestPath>
+  </PropertyGroup>
+
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


weshaggard:
Does this depend on changes in dir.props in the individual repo's? If so can we be sure to setup the TestPath over here as part of that so that we don't end up breaking folks like WCF.

joperezr:
That's a good point, yes it does depend on TestPath, but the problem is that we can't really move that down to Build.Common.props because TargetGroup and TestTFM need to be set already. What do you think about placing it in tests.targets?

weshaggard:
>What do you think about placing it in tests.targets?

Given that this is specific to a given test project configuration I think that should probably be OK. We should do a quick search to ensure TestPath isn't used in any other context first.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -12,6 +12,13 @@
     <TestsSuccessfulSemaphore>tests.passed</TestsSuccessfulSemaphore>
   </PropertyGroup>
 
+  <!-- In case that TestPath is not yet set, default it here -->
+  <PropertyGroup>
+    <TestTargetOutputRelPath Condition="'$(TestTargetOutputRelPath)'=='' And '$(TargetGroup)'!='' And '$(TestTFM)'!=''">$(TargetGroup).$(TestTFM)/</TestTargetOutputRelPath>
+    <TestTargetOutputRelPath Condition="'$(TestTargetOutputRelPath)'=='' And '$(TargetGroup)'=='' And '$(TestTFM)'!=''">default.$(TestTFM)/</TestTargetOutputRelPath>
+    <TestPath Condition="'$(TestPath)'==''">$(TestWorkingDir)$(OSPlatformConfig)/$(MSBuildProjectName)/$(TestTargetOutputRelPath)</TestPath>
```
joperezr: @weshaggard with this change I'll set TestPath in tests.targets in case it is not set yet. It does have a problem though, since I'm only setting it if it is not set before, so in WCF most likely this will already be set incorrectly. What approach do you like better?:
- One option is to remove the condition from this here, so that TestPath gets overriden as dir.props is imported first.
- Keep Condition here, and just coordinate with WCF repo so that they fix dir.props to remove the setting of TestPath since it's going to live here now.

weshaggard: I would keep the condition. I don't believe it should break anything it just might still have the race condition problem correct? Which they can fix by just deleting the property in their repo.

weshaggard:
FYI @naamunds

PR #1139, opened 2016-10-14
Add extension import in packageresolve.targets

alexperovich:
This is to support the source indexer.

/cc: @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
@@ -265,4 +265,6 @@
       <Compile Include="$(BlockReflectionAtributeFile)" />  
     </ItemGroup>
   </Target>
+  
+  <Import Project="$(CustomAfterBuildCommonTargets)" Condition="Exists('$(CustomAfterBuildCommonTargets)')"/>
 </Project>


weshaggard:
What do you need to plug into this spot from the source indexer?

alexperovich:
The targets that rewrite package references to live built binaries.

weshaggard:
Why does it need an extension point here then? You should be able to include that target anywhere and just hook it into the correct target chain.

alexperovich:
Well it needs at least one extension point. I put it here because it is logically a modification of packageresolve.targets.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -183,4 +183,7 @@
     <Message Importance="Low"
              Text="Removed all ResolvedTagetingPackReferences that were not specified explicitly as a TargetingPackReference=[@(TargetingPackReference)]. PackageReferencesToRemove=[@(PackageReferencesToRemove)]." />
   </Target>
+  
+  <Import Project="$(CustomAfterPackageResolveTargets)" Condition="'$(CustomAfterPackageResolveTargets)' != '' AND Exists('$(CustomAfterPackageResolveTargets)')"/>
```
weshaggard: Lets make this a little more general and put it in build.common.targets.

Also the Exists check should be enough as you don't need to check for empty as well.

PR #1145, opened 2016-10-18
Add VSTS Api package for multiplexing checked in build definitions

chcosta:
Implementation for API discussed in https://github.com/chcosta/buildtools/blob/vstsapi_proposal/Documentation/VSTSBuildManagement.md.

This allows us to multiplex build definitions based on a collection id which build orchestrators will utilize based on convention.  

@weshaggard @dagood @dleeapho @AlfredoMS @markwilkie

### DIFF
--- a/src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildClient.cs
+++ b/src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildClient.cs
@@ -122,7 +122,7 @@ namespace VstsBuildsApi
             string requestUri = $"DefaultCollection/{VstsBuildsApiBase(definition)}/_apis/build/definitions/{definition["id"].ToString()}?api-version=2.0";
 
             HttpContent content = new StringContent(JsonConvert.SerializeObject(definition), System.Text.Encoding.UTF8, "application/json");
-            HttpResponseMessage response = await GetClient(definition["project"]["url"].ToString()).PutAsync(requestUri, content);
+            HttpResponseMessage response = await GetClient(GetVstsBuildsApiUrl(definition)).PutAsync(requestUri, content);
             ProcessResponseStatusCode(response);
             return JObject.Parse(await response.Content.ReadAsStringAsync());
         }
@@ -138,7 +138,7 @@ namespace VstsBuildsApi
             string requestUri = $"DefaultCollection/{VstsBuildsApiBase(definition)}/_apis/build/definitions?api-version=2.0";
 
             HttpContent content = new StringContent(JsonConvert.SerializeObject(definition), Encoding.UTF8, "application/json");
-            HttpResponseMessage response = await GetClient(definition["project"]["url"].ToString()).PostAsync(requestUri, content);
+            HttpResponseMessage response = await GetClient(GetVstsBuildsApiUrl(definition)).PostAsync(requestUri, content);
             ProcessResponseStatusCode(response);
             return JObject.Parse(await response.Content.ReadAsStringAsync());
 
@@ -151,7 +151,7 @@ namespace VstsBuildsApi
         {
             string requestUri = $"DefaultCollection/{VstsBuildsApiBase(definition)}/_apis/build/definitions/{definition["id"].ToString()}?api-version=2.0";
 
-            HttpResponseMessage response = await GetClient(definition["project"]["url"].ToString()).GetAsync(requestUri);
+            HttpResponseMessage response = await GetClient(GetVstsBuildsApiUrl(definition)).GetAsync(requestUri);
             ProcessResponseStatusCode(response);
             return JObject.Parse(await response.Content.ReadAsStringAsync());
         }
@@ -163,7 +163,7 @@ namespace VstsBuildsApi
         {
             string requestUri = $"DefaultCollection/{VstsBuildsApiBase(definition)}/_apis/build/definitions?api-version=2.0&name={definition["name"].ToString()}";
 
-            HttpResponseMessage response = await GetClient(definition["project"]["url"].ToString()).GetAsync(requestUri);
+            HttpResponseMessage response = await GetClient(GetVstsBuildsApiUrl(definition)).GetAsync(requestUri);
             ProcessResponseStatusCode(response);
             string json = await response.Content.ReadAsStringAsync();
             JObject definitionsJObject = JObject.Parse(json);
@@ -186,6 +186,11 @@ namespace VstsBuildsApi
         private string VstsBuildsApiBase(JObject definition)
         {
             return definition["project"]["name"].ToString();
+        }
+
+        private string GetVstsBuildsApiUrl(JObject definition)
+        {
+            return definition["project"]["url"].ToString();
         }
 
         private HttpClient GetClient(string url)
--- a/src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildClient.cs
+++ b/src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildClient.cs
@@ -1,0 +1,237 @@
+﻿// Licensed to the .NET Foundation under one or more agreements. 
+// The .NET Foundation licenses this file to you under the MIT license. 
+// See the LICENSE file in the project root for more information. 
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace VstsBuildsApi
+{
+    public class VstsBuildClient
+    {
+        private string _collectionIdentifier;
+        private string _credentials;
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="collectionIdentifier">collection identifer for build definitions</param>
+        /// <param name="credentials">credentials used for basic rest api authentication</param>
+        public VstsBuildClient(string collectionIdentifier, string credentials)
+        {
+            if(string.IsNullOrWhiteSpace(collectionIdentifier))
+            {
+                throw new Exception("Required parameter (collectionIdentifier) cannot be null or white space.");
+            }
+            if(!Uri.IsWellFormedUriString(collectionIdentifier, UriKind.Relative))
+            {
+                throw new Exception(string.Format("collectionIdentifier '{0}' contains invalid characters.  collectionIdentifier must contain only valid uri characters.", collectionIdentifier));
+            }
+
+            _collectionIdentifier = collectionIdentifier;
+            _credentials = credentials;
+        }
+
+        /// <summary>
+        /// Load a local build definition, combine the uniqueIdentifier with the definition name, 
+        /// check VSTS for a matching definition name, if present, update that definition with the
+        /// local build definition, otherwise create a new definition.
+        /// </summary>
+        /// <param name="stream">Stream to a VSTS build definition</param>
+        /// <returns>Created or updated build Id</returns>
+        public async Task<string> CreateOrUpdateDefinitionAsync(Stream stream)
+        {
+            JObject definition = GetDefinition(stream);
+            return await CreateOrUpdateDefinitionAsync(definition).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Load a local build definition, combine the uniqueIdentifier with the definition name, 
+        /// check VSTS for a matching definition name, if present, update that definition with the
+        /// local build definition, otherwise create a new definition.
+        /// </summary>
+        /// <param name="definition">VSTS build definition JSON object</param>
+        /// <returns>Created or updated build definition id</returns>
+        private async Task<string> CreateOrUpdateDefinitionAsync(JObject definition)
+        {
+            var key = string.Join("_", _collectionIdentifier, definition["name"].ToString());
+            definition["name"] = key;
+            var vstsDefinitions = await VstsRetrieveDefinitionsListByNameAndPathAsync(definition).ConfigureAwait(false);
+
+            if (vstsDefinitions.Count == 0)
+            {
+                /* Create */
+                var vstsDefinition = await VstsCreateDefinitionAsync(definition).ConfigureAwait(false);
+                return vstsDefinition["id"].ToString();
+            }
+            else if(vstsDefinitions.Count == 1)
+            {
+                var vstsDefinition = await VstsRetrieveDefinitionByIdAsync(vstsDefinitions[0]).ConfigureAwait(false);
+
+                /* Update */
+                if (!IsSameContents(definition, vstsDefinition))
+                {
+                    definition["revision"] = vstsDefinition["revision"];
+                    definition["id"] = vstsDefinition["id"];
+                    definition["uri"] = vstsDefinition["uri"];
+                    definition["path"] = vstsDefinition["path"];
+                    definition["createdDate"] = vstsDefinition["createdDate"];
+                    definition["url"] = vstsDefinition["url"];
+                    definition["comment"] = "Automated update";
+
+                    vstsDefinition = await VstsUpdateDefinitionAsync(definition).ConfigureAwait(false);
+                }
+                return vstsDefinition["id"].ToString();
+            }
+            else
+            {
+                throw new Exception(string.Format("Obtained multiple {0} definitions with the same 'name' ({1}) and 'path' ({2}) properties.  This should not be possible.", vstsDefinitions.Count, vstsDefinitions[0]["name"].ToString(), vstsDefinitions[0]["path"].ToString()));
+            }
+        }
+
+        private JObject GetDefinition(Stream stream)
+        {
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,228 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+
+namespace VSTSBuildsApi
+{
+    public class VSTSBuildClient
```
dagood: Style guides I've seen say acronyms over two chars should be considered a word, so `VstsBuildClient` and `VstsBuildsApi`. This also lets you rename the private methods from (e.g.) `VSTS_UpdateDefinition` to `VstsUpdateDefinition`, unless there was some other reason for the underscore than readability.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,228 @@
+﻿using Newtonsoft.Json;
```
dagood: Missing license header.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        private HttpClient GetClient(string url)
+        {
+            HttpClient client = new HttpClient();
+            Uri uri = new Uri(url); 
+            client.BaseAddress = new Uri(string.Format("{0}://{1}/", uri.Scheme, uri.Authority));
+            client.DefaultRequestHeaders.Clear();
+            client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", _credentials);
+            return client;
```
dagood: Would the VSTS NuGet package be useful here, or is there a reason to implement the API "raw"?

src/Microsoft.DotNet.Build.VSTSBuildsApi/project.json
```diff
@@ -0,0 +1,11 @@
+{
+  "dependencies": {
+    "Newtonsoft.Json": "9.0.1",
+    "NETStandard.Library": "1.6.0"
+  },
+  "frameworks": {
+    "netstandard1.5": {
+      "imports": "dnxcore50"
```
dagood: Is the `imports` needed?

src/nuget/Microsoft.DotNet.BuildTools.VSTSBuildsApi.nuspec
```diff
@@ -0,0 +1,28 @@
...
+    <tags></tags>
+    <dependencies>
+      <group targetFramework="netstandard1.5">
+        <dependency id="NETStandard.Library" version="1.6.0" />
+        <dependency id="Newtonsoft.Json" version="9.0.1" />
+      </group>
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Microsoft.DotNet.Build.VSTSBuildsApi\*.dll" target="lib" />
```
dagood: Should this target `lib/netstandard1.5` rather than `lib`?

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// <summary>
+        /// Update VSTS Build Definition
+        /// </summary>
+        /// <param name="definition"></param>
+        private void VSTS_UpdateDefinition(JObject definition)
+        {
+            string requestUri = $"DefaultCollection/DevDiv/_apis/build/definitions/{definition["id"].ToString()}?api-version=2.0";
+
+            HttpContent content = new StringContent(JsonConvert.SerializeObject(definition), System.Text.Encoding.UTF8, "application/json");
+            ProcessResponseStatusCode(GetClient(definition["url"].ToString()).PutAsync(requestUri, content).Result);
```
dagood: Could `async` be weaved through this, rather than using the `.Result`s? I imagine the build pipeline would make use of that.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,228 @@
...
+
+                /* Update */
+                if (!IsSameContents(definition, vstsDefinition))
+                {
+                    definition["revision"] = vstsDefinition["revision"];
+                    definition["id"] = vstsDefinition["id"];
+                    definition["uri"] = vstsDefinition["uri"];
+                    definition["path"] = vstsDefinition["path"];
+                    definition["createdDate"] = vstsDefinition["createdDate"];
+                    definition["url"] = vstsDefinition["url"];
```
dagood: I'm wondering if it's worth pulling out the key names into arrays here and the other places in the file. It looks like they're a little different in each case, though.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
chcosta: You're right, I'll fix this.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
chcosta: Which package are you referring too.  Other implementations of the API which we are using currently in PipeBuild didn't not allow the same amount of control over build definitions.  The raw REST API gave us the greatest amount of flexibility.  If there is something else which allows us to download, list, update, and create full definitions, then I'm not opposed to investigating them.

src/nuget/Microsoft.DotNet.BuildTools.VSTSBuildsApi.nuspec
chcosta: I copied this from other buildtools packages.

src/Microsoft.DotNet.Build.VSTSBuildsApi/project.json
chcosta: I copied this from other buildtools packages

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
dagood: It's not a different package, but https://www.nuget.org/packages/Microsoft.TeamFoundationServer.Client/15.106.0-preview does let you do these things. The data models are strongly typed, but they have `DataContract` attributes so I think they'll work with json serialization. I'm not familiar with that, though, so might not be worth it.

src/Microsoft.DotNet.Build.VSTSBuildsApi/project.json
dagood: We want to only include it when needed, so try removing it--if it fails, it's a restore-time failure so it's hard to miss.

src/nuget/Microsoft.DotNet.BuildTools.VSTSBuildsApi.nuspec
dagood: Looks like it's done differently between nuspecs, but `lib/netstandard1.5` to match the framework is more typical for packages.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
chcosta: PipeBuild wasn't designed with `async` in mind, and with the current implementation, I don't want to rewrite a bunch of stuff (in PipeBuild) to support it, but I agree with the principle, and some other tool may want to take advantage of it.  I'll make that change.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,234 @@
...
+        /// Load a local build definition, combine the uniqueIdentifier with the definition name, 
+        /// check VSTS for a matching definition name, if present, update that definition with the
+        /// local build definition, otherwise create a new definition.
+        /// </summary>
+        /// <param name="stream">Stream to a VSTS build definition</param>
+        /// <returns>Created or updated build Id</returns>
+        public async Task<string> CreateOrUpdateDefinitionAsync(Stream stream)
+        {
+            JObject definition = GetDefinition(stream);
+            return await CreateOrUpdateDefinitionAsync(definition);
```
AlfredoMS: Consider using .ConfigureAwait(false)
This is a recommended practice for libraries that expose async methods and don't need to flow context.

AlfredoMS: other awaits in your class could use it too, but since this is a public method, at least this one should have it.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        /// <summary>
+        /// Update VSTS Build Definition
+        /// </summary>
+        /// <param name="definition"></param>
+        /// <returns>updated definitions id</returns>
+        private async Task<JObject> VstsUpdateDefinitionAsync(JObject definition)
+        {
+            string requestUri = $"DefaultCollection/DevDiv/_apis/build/definitions/{definition["id"].ToString()}?api-version=2.0";
```
chcosta: I preferred to derive this from the build definitions "url", because it is possible that we would want this to apply to other accounts, like "mseng".

dagood: I think Alfredo's talking about the `DefaultCollection/DevDiv` part here--which is also in the url, but it looks like `GetClient` only uses the `Authority` part of the url.

chcosta: Ah, that makes sense.  Good catch guys.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            HttpResponseMessage response = await GetClient(definition["url"].ToString()).GetAsync(requestUri);
+            ProcessResponseStatusCode(response);
+            return JObject.Parse(await response.Content.ReadAsStringAsync());
+        }
+
+        /// <summary>
+        /// Retrieve a VSTS Build Definition by name and VSTS folder path
+        /// </summary>
+        /// <param name="name"></param>
```
dagood: Name is inaccurate--I usually just write a summary unless I have things to say about the params. Other comments have this issue too.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        /// <summary>
+        /// Update VSTS Build Definition
+        /// </summary>
+        /// <param name="definition"></param>
+        /// <returns>updated definitions id</returns>
+        private async Task<JObject> VstsUpdateDefinitionAsync(JObject definition)
+        {
+            string requestUri = $"DefaultCollection/{definition["project"]["name"].ToString()}/_apis/build/definitions/{definition["id"].ToString()}?api-version=2.0";
```
dagood: `DefaultCollection/{definition["project"]["name"].ToString()}/_apis/build` is here 4 times, maybe worth extracting. `string VstsBuildApiBase(JObject definition) => $"...";`?

src/nuget/Microsoft.DotNet.BuildTools.VSTSBuildsApi.nuspec
```diff
@@ -0,0 +1,28 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.8.1">
+    <id>Microsoft.DotNet.BuildTools.VSTSBuildsApi</id>
```
dagood: Nitpicky, but mind changing to `Microsoft.DotNet.BuildTools.VstsBuildsApi`, and changing filenames? I like package ID to match the namespace.

src/Microsoft.DotNet.Build.VSTSBuildsApi/VSTSBuildClient.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// check VSTS for a matching definition name, if present, update that definition with the
+        /// local build definition, otherwise create a new definition.
+        /// </summary>
+        /// <param name="definition">VSTS build definition JSON object</param>
+        /// <returns>Created or updated build definition id</returns>
+        private async Task<string> CreateOrUpdateDefinitionAsync(JObject definition)
+        {
+            var key = string.Join("_", _collectionIdentifier, definition["name"].ToString());
+            definition["name"] = key;
+            var vstsDefinitions = await VstsRetrieveDefinitionsListByNameAndPathAsync(definition).ConfigureAwait(false);
```
AlfredoMS: NitPick, using 'var' when the type is not obvious.
According to corefx style conventions: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

>  We only use var when it's obvious what the variable type is (i.e. var stream = new FileStream(...) not var stream = OpenStandardInput()).

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildClient.cs
```diff
@@ -0,0 +1,237 @@
...
+        /// check VSTS for a matching definition name, if present, update that definition with the
+        /// local build definition, otherwise create a new definition.
+        /// </summary>
+        /// <param name="definition">VSTS build definition JSON object</param>
+        /// <returns>Created or updated build definition id</returns>
+        private async Task<string> CreateOrUpdateDefinitionAsync(JObject definition)
+        {
+            var key = string.Join("_", _collectionIdentifier, definition["name"].ToString());
+            definition["name"] = key;
+            var vstsDefinitions = await VstsRetrieveDefinitionsListByNameAndPathAsync(definition).ConfigureAwait(false);
```
AlfredoMS: NitPick, using 'var' when the type is not obvious.
According to corefx style conventions: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

 

> We only use var when it's obvious what the variable type is (i.e. var stream = new FileStream(...) not var stream = OpenStandardInput()).

chcosta: Intellisense makes everything obvious!  Your point, however, is valid.  That is our coding-guideline, thanks for pointing this out.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildClient.cs
```diff
@@ -0,0 +1,237 @@
...
+        /// <summary>
+        /// Load a local build definition, combine the uniqueIdentifier with the definition name, 
+        /// check VSTS for a matching definition name, if present, update that definition with the
+        /// local build definition, otherwise create a new definition.
+        /// </summary>
+        /// <param name="definition">VSTS build definition JSON object</param>
+        /// <returns>Created or updated build definition id</returns>
+        private async Task<string> CreateOrUpdateDefinitionAsync(JObject definition)
+        {
+            var key = string.Join("_", _collectionIdentifier, definition["name"].ToString());
```
weshaggard: Why are you using Join to concat 2 strings separated by a "_"? I would just concat them `_collectionIdentifier + "_" + definition["name"].ToString()`

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildClient.cs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        /// <summary>
+        /// Retrieve a VSTS Build Definition by name and VSTS folder path
+        /// </summary>
+        private async Task<IReadOnlyList<JObject>> VstsRetrieveDefinitionsListByNameAndPathAsync(JObject definition)
+        {
+            string requestUri = $"DefaultCollection/{VstsBuildsApiBase(definition)}/_apis/build/definitions?api-version=2.0&name={definition["name"].ToString()}";
+
+            HttpResponseMessage response = await GetClient(definition["project"]["url"].ToString()).GetAsync(requestUri);
```
weshaggard: nit: Might want to create a helper method for "project" "url" as it appears to be duplicated a few places, simila to VstsBuildsApiBase.

PR #1156, opened 2016-10-25
Report fixes

ericstj:
A couple fixes for build issues that cropped up after recent changes to always produce package reports.

/cc @weshaggard


weshaggard:
LGTM

Do you plan on updating buildtools in corefx? If not I will do that.

ericstj:
I figured you could use it locally, but if you need it updated in official builds go ahead.

PR #1158, opened 2016-10-25
Update NETCore platform versions and add support for fedora 24

mellinoe:
I'd like to enable testing on Fedora 24 in corefx. To do that, we need to add the RID into our TestSuite `runtime.json`. I've also added support for "regular" building on Fedora 24 by making the tool-runtime project.json compatible with the fedora.24-x64 rid. Previously, you needed to set your __PUBLISH_RID to fedora.23-x64 to successfully restore the tools.

@dagood


mellinoe:
@weshaggard @joperezr Does this look okay? I'd like to get it in so I can enable Fedora 24 testing starting with corefx.

PR #1159, opened 2016-10-25
Add FxCop analyzers to build

alexperovich:


bartonjs:
Have the suppressions already been added to the projects that needed them?  It's not good when buildtools upgrades get blocked.

morganbr:
The list of checks looks good. I don't know enough about the build system to review that part.

alexperovich:
The suppressions were added in dotnet/corefx#13014

weshaggard:
cc @AlexGhiondea

mellinoe:
Is there much of an impact on build time with this?

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -21,7 +21,18 @@
...
+    <!-- %24 = $ -->
+    <EnableFxCopAnalyzers Condition="$([System.Text.RegularExpressions.Regex]::IsMatch($(MSBuildProjectDirectory), 'src%24'))">true</EnableFxCopAnalyzers>
+    <EnableFxCopAnalyzers Condition="'$(EnableFxCopAnalyzers)' != 'true'">false</EnableFxCopAnalyzers>
+  </PropertyGroup>
+
+  <ItemGroup Condition="$(EnableFxCopAnalyzers)">
+    <Analyzer Include="$(CodeAnalyzerDir)*.dll" Exclude="$(CodeAnalyzerDir)Microsoft.DotNet.CodeAnalysis.dll"/>
+  </ItemGroup>
+  <PropertyGroup Condition="$(EnableFxCopAnalyzers)">
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)Default.ruleset</CodeAnalysisRuleSet>
```
AlexGhiondea: If we add a condition on the CodeAnalysisRuleSet being already set you can have per-project overrides of the ruleset, which might come in handy.

AlexGhiondea: Why not keep the existing ruleset file and add to it the rules we want to have enabled?

alexperovich: I will do that.

alexperovich: It had no rules in it and the name was too specific. I don't care much either way though.

PR #1161, opened 2016-10-26
Add EchoOff to commands in versioning targets

alexperovich:
This change adds EchoOff to the Exec tasks in versioning.targets.
This makes the commands not be written to the log when building the project.
A detailed log will still have the commands in it.


weshaggard:
While I like the general idea I think this just hides the fact that we are actually calling these all the time. I would instead like for us to figure out how to not call these all the time. 

cc @joperezr

joperezr:
I agree, it should only ever be calling it once when it produces the props file, and then it should not call it any longer, which I think was the behavior originally but maybe something else broke it. If you want you can log an issue to me to investigate.

PR #1164, opened 2016-10-26
Use environment INIT_TOOLS_RESTORE_ARGS variable when bootstrapping

dagood:
Like [init-tools.cmd](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd) and [init-tools.sh](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh), makes `bootstrap.sh/cmd` pick up on the `INIT_TOOLS_RESTORE_ARGS` variable to pass as extra arguments to `dotnet restore`.

For example, the variable can be set to `--disable-parallel` or `--verbosity debug`.

Since bootstrap.sh only downloads a few packages, this PR shouldn't actually matter much for the overall `--disable-parallel` fix. I was thinking a bootstrap PR was necessary to pass along the value to the buildtools init-tools script, but that isn't the case since it's an environment variable. I'm submitting this anyway for consistency, but this PR shouldn't block fixing the CLI build.

@crummel @chcosta
/cc @weshaggard

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -90,7 +90,7 @@ if ($env:buildtools_source -ne $null)
 }
 $packagesPath = Join-Path $RepositoryRoot "packages"
 $dotNetExe = Join-Path $cliLocalPath "dotnet.exe"
-$restoreArgs = "restore $projectJson --packages $packagesPath --source $buildToolsSource --source $nugetOrgSource"
+$restoreArgs = "restore $projectJson --packages $packagesPath --source $buildToolsSource --source $nugetOrgSource $($env:init_tools_restore_args)"
 $process = Start-Process -Wait -NoNewWindow -FilePath $dotNetExe -ArgumentList $restoreArgs -PassThru
 if ($process.ExitCode -ne 0)
 {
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap.ps1
```diff
@@ -90,7 +90,7 @@ if ($env:buildtools_source -ne $null)
 }
 $packagesPath = Join-Path $RepositoryRoot "packages"
 $dotNetExe = Join-Path $cliLocalPath "dotnet.exe"
-$restoreArgs = "restore $projectJson --packages $packagesPath --source $buildToolsSource --source $nugetOrgSource"
+$restoreArgs = "restore $projectJson --packages $packagesPath --source $buildToolsSource --source $nugetOrgSource $($env:init_tools_restore_args)"
```
weshaggard: I wouldn't change the name to BOOT_STRAP_RESTORE_ARGS, also I forget exactly where this is used but could we instead make this an optional parameter to both scrips instead of an environment variable?

dagood: Do you mean you would change the name? (I prefer `BOOTSTRAP_RESTORE_ARGS` to match filename.)

The main point of this variable is https://github.com/dotnet/buildtools/blob/7f31a1d788492615f03ab5b9337ed1c72028edd0/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh#L12-L13, so this could be changed to a positional arg in `init-tools.sh/cmd` (since it doesn't have "proper" arg handling) and an optional param in `bootstrap.sh/ps1` that passes it through.

PR #1170, opened 2016-10-27
Fix a couple minor bootstrap issues.

crummel:
- bootstrap.sh did not check for existence of the Shared Framework symlink before trying to create it, causing bootstrap to bomb out.
- When using a specific framework version instead of auto version, the shared framework path would be incorrect.

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -56,9 +56,9 @@ if ($LastExitCode -ne 0)
 
 # create a junction to the shared FX version directory. this is
 # so we have a stable path to dotnet.exe regardless of version.
+$runtimesPath = Join-Path $CliLocalPath "shared\Microsoft.NETCore.App"
 if ($SharedFrameworkVersion -eq "<auto>")
 {
-    $runtimesPath = Join-Path $CliLocalPath "shared\Microsoft.NETCore.App"
     $SharedFrameworkVersion = Get-ChildItem $runtimesPath -Directory | % { New-Object System.Version($_) } | Sort-Object -Descending | Select-Object -First 1
 }
 $junctionTarget = Join-Path $runtimesPath $SharedFrameworkVersion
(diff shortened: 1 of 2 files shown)


crummel:
@dagood for review please.

dagood:
I just found another that I was going to mention: the `force` logic is broken, so bootstrap always reinitializes:

```
force=false
...
if [[ $force && -f $bootstrapComplete ]]; then
```

The string "false" is never the null string, so the first condition's always true. (`man test`.) Two fix options, not sure which is preferred:
```
force=
...
if [[ $force && -f $bootstrapComplete ]]; then
```

```
force=false
...
if [[ $force = true && -f $bootstrapComplete ]]; then
```

bootstrap/bootstrap.sh
```diff
@@ -188,8 +188,8 @@ if [ $forcedCliLocalPath = "<none>" ]; then
     fi
 fi
 
+runtimesPath="$cliInstallPath/shared/Microsoft.NETCore.App"
 if [ $sharedFxVersion = "<auto>" ]; then
-    runtimesPath="$cliInstallPath/shared/Microsoft.NETCore.App"
```
dagood: Could you change `cliInstallPath` to the same as the (I think) equivalent powershell `CliLocalPath`?

crummel: Changed as suggested.

crummel:
Fixed the issue with --force and the rename, will merge after build passes.

bootstrap/bootstrap.ps1
```diff
@@ -67,7 +67,10 @@ if (-Not (Test-Path $junctionParent))
 {
     mkdir $junctionParent | Out-Null
 }
-cmd.exe /c mklink /j $SharedFrameworkSymlinkPath $junctionTarget | Out-Null
+if (-Not (Test-Path $SharedFrameworkSymlinkPath))
```
weshaggard: Are there cases were this symlink already exists but points to the wrong junction? I would think that is possible if it already exists. I wonder if we should always recreate?


Are there scenarios were we call this script multiple times? I thought the semaphore would block us from calling it multiple times.

crummel: Peter actually ran into this case (he was updating .cliversion in a client repo), I logged https://github.com/dotnet/buildtools/issues/1192 to take care of it.

PR #1173, opened 2016-10-28
Add APIs from httpapi.lib to the p/invoke list for OneCore.

ericeil:
HttpListener depends on these APIs.

@AlexGhiondea


stephentoub:
I assume we've verified these are in OneCore?  Assuming so, LGTM. Thanks.

ericeil:
I've verified that these are in Nano Server.  See https://msdn.microsoft.com/en-us/library/mt588480(v=vs.85).aspx#httpapi.dll.

PR #1174, opened 2016-10-28
Call non-prerelease builds "stable", not ""

dagood:
This avoids adding empty files to the versions repo when releasing stable builds. This flows to auto-update PR titles looking nicer and avoids a GitHub bug where accessing an empty file through the raw API for the first time causes a 503 response.

I don't think this is useful enough to take for the current release, but it will make prerelease => stable auto-PRs smoother in the future.

/cc @gkhanna79 @weshaggard


gkhanna79:
LGTM

PR #1175, opened 2016-10-31
Pickup newer Roslyn build

jkotas:


jkotas:
cc @jaredpar

jaredpar:
Only saw one reference to Roslyn here.  Based on previous conversations I was under the impression CoreFx at least was using two versions of roslyn: stable and latest.  If it's only one version it may complicate the ability to up take span changes.

jkotas:
I agree that we will need SxS fast-track Roslyn version for Span. This change is not fixing it. This is just general goodness update to collect millage on latest Roslyn beta.

I think that the best way to pick up the fast-track Roslyn version for the projects that need it is to skip buildtools and just download it directly from the tool setup script of repo that needs it. (I had it done that way earlier - it was not too bad.)

weshaggard:
cc @chcosta 

Looks good to me assuming we have attempted to build corefx and coreclr with this version?

jkotas:
Unfortunately, this breaks corefx build on a clean machine:

```
Microsoft.CSharp.Core.targets(71,5): error : System.IO.FileNotFoundException: Could not load file or assembly 'System.Security.Principal.Windows, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. [D:\corefx\src\mscorlib.WinRT-Facade\mscorlib.WinRT-Facade.csproj]
Microsoft.CSharp.Core.targets(71,5): error : File name: 'System.Security.Principal.Windows, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' [D:\corefx\src\mscorlib.WinRT-Facade\mscorlib.WinRT-Facade.csproj]
Microsoft.CSharp.Core.targets(71,5): error : at Microsoft.CodeAnalysis.CommandLine.DesktopBuildClient.GetPipeNameForPath(String compilerExeDirectory) [D:\corefx\src\mscorlib.WinRT-Facade\mscorlib.WinRT-Facade.csproj]
Microsoft.CSharp.Core.targets(71,5): error : at Microsoft.CodeAnalysis.CommandLine.DesktopBuildClient.RunServerCompilation(RequestLanguage language, List`1 arguments, BuildPaths buildPaths, String keepAlive, String libEnvVariable, CancellationToken cancellationToken) [D:\corefx\src\mscorlib.WinRT-Facade\mscorlib.WinRT-Facade.csproj]
Microsoft.CSharp.Core.targets(71,5): error : at Microsoft.CodeAnalysis.BuildTasks.ManagedCompiler.ExecuteTool(String pathToTool, String responseFileCommands, String commandLineCommands) [D:\corefx\src\mscorlib.WinRT-Facade\mscorlib.WinRT-Facade.csproj]
```

The problem is that Roslyn switched to building against netstandard1.X assemblies, but it does not bundle all required OOB facades. The list of missing OOB facades (may be incomplete):

- System.Security.Principal.Windows
- System.Runtime.InteropServices.RuntimeInformation
- System.Security.Cryptography.Algorithms
- System.Security.Cryptography.Primitives
- System.IO.Pipes

@jaredpar Would it be possible to fix this and re-spin the `2.0.0-beta5` package with the fix?

weshaggard:
We should be running the compiler against the shared framework which I believe contains all those (with the exception of System.IO.Pipes). Or is it running against desktop in this case?

jkotas:
Yes, most of the corefx build is running against desktop.

weshaggard:
>Yes, most of the corefx build is running against desktop.

That is something we should probably look into. Ideally it would run against core like we have to on other platforms. However I do realize that is difficult when people work in VS or in a VS command shell.

jaredpar:
> Yes, most of the corefx build is running against desktop.

In that case you are using the wrong NuGet package.  The compiler produces two packages:

- Microsoft.Net.Compilers: for running compilers on the desktop. 
- Microsoft.Net.Compilers.netcore: for running compilers on the shared framework.

The latter though is really geared towards consumption in CLI.  It bakes in a couple of assumptions about their infrastructure.  It can be a general purpose NuGet package but it's never been thoroughly tested as such.

CC @agocke to correct my mistakes above.

jkotas:
We are using the right package: Microsoft.Net.Compilers on Windows with desktop, and Microsoft.Net.Compilers.netcore on Unix with .NET Core. 

The problem that I have described is in Microsoft.Net.Compilers package. Download https://www.nuget.org/packages/Microsoft.Net.Compilers/2.0.0-beta5 and run ildasm on tools\Microsoft.Build.Tasks.CodeAnalysis.dll. You should see that it is referencing System.IO.Pipes that is nowhere to be found.

This problem is recent regression. It does not exist in 2.0.0-beta3. Also, it seems to be limited to compiler service that is disabled by default if you reference the compiler via nuget package. It may explain why you have not noticed it. However, disabling compiler service for corefx repo would have prohibitive impact on build throughput.

jaredpar:
> You should see that it is referencing System.IO.Pipes that is nowhere to be found.

Yeah @agocke saw that in parallel with this thread.  We've tracked down the problem and found a way to verify our NuSpec at CI time.  Going to get that implemented today.

> This problem is recent regression. It does not exist in 2.0.0-beta3

That's because in Beta3 we were still using my ultra-hacky reflection load for facades.  Now that we've deleted it and are using real facades we're seeing the gap in the NuSpec definition.  

Working on a solution.

jkotas:
I gave it another try with 2.0.0-rc Microsoft.Net.Compilers package. It works much better. There are two issues when trying to compile corefx repo with these compilers:

- Expected build breaks in ValueTuple tests: Cannot reference 'System.Runtime.CompilerServices.TupleElementNamesAttribute' explicitly. Use the tuple syntax to define tuple names.
- Bogus compiler error (https://github.com/dotnet/roslyn/issues/15553). Easy to workaround.

jkotas:
@weshaggard Is it good to merge? (making sure that it won't collide with other engineering work in corefx)

weshaggard:
If it works I'm good with it. It shouldn't interfere with the engineering work going on in corefx.

stephentoub:
> It shouldn't interfere with the engineering work going on in corefx.

@weshaggard, once this change is merged and consumed into corefx, any reason we can't start using available C# 7 features in any corefx project?

jaredpar:
@stephentoub yes please use C# 7.0 features 😄

weshaggard:
>@weshaggard, once this change is merged and consumed into corefx, any reason we can't start using available C# 7 features in any corefx project?

I'm not aware of any. There is potential issues when working in VS which might require a newer version but I'm not sure that is the case or not. Please do some basic testing and if it works in VS and CI passes go for it.

stephentoub:
> Please do some basic testing and if it works in VS and CI passes go for it.

It "works" in VS in that doing a build completes successfully.  But with VS2015, you still get red squigglies, as the older compiler appears to be used for the IDE.

jaredpar:
It's possible to eliminate the IDE squiggles by installing our insider preview VSIX.  Instructions are listed here:

https://github.com/dotnet/roslyn/blob/master/docs/contributing/Building,%20Debugging,%20and%20Testing%20on%20Windows.md

This will stop being a solution at some point in the future as we get closer to RTM and can't maintain our bits on VS2015 anymore.

PR #1176, opened 2016-10-31
Allow for ContractProject to be specified as item

### LINKED ISSUE #1041: ResolveContract.targets doesn't allow for specifying metadata like TargetGroup for ContractProject
/cc @JeremyKuhne @weshaggard

ericstj:
This allows folks to define metadata on that item like TargetGroup.

Fixes #1041 

I needed to fix this in order to make all the projects contributing to NETCoreApp as TargetGroup='' (/cc @ellismg @chcosta)

/cc @weshaggard @JeremyKuhne


ericstj:
I had to make one additional fix to set AdditionalProperties since we were consuming the item in ValidateApiCompatForSrc.

ericstj:
> Seems like we might need a way to flow this information automatically but for now I think what you have will do.

Yeah, this will look different in the future when we don't build netstandard refs for most of these... so I expect the distinction will be less important.

PR #1178, opened 2016-10-31
If you have a space inbetween args in your model.xml it doesn't work …

Maoni0:
…and the error is very non descriptive.


Maoni0:
@jkotas PTAL

PR #1182, opened 2016-11-01
Undo default for BuildTestsAgainstPackages

ericstj:
https://github.com/dotnet/buildtools/commit/2ba3679dbccd7ee0a79ae8d18fede36232417aff#diff-c368ca147f7a6d95f33280f334ff9e84R109
incorrectly turned on BuildTestsAgainstPackages (partially) for all
tests consuming BuildTools.  I understand that folks want that to be the
default for CoreFx but it should not be defined that way in buildtools.

/cc @weshaggard @karajas


ericstj:
Merging this since BuildTools consumption is blocked on it.

PR #1191, opened 2016-11-02
Persist version information to disk

alexperovich:
This change stops the build from running git commands during every build.
It stops the `git rev-parse` and `git show` commands by persisting their output to a file in the bin directory.

/cc: @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -347,7 +347,7 @@
     </PropertyGroup>
     
     <!-- Since by default the file will get dropped at the obj dir, make sure that the dir is created already or else WriteLinesToFile will error. -->
-    <MakeDir Condition="!Exists('$(BuildVersionFilePath)')" Directories="$(BuildVersionFilePath)" />
+    <MakeDir Directories="$(BuildVersionFilePath)" />
       
     <WriteLinesToFile
       ContinueOnError="WarnAndContinue"
@@ -366,6 +366,7 @@
     </ItemGroup>
 
     <PropertyGroup>
+      <!-- Redefine AssemblyFileVersion because we just computed it. -->
       <AssemblyFileVersion>$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)</AssemblyFileVersion>
     </PropertyGroup>
   </Target>
@@ -373,8 +374,6 @@
   <!-- This target will only be executed if BuildVersion.props doesn't exist yet -->
   <Target Name="CreateVersionFileDuringBuild" Condition="'$(SkipVersionGeneration)'!='true' AND '$(ShouldCreateVersionFileDuringBuild)'=='true'" DependsOnTargets="CreateOrUpdateCurrentVersionFile">
     <PropertyGroup Condition="'$(SkipVersionGeneration)'!='true'">
-      <BuildNumberMajor >$(GeneratedBuildNumberMajor)</BuildNumberMajor>
-      <BuildNumberMinor>$(GeneratedBuildNumberMinor)</BuildNumberMinor>
       <VersionSuffix Condition="'$(PreReleaseLabel)' != ''">-$(PreReleaseLabel)</VersionSuffix>
       <VersionSuffix Condition="'$(IncludeBuildNumberInPackageVersion)' == 'true'">$(VersionSuffix)-$(BuildNumberMajor)-$(BuildNumberMinor)</VersionSuffix>
     </PropertyGroup>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -42,7 +42,7 @@
       Inputs="$(MSBuildProjectFile)"
       Outputs="$(AssemblyInfoFile)"
       Condition="'$(GenerateAssemblyInfo)'=='true'"
-      DependsOnTargets="CreateVersionFileDuringBuild;GetLatestCommitHash">
+      DependsOnTargets="CreateVersionFileDuringBuild">
 
     <Error Condition="!Exists('$(IntermediateOutputPath)')" Text="GenerateAssemblyInfo failed because IntermediateOutputPath isn't set to a valid directory" />
 
@@ -124,7 +124,7 @@
   <Target Name="GenerateVersionHeader"
       Inputs="$(MSBuildProjectFile)"
       Outputs="$(NativeVersionHeaderFile)"
-      DependsOnTargets="CreateVersionFileDuringBuild;GetLatestCommitHash"
+      DependsOnTargets="CreateVersionFileDuringBuild"
       Condition="'$(NativeVersionHeaderFile)'!='' and '$(GenerateVersionHeader)'=='true' and !Exists($(NativeVersionHeaderFile))">
 
     <ItemGroup>
@@ -182,7 +182,7 @@
   <Target Name="GenerateVersionSourceFile"
       Inputs="$(MSBuildProjectFile)"
       Outputs="$(NativeVersionSourceFile)"
-      DependsOnTargets="CreateVersionFileDuringBuild;GetLatestCommitHash"
+      DependsOnTargets="CreateVersionFileDuringBuild"
       Condition="'$(NativeVersionSourceFile)'!='' and '$(GenerateVersionSourceFile)'=='true'">
 
     <ItemGroup>
@@ -204,17 +204,6 @@
     </ItemGroup>
   </Target>
 
-  <Target Name="GetLatestCommitHash" Condition="'$(LatestCommit)'==''">
-    <Exec Command="git rev-parse HEAD 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
-      <Output TaskParameter="ConsoleOutput" PropertyName="LatestCommit" />
-      <Output TaskParameter="ExitCode" PropertyName="LatestCommitExitCode" />
-    </Exec>
-    <!-- We shouldn't fail the build if we can't retreive the commit hash, so in this case just set it to N/A -->
-    <PropertyGroup Condition="'$(LatestCommitExitCode)'!='0'">
-      <LatestCommit>N/A</LatestCommit>
-    </PropertyGroup>
-  </Target>
-
   <PropertyGroup>
     <GetNuGetPackageVersionsDependsOn>$(GetNuGetPackageVersionsDependsOn);CreateVersionInfoFile</GetNuGetPackageVersionsDependsOn>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -16,6 +16,68 @@
   <Import Project="$(BuildNumberTarget)" Condition="Exists('$(BuildNumberTarget)')" />
 
   <!-- #################################### -->
+  <!-- Version Info -->
+  <!-- #################################### -->
+
+  <ItemGroup>
+    <VersionInfoFile Include="$(BinDir)VersionInfo.json"/>
```
weshaggard: If you were going to go as far as writing it to a file you should just write it to a props file that can be directly imported.

joperezr: ><VersionInfoFile Include="$(BinDir)VersionInfo.json"/> [](start = 4, length = 54)

One problem that this has is that if people  don't clean their bin\ dir often the version will be just wrong and outdated most of the time.

alexperovich: It will be rewritten on a Clean/Rebuild because it is added to FileWrites. If the versions are slightly off locally that isn't a problem.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -317,27 +364,6 @@
...
-      <Output TaskParameter="ExitCode" PropertyName="GitCommandExitCode" />
-    </Exec>
-    <PropertyGroup Condition="'$(GitCommandExitCode)'!='0'">
-      <VersionSeedDate>%(VersionTargetsFile.ModifiedTime)</VersionSeedDate>
-      <VersionSeedSourceComment>VersionSeedDate was produced by getting the timestamp of versioning.targets.</VersionSeedSourceComment>
-    </PropertyGroup>
-  </Target>
-
   <!-- This target will only be executed if BuildVersion.props doesn't exist yet -->
   <Target Name="CreateVersionFileDuringBuild" Condition="'$(SkipVersionGeneration)'!='true' AND '$(ShouldCreateVersionFileDuringBuild)'=='true'" DependsOnTargets="CreateOrUpdateCurrentVersionFile">
```
weshaggard: We should consider putting the commit sha in this versions prop file we generate and so we can use the same mechanism for invalidating it.

alexperovich: The commit sha in the VersionInfo file isn't being used to invalidate it. The file is "invalidated" when it gets deleted in a clean. I can't see the value in putting the commit sha in the BuildVersion.props file.

weshaggard: While I agree it is loosely coupled the build version props file gets invalidated whenever you clean or roll over to the next day so it gets invalidated more often then this new VersionInfo.props file would. My preference would still be to combine them.

src/Microsoft.DotNet.Build.Tasks/GenerateVersionInfoFile.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                string path = InfoFile.GetMetadata("FullPath");
+                string directory = Path.GetDirectoryName(path);
+                if (!Directory.Exists(directory))
+                {
+                    Directory.CreateDirectory(directory);
+                }
+                File.WriteAllLines(path, new []
+                {
+                    @"<?xml version=""1.0"" encoding=""utf-8""?>",
```
weshaggard: Any reason you need a custom task to do this? You can just call WriteLinesToFile task like the other props file https://github.com/dotnet/buildtools/pull/1191/files#diff-09dfec4a89fd90309644b01711f96159R307.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -312,29 +321,52 @@
...
+    <LatestCommit Condition="'%24(LatestCommit)' == ''">$(LatestCommit)</LatestCommit>
+    <BuiltByString Condition="'%24(BuiltByString)' == ''">$(BuiltByString)</BuiltByString>
+  </PropertyGroup>
+</Project>
+]]>
+      </VersionFileContent>
     </PropertyGroup>
+    
+    <!-- Since by default the file will get dropped at the obj dir, make sure that the dir is created already or else WriteLinesToFile will error. -->
+    <MakeDir Condition="!Exists('$(BuildVersionFilePath)')" Directories="$(BuildVersionFilePath)" />
```
ericstj: You don't need the condition, typically folks leave it out and the task has a fast path.  That's the pattern followed by common targets.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -225,7 +214,7 @@
   </PropertyGroup>
   
   <Target Name="CreateVersionInfoFile"
-          DependsOnTargets="GetLatestCommitHash"
+          DependsOnTargets="CreateVersionFileDuringBuild"
           BeforeTargets="BuildAllProjects"
           Inputs="$(LatestCommit)"
```
ericstj: Does this actually do anything?  It looks to me like $(LatestCommit) is a hash.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -238,70 +227,90 @@
...
+    Outputs:
+      $(LatestCommit)
+      $(BuiltByString)
+      $(BuildNumberMajor)
+      $(BuildNumberMinor)
+      $(AssemblyFileVersion)
+  -->
+  <Target Name="CreateOrUpdateCurrentVersionFile"
+      BeforeTargets="ResolveProjectReferences"
+      Condition="'$(SkipVersionGeneration)' != 'true' AND '$(VersionPropsImported)' != 'true'">
```
ericstj: If I understand correctly this condition: `'$(VersionPropsImported)' != 'true'` is the case where the file exists.  If that is true, when will you ever "Update" the Version file?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -312,29 +321,52 @@
...
+    <ItemGroup>
+      <OldBuildVersionFiles Include="$(BuildVersionFilePath)BuildVersions-*.props" Exclude="$(BuildVersionFile)" />
+    </ItemGroup>
+    <Delete Files="@(OldBuildVersionFiles)" TreatErrorsAsWarnings="true"/>
+    <ItemGroup>
+      <OldBuildVersionFiles Remove="@(OldBuildVersionFiles)"/>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <AssemblyFileVersion>$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)</AssemblyFileVersion>
```
ericstj: Can you add a comment here that says we need to redefine AssemblyFileVersion because we just calculated it?
Have you looked for any other derived properties that need to be reset?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -343,7 +375,6 @@
     <PropertyGroup Condition="'$(SkipVersionGeneration)'!='true'">
       <BuildNumberMajor >$(GeneratedBuildNumberMajor)</BuildNumberMajor>
```
ericstj: Won't this clear BuildNumberMajor from the imported props file?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
alexperovich: It may only do something in the case where a file happens to have the same name as the commit hash. I can remove it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
alexperovich: Ill remove it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
alexperovich: This only runs when the file doesn't exist. Though these are now redundant. I will remove them.

ericstj: What's the mechanism that makes it only run when the file doesn't exist?  Actually I don't see anywhere that ShouldCreateVersionFileDuringBuild is set, so maybe it was never even running.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
ericstj: What about VersionSuffix?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
ericstj: Nevermind, I see it here: https://github.com/alexperovich/buildtools/blob/daac863e9ea367338d6295c42264a58f4eb5bf1e/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets

alexperovich:
@ericstj I have compared diag logs before and after for System.Runtime and there are not any differences in the generated assembly info files.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -352,6 +382,6 @@
   <Target Name="NativeResourceCompile" DependsOnTargets="$(BeforeResourceCompileTargets)" Inputs="$(MsBuildThisFileDirectory)NativeVersion.rc" Outputs="$(Win32Resource)">
     <Error Condition="!Exists('$(RCPATH)')" Text="NativeResourceCompile failed because RCPath is set to an non-existing rc.exe path." />
 
-    <Exec Command="&quot;$(RCPath)&quot; /i $(BaseIntermediateOutputPath) /i $(IntermediateOutputPath) /i $(WindowsSDKPath)\inc /i $(VCSDKPath)\Include /D _UNICODE /D UNICODE /l&quot;0x0409&quot; /r /fo &quot;$(Win32Resource)&quot; &quot;$(MsBuildThisFileDirectory)NativeVersion.rc&quot;" />
+    <Exec Command='"$(RCPath)" /i $(BaseIntermediateOutputPath) /i $(IntermediateOutputPath) /i $(WindowsSDKPath)\inc /i $(VCSDKPath)\Include /D _UNICODE /D UNICODE /l"0x0409" /r /fo "$(Win32Resource)" "$(MsBuildThisFileDirectory)NativeVersion.rc"' />
```
alexperovich: Not really any reason other that &quot; looks bad and is more characters.

alexperovich:
@dotnet-bot test this please

alexperovich:
@dotnet-bot test this please

alexperovich:

PR #1197, opened 2016-11-04
Use Version Metadata in buildagainstpackages if specified and remove AssemblyVersion

karajas:
Ensure version returned contains prerelease label.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -374,21 +374,22 @@ namespace Microsoft.DotNet.Build.Tasks
                 // Don't add a new dependency if one already exists.
                 if (!returnDependenciesList.ContainsKey(name))
                 {
-                    NuGetVersion nuGetVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
-                    PackageItem packageItem = new PackageItem(name, nuGetVersion);
-                    bool? forceVersion = dependency.GetMetadata("ForceVersion")?.Equals("true", StringComparison.OrdinalIgnoreCase);
-                    string version = packageItem.GetVersionString();
+                    string versionFromMetadata = dependency.GetMetadata("Version");
+                    NuGetVersion nuGetVersion = null;
+                    if (versionFromMetadata != string.Empty)
+                    {
+                        nuGetVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
+                    }
 
                     // a package version was provided, use its version information.
-                    if (packageInformation.ContainsKey(name) && (bool) !forceVersion)
+                    if (packageInformation.ContainsKey(name))
                     {
-                        version = packageInformation[name].Version.ToString();
+                        nuGetVersion = packageInformation[name].Version;
                     }
+                    PackageItem packageItem = new PackageItem(name, nuGetVersion);
+                    string version = packageItem.GetVersionString();
+
                     JProperty property = new JProperty(name, version);
-                    if (forceVersion == true && framework != null)
-                    {
-                        continue;
-                    }
                     returnDependenciesList.Add(name, property);
                 }
                 else
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -376,14 +376,19 @@ namespace Microsoft.DotNet.Build.Tasks
                 {
                     NuGetVersion nuGetVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
                     PackageItem packageItem = new PackageItem(name, nuGetVersion);
+                    bool? forceVersion = dependency.GetMetadata("ForceVersion")?.Equals("true", StringComparison.OrdinalIgnoreCase);
                     string version = packageItem.GetVersionString();
 
                     // a package version was provided, use its version information.
-                    if (packageInformation.ContainsKey(name))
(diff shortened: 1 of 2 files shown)


karajas:
@weshaggard PTAL.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -358,14 +358,19 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
                 {
                     NuGetVersion nuGetVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
                     PackageItem packageItem = new PackageItem(name, nuGetVersion);
+                    bool? forceVersion = dependency.GetMetadata("ForceVersion")?.Equals("true", StringComparison.OrdinalIgnoreCase);
                     string version = packageItem.GetVersionString();
 
                     // a package version was provided, use its version information.
-                    if (packageInformation.ContainsKey(name))
+                    if (packageInformation.ContainsKey(name) && forceVersion!=true)
```
weshaggard: I think I would prefer `!forceVersion` as it is a little clearer.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -358,14 +358,19 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
                 {
                     NuGetVersion nuGetVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
                     PackageItem packageItem = new PackageItem(name, nuGetVersion);
+                    bool? forceVersion = dependency.GetMetadata("ForceVersion")?.Equals("true", StringComparison.OrdinalIgnoreCase);
```
weshaggard: I'm not sure I fully understand what this is going to do. It isn't clicking in my head where ForceVersion will show-up and how it actually applies to the rewritting. @chcosta should also look at this change.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -458,7 +463,7 @@ public NuGetVersion Version
 
         public string GetVersionString()
         {
-            return string.Join(".", _version.Major, _version.Minor, _version.Patch);
```
chcosta: Do you no longer require a 3-part version string?  older versions of NuGet used a 4 part version string.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: What are you forcing the version to be and why?  ie, why would you not want to use the version provided by a package drop or version file and would instead prefer the assembly version?

chcosta: I see, this is related to https://github.com/dotnet/corefx/pull/13128/files...

chcosta: I wonder if it makes more sense to remove https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets#L132 and instead always use "Version" if it is present.

karajas: This was part of the workaround to get all the runtime dependencies with the right version (LKG). I'm ok with using Version if present if that's cleaner.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
karajas: I wanted to include the prerelease label. Will do some more testing to make sure we don't break.

karajas:
@chcosta I wanted an explicit way to test against packages that aren't built locally and require the use of an LKG. Currently the generated project.json will always consume the live version of a package with all live subdependencies. This makes sense in the context of build pipeline since we build each leg and publish and then consume these changes. Locally we want to use the LKG version. This seemed like the cheapest way around this issue for the generated proejct.json.

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
chcosta: I'd prefer to use Version if it's present.

weshaggard:
I believe with our current plans we can close this PR now.

karajas:
@weshaggard @chcosta Updated this to have the regex fixes

src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -374,15 +374,21 @@ private JObject GenerateDependencies(JObject projectJsonRoot, ITaskItem[] extern
...
                 if (!returnDependenciesList.ContainsKey(name))
                 {
-                    NuGetVersion nuGetVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
-                    PackageItem packageItem = new PackageItem(name, nuGetVersion);
-                    string version = packageItem.GetVersionString();
+                    string versionFromMetadata = dependency.GetMetadata("Version");
+                    NuGetVersion nuGetVersion = null;
+                    if (versionFromMetadata != string.Empty)
+                    {
+                        nuGetVersion = NuGetVersion.Parse(dependency.GetMetadata("Version"));
```
weshaggard: dependency.GetMetadata("Version") -> versionFromMetadata

karajas:
Abandoning this for now. Will cleanup later.

Issue #1201, opened 2016-11-07
Init-tools `find` command failing on Fedora and Alpine


ericstj:
https://github.com/dotnet/buildtools/commit/2d3d95068daf1b3edb6f7eb7001c6030bc23f4f3 introduced a dependency on `find` which isn't working on Fedora or Alpine.

On alpine this is failing as follows:
```
2016-11-07T14:03:47.4123920Z find: unrecognized: -iwholename
2016-11-07T14:03:47.4149380Z BusyBox v1.24.2 (2016-08-12 14:38:34 GMT) multi-call binary.
2016-11-07T14:03:47.4153070Z 
2016-11-07T14:03:47.4159280Z Usage: find [-HL] [PATH]... [OPTIONS] [ACTIONS]
2016-11-07T14:03:47.4162980Z 
2016-11-07T14:03:47.4168500Z Search for files and perform actions on them.
2016-11-07T14:03:47.4174260Z First failed action stops processing of current file.
2016-11-07T14:03:47.4180570Z Defaults: PATH is current directory, action is '-print'
2016-11-07T14:03:47.4184140Z 
2016-11-07T14:03:47.4191390Z 	-L,-follow	Follow symlinks
2016-11-07T14:03:47.4198250Z 	-H		...on command line only
2016-11-07T14:03:47.4205290Z 	-xdev		Don't descend directories on other filesystems
2016-11-07T14:03:47.4212470Z 	-maxdepth N	Descend at most N levels. -maxdepth 0 applies
2016-11-07T14:03:47.4218660Z 			actions to command line arguments only
2016-11-07T14:03:47.4224820Z 	-mindepth N	Don't act on first N levels
2016-11-07T14:03:47.4231920Z 	-depth		Act on directory *after* traversing it
2016-11-07T14:03:47.4235480Z 
2016-11-07T14:03:47.4240520Z Actions:
2016-11-07T14:03:47.4246230Z 	( ACTIONS )	Group actions for -o / -a
2016-11-07T14:03:47.4252800Z 	! ACT		Invert ACT's success/failure
2016-11-07T14:03:47.4260510Z 	ACT1 [-a] ACT2	If ACT1 fails, stop, else do ACT2
2016-11-07T14:03:47.4267510Z 	ACT1 -o ACT2	If ACT1 succeeds, stop, else do ACT2
2016-11-07T14:03:47.4274660Z 			Note: -a has higher priority than -o
[...]

PR #1202, opened 2016-11-07
Simplify case-insensitive target file copy in init-tools.sh

dagood:
No longer depend on `find`, which isn't available in our Fedora builds, and `-iwholename`, which isn't available in our Alpine builds. Allow lowercased package ids using character groups in the path glob.

Undoes the way https://github.com/dotnet/buildtools/pull/1114 solved the problem, which added full insensitivity. We only need to allow for proper-cased and lowercased paths, not any case. I also found `shopt -s nocaseglob` which could help in any other cases if real insensitivity is needed.

Should fix https://github.com/dotnet/buildtools/issues/1201. I haven't tested the full init-tools flow yet.

@ericstj
/cc @ellismg @chcosta @mellinoe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -98,10 +98,10 @@ if [ "$?" != "0" ]; then
     echo "ERROR: An error ocurred when running: '$__DOTNET_CMD restore \"${__PORTABLETARGETS_PROJECTJSON}\"'. Please check above for more details."
     exit 1
 fi
-find $__PACKAGES_DIR -type d \( \
-    -iwholename "${__PACKAGES_DIR}/Microsoft.Portable.Targets/${__PORTABLETARGETS_VERSION}/contentFiles/any/any" -or \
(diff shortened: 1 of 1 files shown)


ericstj:
LGTM, but I'm by no means an expert here.

ellismg:
LGTM as well.

dagood:
Tried it out with an actual built buildtools package, and it looks like the quotes or the variable force the path to be treated literally, so it doesn't work. Working on it.

dagood:
Got it--had to unquote the glob part to avoid the shell taking it literally. Left the rest quoted in case there are spaces in the path.

dagood:
@ericstj Should I revert https://github.com/dotnet/buildtools/pull/1200 so that the build fails early if this still doesn't work?

ericstj:
Sure, you can.  That's a bit secondary though.  I'd prefer that the builds fail in init-tools before we get to the point of calling msbuild.

dagood:
Agreed. Remembered that `bootstrap.sh` has some interesting things it sets that would fix this: https://github.com/dotnet/buildtools/blob/38637b1efad73189f4d8fb0bcc09ce089f306798/bootstrap/bootstrap.sh#L3-L9. Specifically `set -e` (exit on nonzero error code, rather than returning error code of last command to run), but I think we should just include the whole block.

PR #1206, opened 2016-11-08
Use "set -e" rather than catching NZECs manually

dagood:
Copy over "strict" settings from `bootstrap.sh` to `init-tools.sh`. The main advantage is that the script will now exit as soon as any command gets a non-zero exit code (NZEC), so `init-tools.sh` can't silently fail like it did for #1201.

Tested on Alpine docker container.

See https://github.com/dotnet/buildtools/pull/1202.

@ericstj @ellismg @crummel 
/cc @chcosta


dagood:
Ping--do the "strict" settings here seem reasonable?

ellismg:
LGTM.  Sorry for the delay.  Thanks for this improvement!

PR #1210, opened 2016-11-09
Automatically detect non-Windows test RIDs

mellinoe:
This adds an additional output to the `GetTargetMachineInfo` task: the nuget runtime ID of the machine running the build. We use this rid as the default test rid for x-plat builds. This means that corefx can stop using `ubuntu.14.04-x64` as the default rid for all Linux distros, and instead just rely on a reasonable default being available.

@weshaggard @Priya91 

Related to https://github.com/dotnet/corefx/issues/12461

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -281,6 +281,19 @@
             Files="$(TestsSuccessfulSemaphore)" />
   </Target>
 
+  <Target Name="GetDefaultTestRid">
+    <GetTargetMachineInfo>
+      <Output TaskParameter="RuntimeIdentifier" PropertyName="DefaultTestNugetRuntimeId" />
+    </GetTargetMachineInfo>
+    <!-- On Windows, we always use win7-x64 as the default test RID because the build context,
+         usually 32-bit full-framework MSBuild, is not a good default test context. -->
+    <PropertyGroup>
+      <DefaultTestNugetRuntimeId Condition="$(DefaultTestNugetRuntimeId.StartsWith('win'))">win7-x64</DefaultTestNugetRuntimeId>
+      <TestArchitecture Condition="'$(TestArchitecture)' == ''">x64</TestArchitecture>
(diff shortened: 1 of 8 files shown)


mellinoe:
I needed to move where the default values are defined for `TestArchitecture` and `TestNugetRuntimeId` as part of this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -281,6 +281,19 @@
...
   </Target>
 
+  <Target Name="GetDefaultTestRid">
+    <GetTargetMachineInfo>
+      <Output TaskParameter="RuntimeIdentifier" PropertyName="DefaultTestNugetRuntimeId" />
+    </GetTargetMachineInfo>
+    <!-- On Windows, we always use win7-x64 as the default test RID because the build context,
+         usually 32-bit full-framework MSBuild, is not a good default test context. -->
+    <PropertyGroup>
+      <DefaultTestNugetRuntimeId Condition="$(DefaultTestNugetRuntimeId.StartsWith('win'))">win7-x64</DefaultTestNugetRuntimeId>
```
Priya91: why not win10-x64?

mellinoe: * win10-x64 isn't really valid to use on downlevel OS's (8.1, 8, 7)
* win7-x64 is what we are using before the change
* There's no material difference between the two even aside from the above two points. We don't have different assets between win7-x64 and win10-x64 for the scenarios that this default will actually affect.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -281,6 +281,19 @@
...
+      <TestArchitecture Condition="'$(TestArchitecture)' == ''">x64</TestArchitecture>
```
weshaggard: It doesn't look like we are even using the TestArchitecture any more. I wonder if that should be an input the getting the default and thus in can override? Or perhaps we should get the OS and arch as separate parameters and combine them here?

PR #1214, opened 2016-11-09
Automatically detect TestNugetRuntimeId

mellinoe:
Re-submitting #1210 with slight changes. This fixes an issue in the `GetApplicableAssetsFromPackages` task if no target runtime is passed in.

@weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -281,6 +281,23 @@
             Files="$(TestsSuccessfulSemaphore)" />
   </Target>
 
+  <Target Name="GetDefaultTestRid">
+    <GetTargetMachineInfo>
+      <Output TaskParameter="RuntimeIdentifier" PropertyName="DefaultTestNugetRuntimeId" />
+    </GetTargetMachineInfo>
+    <!-- On Windows, we always use win7-x64 as the default test RID because the build context,
+         usually 32-bit full-framework MSBuild, is not a good default test context. -->
+    <PropertyGroup>
+      <DefaultTestNugetRuntimeId Condition="$(DefaultTestNugetRuntimeId.StartsWith('win'))">win7-x64</DefaultTestNugetRuntimeId>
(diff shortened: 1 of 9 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -281,6 +281,23 @@
...
+         usually 32-bit full-framework MSBuild, is not a good default test context. -->
+    <PropertyGroup>
+      <DefaultTestNugetRuntimeId Condition="$(DefaultTestNugetRuntimeId.StartsWith('win'))">win7-x64</DefaultTestNugetRuntimeId>
+      <TestArchitecture Condition="'$(TestArchitecture)' == ''">x64</TestArchitecture>
+      <TestNugetRuntimeId Condition="'$(TestNugetRuntimeId)' == ''">$(DefaultTestNugetRuntimeId)</TestNugetRuntimeId>
+    </PropertyGroup>
+  </Target>
+
+  <PropertyGroup>
+    <ResolvePkgProjReferencesDependsOn>GetDefaultTestRid;$(ResolvePkgProjReferencesDependsOn)</ResolvePkgProjReferencesDependsOn>
```
weshaggard: I'm not sure which dependency direction I like better but is there a reason you decided to depend on ResolvePkgProjReferencesDependsOn here as opposed to depending on GetDefaultTestRid from the other direction? I think I would have leaned the other way but I don't think it matters too much.

mellinoe: Could you clarify what you mean? I'm adding to the `ResolvePkgProjReferencesDependsOn` list here.

weshaggard: I think I would have made this change over https://github.com/dotnet/buildtools/blob/bd11b84d8d7af2c599d52d51a150c4d06e7931b2/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets#L28 instead so that it is clearer that ResolvePkgProjReferences depends on this as opposed to having to keep track of all the various tasks that might depend on it in this file. I think it is minor but I like to think of the dependency from the other direction.

PR #1215, opened 2016-11-10
Hook GetDefaultTestRid before BeforeResolveReferences

mellinoe:
Many of our test projects use P2P references to pkgproj's. In order for assets to be resolved correctly out of them for the test run, we need the `TestNugetRuntimeId` to already be set. So I've hooked the new target to before those project references are resolved so that the property is already set when they are built.

@weshaggard @karajas

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
@@ -25,7 +25,7 @@
   <PropertyGroup>
     <!-- Resolve from pkgProjs before nuget packages -->
     <ResolveNugetPackagesDependsOn>$(ResolveNugetPackagesDependsOn);ResolvePkgProjReferences</ResolveNugetPackagesDependsOn>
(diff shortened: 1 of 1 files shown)


ericstj:
The change looks OK to me.  These sequencing problems usually require more testing rather than just code review, so assuming you've tested and it does the right thing I'm OK with it.

ericstj:
Per our offline conversation the reason this wasn't working before is because you were setting ResolvePkgProjReferencesDependsOn, but that was getting overwritten by https://github.com/dotnet/buildtools/blob/bd11b84d8d7af2c599d52d51a150c4d06e7931b2/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets#L28.

So you could either make this change you have here, or fix PackageLibs.targets to append to the ResolvePkgProjReferencesDependsOn property rather than just setting it.  I tend to prefer DependsOn because it is more resilient to the sequence changing.

mellinoe:
I did this:

> fix PackageLibs.targets to append to the ResolvePkgProjReferencesDependsOn property rather than just setting it

but I'm still testing it on corefx.

mellinoe:
Alright, I verified that it fixes at least one of the problematic projects in corefx. All the other projects should be fixed too, but I'm going to run a full build now and merge this afterwards.

mellinoe:
Okay, full windows build succeeded. Merging this in.

PR #1219, opened 2016-11-11
Fix message to be accurate

danmosemsft:
Space does not dismiss a Console.ReadLine()

### DIFF
--- a/src/Run/CommandLine.cs
+++ b/src/Run/CommandLine.cs
@@ -1651,7 +1651,7 @@ class CommandLine
         }
         /// <summary>
         /// prints a string to the console in a nice way.  In particular it
-        /// displays a screen full of data and then as user to type a space to continue.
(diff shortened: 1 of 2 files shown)


src/Run/CommandLine.cs
```diff
@@ -1675,7 +1675,7 @@ private static void DisplayStringToConsole(string helpString)
                 Console.Write(helpString.Substring(pos, nextPos - pos));
                 if (nextPos == helpString.Length)
                     break;
-                Console.Write("[Press space to continue...]");
+                Console.Write("[Press Enter to continue...]");
                 Console.Read();
```
zabulus: Here is ```Console.Read``` not ```Console.ReadLine```

danmosemsft: Thank you. Same change though.

danmosemsft:
@joperezr ok?

PR #1222, opened 2016-11-11
Add a GenFacades MSBuild task

mellinoe:
This adds a build task which directly invokes the core logic of GenFacades, without launching a separate process. This was accomplished with the following steps:

* Split the core logic of GenFacades into a new assembly, "GenFacades.Core". This contains the same logic of the old assembly, minus the argument-parsing code. Arguments are passed in in an already-processed format.
* Add a Console application, called GenFacades.exe for compatibility. This contains the stripped-out argument parsing from the original assembly, and then calls into GenFacades.Core. This program is identical in interface and functionality to the program prior to this change.
* Add an MSBuild task "GenFacadesTask", which simply calls into GenFacades.Core, passing arguments from the Task's properties.
* Modify partialfacades.targets to handle invoking GenFacades either via the new build task, or via the old executable, based on whether `UseLegacyGenFacadesExe` is set to `true`.

I haven't re-profiled this yet, but my previous attempt showed us spending 1/3 to 1/5th the time in GenFacades after the change.

A caveat:

* GenFacades uses Microsoft.Cci, which only targets `netstandard1.3`. This means that we cannot use it in projects targeting `net45`, and therefore couldn't use it in our current build tasks assembly. This limitation has historically existed because we ran our TFS builds in a special environment did not support `net46`. As part of this change, I've retargeted our "desktop" tool-runtime to `net46`, retargeted the build tasks assembly, and made GenFacades.Core target `netstandard1.3` (which supports `net46`). This means our build is no longer guaranteed to work in that internal environment anymore.

@weshaggard @ericstj 

We can hold off on this until we are at a stable enough point to take the churn, if we feel it's necessary.


src/GenFacades.Desktop/GenFacades.Desktop.csproj
```diff
@@ -1,54 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
weshaggard: I do think we want a version of GenFacades that runs on desktop.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -28,9 +28,13 @@
         "portable-net45+win8"
       ]
     },
-    "net45": {
+    "net46": {
```
weshaggard: If we are going to move to net46 I would like to update the packages and targets that currently reference net45 to be updated to net46 now so we have consistency.

mellinoe: I changed the nuget package's folder name and adjusted build targets to account for it. I haven't re-targeted all of our desktop assets to net46 unless there was a need, but we can do that in the future without problems now.

mellinoe:
I also added a desktop project for GenFacades. The corefx changes that will be needed to consume this are here:

https://github.com/mellinoe/corefx/commit/244ccf4b12b07f481e3cda594a291067a0fdd3c9

ericstj:
LGTM pending mirror being stopped.

src/GenFacades/GenFacades.Console/GenFacades.Console.net46.csproj
```diff
@@ -4,25 +4,23 @@
 
   <PropertyGroup>
```
weshaggard: Is keeping the net46 version in the same directory going to cause any issues? Hopefully not but before were were keeping it in a separate directory similar to all the other projects.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -28,7 +28,7 @@
     <ToolsDir Condition="'$(ToolsDir)'==''">$(MSBuildThisFileDirectory)</ToolsDir>
     <DotnetCliPath Condition="'$(DotnetCliPath)'==''">$(ToolRuntimePath)dotnetcli/</DotnetCliPath>
     <OverrideToolHost>$(DotnetCliPath)dotnet</OverrideToolHost>
-    <BuildToolsTaskDir Condition="'$(BuildToolsTargets45)' == 'true'">$(ToolsDir)net45/</BuildToolsTaskDir>
+    <BuildToolsTaskDir Condition="'$(BuildToolsTargetsDesktop)' == 'true'">$(ToolsDir)net46/</BuildToolsTaskDir>
```
weshaggard: As an FYI changing this to net46 is going to likely cause issues with the next set of folks picking up this change. If we find any people hitting that instead of fixing that spot I think we should remove it and have it use the settings from this common props.

Also it might be worth defaulting BuildToolsTargetsDesktop to true if BuildToolsTargets45 is set to help ease the transition. See https://github.com/dotnet/corefx/blob/master/dir.props#L30 and I assume the same is in coreclr/wcf repos.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -9,7 +9,7 @@ if [%PACKAGES_DIR%] == [] set PACKAGES_DIR=%TOOLRUNTIME_DIR%
 :: Remove quotes to the packages directory
 set PACKAGES_DIR=%PACKAGES_DIR:"=%
 IF [%BUILDTOOLS_TARGET_RUNTIME%]==[] set BUILDTOOLS_TARGET_RUNTIME=win7-x64
-IF [%BUILDTOOLS_NET45_TARGET_RUNTIME%]==[] set BUILDTOOLS_NET45_TARGET_RUNTIME=win7-x86
```
weshaggard: Any idea where this gets set?

src/GenFacades/GenFacades.Console/GenFacades.Console.net46.csproj
mellinoe: I don't think it should be problematic. The output gets put into the "GenFacades.Console.net46" folder, and the project.json in the directory supports both targets. So it should be okay.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
mellinoe: I don't think this gets set anywhere, to be honest. I think it always takes this default value.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <RoslynPropsFile Condition="'$(RoslynPropsFile)' == '' and '$(OsEnvironment)' != 'Unix'">$(ToolRuntimePath)/net45/roslyn/build/Microsoft.Net.Compilers.props</RoslynPropsFile>
+    <RoslynPropsFile Condition="'$(RoslynPropsFile)' == '' and '$(OsEnvironment)' != 'Unix'">$(ToolRuntimePath)/net46/roslyn/build/Microsoft.Net.Compilers.props</RoslynPropsFile>
```
mellinoe: Hopefully. I'll try it out.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
weshaggard: Yes I think it is worth setting this default in buildtools, but still allowing individual repos to override it as necessary because I do think we will want to allow them to not run on desktop in some cases.

mellinoe:
Added a couple more cleanup commits. Now we shouldn't need any changes to other repos except in places where they were passing custom command-line arguments to GenFacades. And that is only corefx, as far as I know.

weshaggard:
> And that is only corefx, as far as I know.

The dotnet/standard repro is also passing "-ignoreMissingTypes" but that isn't a big deal I will take of it when we update buildtools there.

mellinoe:
@weshaggard I rebased this just now; there were no conflicts since most everything was contained to new targets. Do you have any objections to getting this in now? I'm still seeing that this results in the GenFacades portions of the build completing in about 1/3 of the previous time. I have a branch in corefx ready to consume this as well: https://github.com/mellinoe/corefx/commit/1f6a4e910684fe0c92d4aa10d6a87339c8a657a8. The changes are very minimal.

weshaggard:
LGTM

src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
```diff
@@ -82,9 +83,81 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
-    <None Include="PackageFiles\**\*.*" />
+    <None Include="PackageFiles\ApiCompat.runtimeconfig.json" />
```
dagood: I don't personally mind breaking the `**\*.*` into its parts, since I use VS, but was this intentional?

mellinoe: This was almost certainly because I opened the project in VS and it automatically did this.

PR #1223, opened 2016-11-12
Allow Filtering of rids from SupportsFile

karajas:
/cc @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
@@ -1,7 +1,9 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
 using System.Collections.Generic;
 using Newtonsoft.Json.Linq;
-using NuGet.Frameworks;
 using NuGet.RuntimeModel;
 
 namespace Microsoft.DotNet.Build.Tasks
@@ -37,8 +39,7 @@ namespace Microsoft.DotNet.Build.Tasks
         public static List<TfmRidPair> FilterForApplicableTFMRIDPairs(List<TfmRidPair> tfmRidPairs,
             string framework, string runtimeIdentifier)
         {
-            tfmRidPairs.RemoveAll(x => !framework.Equals(x.framework));
-            tfmRidPairs.RemoveAll(x => !runtimeIdentifier.Equals(x.runtimeIdentifier));
+            tfmRidPairs.RemoveAll(x => !framework.Equals(x.framework) || !runtimeIdentifier.Equals(x.runtimeIdentifier));
             
             return tfmRidPairs;
         }
--- a/src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
@@ -1,0 +1,294 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Framework;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet.Frameworks;
+using NuGet.RuntimeModel;
+using NuGet.Versioning;
+using Task = Microsoft.Build.Utilities.Task;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class FilterRuntimesFromSupports : Task
+    {
+        public string[] Frameworks { get; set; }
+
+        [Required]
+        public string[] RuntimeIds { get; set; }
+
+        [Required]
+        public string RuntimeGraphFile { get; set; }
+
+        [Required]
+        public string SupportsFile { get; set; }
+
+        [Required]
+        public string InputProjectJson { get; set; }
+
+        public override bool Execute()
+        {
+            if (!File.Exists(RuntimeGraphFile))
+            {
+                Log.LogError("Cannot find specified runtime.json - '{0}'", RuntimeGraphFile);
+                return false;
+            }
+
+            if (!File.Exists(SupportsFile))
+            {
+                Log.LogError("Cannot find specified runtime.json - '{0}'", RuntimeGraphFile);
+                return false;
+            }
+
+            RuntimeGraph runtimeGraph = ReadRuntimeGraph(RuntimeGraphFile);
+            RuntimeGraph supportsGraph = ReadRuntimeGraph(SupportsFile);
+
+            List<CompatibilityProfile> filteredSupports = new List<CompatibilityProfile>();
+            List<CompatibilityProfile> filteredRuntimes = new List<CompatibilityProfile>();
+
+            if (Frameworks!=null && Frameworks.Length > 0)
+            {
+                foreach (var key in supportsGraph.Supports.Keys)
+                {
+                    var compat = DefaultCompatibilityProvider.Instance;
+                    CompatibilityProfile compatibilityProfile = null;
+                    supportsGraph.Supports.TryGetValue(key, out compatibilityProfile);
+                    bool isFrameworkFiltered = false;
+                    foreach (var tfmRidPair in compatibilityProfile.RestoreContexts)
+                    {
+                        foreach (string tfm in Frameworks)
+                        {
+                            if (compat.IsCompatible(NuGetFramework.Parse(tfm), tfmRidPair.Framework))
+                            {
+                                isFrameworkFiltered = true;
+                                filteredSupports.Add(compatibilityProfile);
+                                break;
+                            }
+                        }
+                        if (isFrameworkFiltered)
+                        {
+                            break;
+                        }
+                    }
+                }
+            }
+            else
+            {
+                foreach (var key in supportsGraph.Supports.Keys)
+                {
+                    CompatibilityProfile compatibilityProfile = null;
+                    supportsGraph.Supports.TryGetValue(key, out compatibilityProfile);
+                    filteredSupports.Add(compatibilityProfile);
+                }
+            }
+
+            if (RuntimeIds.Length > 0)
+            {
+                foreach (var compatibilityProfile in filteredSupports)
+                {
+                    List<bool> toRemove = new List<bool>();
+                    List<FrameworkRuntimePair> frameworkRuntimePairs = new List<FrameworkRuntimePair>();
+                    foreach (var tfmRidPair in compatibilityProfile.RestoreContexts)
+                    {
+                        bool isRuntimeFiltered = false;
+                        foreach (string runtimeId in RuntimeIds)
+                        {
+                            if (runtimeGraph.AreCompatible(runtimeId, tfmRidPair.RuntimeIdentifier))
+                            {
+                                isRuntimeFiltered = true;
+                                break;
+                            }
+                        }
+                        toRemove.Add(isRuntimeFiltered);
+                    }
+                    for (int i = 0, k = 0; i < toRemove.Count; i++)
+                    {
+                        if (!toRemove[i])
+                        {
+                            compatibilityProfile.RestoreContexts.RemoveAt(k);
+                        }
+                        else
+                        {
+                            k++;
+                        }
+                    }
+                    filteredRuntimes.Add(compatibilityProfile);
+                }
+            }
+            else
+            {
+                filteredRuntimes = filteredSupports;
+            }
+
+            JObject projectJson = ReadProject(InputProjectJson);
+            JObject supportsClause = new JObject();
+            foreach (var compatibilityProfile in filteredRuntimes)
+            {
+                Dictionary<string, List<string>> map = new Dictionary<string, List<string>>();
+                foreach (var tfmRidPair in compatibilityProfile.RestoreContexts)
+                {
+                    if (map.ContainsKey(tfmRidPair.Framework.GetShortFolderName()))
+                    {
+                        List<string> str = new List<string>();
+                        map.TryGetValue(tfmRidPair.Framework.GetShortFolderName(), out str);
+                        str.Add(tfmRidPair.RuntimeIdentifier);
+                    }
+                    else
+                    {
+                        map.Add(tfmRidPair.Framework.GetShortFolderName(), new List<string>() { tfmRidPair.RuntimeIdentifier });
+                    }
+                }
+                var tfmRidsForSupports = new JObject();
+                foreach (var item in map)
+                {
+                    tfmRidsForSupports[item.Key] = new JArray(item.Value);
+                }
+                supportsClause[compatibilityProfile.Name] = tfmRidsForSupports;
+            }
+
+            if (projectJson["supports"]!=null)
+            {
+                projectJson["supports"] = supportsClause;
+            }
+
+            string projectJsonRoot = JsonConvert.SerializeObject(projectJson, Formatting.Indented) + Environment.NewLine;
+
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -0,0 +1,294 @@
...
+                                break;
+                            }
+                        }
+                        toRemove.Add(isRuntimeFiltered);
+                    }
+                    for (int i = 0, k = 0; i < toRemove.Count; i++)
+                    {
+                        if (!toRemove[i])
+                        {
+                            compatibilityProfile.RestoreContexts.RemoveAt(k);
```
weshaggard: You can avoid this boolean array by using a for loop over compatibilityProfile.RestoreContexts in reverse order and calling RemoveAt(i).

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -0,0 +1,294 @@
...
+                    filteredSupports.Add(compatibilityProfile);
+                }
+            }
+
+            if (RuntimeIds.Length > 0)
+            {
+                foreach (var compatibilityProfile in filteredSupports)
+                {
+                    List<bool> toRemove = new List<bool>();
+                    List<FrameworkRuntimePair> frameworkRuntimePairs = new List<FrameworkRuntimePair>();
```
weshaggard: Looks like this is unused.

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -0,0 +1,294 @@
...
+                }
+                var tfmRidsForSupports = new JObject();
+                foreach (var item in map)
+                {
+                    tfmRidsForSupports[item.Key] = new JArray(item.Value);
+                }
+                supportsClause[compatibilityProfile.Name] = tfmRidsForSupports;
+            }
+
+            if (projectJson["supports"]!=null)
```
weshaggard: Shouldn't we always set this? Why do we need to check for null?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -3,9 +3,11 @@
   <UsingTask TaskName="AddDependenciesToProjectJson" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="GetPackageNumberFromPackageDrop" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="DumpItem" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="FilterRuntimesFromSupports" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <ItemGroup>
     <_PackagesDrops Include="$(PackagesDrops)" />
+    <SupportsDefinitionFile Condition="'$(SupportsDefinitionFile)' == '' AND Exists('$(SrcDir)Microsoft.DotNet.BuildTools.TestSuite/runtime.json')">$(SrcDir)Microsoft.DotNet.BuildTools.TestSuite/runtime.json</SupportsDefinitionFile>
```
weshaggard: Where is SrcDir set?

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -0,0 +1,294 @@
...
+                foreach (var compatibilityProfile in filteredSupports)
+                {
+                    List<bool> toRemove = new List<bool>();
+                    List<FrameworkRuntimePair> frameworkRuntimePairs = new List<FrameworkRuntimePair>();
+                    foreach (var tfmRidPair in compatibilityProfile.RestoreContexts)
+                    {
+                        bool isRuntimeFiltered = false;
+                        foreach (string runtimeId in RuntimeIds)
+                        {
+                            if (runtimeGraph.AreCompatible(runtimeId, tfmRidPair.RuntimeIdentifier))
```
weshaggard: @ericstj is there any benefit to keep the compatible runtimes as opposed to just always replacing the runtimes with the passed in runtimes?

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -0,0 +1,294 @@
...
+            using (TextReader projectFileReader = File.OpenText(projectJsonPath))
+            {
+                var projectJsonReader = new JsonTextReader(projectFileReader);
+                var serializer = new JsonSerializer();
+                return serializer.Deserialize<JObject>(projectJsonReader);
+            }
+        }
+
+
+        public static RuntimeGraph ReadRuntimeGraph(string filePath)
```
ericstj: Why didn't you use [JsonRuntimeFormat.ReadRuntimeGraph](https://github.com/NuGet/NuGet.Client/blob/88330551e465e5673505c525515e74129f73d98c/src/NuGet.Core/NuGet.RuntimeModel/JsonRuntimeFormat.cs#L17)?

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
ericstj: There are cases where we expect specific RIDs to be compatible.  For instance desktop should run on win/win-x64/win-x86.  If we used win7-x64 we might miss a packaging problem.  Similarly core should run on win7-x64/win7-x86, if we used win10-x64 we might miss a packaging problem.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
karajas: SrcDir was in corefx, but this is the wrong default. It should be the runtime.json in the packagesdir.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -162,6 +164,21 @@
...
   </Target>
+  
+  <Target Name="FilterRuntimesFromSupports"
+          DependsOnTargets="AddDependenciesToProjectJson;GatherProjectReferenceForProjectJson;DetermineProjectJsonPath"
+          Condition="'@(_InjectProjectReferenceDependency)' != '' and '$(RestoreDuringBuild)' != 'false'">
+    <!-- Filter Runtimes Based on the current platform -->
+    <ItemGroup>
+      <Runtimes Include="$(TestNugetRuntimeId)" />
+    </ItemGroup>
+    <FilterRuntimesFromSupports
```
karajas: Will remove.

karajas:
@weshaggard @ericstj PTAL at this change with https://github.com/dotnet/corefx/compare/master...karajas:defaultBAP

karajas:
Where should the default TestTFM come from? I believe the TestNugetRuntimeId is now correctly detected depending on the platform.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -144,9 +144,17 @@
 
     <!-- If VersionsFiles were passed on the command-line, convert them to an item group -->
     <ItemGroup>
-      <_VersionsFiles Include="$(VersionsFiles)" />
+      <_VersionsFiles Include="$(VersionsFiles)" />    
+      <SupportsDefinitionFileGroup Condition="'$(SupportsDefinitionFile)' == ''" Include="$(MSBuildThisFileDirectory)../packages/Microsoft.DotNet.BuildTools.TestSuite/**/runtime.json" />
```
ericstj: Can we just get rid of this and default it to the test-runtime\project.json?

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -0,0 +1,89 @@
...
+                        map.Add(tfmRidPair.Framework.GetShortFolderName(),
+                            new List<string>() {tfmRidPair.RuntimeIdentifier});
+                    }
+                }
+                var tfmRidsForSupports = new JObject();
+                foreach (var item in map)
+                {
+                    tfmRidsForSupports[item.Key] = new JArray(item.Value);
+                }
+                supportsClause[compatibilityProfile.Name] = tfmRidsForSupports;
```
ericstj: It might be cleaner if we just use the TFM:RID tuples and put them under a single inline supports clause, rather than preserving the compatibilityProfile names and subsetting them.  I don't trust NuGet to handle conflicts correctly and at a minimum it is confusing.

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -0,0 +1,89 @@
...
+using NuGet.RuntimeModel;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public partial class FilterRuntimesFromSupports
+    {
+        public static List<CompatibilityProfile> GetAllTfmRidPairs(string supportsFile)
+        {
+            RuntimeGraph supportsGraph = JsonRuntimeFormat.ReadRuntimeGraph(supportsFile);
+            List<CompatibilityProfile> tfmRidPairs = new List<CompatibilityProfile>();
```
ericstj: I was actually imagining this returning FrameworkRuntimePairs rather than trying to preserve the actual names.

weshaggard:
>Where should the default TestTFM come from? I believe the TestNugetRuntimeId is now correctly detected depending on the platform.

I think you should use FiltertoTestTFM for this purpose https://github.com/dotnet/corefx/blob/master/dir.props#L496

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
karajas: This makes sense, will change.

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
karajas: Will do.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
ericstj: I think we already have a property TestRuntimeProjectJson.  That seems like a better default.

weshaggard: Yes I would prefer we don't do wildcard searching under the packages folder. Do we think TestRuntimeProjectJson file is going to be the right file for this case? It has a different set of supports clauses.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -144,9 +144,17 @@
...
     </ItemGroup>
 
+    <PropertyGroup>
+      <SupportsDefinitionFile>@(SupportsDefinitionFileGroup)</SupportsDefinitionFile>
+    </PropertyGroup>
+
+    
+    <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] KARTHIKDEBUG supprots '$(SupportsDefinitionFile)' '$(TestNugetRuntimeId)' '$(TestTFM)'" />
```
weshaggard: reminder you should remove this before merging.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
weshaggard: We could also consider including the supports runtime.json file directly in the BuildTools package given that the source of truth is in the BuildTools repo anyway.

karajas: This seems like a property referring to the buildtools variable. The property in corefx is CommonTestProjectJson, which has the most recent supports clauses.

ericstj: Whatever we use I want it to be configurable by repos so we don't have to update buildtools to change it.  I'd also like to minimize the boilerplate for configuring it.

karajas: I think an issue I had with this is that, since this file is imported only during a msbuild process of a single project it needs to inherit a global property from the respective repos dir.props.

karajas: Think the way it's hooked up now is during a corefx build, the generated common project.json is set to CommonTestProjectJson, which will keep the non-common project.json's in check. So if you needed to update supports you will now just update the common project.json

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -141,11 +141,19 @@
...
+      <Output TaskParameter="RuntimeIdentifier" PropertyName="SyncNugetRuntimeId" />
+    </GetTargetMachineInfo>
 
     <!-- If VersionsFiles were passed on the command-line, convert them to an item group -->
     <ItemGroup>
       <_VersionsFiles Include="$(VersionsFiles)" />
     </ItemGroup>
+    <PropertyGroup>
+      <SupportsDefinitionFile Condition="'$(SupportsDefinitionFile)' == ''">$(CommonTestProjectJson)</SupportsDefinitionFile>
+      <SyncNugetRuntimeId Condition="$(SyncNugetRuntimeId.StartsWith('win'))">win7-x64</SyncNugetRuntimeId>
```
weshaggard: We should not introduce another property.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -141,11 +141,19 @@
...
+    <GetTargetMachineInfo>
+      <Output TaskParameter="RuntimeIdentifier" PropertyName="SyncNugetRuntimeId" />
+    </GetTargetMachineInfo>
 
     <!-- If VersionsFiles were passed on the command-line, convert them to an item group -->
     <ItemGroup>
       <_VersionsFiles Include="$(VersionsFiles)" />
     </ItemGroup>
+    <PropertyGroup>
+      <SupportsDefinitionFile Condition="'$(SupportsDefinitionFile)' == ''">$(CommonTestProjectJson)</SupportsDefinitionFile>
```
weshaggard: We should condition on the existence of CommonTestProjectJson.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -280,24 +280,7 @@
     <Delete Condition="'$(ForceRunTests)'=='true' And Exists($(TestsSuccessfulSemaphore))"
             Files="$(TestsSuccessfulSemaphore)" />
```
weshaggard: remove using task in this file.

PR #1226, opened 2016-11-15
Bootstrap needs cli subdirectory to exist

JeremyKuhne:
@weshaggard @danmosemsft

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -45,6 +45,11 @@ Invoke-WebRequest "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/script
 $rootCliVersion = Join-Path $RepositoryRoot ".cliversion"
 $dotNetCliVersion = Get-Content $rootCliVersion
 
+if (-Not (Test-Path $CliLocalPath))
+{
+    mkdir $CliLocalPath | Out-Null
+}
+
 # now execute the script
 Write-Host "$dotnetInstallPath -Version $dotNetCliVersion -InstallDir $CliLocalPath -Architecture ""$Architecture"""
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap.ps1
```diff
@@ -45,6 +45,11 @@ Invoke-WebRequest "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/script
 $rootCliVersion = Join-Path $RepositoryRoot ".cliversion"
 $dotNetCliVersion = Get-Content $rootCliVersion
 
+if (-Not (Test-Path $CliLocalPath))
```
weshaggard: Do we have the same issue in bootstrap.sh?

weshaggard:
@crummel

bootstrap/bootstrap.ps1
JeremyKuhne: Probably- I'll follow up. This won't pass CI without the fix as it pulls directly from the repo. I assume I can put this in?

JeremyKuhne: We should probably have this repo use the local script instead of pulling from the net. Thoughts?

crummel: We could do an override environment variable similar to BUILDTOOLS_SOURCE.

crummel:
Looks good to me.  I do think the Linux version will have the same issue.

crummel:
Shell script change looks good too, thanks for the fix.

bootstrap/bootstrap.ps1
weshaggard: I would prefer we build in the same way other repo's build unless there is a strong reason.

PR #1227, opened 2016-11-15
Add WarnAsError Logger

JeremyKuhne:
We want to allow hard failing the build if any warnings
are raised. Starting simple- we can allow configuring for
specific warnings as we hit the need.

@weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/WarnAsErrorLogger.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/WarnAsErrorLogger.cs
@@ -25,7 +25,7 @@ namespace Microsoft.DotNet.Build.Tasks
         {
             if (_warningsLogged)
             {
-                throw new Exception("Warings were logged, failing the build.");
+                throw new Exception("Warnings were logged, failing the build.");
             }
         }
     }
--- a/src/Microsoft.DotNet.Build.Tasks/WarnAsErrorLogger.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/WarnAsErrorLogger.cs
@@ -1,0 +1,32 @@
+﻿
+using Microsoft.Build.Framework;
+using System;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class WarnAsErrorLogger : ILogger
+    {
+        private bool _warningsLogged;
+
+        public string Parameters { get; set; }
+        public LoggerVerbosity Verbosity { get; set; }
+
+        public void Initialize(IEventSource eventSource)
+        {
+            eventSource.WarningRaised += EventSource_WarningRaised;
+        }
+
+        private void EventSource_WarningRaised(object sender, BuildWarningEventArgs e)
+        {
+            _warningsLogged = true;
+        }
+
+        public void Shutdown()
+        {
+            if (_warningsLogged)
+            {
+                throw new Exception("Warings were logged, failing the build.");
+            }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/WarnAsErrorLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+        private void EventSource_WarningRaised(object sender, BuildWarningEventArgs e)
+        {
+            _warningsLogged = true;
+        }
+
+        public void Shutdown()
+        {
+            if (_warningsLogged)
+            {
+                throw new Exception("Warings were logged, failing the build.");
```
weshaggard: warings -> warnings

ericstj:
I think we need configuration from the get-go.  We won't be able to wait for a build tools change when someone makes a code change to introduce a warning and we decide its suppressible.

weshaggard:
>I think we need configuration from the get-go. We won't be able to wait for a build tools change when someone makes a code change to introduce a warning and we decide its suppressible.

I agree but we would need to figure out what such a suppression would look like.

danmosemsft:
While we may not want anyone to get through CI with warnings, it's not uncommon to have warnings you're ignoring during development (such as unused variable) and it would be tempting to disable warnings entirely for your project in order to get unblocked rather than fix them. We should probably disable warning as error for some of those.

As for the approach, while throwing on shutdown may be good enough for us, a more pretty solution would be to subclass the MSBuild logger itself, then you have the opportunity to turn warnings into actual errors, so they get logged in the right place and in the standard format. Not suggesting we do that right now.

weshaggard:
>As for the approach, while throwing on shutdown may be good enough for us, a more pretty solution would be to subclass the MSBuild logger itself, then you have the opportunity to turn warnings into actual errors, so they get logged in the right place and in the standard format. Not suggesting we do that right now.

I was under the impression you couldn't generate an error from a logger but if you can I would highly suggest we do that instead of the hack of throwing at the end of the logger.

danmosemsft:
You can't generate an error from a logger however a logger can itself treat a warning as an error. The only thing it can't do is fail the build, if that is important it would have to throw as this one is doing, but at least the "warnings" would appear in the .err log and if you are selective about the warnings you treat as errors it will be visible which are being treated as errors and which are warnings.

JeremyKuhne:
> I think we need configuration from the get-go. We won't be able to wait for a build tools change when someone makes a code change to introduce a warning and we decide its suppressible.

We can simply unhook the logger to disable it. It may be that we want to be more or less configurable once we see a real world warning we need to ignore. For example:

- Do we just allow skipping specific warning codes?
- Do we allow skipping specific projects?
- Do we go as far as specific files, line numbers, etc?

It is also possible we'll discover scenarios where we have things we either can't fix or can't fix easily. It would be nice to be as focused as possible if these sort of scenarios come up.

I'd prefer to be more informed and grow organically as we can simply disable it at this point.

danmosemsft:
I noticed that currently MSBuild folks have a PR out for "disable warning X" and "warning Y as error". That woudl be built into the engine. Given we use the VS-installed copy of MSBuild when we're on Windows, we could use that about a year from now :(

JeremyKuhne:
> Given we use the VS-installed copy of MSBuild when we're on Windows

Shouldn't we ultimately be using a package for it?

stephentoub:
@JeremyKuhne, should this PR from 2016 be closed?

PR #1232, opened 2016-11-17
Add $(TestDebugger) flag to easily debug unittests

danmosemsft:
msbuild /t:testdebugger=c:\debuggers\windbg.exe automatically runs the next unit tests under the debugger. 

Should work with devenv.exe too.

@joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -101,7 +101,8 @@
     <TestCommandLine Condition="'$(Performance)'!='true'">$(TestProgram) $(TestArguments) {XunitTraitOptions}</TestCommandLine>
 
     <!-- set $(TestDebugger) to eg c:\debuggers\windbg.exe to run tests under a debugger -->
-    <TestCommandLine Condition="'$(TestDebugger)' != ''">$(TestDebugger) $(TestCommandLine)</TestCommandLine>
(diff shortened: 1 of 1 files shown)


danmosemsft:
@joperezr LGTU?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -99,6 +99,10 @@
     <TestArguments Condition="'$(TestHostExecutable)'==''">$(XunitArguments)</TestArguments>
 
     <TestCommandLine Condition="'$(Performance)'!='true'">$(TestProgram) $(TestArguments) {XunitTraitOptions}</TestCommandLine>
+
+    <!-- set $(TestDebugger) to eg c:\debuggers\windbg.exe to run tests under a debugger -->
+    <TestCommandLine Condition="'$(TestDebugger)' != '' and !$(TestDebugger.Contains('devenv'))">$(TestDebugger) $(TestCommandLine)</TestCommandLine>
+    <TestCommandLine Condition="'$(TestDebugger)' != '' and  $(TestDebugger.Contains('devenv'))">$(TestDebugger) /debugexe $(TestCommandLine)</TestCommandLine>
```
weshaggard: Why not just force folks to pass the full debugger command line instead of having this hack here?

danmosemsft: Since the hack is on one line and makes it much simpler to type

PR #1233, opened 2016-11-18
Change _RemovePackageProjectReference condition to work on non-Windows platforms.

karajas:
/cc @chcosta @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
@@ -82,7 +82,7 @@
     <ItemGroup Condition="'$(KeepAllProjectReferences)' != 'true'">
       <!-- Don't alter the ProjectReference's from the project during the extraction phase -->
       <_ClonedProjectReference Include="@(ProjectReference)" />
-      <_RemovePackageProjectReference Condition="'%(_ClonedProjectReference.KeepProjectReference)' != 'true' and !$([System.Text.RegularExpressions.Regex]::IsMatch('%(_ClonedProjectReference.FullPath)', '\\tests\\', System.Text.RegularExpressions.RegexOptions.IgnoreCase))"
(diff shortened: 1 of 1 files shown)


karajas:
Merging this in as a nominal change.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/buildagainstpackages.targets
```diff
@@ -82,7 +82,7 @@
     <ItemGroup Condition="'$(KeepAllProjectReferences)' != 'true'">
       <!-- Don't alter the ProjectReference's from the project during the extraction phase -->
       <_ClonedProjectReference Include="@(ProjectReference)" />
-      <_RemovePackageProjectReference Condition="'%(_ClonedProjectReference.KeepProjectReference)' != 'true' and !$([System.Text.RegularExpressions.Regex]::IsMatch('%(_ClonedProjectReference.FullPath)', '\\tests\\', System.Text.RegularExpressions.RegexOptions.IgnoreCase))"
+      <_RemovePackageProjectReference Condition="'%(_ClonedProjectReference.KeepProjectReference)' == 'true' or $([System.Text.RegularExpressions.Regex]::IsMatch('%(_ClonedProjectReference.FullPath)', '\\tests\\', System.Text.RegularExpressions.RegexOptions.IgnoreCase)) or $([System.Text.RegularExpressions.Regex]::IsMatch('%(_ClonedProjectReference.FullPath)', '//tests//', System.Text.RegularExpressions.RegexOptions.IgnoreCase))"
```
chcosta: since this is a regex expression, could you do a char set instead?  ie, '[\\\/]tests[\\\/]'

note that I haven't tested that regex expression, but i'd think something like that would be sufficient.

karajas: Sorry for the quick merge, will fix in another pr.

PR #1236, opened 2016-11-22
Changing the test nuget moniker when TestArchitecture is passed in

joperezr:
cc: @naamunds @weshaggard 

Fixes dotnet/corefx#13739

With this, people running `build.cmd -buildArch:x86` will actually have the tests running for x86


joperezr:
Merging since Wes and I had already talked about making this change. I'll address any feedback that comes after this later.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
```diff
@@ -115,6 +115,7 @@
     <PropertyGroup>
       <DefaultTestNugetRuntimeId Condition="$(DefaultTestNugetRuntimeId.StartsWith('win'))">win7-x64</DefaultTestNugetRuntimeId>
       <TestArchitecture Condition="'$(TestArchitecture)' == ''">x64</TestArchitecture>
+      <TestNugetRuntimeId Condition="'$(TestNugetRuntimeId)' == '' And '$(TestArchitecture)'=='x86'">win7-x86</TestNugetRuntimeId>
```
weshaggard: While this works right now I don't like this approach. I would much prefer we only override the TestArchitecture so we future proof ourselves for other x86 platforms as well.

PR #1237, opened 2016-11-23
Fixing excluding duplicates between project.json closure and project …

SedarG:
…references in ReferenceCopyLocalPaths


SedarG:
@ericstj PTAL, It looks to me like there is a bug in excluding the entries that already exists in project references from resolved NuGet package assets (for CopyLocal items)

SedarG:
I did some testing on CoreFX repo and all looks ok. Merging the changes.

PR #1239, opened 2016-11-30
Allow __PUBLISH_RID detection in strict mode

dagood:
https://github.com/dotnet/buildtools/pull/1206 broke the condition that checked to see if `__PUBLISH_RID` has no value. This fix uses an empty string if `__PUBLISH_RID` is undefined rather than erroring out. (Mentioned at top of file.)

I didn't catch it because I only tested without auto-detection--I got an error when trying OSX.

@ellismg Could you take a quick look? Tiny fix.


ellismg:
LGTM

PR #1241, opened 2016-11-30
Make Supports filtering optional

karajas:
/cc @naamunds @weshaggard 

This should unblock official builds.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
@@ -68,14 +68,11 @@ namespace Microsoft.DotNet.Build.Tasks
         // The directory to put the generated project.json in
         [Required]
         public string OutputProjectJson { get; set; }
-
-        [Required]
+        
         public string SupportsFile { get; set; }
 
-        [Required]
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/AddDependenciesToProjectJson.cs
```diff
@@ -145,10 +142,13 @@ public override bool Execute()
                 projectRoot = UpdateDependenciesProperty(projectRoot, dependencies, Frameworks[i]);
             }
 
-            var definedTfmRidPairs = FilterRuntimesFromSupports.GetAllTfmRidPairs(SupportsFile);
-            var applicableTfmRidPairs = FilterRuntimesFromSupports.FilterForApplicableTFMRIDPairs(definedTfmRidPairs, TestTargetFramework, TestRuntime);
-            projectRoot["supports"] = FilterRuntimesFromSupports.GenerateCustomSupportsClause(applicableTfmRidPairs);
-
+            if (SupportsFile != null)
```
karajas: I can switch the equals statement around and I think that should be enough.

karajas: resolved.

PR #1243, opened 2016-12-04
Coverage flags

danmosemsft:
Allow /p:PopCoverageReport=true to cause the coverage report to open in your browser as soon as it's ready.

Allow eg /p:corelibroot=\git\coreclr.dan\bin\Product\Windows_NT.x64.Release to cause test runs to get your privately built corelib and symbols. This is an easy workaround until we have a fix for https://github.com/dotnet/corefx/issues/13094 (and useful anyway). Disables native images also, so it's not necessary to copy those.

@stephentoub

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -62,6 +62,10 @@
 
     <Exec Command="$(CoverageReportGeneratorCommandLine) -reports:$(CoverageOutputFilePath)"
           ContinueOnError="ErrorAndContinue" />
+
+    <Exec Command="start $(CoverageReportDir)index.htm"
+          Condition="'$(PopCoverageReport)' == 'true'" />
+
   </Target>
 
   <PropertyGroup Condition="'$(GenerateFullCoverageReport)'=='true'">
@@ -78,6 +82,9 @@
     <Exec Command="$(CoverageReportGeneratorCommandLine) -reports:$(CoverageReportDir)\*.coverage.xml"
           ContinueOnError="ErrorAndContinue"
           WorkingDirectory="$(ProjectDir)" />
+
+    <Exec Command="start $(CoverageReportDir)index.htm"
+          Condition="'$(PopCoverageReport)' == 'true'" />          
 
     <PropertyGroup>
       <CoverallsUploaderCommandLine>$(PackagesDir)coveralls.io.$(CoverallsUploaderVersion)\tools\coveralls.net.exe</CoverallsUploaderCommandLine>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -84,6 +84,18 @@
       <TestCopyLocal Include="@(OverridenRuntimeContent)" />
      </ItemGroup>
 
+    <ItemGroup Condition="'$(CoreLibRoot)' != ''">
```
stephentoub: You included this in a PR related to coverage, but this isn't specific to coverage, right?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -62,6 +62,10 @@
 
     <Exec Command="$(CoverageReportGeneratorCommandLine) -reports:$(CoverageOutputFilePath)"
           ContinueOnError="ErrorAndContinue" />
+
+    <Exec Command="start $(CoverageReportDir)index.htm"
+          Condition="'$(PopCoverageReport)' == 'true'" />
```
danmosemsft: Given it's an untyped language I'm leaning towards leaving it as a boolean so nobody using it someplace new could trip up and not realize there's two versions of true.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
danmosemsft: No, but right now it's why I need it. Perhaps I should have put it in separately but it doesn't seem worth separating now.

danmosemsft:
Attempted to update this PR but I guess you can't. Opening a new one.

weshaggard:
Once this is in Dan can you please update the docs in corefx to list this option, or ideally add it to the project.json file so that it is documented and folks will be able to find it again without reading all our targets files.

danmosemsft:
@weshaggard that goes into @(TestCopyLocal) rather than @(RunTestsForProjectInputs) which I switched to in https://github.com/dotnet/buildtools/pull/1244. 

@(RunTestsForProjectInputs) is easy to understand, it just get copied into the test dir.

@(TestCopyLocal) paths get ultimately written into dependencylist.txt which goes to the <GenerateAssemblyList> task  which only seems to run if $(UseLegacyXunitPerfRunner) is true.  That is set externally - I don't know what sets it. 

I also see that @(OverridenRuntimeContent) comes from $(BUILDTOOLS_OVERRIDE_RUNTIME) and those files are also copied in init-tools.sh - perhaps should be copied in init-tools.cmd also.

I wonder if all this could be simplified. Maybe @naamunds understands this.

naamunds:
@danmosemsft The `OverridenRuntimeContent` items are always copied to the test directory also.  This is clearer with the changes in https://github.com/dotnet/buildtools/pull/1228.  I would suggest using `RunTestsForProjectInputs` only if these items need to be Inputs to the RunTestsForProject target.

danmosemsft:
@naamunds OK, I'll wait for your change #1228

naamunds:
@danmosemsft I've merged #1228, so you can go ahead and update #1244 now.

PR #1246, opened 2016-12-05
Use full name to compare nuget frameworks instead of shortname in FilterRuntimesFromSupports.

karajas:
This is to ensure we pass in TestNugetTargetMoniker instead of TestTFM for filtering runtime identifiers from supports.

/cc @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
@@ -4,6 +4,7 @@
 
 using System.Collections.Generic;
 using Newtonsoft.Json.Linq;
+using NuGet.Frameworks;
 using NuGet.RuntimeModel;
 
 namespace Microsoft.DotNet.Build.Tasks
@@ -15,6 +16,7 @@ namespace Microsoft.DotNet.Build.Tasks
             framework = tfm;
             runtimeIdentifier = rid;
         }
+        //This should be the long form of the Nuget Framework moniker.
         public string framework { get; set; }
         public string runtimeIdentifier { get; set; }
     }
@@ -30,7 +32,7 @@ namespace Microsoft.DotNet.Build.Tasks
                 supportsGraph.Supports.TryGetValue(key, out compatibilityProfile);
                 foreach (var tfmRidPair in compatibilityProfile.RestoreContexts)
                 {
-                    tfmRidPairs.Add(new TfmRidPair(tfmRidPair.Framework.GetShortFolderName(), tfmRidPair.RuntimeIdentifier));
+                    tfmRidPairs.Add(new TfmRidPair(tfmRidPair.Framework.ToString(), tfmRidPair.RuntimeIdentifier));
                 }
             }
             return tfmRidPairs;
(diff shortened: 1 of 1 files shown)


weshaggard:
The change looks reasonable but I don't know why we need to always use a full name.

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -4,18 +4,20 @@
...
     public class TfmRidPair
     {
-        public TfmRidPair(string tfm, string rid)
+        public TfmRidPair(NuGetFramework tfm, string rid)
         {
             framework = tfm;
             runtimeIdentifier = rid;
         }
-        public string framework { get; set; }
+        //This should be the long form of the Nuget Framework moniker.
```
ericstj: I would have expected you to continue to use a string, but just make sure that you're passing in the long form.

karajas:
This was in the case of netcore50 builds. We do a string compare on netcore50 but the tfm for netcore50 is uap10.0. The issue is that we shouldn't be using FilterToTestTFM (which is a user defined string) but TestNugetTargetFramework which is a full framework specifier. 

https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L76

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
karajas: I can revert to using a string, don't really have a preference other than possibly in the future the compat check may change.

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -41,7 +43,7 @@ public static List<TfmRidPair> GetAllTfmRidPairs(string supportsFile)
         {
             if (framework != null)
             {
-                tfmRidPairs.RemoveAll(x => !framework.Equals(x.framework));
+                tfmRidPairs.RemoveAll(x => !framework.Equals(x.framework.ToString()));
```
karajas: This is ToString()'d but I can change the type back to string for clarity.

karajas:
Changed type back to string.

src/Microsoft.DotNet.Build.Tasks/FilterRuntimesFromSupports.cs
```diff
@@ -57,7 +59,7 @@ public static JObject GenerateCustomSupportsClause(List<TfmRidPair> applicableTf
             var tfmRidsForSupports = new JObject();
             foreach (var tfmRidPair in applicableTfmRidPairs)
             {
-                tfmRidsForSupports[tfmRidPair.framework] = new JArray(tfmRidPair.runtimeIdentifier);
+                tfmRidsForSupports[NuGetFramework.Parse(tfmRidPair.framework).GetShortFolderName()] = new JArray(tfmRidPair.runtimeIdentifier);
```
ericstj: I don't think you'll hit it since you are filtering to a single RID, but technically this could be overwriting entries in tfmRidsForSupports if you have multiple tuples with the same framework and different RIDs.

Also, I don't think this change is technically needed.  I think NuGet will handle the long form here.

karajas: I left this change in for consistency when people want to debug generated project.json so that they can determine what changed and not have to look up tfm short form -> long form mappings.

karajas:
Will resolve the multiple rid issue in a future fix. Will merge this now.

PR #1249, opened 2016-12-06
Upgrade net45 System.Runtime.InteropServices.RuntimeInformation

dagood:
Upgrade `System.Runtime.InteropServices.RuntimeInformation` to take a change that removed the need for api-set DLLs, described in https://github.com/dotnet/corefx/issues/13566. This was causing win7 runs to fail. A repro I had of the issue goes away when I use a buildtools package produced with this change.

The only package diff from this is `lib/net45/System.Runtime.InteropServices.RuntimeInformation.dll`.

Is depending on the prerelease package during the buildtools build ok, or should we avoid it at all costs?

Fixes (when flowed through) https://github.com/dotnet/corefx/issues/14165.

@weshaggard @JohnChen0 @stephentoub 
/cc @dleeapho

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.net45/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/project.json
@@ -8,7 +8,7 @@
     "NuGet.Versioning": "3.6.0-rc-1987",
     "System.Collections.Immutable": "1.1.37",
     "System.Reflection.Metadata": "1.1.0",
-    "System.Runtime.InteropServices.RuntimeInformation": "4.0.0"
+    "System.Runtime.InteropServices.RuntimeInformation": "4.4.0-beta-24813-03"
   },
   "frameworks": {
     "net45": {}
--- a/dependencies.props
+++ b/dependencies.props
@@ -19,10 +19,17 @@
     <CurrentRefXmlPath>$(MSBuildThisFileFullPath)</CurrentRefXmlPath>
   </PropertyGroup>
 
+  <!-- Packages that don't take versions from the corefx build-info. -->
+  <PropertyGroup>
+    <!-- Take new version of RuntimeInformation that doesn't use api-set dlls to fix win7 build. -->
+    <CoreFxBuildInfoDisabledPackages>System.Runtime.InteropServices.RuntimeInformation;$(CoreFxBuildInfoDisabledPackages)</CoreFxBuildInfoDisabledPackages>
+  </PropertyGroup>
+
   <ItemGroup>
     <RemoteDependencyBuildInfo Include="CoreFx">
       <BuildInfoPath>$(BaseDotNetBuildInfo)corefx/release/1.1.0</BuildInfoPath>
       <CurrentRef>$(CoreFxCurrentRef)</CurrentRef>
+      <DisabledPackages>$(CoreFxBuildInfoDisabledPackages)</DisabledPackages>
     </RemoteDependencyBuildInfo>
     <RemoteDependencyBuildInfo Include="CoreClr">
       <BuildInfoPath>$(BaseDotNetBuildInfo)coreclr/$(DependencyBranch)</BuildInfoPath>


src/Microsoft.DotNet.Build.Tasks.net45/project.json
```diff
@@ -8,7 +8,7 @@
     "NuGet.Versioning": "3.6.0-rc-1987",
     "System.Collections.Immutable": "1.1.37",
     "System.Reflection.Metadata": "1.1.0",
-    "System.Runtime.InteropServices.RuntimeInformation": "4.0.0"
+    "System.Runtime.InteropServices.RuntimeInformation": "4.4.0-beta-24806-04"
```
weshaggard: While this may fix the api-set issue it might actually cause other issues. In particular it will not be pulling in the closure of other prerelease packages. Can you please ensure that those other packages don't cause issue.

weshaggard:
Another potential way to fix this is to pull in the api-set into the closure if the other packages in the closure cause issues.

src/Microsoft.DotNet.Build.Tasks.net45/project.json
dagood: Are you saying that not pulling in the closure might cause issues, or that packages pulled in by the closure might cause issues? I have confirmed that the only file that changes in the resulting buildtools package is `lib/net45/System.Runtime.InteropServices.RuntimeInformation.dll`, if that is expected. Since this upgrade is only on the net45 side (is that an insufficient fix?), I don't think there is any extended closure.

weshaggard: The package closure isn't the real trouble necessarily but it might point to any new dependencies that the assembly might have. I don't know if there are any but that is what I'm worried about. If the assembly it is including now has all its dependencies satisfied already then I'm good with this change.

dagood: This is what I mean by net45 not having an extended closure for that package:
```
<group targetFramework=".NETFramework4.5" />
<group targetFramework=".NETStandard1.1">
	<dependency id="runtime.native.System" version="4.4.0-beta-24807-03" exclude="Compile" />
	<dependency id="System.Reflection" version="4.4.0-beta-24807-03" exclude="Compile" />
	<dependency id="System.Reflection.Extensions" version="4.4.0-beta-24807-03" exclude="Compile" />
	<dependency id="System.Resources.ResourceManager" version="4.4.0-beta-24807-03" exclude="Compile" />
	<dependency id="System.Runtime" version="4.4.0-beta-24807-03" />
	<dependency id="System.Runtime.InteropServices" version="4.4.0-beta-24807-03" exclude="Compile" />
	<dependency id="System.Threading" version="4.4.0-beta-24807-03" exclude="Compile" />
</group>
```
I suppose all its dependencies are satisfied because there aren't any. :smile: I can see why the dep closure could be a problem if we need to upgrade this package on the .NET Core side too, though.

dagood:
I realized that the package I'm upgrading to actually wasn't part of a clean build (publishing dotnet/versions info failed due to bot account token being out of date) so I'm holding off on this. There's work being done to fix the corefx master build, and I'll update the version here when we get a good build.

src/Microsoft.DotNet.Build.Tasks.net45/project.json
dagood: Ah! I didn't catch the distinction, that makes sense. Here's old vs. new with what ildasm shows me: https://gist.github.com/dagood/1e9a84cfb4b1ba52d883765e4c8d6468.

That confirms that the api-set DLLs are no longer present and there's the `.module extern kernel32.dll` as expected, but there's also a new dependency:
```
.assembly extern System.Reflection.Extensions
{
  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
  .ver 4:0:0:0
}
```
It's in my GAC, but I don't know if that means it's ok.

dagood:
Rebased and changed to `4.4.0-beta-24813-03` which was part of a good build.

PR #1252, opened 2016-12-07
Update to NuGet that has netstandard2.0

ericstj:
Use a private build of NuGet with netcoreapp1.2 > netstandard2.0
compatibility mapping.

https://github.com/NuGet/NuGet.Client/commit/fa1a7540cceef33e6e519829fd52f2265f18824e

Once this is in we can pick it up in corefx and our CLI will get the update as well due to the init-tools hack: https://github.com/dotnet/corefx/commit/013ad32b3e871bd86745e380088c55d877748356

/cc @weshaggard @danmosemsft @karelz


ericstj:
/cc @alexperovich 

I've tested picking this up in corefx and hit validation errors due to the mapping of net461 > nca2.0.  We'll need to have a fix ready those before we merge this.

ericstj:
Here are the remaining errors in corefx.
https://gist.github.com/ericstj/84dd13b356d353b18a412fe61737b1b8

All are due to netstandard remapping for net461.  The fixes will be to ensure we have a net461 facade configuration that supports the latest API.  @alexperovich can you handle this along with the rename?

ericstj:
We have to change the version of NuGet we're using.  The previous version (`xprivate-...`) broke NuGet's CI when I published to build-tools, because apparently that's in their list of sources and they're using floating versions and suddenly started getting my private build. 😕

ericstj:
I've fixed all the desktop builds so that we can make this change:
https://github.com/ericstj/corefx/commit/ea62c246f30d4c6c63c300312a0d657c1b0fa508

ericstj:
Ok, now we have the updated mapping for netcoreapp2.0.  Let me make sure the corefx side is ready to go then we can merge this.

ericstj:
CoreFx is ready.  Merging

PR #1255, opened 2016-12-09
Have GenAPI also include exported types when writing typeforward list

weshaggard:
cc @ericstj


ericstj:
LGTM

weshaggard:
turned out we already had support for this that I overlooked. I updated to use that instead of what I did originally.

ericstj:
Doh, that looks better.

Issue #1092, opened 2016-10-05
Handle assembly binding when running tests against full framework


tarekgh:
When running the tests against full framework (TestTFM is net46x), we run into assembly binding issues by the loader because the loader can load the older version of the assembly which cause runtime error like: 

```
System.IO.IsolatedStorage.Tests.HelperTests.GetNormalizedStrongNameHash [FAIL]
        System.IO.FileLoadException : Could not load file or assembly 'System.Security.Cryptography.Primitives, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040) 
```

I believe the way to fix this is to add the following to tests.targets:

``` XML
  <PropertyGroup>
   <!-- Add the following properties -->
    <_IsNet4XTFM Condition="$(TestTFM.StartsWith('net4'))">true</_IsNet4XTFM>
    <AutoGenerateBindingRedirects Condition="'$(_IsNet4XTFM)' =='true'">true</AutoGenerateBindingRedirects>
    <GenerateBindingRedirectsOutputType Condition="'$(_IsNet4XTFM)' =='true'">true</GenerateBindingRedirectsOutputType>
  </PropertyGroup>


  <Target Name="DiscoverTestInputs" DependsOnTargets="ResolveProjectReferences;ResolveAssemblyReferences;GetCopyToOutputDirectoryItems">
    <ItemGroup>
     ….

      <!-- Add the following properties -->
      <RunTestsForProjectInputs Include="$(TargetPath).config" /> 
    </ItemGroup>
  </Target>

  <PropertyGroup>
    <XunitResultsFileName>testResults.xml</XunitResultsFileName>

    <!-- Add $(TargetFileName).config to xunit runner parameters -->
    <XunitOptions Condition="'$(_IsNet4XTFM)' == 'true'">$(XunitOptions) $(TargetFileName).config -noshadow </XunitOptions>
…
  </PropertyGroup>
```

tarekgh:
CC @ericstj @JeremyKuhne

ericstj:
/cc @weshaggard 
For this to work ResolveAssemblyReferences (RAR) will need to see the desktop assemblies.  Not sure if that will necessarily happen in the cases where we compile against netstandard but run against net4*.

jiangzeng:
[...]

PR #1257, opened 2016-12-09
Add binding redirect generation for desktop tests

karajas:
/cc @danmosemsft @AlexGhiondea @ericstj @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using Microsoft.Build.Framework;
+using Task = Microsoft.Build.Utilities.Task;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GenerateBindingRedirect : Task
+    {
+        [Required]
+        public ITaskItem[] AssemblyIdentities { get; set; }
+
+        [Required]
+        public ITaskItem[] Executables { get; set; }
+
+        [Required]
+        public string OutputPath { get; set; }
+
+        private static XNamespace ns { get; set; }
+
+        public override bool Execute()
+        {
+            ns = "urn:schemas-microsoft-com:asm.v1";
+            XElement bindingRedirectAssemblies = new XElement(ns + "assemblyBinding");
+            foreach (ITaskItem assembly in AssemblyIdentities)
+            {
+                string publicKeyToken = assembly.GetMetadata("PublicKeyToken");
+                string assemblyVersion = assembly.GetMetadata("Version");
+                string assemblyName = assembly.GetMetadata("Name");
+                if (string.IsNullOrEmpty(publicKeyToken))
+                {
+                    Log.LogWarning($"Empty publicKeyToken for {assemblyName} {assemblyVersion}");
+                }
+                string culture = string.IsNullOrEmpty(assembly.GetMetadata("Culture"))
+                    ? "neutral"
+                    : assembly.GetMetadata("Culture");
+                XElement assemblyIdentity = new XElement(ns + "assemblyIdentity",
+                    new XAttribute("name", assemblyName),
+                    new XAttribute("publicKeyToken", publicKeyToken),
+                    new XAttribute("culture", culture));
+                XElement bindingRedirect = new XElement(ns + "bindingRedirect",
+                    new XAttribute("oldVersion", $"0.0.0.0-{assemblyVersion}"),
+                    new XAttribute("newVersion", assemblyVersion));
+                XElement dependentAssembly = new XElement(ns + "dependentAssembly",
+                    assemblyIdentity,
+                    bindingRedirect);
+                bindingRedirectAssemblies.Add(dependentAssembly);
+            }
+            XDocument doc = new XDocument(new XElement("configuration", new XElement("runtime", bindingRedirectAssemblies)));
+            foreach (ITaskItem executable in Executables)
+            {
+                string executableName = Path.GetFileName(executable.ItemSpec);
+                using (FileStream fs = new FileStream(Path.Combine(OutputPath, executableName + ".config"), FileMode.Create))
+                {
+                    doc.Save(fs);
+                }
+            }
+            
+            return !Log.HasLoggedErrors;
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
@@ -129,6 +129,16 @@
       <SourcesToCopyToTestDir Include="@(OverridenRuntimeContent)" />
       <DestinationsForTestDir Include="@(OverridenRuntimeContent->'$(TestPath)\%(Filename)%(Extension)')" />
     </ItemGroup>
+    
+    <!-- Removing assets from TestCopyLocal that come in from P2P references in SourcesToCopyToTestDir -->
+    <ItemGroup>
+        <_SourcesToCopyToTestDirByFileName Include="@(SourcesToCopyToTestDir->'%(Filename)%(Extension)')">
+          <SourcePath>%(SourcesToCopyToTestDir.Identity)</SourcePath>
+        </_SourcesToCopyToTestDirByFileName>
+        <!-- This is the item group the contains the final test assemblies(from both p2p and package resolve) that are present in the test path in bin dir -->
+        <_FinalListOfTestAssetsByFileName Include="@(_TestCopyLocalByFileNameWithoutDuplicates)" Exclude="@(_SourcesToCopyToTestDirByFileName)" />
+        <_FinalListOfTestAssetsByFileName Include="@(_SourcesToCopyToTestDirByFileName)" />
+    </ItemGroup>
 
     <Copy
       SourceFiles="@(SourcesToCopyToTestDir)"
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -4,6 +4,7 @@
   <!-- This is the target that copies the test assets to the test output -->
   <Import Project="$(MSBuildThisFileDirectory)publishtest.targets" />
   <UsingTask TaskName="GenerateTestExecutionScripts" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="GenerateBindingRedirect" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <!-- Which categories of tests to run by default -->
   <PropertyGroup>
     <TestDisabled>false</TestDisabled>
@@ -52,8 +53,10 @@
     <XunitExecutable>xunit.console.netcore.exe</XunitExecutable>
   </PropertyGroup>
 
+  <!-- If not Core or AppX, test against desktop -->
   <PropertyGroup Condition="'$(TestWithCore)' != 'true' And '$(TestAppX)' != 'true'">
     <XunitExecutable Condition="'$(XunitExecutable)' == ''">xunit.console.exe</XunitExecutable>
+    <TestAgainstDesktop>true</TestAgainstDesktop>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TestAppX)' == 'true'">
@@ -206,6 +209,21 @@
       OutputListLocation="$(OutDir)assemblylist.txt"
       Condition="'$(UseLegacyXunitPerfRunner)'=='true'"
      />
+     
+    <!-- For Desktop Tests, we need to generate Binding Redirects. -->
+    <ItemGroup>
+      <_FilteredAssembliesForBindingRedirect Include="%(_FinalListOfTestAssetsByFileName.SourcePath)" Condition="'%(_FinalListOfTestAssetsByFileName.Extension)'=='.dll' or '%(_FinalListOfTestAssetsByFileName.Extension)'=='.exe'" />
+      <ExecutablesForBindingRedirect Include="%(_SourcesToCopyToTestDirByFileName.SourcePath)" Condition="'%(_SourcesToCopyToTestDirByFileName.Extension)'!='.pdb'"/>
+    </ItemGroup>
+    <GetAssemblyIdentity Condition="'$(TestAgainstDesktop)' == 'true'" AssemblyFiles="@(_FilteredAssembliesForBindingRedirect)">
+      <Output TaskParameter="Assemblies" ItemName="TestAssemblyIdentities"/>
+    </GetAssemblyIdentity>
+    
+    <GenerateBindingRedirect Condition="'$(TestAgainstDesktop)' == 'true' and '$(LocalAppConfigFile)' == ''"
+                             Executables="@(ExecutablesForBindingRedirect)"
+                             AssemblyIdentities="@(TestAssemblyIdentities)"
+                             OutputPath="$(TestPath)"
+    />
 
     <!-- For .NET Native compilation, we first need to generate a native executable if possible. -->
     <ItemGroup Condition="'$(UseDotNetNativeToolchain)' == 'true' AND '$(Performance)'!='true'" >
--- a/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks/Microsoft.DotNet.Build.Tasks.csproj
@@ -26,6 +26,7 @@
     <Compile Include="GenerateAssemblyList.cs" />
     <Compile Include="AddDependenciesToProjectJson.cs" />
     <Compile Include="FilterRuntimesFromSupports.cs" />
+    <Compile Include="GenerateBindingRedirect.cs" />
     <Compile Include="GenerateTestExecutionScripts.cs" />
     <Compile Include="GenerateCurrentVersion.cs" />
     <Compile Include="GenerateResourcesCode.cs" />


src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,61 @@
...
+            XDocument doc = new XDocument(new XElement("configuration", new XElement("runtime", bindingRedirectAssemblies)));
+            foreach (ITaskItem executable in Executables)
+            {
+                string executableName = Path.GetFileName(executable.ItemSpec);
+                using (FileStream fs = new FileStream(Path.Combine(OutputPath, executableName+".config"), FileMode.Create))
+                {
+                    doc.Save(fs);
+                }
+            }
+            return true;
```
danmosemsft: as general good practice Execute() on all MSBuild tasks should return thus
`return !Log.HasLoggedErrors;`

In this case it will be true anyway -- but are there any exceptions/error cases to handle and log here?

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,61 @@
...
+        public ITaskItem[] AssemblyIdentities { get; set; }
+
+        [Required]
+        public ITaskItem[] Executables { get; set; }
+
+        [Required]
+        public string OutputPath { get; set; }
+
+        private static XNamespace ns { get; set; }
+        public override bool Execute()
```
danmosemsft: line spacing

karajas:
Fixes issue https://github.com/dotnet/buildtools/issues/1092

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -206,6 +209,28 @@
       OutputListLocation="$(OutDir)assemblylist.txt"
       Condition="'$(UseLegacyXunitPerfRunner)'=='true'"
      />
+     
+     <!-- For Desktop Tests, we need to generate Binding Redirects. -->
+     <ItemGroup>
+        <_FileNamesInReferenceCopyLocalPaths Include="@(ReferenceCopyLocalPaths->'%(Filename)%(Extension)')">
+          <SourcePath>%(ReferenceCopyLocalPaths.Identity)</SourcePath>
+        </_FileNamesInReferenceCopyLocalPaths>
+        <_FileNamesInTestCopyLocalByFileNameWithoutDuplicates Include="@(_TestCopyLocalByFileNameWithoutDuplicates->'%(Filename)%(Extension)')">
```
danmosemsft: Isn't this 
```
+        <_FileNamesInTestCopyLocalByFileNameWithoutDuplicates Include="@(_TestCopyLocalByFileNameWithoutDuplicates->'%(Filename)%(Extension)')">
 +          <SourcePath>%(_TestCopyLocalByFileNameWithoutDuplicates.SourcePath)</SourcePath>
 +        </_FileNamesInTestCopyLocalByFileNameWithoutDuplicates>
```

the same as 
```
<_FileNamesInTestCopyLocalByFileNameWithoutDuplicates  Include="@(_TestCopyLocalByFileNameWithoutDuplicates)"/>
```

ie., metadata like `SourcePath` should be copied automatically when you populate one list from anotehr using an item transform (an @ expression)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -206,6 +209,28 @@
...
+          <SourcePath>%(_TestCopyLocalByFileNameWithoutDuplicates.SourcePath)</SourcePath>
+        </_FileNamesInTestCopyLocalByFileNameWithoutDuplicates>
+        <_FilteredListForBindingRedirect Include="@(_FileNamesInReferenceCopyLocalPaths)" Exclude="@(_FileNamesInTestCopyLocalByFileNameWithoutDuplicates)" />
+        <_FilteredListForBindingRedirect Include="@(_FileNamesInReferenceCopyLocalPaths)" />
```
danmosemsft: I'm not sure what this accomplishes
```
+        <_FilteredListForBindingRedirect Include="@(_FileNamesInReferenceCopyLocalPaths)" Exclude="@(_FileNamesInTestCopyLocalByFileNameWithoutDuplicates)" />
 +        <_FilteredListForBindingRedirect Include="@(_FileNamesInReferenceCopyLocalPaths)" />
```

It includes one list excepting some things, then includes it again entirely.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            ns = "urn:schemas-microsoft-com:asm.v1";
+            XElement bindingRedirectAssemblies = new XElement(ns + "assemblyBinding");
+            foreach (ITaskItem assembly in AssemblyIdentities)
+            {
+                string assemblyVersion = assembly.GetMetadata("Version");
+                XElement assemblyIdentity = new XElement("assemblyIdentity",
+                                                         new XAttribute("name", assembly.GetMetadata("Name")),
+                                                         new XAttribute("publicKeyToken", assembly.GetMetadata("PublicKeyToken")),
```
AlexGhiondea: why not do the same for PublicKeyToken and provide a default if we don't have one?

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,61 @@
...
+                                                         new XAttribute("culture", string.IsNullOrEmpty(assembly.GetMetadata("Culture")) ? "neutral" : assembly.GetMetadata("Culture")));
```
AlexGhiondea: Not sure how expensive `GetMetadata("Culture")` is, but if it is expensive, you should cache the call and then check its value.

danmosemsft: It's a lookup in a small dictionary so no need. Some built-in metadata like %(FullPath) is expensive, but it's cached by MSBuild in that dictionary.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
karajas: Will do so, probably a better idea anyways. I think the only other exceptions we can expect are possible missing metadata and xml write exceptions. Should I try/catch the whole thing? Don't think I can log and continue.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
karajas: So ReferenceCopyLocalPaths do not have SourcePath, I just wanted to standardize the metadata that I wanted in both groups.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
karajas: So this is the case where an assembly has been placed in the test path prior to resolving references, so I want to remove files from my list of files from resolve and include the ones that have been copied instead with the appropriate metadata.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,61 @@
...
+                string executableName = Path.GetFileName(executable.ItemSpec);
+                using (FileStream fs = new FileStream(Path.Combine(OutputPath, executableName+".config"), FileMode.Create))
+                {
+                    doc.Save(fs);
+                }
+            }
+            return true;
+        }
+
+        private XElement RemoveNamespace(XElement node)
```
AlexGhiondea: Why is this needed?

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
karajas: Sure, is there a source from which I can grab the default PublicKeyToken?

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
karajas: Will fix.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,61 @@
...
+                                                        new XAttribute("newVersion", assemblyVersion));
+                bindingRedirect = RemoveNamespace(bindingRedirect);
+                XElement dependentAssembly = new XElement("dependentAssembly",
+                                                          assemblyIdentity,
+                                                          bindingRedirect);
+                dependentAssembly = RemoveNamespace(dependentAssembly);
+                bindingRedirectAssemblies.Add(dependentAssembly);
+            }
+            XDocument doc = new XDocument(new XElement("configuration", new XElement("runtime", bindingRedirectAssemblies)));
+            foreach (ITaskItem executable in Executables)
```
karajas: I don't believe any of the corefx tests are built in this fashion. If we do want to include a local app.config, maybe I can add a override that can be passed through the new test csproj.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
karajas: I was having some difficulty understanding why empty namespaces were being added to the elements. Will try to find a cleaner way.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
AlexGhiondea: I don't think so. I was thinking more about skipping if the assembly does not have a publickeytoken.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -206,6 +209,28 @@
...
+        <_FilteredAssembliesForBindingRedirect Include="%(_FilteredListForBindingRedirect.SourcePath)" Condition="$([System.String]::Copy('%(_FilteredListForBindingRedirect.SourcePath)').EndsWith('.dll')) or $([System.String]::Copy('%(Remaining.SourcePath)').EndsWith('.exe'))" />
```
danmosemsft: This String.Copy trick is so horrible I opened a suggestion https://github.com/Microsoft/msbuild/issues/1465

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -206,6 +209,28 @@
...
+        <ExecutablesForBindingRedirect Include="%(IncludedFileForRunnerScript.SourcePath)" Condition="$([System.String]::Copy('%(IncludedFileForRunnerScript.SourcePath)').EndsWith('.exe'))"/>
```
danmosemsft: Suggestion, instead of all these item manipulation expressions, is it possible to just give @(ReferenceCopyLocalPaths) to the task ? It can filter out all but exe and dll, and can consult %(SourcePath) if needed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
karajas: I could probably just compare on Extension in that case.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,95 @@
...
+                    string assemblyVersion = assembly.GetMetadata("Version");
+                    string assemblyName = assembly.GetMetadata("Name");
+                    string culture = string.IsNullOrEmpty(assembly.GetMetadata("Culture"))
+                        ? "neutral"
+                        : assembly.GetMetadata("Culture");
+                    XElement assemblyIdentity = new XElement("assemblyIdentity",
+                        new XAttribute("name", assemblyName),
+                        new XAttribute("publicKeyToken", publicKeyToken),
+                        new XAttribute("culture", culture));
+                    assemblyIdentity = RemoveNamespace(assemblyIdentity);
```
ericstj: Use `ns +` to avoid this.

karajas:
Removed deduping from task and fixed up publishtests.targets to have a variable with the right test assets. PTAL.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public override bool Execute()
+        {
+            ns = "urn:schemas-microsoft-com:asm.v1";
+            XElement bindingRedirectAssemblies = new XElement(ns + "assemblyBinding");
+            List<XElement> dependentAssemblyList = new List<XElement>();
+            foreach (ITaskItem assembly in AssemblyIdentities)
+            {
+                string publicKeyToken = assembly.GetMetadata("PublicKeyToken");
+                if (!string.IsNullOrEmpty(publicKeyToken))
```
AlexGhiondea: I typically try to avoid indentation when not needed.

You would write this as:
```csharp
if (string.IsNullOrEmpty(publicKeyToken))
{
    // maybe log a warning
    continue;
}

// the rest of the code.
```

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,69 @@
...
+                    XElement bindingRedirect = new XElement(ns + "bindingRedirect",
+                        new XAttribute("oldVersion", $"0.0.0.0-{assemblyVersion}"),
+                        new XAttribute("newVersion", assemblyVersion));
+                    XElement dependentAssembly = new XElement(ns + "dependentAssembly",
+                        assemblyIdentity,
+                        bindingRedirect);
+                    dependentAssemblyList.Add(dependentAssembly);
+                }
+            }
+            foreach (XElement element in dependentAssemblyList)
```
AlexGhiondea: Why do you need this additional `foreach`? Why not directly add them to `bindingRedirectAssemblies` ?

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                bindingRedirectAssemblies.Add(element);
+            }
+            XDocument doc = new XDocument(new XElement("configuration", new XElement("runtime", bindingRedirectAssemblies)));
+            foreach (ITaskItem executable in Executables)
+            {
+                string executableName = Path.GetFileName(executable.ItemSpec);
+                using (FileStream fs = new FileStream(Path.Combine(OutputPath, executableName + ".config"), FileMode.Create))
```
AlexGhiondea: Any reason not to use the method that takes a string as the path to the file?

```csharp
doc.Save(Path.Combine(OutputPath, executableName + ".config"))
```

I am fine with either approach.

karajas: We're still dependent on a old version of XML, I can follow up on updating it.

karajas:
PTAL

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -129,6 +129,18 @@
...
       <DestinationsForTestDir Include="@(OverridenRuntimeContent->'$(TestPath)\%(Filename)%(Extension)')" />
     </ItemGroup>
+    
+    <!-- Removing assets from TestCopyLocal that come in from P2P references in SourcesToCopyToTestDir -->
+    <ItemGroup>
+        <_FileNamesInSourcesToCopyToTestDir Include="@(SourcesToCopyToTestDir->'%(Filename)%(Extension)')">
+          <SourcePath>%(SourcesToCopyToTestDir.Identity)</SourcePath>
+        </_FileNamesInSourcesToCopyToTestDir>
+        <_FileNamesInTestCopyLocalByFileNameWithoutDuplicates Include="@(_TestCopyLocalByFileNameWithoutDuplicates->'%(Filename)%(Extension)')">
+          <SourcePath>%(_TestCopyLocalByFileNameWithoutDuplicates.SourcePath)</SourcePath>
```
ericstj: This line does nothing.

ericstj: Moreover I don't think you need this item.  I believe _TestCopyLocalByFileNameWithoutDuplicates already has Identity of FileName+Extension and has SourcePath metadata.

karajas: removed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -206,6 +209,21 @@
       OutputListLocation="$(OutDir)assemblylist.txt"
       Condition="'$(UseLegacyXunitPerfRunner)'=='true'"
      />
+     
+    <!-- For Desktop Tests, we need to generate Binding Redirects. -->
+    <ItemGroup>
+      <_FilteredAssembliesForBindingRedirect Include="%(_DeduppedListOfTestAssets.SourcePath)" Condition="'%(_DeduppedListOfTestAssets.Extension)'=='.dll' or '%(IncludedFileForRunnerScript.Extension)'=='.exe'" />
```
karajas: typo will fix.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -129,6 +129,15 @@
...
       <DestinationsForTestDir Include="@(OverridenRuntimeContent->'$(TestPath)\%(Filename)%(Extension)')" />
     </ItemGroup>
+    
+    <!-- Removing assets from TestCopyLocal that come in from P2P references in SourcesToCopyToTestDir -->
+    <ItemGroup>
+        <_FileNamesInSourcesToCopyToTestDir Include="@(SourcesToCopyToTestDir->'%(Filename)%(Extension)')">
+          <SourcePath>%(SourcesToCopyToTestDir.Identity)</SourcePath>
+        </_FileNamesInSourcesToCopyToTestDir>
+        <_DeduppedListOfTestAssetsByFileName Include="@(_TestCopyLocalByFileNameWithoutDuplicates)" Exclude="@(_FileNamesInSourcesToCopyToTestDir)" />
+        <_DeduppedListOfTestAssetsByFileName Include="@(_FileNamesInSourcesToCopyToTestDir)" />
```
ericstj: /cc @MattGal now that we have a correct merged list of files we should look at removing the hacks from the test script and the VS target that only copy over if the file doesn't already exist.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -129,6 +129,15 @@
       <SourcesToCopyToTestDir Include="@(OverridenRuntimeContent)" />
       <DestinationsForTestDir Include="@(OverridenRuntimeContent->'$(TestPath)\%(Filename)%(Extension)')" />
     </ItemGroup>
+    
+    <!-- Removing assets from TestCopyLocal that come in from P2P references in SourcesToCopyToTestDir -->
+    <ItemGroup>
+        <_FileNamesInSourcesToCopyToTestDir Include="@(SourcesToCopyToTestDir->'%(Filename)%(Extension)')">
```
ericstj: nit: To match the convention this should also be named _SourceToCopyToTestDirByFileName

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
```diff
@@ -0,0 +1,65 @@
...
+                XElement dependentAssembly = new XElement(ns + "dependentAssembly",
+                    assemblyIdentity,
+                    bindingRedirect);
+                bindingRedirectAssemblies.Add(dependentAssembly);
+            }
+            XDocument doc = new XDocument(new XElement("configuration", new XElement("runtime", bindingRedirectAssemblies)));
+            foreach (ITaskItem executable in Executables)
+            {
+                string executableName = Path.GetFileName(executable.ItemSpec);
+                using (FileStream fs = new FileStream(Path.Combine(OutputPath, executableName + ".config"), FileMode.Create))
```
ericstj: Did you look at generating <testAssembly>.dll.config?  I believe that convention is supported by XUnit and would constrain the bindingRedirects to the isolated AppDomain that the tests run in.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
```diff
@@ -129,6 +129,15 @@
...
+          <SourcePath>%(SourcesToCopyToTestDir.Identity)</SourcePath>
+        </_FileNamesInSourcesToCopyToTestDir>
+        <_DeduppedListOfTestAssetsByFileName Include="@(_TestCopyLocalByFileNameWithoutDuplicates)" Exclude="@(_FileNamesInSourcesToCopyToTestDir)" />
```
ericstj: nit: to me a more useful name might be "finalTestAssetsByFileName".  Something to make it clear that this is the list that folks ought to rely on for finding all the final test assemblies.  Who knows if that will actually prevent folks from further forking this in the future... but at least its an attempt.

src/Microsoft.DotNet.Build.Tasks/GenerateBindingRedirect.cs
karajas: I did try a version with this, I couldn't determine whether the redirect was being used from fuslog. It should be just adding the dlls to the ExecutablesForBindingRedirect itemgroup

karajas: This would be applicable only for the Test.dll.config I believe? Should I go ahead and do that instead of xunit.console.exe.config?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/publishtest.targets
naamunds: @ericstj See #1228.

karajas:
This still has an issue with filtering native assemblies

PR #1258, opened 2016-12-10
Add Win32 API support to PInvoke Analyzer

JohnChen0:
Add two new API lists that use Win32 DLL names instead of API set names. This commit is needed for dotnet/corefx#13738.


gkhanna79:
@JohnChen0 Can you please remind me why we need two lists? What is the difference between them?

JohnChen0:
@gkhanna79 One of the list is used to check desktop apps, the other list is used to check UWP apps.

JohnChen0:
@weshaggard Could you please take a look at this PR and let me know if you have any concerns? I have changes in other repos that depend on this PR, so I'd like to get this through quickly if possible. Thanks.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
```diff
@@ -18,7 +18,9 @@
     -->
 
     <ValidPinvokeMappings>$(MSBuildThisFileDirectory)PinvokeAnalyzer_OneCoreApis.txt</ValidPinvokeMappings>
+    <ValidPinvokeMappings Condition="'$(UseWin32Apis)'=='true'">$(MSBuildThisFileDirectory)PinvokeAnalyzer_Win32Apis.txt</ValidPinvokeMappings>
```
weshaggard: Where do we expect UseWin32Apis to be set?

weshaggard:
Other then the one question the change looks good to me.

src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets
JohnChen0: This will be set by the project that uses the buildtools. For example, in dotnet/corefx, this will be set in https://github.com/dotnet/corefx/blob/master/dir.props by an upcoming PR.

Eventually, if most or all projects start using this option, we may consider making it the default. However to avoid a breaking change, for now projects need to opt in.

weshaggard: sounds reasonable.

PR #1266, opened 2016-12-15
Fix GenerateTestBindingRedirects merge

naamunds:
Revert GenerateTestBindingRedirects to use _CombinedListOfTestAssets and
remove the Conditions from the task calls.  The only intended changes to
this target in #1263 were to add the call to FilterNativeAssemblies and
use the output of that as the input to GetAssemblyIdentity.

/cc @karajas @MattGal @ericstj


karajas:
LGTM

karajas:
@naamunds Actually I have another fix that I can add to this that removes FilterNativeAssemblies

PR #1270, opened 2016-12-17
Add checked-in VSTS Release functionality to VstsBuildsApi

dagood:
Adds ability to create/update VSTS Release definitions using VstsBuildsApi. To be used for symbol auto-archive.

Refactors `VstsBuildClient` into `VstsDefinitionClient` to do the same sorts of operations over the two APIs/schemas. Unfortunately git didn't count this as a move.

I've tested build and release definition creating/updating on my own VSTS instance so far. I'd like to at *least* test the build definition side of this on the DevDiv VSTS instance with the CoreFX build definitions before merging.

@chcosta could you take a look?


src/Microsoft.DotNet.Build.VstsBuildsApi/JsonHttpClient.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// is because we can not properly handle "non-authoritative" responses which are reported
+        /// as "success".
+        /// </summary>
+        private static void ProcessResponseStatusCode(HttpResponseMessage response)
+        {
+            if (response.StatusCode == System.Net.HttpStatusCode.OK)
+            {
+                return;
+            }
+            throw new HttpRequestException(string.Format("Response code {0} received from {1} is not a valid response.", response.StatusCode, response.RequestMessage.RequestUri));
```
MichaelSimons: >string.Format("Response code {0} received from {1} is not a valid response.", response.StatusCode, response.RequestMessage.RequestUri [](start = 43, length = 133)

Consider using an interpolated string.  I find it to be more readable and succinct.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsDefinitionClient.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="collectionIdentifier">collection identifer for build and release definitions</param>
+        /// <param name="credentials">credentials used for basic rest api authentication</param>
+        public VstsDefinitionClient(string collectionIdentifier, string credentials)
+        {
+            if (string.IsNullOrWhiteSpace(collectionIdentifier))
+            {
+                throw new Exception("Required parameter (collectionIdentifier) cannot be null or white space.");
```
MichaelSimons: >Exception [](start = 26, length = 9)

Seems like these exceptions should be a type of ArgumentException.

src/Microsoft.DotNet.Build.VstsBuildsApi/JsonHttpClient.cs
dagood: Ah, yeah, I'll change that. A lot of this PR is refactoring `VstsBuildClient.cs` and I didn't tend to change things like this. Too bad the diff didn't detect any of these changes as a move.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsDefinitionClient.cs
```diff
@@ -0,0 +1,178 @@
...
+                if (!IsDefinitionContentSubsetEquivalent(client, definition, vstsDefinition))
+                {
+                    CopyIdentifiableInformation(client, vstsDefinition, definition);
+                    vstsDefinition = await client.VstsUpdateDefinitionAsync(definition).ConfigureAwait(false);
+                }
+                return vstsDefinition["id"].ToString();
+            }
+            else
+            {
+                throw new Exception(string.Format(
```
MichaelSimons: >Exception [](start = 26, length = 9)

Consider something like throwing an InvalidOperationException.

src/Microsoft.DotNet.Build.VstsBuildsApi/JsonHttpClient.cs
MichaelSimons: I realized that after looking at the whole change :)

---
In reply to: [93263197](https://github.com/dotnet/buildtools/pull/1270#discussion_r93263197) [](ancestors = 93263197)

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsDefinitionClient.cs
```diff
@@ -0,0 +1,178 @@
...
+                /* Update */
+                if (!IsDefinitionContentSubsetEquivalent(client, definition, vstsDefinition))
+                {
+                    CopyIdentifiableInformation(client, vstsDefinition, definition);
+                    vstsDefinition = await client.VstsUpdateDefinitionAsync(definition).ConfigureAwait(false);
+                }
+                return vstsDefinition["id"].ToString();
+            }
+            else
+            {
```
MichaelSimons: Given the way the code is returning in the previous if blocks, the else is not needed and adds a small degree of confusion.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
```diff
@@ -0,0 +1,95 @@
...
+
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Build.VstsBuildsApi
+{
+    internal class VstsBuildHttpClient : IVstsDefinitionHttpClient
```
MichaelSimons: >VstsBuildHttpClient [](start = 19, length = 19)

Noticeable amount of duplicated logic between this and the release client.  What were the deciding factors in using an interface vs a base class?

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsReleaseHttpClient.cs
```diff
@@ -0,0 +1,133 @@
...
+                .Where(property => property != null);
+        }
+
+        private static string GetRequestUri(JObject definition, string command, Dictionary<string, string> args = null)
+        {
+            if (args == null)
+            {
+                args = new Dictionary<string, string>();
+            }
+            args["api-version"] = ApiVersion;
```
chcosta: May consider applying this same pattern to the Builds definitions

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsReleaseHttpClient.cs
```diff
@@ -0,0 +1,133 @@
...
+            return await _jsonClient.PostAsync(requestUri, definition);
+        }
+
+        public async Task<JObject> VstsRetrieveDefinitionByIdAsync(JObject definition)
+        {
+            string requestUri = GetRequestUri(definition, $"definitions/{definition["id"]}");
+            return await _jsonClient.GetAsync(requestUri);
+        }
+
+        public async Task<IReadOnlyList<JObject>> VstsRetrieveDefinitionsListByNameAndPathAsync(JObject definition)
```
chcosta: This doesn't actually take into account the path, so that may be a little confusing (I don't know if Path is a concept for Release definitions).  In the Builds definitions, there needs to be some refactoring, we currently retain the Path in the local json definitions, but you can't actually create or update a definition with the Path information, and we strip it out when creating / updating.  We should consider renaming this to `VstsRetrieveDefinitionsListByName` and explicitly setting the path to "\\" to prevent https://github.com/dotnet/core-eng/issues/163.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
```diff
@@ -0,0 +1,95 @@
...
+                        definition["path"].ToString() == childObject["path"].ToString())
+                    {
+                        definitions.Add(childObject);
+                    }
+                }
+            }
+            return definitions;
+        }
+
+        public IEnumerable<JProperty> FindIdentifiableDefinitionProperties(JObject definition)
```
chcosta: internal?

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
```diff
@@ -0,0 +1,95 @@
...
+            return await _jsonClient.PostAsync(requestUri, definition);
+        }
+
+        public async Task<JObject> VstsRetrieveDefinitionByIdAsync(JObject definition)
+        {
+            string requestUri = $"DefaultCollection/{VstsBuildsApiBase(definition)}/_apis/build/definitions/{definition["id"].ToString()}?api-version=2.0";
+            return await _jsonClient.GetAsync(requestUri);
+        }
+
+        public async Task<IReadOnlyList<JObject>> VstsRetrieveDefinitionsListByNameAndPathAsync(JObject definition)
```
chcosta: Are you expecting these methods to be called publically?

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsReleaseHttpClient.cs
```diff
@@ -0,0 +1,133 @@
...
+            "createdBy",
+            "createdOn",
+            "deployStep",
+            "modifiedBy",
+            "modifiedOn",
+            "owner",
+            "id",
+            "lastRelease",
+            "revision",
+        };
```
chcosta: I wonder if we should move these (and the ApiVersion) outside of the classes and into an external file, I guess it's not clear how / if this information would change in a different api version, but editing a config file would be preferable to recompiling or ensuring you had the right toolset for the right api version.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsDefinitionClient.cs
dagood: Agreed, removing.

---
In reply to: [93265445](https://github.com/dotnet/buildtools/pull/1270#discussion_r93265445) [](ancestors = 93265445)

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsDefinitionClient.cs
dagood: Kind of a weird one because it depends on what's on the server more than the state of the client itself (as `InvalidOperationException` docs suggest). I don't know of anything for this situation, so IOE sounds fine to me.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
dagood: No good reason. I think I may have originally had the clients inherit `JsonHttpClient` and didn't rethink the interface when I changed that. An abstract base class does sound better.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsReleaseHttpClient.cs
dagood: Good idea, I'll do that while bringing stuff into the common base class.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsReleaseHttpClient.cs
chcosta: Path is stripped because it's not supported for creating or updating a build definition.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsReleaseHttpClient.cs
dagood: I think it makes sense to make this configurable, but if we need a new API version I'd think there would be more changes than just the array and version strings so I don't know if just a file makes sense for that scenario.

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsReleaseHttpClient.cs
dagood: Ah, dang, I guess my personal account's VSTS instance is a bit ahead then. I'll test Release creation very soon on devdiv.visualstudio.com to make sure that isn't different too.

src/Microsoft.DotNet.Build.VstsBuildsApi/JsonHttpClient.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Build.VstsBuildsApi
+{
+    internal class JsonHttpClient
+    {
+        private HttpClient _client;
+
+        public JsonHttpClient(Uri baseAddress, string credentials)
```
dagood: @chcosta I didn't change the `public` stuff in `internal` classes to `internal` because the class being internal means everything inside will implicitly be internal. If the class ever needs to be public, it reduces the changes needed. Does that sound ok?

dagood:
@MichaelSimons @chcosta Could you take another look?

dagood:
Thanks! Merging, and I'll be trying out the official package in PipeBuild next.

PR #1272, opened 2016-12-17
Add usage documentation for test targets

naamunds:
I'm planning to add a separate page documenting the test target design.

/cc @weshaggard @maririos @MattGal

### DIFF
--- a/Documentation/test-targets-usage.md
+++ b/Documentation/test-targets-usage.md
@@ -1,0 +1,103 @@
+# Test Targets
+
+BuildTools provides test targets and related logic for publishing and running tests.  The primary entry points are the [`Test` target](https://github.com/dotnet/buildtools/blob/87422f6cb8/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L340) in Microsoft.DotNet.Build.Tasks and the [`CloudBuild` target](https://github.com/dotnet/buildtools/blob/87422f6cb8/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/CloudTest.targets#L79) in Microsoft.DotNet.Build.CloudTestTasks.  The `Test` target is used to run or archive tests locally, and the `CloudBuild` target is used to run tests in Helix.
+
+## Usage scenarios
+
+*Tips:*
+- If building in a non-Windows environment, call `<repo-root>/Tools/msbuild.sh` instead of just `msbuild`.
+- Dotnet repos typically provide `BuildAndTest` and `RebuildAndTest` targets which can be used to build and run tests from a single command, so you may wish to substitute either of those targets for `Test` in the examples below.
+ 
+#### Run tests for a project with the default options
+
+The following command runs tests for System.Collections.Immutable.Tests.csproj using the default options.
+```
+msbuild /t:Test System.Collections.Immutable.Tests.csproj
+```
+
+#### Run a single XUnit method
+
+You can use the `XUnitOptions` property to override the options used for the XUnit command that runs the tests.  For example, the following command runs only the `System.Security.Cryptography.Pkcs.Tests.CmsRecipientCollectionTests.Twoary` method.
+
+```
+msbuild /t:Test "/p:XunitOptions=-method System.Security.Cryptography.Pkcs.Tests.CmsRecipientCollectionTests.Twoary" System.Security.Cryptography.Pkcs.Tests.csproj
+```
+
+#### Debug tests in Visual Studio
+
+1.  Open the project's solution in Visual Studio.
+2.  Set the test project as the Startup Project.
+3.  Build (or rebuild) the solution.
+4.  Press F5 to debug the project.
+
+To modify the XUnit command which will be executed when debugging, edit the *Command line arguments* in the *Debug* section of the test project's properties, for example to debug just a single test method.  Note that these changes will be overwritten when rebuilding the project.
+
+#### Run tests for a specified TFM and/or RID
+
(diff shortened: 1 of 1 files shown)


Documentation/test-targets-usage.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+#### Build and run tests against packages
+
+By default, the test project's direct dependencies (i.e. assemblies for the projects to which the test project has a ProjectReference, such as the corresponding product project) are copied from each referenced project's `$(TargetPath)`.  The `BuildTestsAgainstPackages` property provides the option to instead copy these references from a packages directory.
+```
+msbuild /t:BuildAndTest /p:BuildTestsAgainstPackages=true /p:"PackagesDrops=C:\Repos\corefx\bin\packages\Debug" System.Collections.Immutable.Tests.csproj
+```
+
+[Building Tests Against Packages](https://github.com/dotnet/corefx/blob/master/Documentation/building/build-tests-against-packages.md) has more information about doing so in CoreFX.
```
naamunds: @karajas @chcosta Is there a better way to describe the BuildTestsAgainstPackages scenario?  Is this missing anything?

Documentation/test-targets-usage.md
```diff
@@ -0,0 +1,108 @@
...
+#### Run tests for a project with the default options
+
+The following command runs tests for System.Collections.Immutable.Tests.csproj using the default options.
+```
+msbuild /t:Test System.Collections.Immutable.Tests.csproj
+```
+
+#### Run a single XUnit method
+
+You can use the `XUnitOptions` property to override the options used for the XUnit command that runs the tests.  For example, the following command runs only the `System.Security.Cryptography.Pkcs.Tests.CmsRecipientCollectionTests.Twoary` method.
```
weshaggard: As a follow-up it might be worth filing an issue to make this option easier.

Documentation/test-targets-usage.md
```diff
@@ -0,0 +1,108 @@
...
+1.  Open the project's solution in Visual Studio.
+2.  Set the test project as the Startup Project.
+3.  Build (or rebuild) the solution.
+4.  Press F5 to debug the project.
+
+You can modify the XUnit command which will be executed when debugging by editing the *Command line arguments* in the *Debug* section of the test project's properties, for example to debug just a single test method.  Note that these changes will be overwritten when rebuilding the project.
+
+#### Run tests for a specified TFM and/or RID
+
+You can specify the target framework moniker (TFM) on which to run tests by using the `TestTFM` property.  Similarly, the runtime ID (RID) can be specified with the `TestNugetRuntimeId` property.
```
weshaggard: Once we finishing our current engineering work "TestTFM" and "TestNugetRuntimeId" will be obsolete properties, but they are still useful for now. It would be good if you can begin understanding the replacements and so you can help update these docs as well as ensure we aren't missing any scenarios.

Documentation/test-targets-usage.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Run tests in Helix
+
+The `CloudBuild` target can be used to queue a test run in Helix.
+
+Here is an example based on [the CoreFX official build](https://github.com/dotnet/corefx/blob/021c590e6cb166c4bfc62b2c9966e317c37c1ed6/buildpipeline/DotNet-CoreFx-Trusted-Windows-Build-Test.json#L215-L216):
+```
+msbuild src\post.builds /t:CloudBuild /p:"BuildTestsAgainstPackages=true" /p:"ConfigurationGroup=Debug" /p:"ArchiveTests=true" /p:"EnableCloudTest=true" /p:"BuildMoniker=[build_moniker]" /p:"TargetQueue=Windows.10.Amd64" /p:"TestProduct=corefx" /p:"Branch=master" /p:"CloudDropAccountName=[cloud_drop_account_name]" /p:"CloudResultsAccountName=[cloud_results_account_name]" /p:"CloudDropAccessToken=[cloud_drop_access_token]" /p:"CloudResultsAccessToken=[cloud_results_access_token]" /p:"Outerloop=true" /p:"FilterToOSGroup=Windows_NT" /p:"TimeoutInSeconds=1200" /p:"TargetOS=Windows_NT" /p:"HelixApiAccessKey=[helix_key]" /p:"CloudTestContainerSuffix=windows10ilc" /p:"OverwriteOnUpload=true" /p:"TargetsWindows=true" /flp:"LogFile=pushtohelix-diag.log;Verbosity=diagnostic" /p:HelixApiEndpoint=https://helix.dot.net/api/2016-06-28/jobs
```
weshaggard: I think I would avoid copying the command line here as you already provided a link which is good. Having the command line will cause people to be tempted to try and copy/paste it.

weshaggard:
I think this is a good start. One scenario I think is worth calling out is how to do these same things on non-windows. Even it if is just a key that says replace msbuild with ./tools/msbuild.sh call.

Documentation/test-targets-usage.md
naamunds: I've removed this section in light of that.

Documentation/test-targets-usage.md
naamunds: Good point; changed as suggested.

Documentation/test-targets-usage.md
weshaggard: We won't have any notion of a TestTFM the BuildConfiguration you are building for will essentially be the TestTFM. For the TestNugetRuntimeId we be replaced with BuildConfiguration and RuntimeOS properties.

PR #1282, opened 2016-12-22
Add ARM64 to nuspec generation list

ianhays:
This is to unblock building managed packages for arm64 alongside the native ones so we don't have to do two separate builds to get the entire package set that we need for testing. Without this change, I have to do full build-managed.cmd's - one with /p:Platform=arm64 and one without.

@karajas @mellinoe this resolves the issue I had yesterday.


ianhays:
nvm, worked around it.

karajas:
So the reason we require this is to preserve the same build.cmd right? Because if you had passed in Platform to build-managed

ianhays:
Sorry, I got my builds mixed up. This is not actually required to include the native packages in the package part of the build-managed.cmd build.

The reason I thought that was because I had not done a build-native.cmd since deleting my bin folder, so build-managed didn't build those packages since there wasn't any dll to package up. I was incorrect in thinking that `/p:Platform=arm64` was required to build the arm64 packages.

PR #1286, opened 2017-01-04
Update shared framework to 1.1.0, change CLI version to 1.0.0-preview2-1-003182

dagood:
Upgrades buildtools shared framework version to 1.1.0. Helps with https://github.com/dotnet/corefx/issues/14215 by letting CoreFX use CLI version `1.0.0-preview2-1-003182.`

If I stopped there, buildtools would be blocked from uptaking a newer version of itself because bootstrap.sh/ps1 don't support fetching CLI `1.0.0-preview2-1-003182` to get the 1.1.0 framework. To fix this I added `DotNetInstallBranch` that configures which CLI branch dotnet-install comes from. To uptake a new buildtools package, the CLI version will have to be changed and `-DotNetInstallBranch "rel/1.0.0-preview2.1"` passed into bootstrap.ps1 by run.ps1.

~~I also added `invoke_annotated` to bootstrap.sh to remove duplicated code.~~

@crummel @ellismg PTAL
/cc @naamunds


dagood:
@crummel I removed the unnecessary parts, could you take a look?

PR #1287, opened 2017-01-04
If no "Latest_Packages" versions repository file exists, create one

dagood:
This allows a build to succeed on a branch that doesn't have a build-info entry. Without this change, the build would fail to find an existing entry and fail with a 404.

Implemented by catching the 404 in the specific case of requesting file contents, and returning `null` contents. A GitHub 404 can also be an authentication error, so I don't want to catch it very broadly.

@StephenBonikowsky This will fix the publish build problems you're hitting with manual builds of new WCF branches.

@weshaggard @gkhanna79 This will also let us stop having to manually create blank build-info files when setting up new release branch builds.

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
@@ -199,7 +199,6 @@ namespace Microsoft.DotNet.VersionTools.Automation
 
             if (latestPackages == null)
             {
-                Trace.TraceInformation($"No versions repository file '{path}' found.");
                 return null;
             }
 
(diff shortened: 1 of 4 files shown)


StephenBonikowsky:
Thanks Davis.

src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
```diff
@@ -188,6 +197,12 @@ public VersionsRepoUpdater(GitHubAuth gitHubAuth, GitHubProject project)
                 path,
                 new GitHubBranch("master", _project));
 
+            if (latestPackages == null)
+            {
+                Trace.TraceInformation($"No versions repository file '{path}' found.");
```
naamunds: What do you think about combining this trace statement with the one on lines 119-121?

dagood: This one's redundant--I guess I was thinking that it would be good to have this info if we use `GetPackagesAsync` again for something else and don't log the null case at that call site. I don't think that's likely, so I'll remove it.

PR #1289, opened 2017-01-06
PInvoke Analyer changes need for CoreCLR

JohnChen0:
Two changes that are required for running PInvoke Analyer in CoreCLR repo:
1. Ignore QCall (internal calls from BCL to runtime)
2. Add some APIs used by System.Private.CoreLib.dll to allowed list


JohnChen0:
cc @AlexGhiondea @gkhanna79 @weshaggard

PR #1294, opened 2017-01-10
Uptake the latest buildtools package, and new CLI

dagood:
The buildtools package uses shared framework 1.1.0 instead of 1.0.0, so switch to a matching CLI package. This is the same CLI version as CoreFX and CoreCLR use.

See https://github.com/dotnet/buildtools/pull/1286

~~Known issues: there are binclashes, and a test failure. I'm submitting this PR as-is as a starting point for making this work. Hopefully anyone who needs to do this upgrade will see this PR first and avoid wasted effort.~~ fixed


dagood:
I couldn't track down what buildtools change started the assumption that `ConfigurationGroup` is always set as a msbuild property arg, but I was able to figure out what was different between buildtools and corefx. corefx's `config.json` explicitly calls out `ConfigurationGroup` in the `defaultValues` block. This caused binclashes because msbuild thought the projectreference wasn't the same configuration as the direct project build, so they would both build and clash.

dagood:
@dotnet-bot test this please

PR #1300, opened 2017-01-12
Update xunit refs to beta4

jamesqo:
This updates all references to xunit to beta4, so the runner can take advantage of some of the bugs fixed since beta2. (e.g. This came up in https://github.com/dotnet/corefx/pull/14798#discussion_r95482278 where a [fixed xunit bug](https://github.com/xunit/assert.xunit/pull/5) was not present in the corefx runner, necessitating ugly workarounds.) I had to suppress the warnings for some types that got obsoleted for performance reasons in a few places; initially I tried to fix the warnings and update those parts to the new xunit codebase, but lots of code has changed since the fork in 2014 so I figured it was better to keep things as they are.

**Related:** https://github.com/dotnet/corefx/pull/14798

/cc @ianhays, @mellinoe


src/common/TestDiscoveryVisitor.cs
```diff
@@ -1,26 +0,0 @@
-﻿using System.Collections.Generic;
```
jamesqo: I had to remove this file because it started conflicting with https://github.com/xunit/xunit/blob/master/src/xunit.runner.utility/Visitors/TestDiscoveryVisitor.cs at the call sites.

ianhays: Is it even used anywhere?

ianhays:
Have you tested CoreFX/CoreCLR with a buildtools version produced with these changes yet?

jamesqo:
@ianhays 

> Have you tested CoreFX/CoreCLR with a buildtools version produced with these changes yet?

Are there instructions on how to do that? I couldn't find any after a quick scan through https://github.com/dotnet/buildtools/tree/master/Documentation, but maybe I missed something.

ianhays:
@jamesqo I don't believe there are any docs, no. I used to just build BuildTools and replace the binaries in the packages in CoreFX with those, but that pipeline may have changed with all of the engineering changes that have gone through recently.

@mellinoe may know more.

hughbe:
Looks like I duplicated your work in #1319 (I've closed that one). Would be good to get this one merged in

danmosemsft:
@safern is this PR redundant now?

safern:
> @safern is this PR redundant now?

Yes, we should definitely update xunit's versions but we should try to keep the same version in Corefx and buildtools. 

This is updating to 2.2.0-beta4, in Corefx we are already in [2.3.0-beta1](https://github.com/dotnet/corefx/blob/master/dependencies.props#L138). I would do it on a new PR as buildtools has change quite a lot since this PR as we can see in the merge conflicts.

jamesqo:
@safern OK, I will close this PR then. Unfortunately, I will not have the time to open a new one so that is up to you guys (or someone from the community who wants to pitch in).

Issue #1308, opened 2017-01-26
Init-tools.sh fails on case-insensitive volume


weshaggard:
When using docker for windows and mounting a windows volume to share he source between a linux docker container and my windows machine I get the following error when running init-tools.

```
Installing dotnet cli...
Initializing BuildTools...
mv: '/d/git/corefx/Tools/Microsoft.CSharp.targets' and '/d/git/corefx/Tools/Microsoft.CSharp.Targets' are the same file
ERROR: An error occured when trying to initialize the tools. Please check '/d/git/corefx/init-tools.log' for more details.
```

cc @chcosta @joperezr 

@joperezr it looks like you added this https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh#L105 but we never got this hack cleaned up. Do you know if it is still necessary?

joperezr:
Seems like so, we get that targets file from the `Microsoft.Build.Targets` package, and it seems that in the one we are using they still have a lower case `t` on `Microsoft.CSharp.targets`. The package we are using though is very old, so it might be that they have fixed this some time ago but we just haven't upraded the version of the package.

weshaggard:
At least on my ubuntu docker image running on windows commenting out the mv command allows the build to continue and succeed. However that is likely an artifact of the windows volume that I've mounted.

joperezr:
yeah that might be it. The problem is that the Portable targets (Microsoft.Portable.CSharp.targets) have the following line:
```xml
<Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.Targets" />
```
given they use a capital T, then full linux systems wouldn't find the targets file.

PR #1310, opened 2017-01-26
Ignore errors in mv command at the end of init-tools.sh

weshaggard:
When running from volumes that aren't case-sensitive like a windows
volumne mounted in a linux docker container, the mv command fails.
To help eliminate issues we will just ignore the error code for
the mv command and if it fails it will print out a message but
won't fail the intitialization of tools.

Fixes https://github.com/dotnet/buildtools/issues/1308

cc @joperezr


mellinoe:
Can you just check if the capitalized version doesn't exist before trying the copy?

weshaggard:
@mellinoe yes I guess I could do that as well.

joperezr:
LGTM

PR #1314, opened 2017-02-01
Update MSBuild package versions and account for breaking changes.

mellinoe:
We are using a very old version of .NET Core MSBuild (from last June). This commit updates the versions we are using. The "MSBuild" and "Microsoft.Build.Targets" packages have been replaced by the "Microsoft.Build.Runtime" package, which I've switched to. There are a couple of path-related breaking changes that I needed to react to in this change, as well.

All of these new packages are available on nuget.org.

@weshaggard @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <RoslynVersion>2.0.0-rc</RoslynVersion>
     <RoslynPackageName>Microsoft.Net.Compilers</RoslynPackageName>
+    <RoslynTargetsPath>$(ToolRuntimePath)</RoslynTargetsPath>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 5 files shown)


danmosemsft:
When I type "msbuild" in my corefx repo, it will still use the one installed by VS, right? We've talked in the past about changing it to use the same xplat one we build with on Unix, presumably by having build.cmd set up an alias.

mellinoe:
> When I type "msbuild" in my corefx repo, it will still use the one installed by VS, right?

If you have run the "Visual Studio Developer Command Prompt" batch script, then yes, it will run the one installed by VS. Otherwise, it will still not do anything.

> changing it to use the same xplat one we build with on Unix

This is part of enabling that scenario, although I don't think we will enable it for a while. I can run the full managed build with my local changes on .NET Core MSBuild on Windows, but the native build is still problematic. We rely on the VC++ targets and compiler on Windows, and there's no easy way for us to enable that with the .NET Core MSBuild as far as I know.

weshaggard:
Do you know if we see any performance changes? + or -?

mellinoe:
I didn't see that there was much of a difference either way, but we should get some bug fixes in here.

PR #1317, opened 2017-02-03
Update to run on Jenkins

DrewScoggins:
I have ripped out most of the guts that were in the PerfTesting.targets
file and moved everything over to run using the dotnet cli and the new
dev/eng model.  This includes removing all of the previous goo that
helped with loading the analysis and runner exes as those have been
replaced by the PerfRunner.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
@@ -1,105 +1,42 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
-  <!-- Perf Runner NuGet package paths -->
-  <PropertyGroup Condition="'$(TargetOS)'=='Linux'">
-    <PerfDotNetCliDir>$PACKAGE_DIR/perf.dotnetcli</PerfDotNetCliDir>
-    <PerfDotNetCliInstaller>$(ToolsDir)/RunnerScripts/xunitrunner-perf/ubuntu-dotnet-local-install.sh</PerfDotNetCliInstaller>
-    <XunitPerfRunnerPackageId>Microsoft.DotNet.xunit.performance.runner.cli</XunitPerfRunnerPackageId>
-  </PropertyGroup>
-  <!-- Perf Runner NuGet package paths -->
-  <PropertyGroup Condition="'$(TargetOS)'=='Windows_NT'">
-    <XunitPerfRunnerPackageId>Microsoft.DotNet.xunit.performance.runner.Windows</XunitPerfRunnerPackageId>
-  </PropertyGroup>
-
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Perf Analysis NuGet package paths -->
-  <PropertyGroup Condition="'$(TargetOS)'=='Linux'">
-    <XunitPerfAnalysisPackageId>Microsoft.DotNet.xunit.performance.analysis.cli</XunitPerfAnalysisPackageId>
-    <XunitPerfAnalysisPath>$(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/lib/netstandard1.3/Microsoft.DotNet.xunit.performance.analysis.cli.dll</XunitPerfAnalysisPath>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetOS)'=='Windows_NT'">
-    <XunitPerfAnalysisPackageId>Microsoft.DotNet.xunit.performance.analysis</XunitPerfAnalysisPackageId>
-    <XunitPerfAnalysisPath>$(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/tools/xunit.performance.analysis.exe</XunitPerfAnalysisPath>
-    <XunitPerfAnalysisDir>$(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/tools</XunitPerfAnalysisDir>
-  </PropertyGroup>
-
-
-  <!-- Installing dotnet cli version that can execute the xunit perf test runner -->
-  <Target Name="InstallDotnetCliForPerfExecution" Condition="'$(TargetOS)'=='Linux' AND '$(Performance)'=='true' AND !Exists('$(PerfDotNetCliDir)/dotnet') AND '$(OsEnvironment)'=='Unix'" AfterTargets="CopyXunitExecutionFiles" >
-    <Exec Command="chmod +x $(PerfDotNetCliInstaller)" />
-    <Exec Command="$(PerfDotNetCliInstaller) -d $(PackagesDir)/perf.dotnetcli -v $(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/lib/netstandard1.3/DotNetCliVersion.txt" ContinueOnError="true"/>
-  </Target>
-
-  <ItemGroup>
-    <XunitPerfRunnerFiles Include ="$(PackagesDir)$(XunitPerfRunnerPackageId)/**/*.*" />
-    <XunitPerfAnalysisFiles Include ="$(PackagesDir)$(XunitPerfAnalysisPackageId)/**/*.*" />
-  </ItemGroup>
-
-  <!-- Copy over the performance runners to the test directory-->
-  <Target Name="CopyXunitExecutionFiles" AfterTargets="CopyTestToTestDirectory" Condition="'$(Performance)' == 'true'">
-    <Error Condition="'$(XunitPerfAnalysisPackageVersion)' == ''" Text="Xunit performance package version was not specified" />
-    <Error Condition="!Exists('$(XunitPerfAnalysisPath)')" Text="Xunit performance package was not found" />
-    <Copy SourceFiles="$(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/lib/netstandard1.3/DotNetCliVersion.txt"
-          DestinationFolder="$(TestWorkingDir)SupplementalPayload/RunnerScripts/xunitrunner-perf"
-          SkipUnchangedFiles="true"
-          Condition="'$(TargetOS)'=='Linux'"/>
-    <!-- The packages Microsoft.DotNet.xunit.performance.runner.Windows and Microsoft.DotNet.xunit.performance.analysis is not being uploaded along with Packages.zip, hence the workaround -->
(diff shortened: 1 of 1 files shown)


DrewScoggins:
PTAL @mellinoe @dpodder @jorive

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -1,123 +1,67 @@
...
-  <!-- Installing dotnet cli version that can execute the xunit perf test runner -->
-  <Target Name="InstallDotnetCliForPerfExecution" Condition="'$(TargetOS)'=='Linux' AND '$(Performance)'=='true' AND !Exists('$(PerfDotNetCliDir)/dotnet') AND '$(OsEnvironment)'=='Unix'" AfterTargets="CopyXunitExecutionFiles" >
-    <Exec Command="chmod +x $(PerfDotNetCliInstaller)" />
-    <Exec Command="$(PerfDotNetCliInstaller) -d $(PackagesDir)/perf.dotnetcli -v $(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/lib/netstandard1.3/DotNetCliVersion.txt" ContinueOnError="true"/>
+  
+  <ItemGroup>
+	<TraceEventNativePath Include="$(PackagesDir)\$(TraceEventPackage)\lib\native\**\*.*" />
+  </ItemGroup>
+  
+  <Target Name ="PublishPerfRunner" Condition="'$(Performance)'=='true'" BeforeTargets="RunTestsForProject">
```
mellinoe: This is kind of unusual; can we just include the sources here and bundle the .exe file and its dependencies into the buildtools package, like we do for the other tool projects here?

DrewScoggins: I think we can solve this is two parts.  The first is moving the PerfRunner to work like RemoteExecutorConsoleApp.  This will remove the bulk of the goo that we are doing here.  We will still however need to copy over the native portions of TraceEvent, the last step in this target.  The way that the code currently works we need to have the native folders in the directory of where the TraceEvent code is being executed ( @brianrob or @vancem please correct me if I am wrong.

brianrob: Yes, you do need to have the native dependencies present next to TraceEvent to control ETW.

mellinoe: Ok, the last part (copying native portions of TraceEvent) look fine to me.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -1,123 +1,61 @@
...
-    <XunitPerfRunnerPackageId>Microsoft.DotNet.xunit.performance.runner.Windows</XunitPerfRunnerPackageId>
-  </PropertyGroup>
-
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Perf Analysis NuGet package paths -->
-  <PropertyGroup Condition="'$(TargetOS)'=='Linux'">
+  <PropertyGroup>
     <XunitPerfAnalysisPackageId>Microsoft.DotNet.xunit.performance.analysis.cli</XunitPerfAnalysisPackageId>
     <XunitPerfAnalysisPath>$(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/lib/netstandard1.3/Microsoft.DotNet.xunit.performance.analysis.cli.dll</XunitPerfAnalysisPath>
+	<TraceEventPackage>Microsoft.Diagnostics.Tracing.TraceEvent\1.0.0-alpha-experimental</TraceEventPackage>
```
weshaggard: nit: Looks like you have  number of places where you are doing tabs or 4 spaces where the rest of the file and the guidelines is 2 spaces.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -1,123 +1,61 @@
...
-  <PropertyGroup Condition="'$(TargetOS)'=='Windows_NT'">
-    <XunitPerfRunnerPackageId>Microsoft.DotNet.xunit.performance.runner.Windows</XunitPerfRunnerPackageId>
-  </PropertyGroup>
-
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Perf Analysis NuGet package paths -->
-  <PropertyGroup Condition="'$(TargetOS)'=='Linux'">
+  <PropertyGroup>
     <XunitPerfAnalysisPackageId>Microsoft.DotNet.xunit.performance.analysis.cli</XunitPerfAnalysisPackageId>
     <XunitPerfAnalysisPath>$(PackagesDir)$(XunitPerfAnalysisPackageId)/$(XunitPerfAnalysisPackageVersion)/lib/netstandard1.3/Microsoft.DotNet.xunit.performance.analysis.cli.dll</XunitPerfAnalysisPath>
```
weshaggard: The package version should be parameterized like the other packages so that we can update them without requiring a full BuildTools update.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -1,123 +1,61 @@
...
+	<BenchviewDir>/home/administrator/tools/Microsoft.BenchView.JSONFormat/tools</BenchviewDir>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetOS)'=='Windows_NT'">
-    <PerfTestCommandDotnetExecutable></PerfTestCommandDotnetExecutable>
+    <PerfTestCommandDotnetExecutable>PerfRunner.exe</PerfTestCommandDotnetExecutable>
     <PerfAnalysisReportCommand>$(XunitPerfAnalysisPath) $(XunitAnalysisReportArgs)</PerfAnalysisReportCommand>
     <PerfAnalysisCsvCommand>xunit.performance.analysis.exe $(XunitAnalysisCsvArgs)</PerfAnalysisCsvCommand>
-    <PerfTestCommand>$(PerfTestCommandDotnetExecutable) xunit.performance.run.exe $(XunitRunnerArgs) $(AdditionalPerfTestArgs)</PerfTestCommand>
+    <PerfTestCommand>%RUNTIME_PATH%\dotnet.exe $(PerfTestCommandDotnetExecutable)</PerfTestCommand>
+	<BenchviewDir>C:\Tools\Microsoft.BenchView.JSONFormat\tools</BenchviewDir>
```
weshaggard: I assume this is a hack that needs to be cleaned out.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
DrewScoggins: Good call. I have made this change and put the package version in the dependencies.props file in the CoreFX repo.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
DrewScoggins: I have one through and cleaned these up, should be good.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
DrewScoggins: Which part?  The definition of the BenchviewDir path?  On all of our perf machines we do a restore of a nuget package that we have that contains all of our Benchview tooling.  The problem is that this package was not built with nuget3 in mind and cannot be restored by dotnet or the command line nuget 3 tooling.  So during a step in the CI we ensure that we have these tools downloaded to this location.

weshaggard: We shouldn't be hard coding paths to "C:\tools" that is not a good idea. We should be isolating our tools somewhere under $(ToolsDir) or $(BinDir)

DrewScoggins: OK.  I can just put the nuget restore step that we do after the first time that we build the product and have the Benchview tools restored into the Tools directory.  Would that be OK?

weshaggard: Yes that would be much preferred. You should do the same for the Linux fork of these properties as well.

DrewScoggins: Yeah, will do.

DrewScoggins:
OK.  I made the change to reference the Benchview tooling from the tools folder.  Any other comments? @weshaggard @mellinoe

DrewScoggins:
I will of course squash this down to one commit.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PerfTesting.targets
```diff
@@ -1,123 +1,61 @@
...
-    <TestCommandLines Include = "if [ -f  $HELIX_CORRELATION_PAYLOAD/RunnerScripts/xunitrunner-perf/postprocessperfresults.py ];"></TestCommandLines>
-    <TestCommandLines Include = "then $HELIX_PYTHONPATH $HELIX_CORRELATION_PAYLOAD/RunnerScripts/xunitrunner-perf/postprocessperfresults.py --csvDir $(XunitPerfAnalysisCsvDir)"></TestCommandLines>
-    <TestCommandLines Include = "fi"></TestCommandLines>
+    <PerfTestCommandLines Include="$(PerfTestCommand)" />
+    <!-- We use the 20* syntax as the XML file created starts with the year and 20xx will be good for a while -->
   </ItemGroup>
   <ItemGroup Condition="'$(TargetOS)'=='Windows_NT'">
-    <TestCommandLines Include = "IF EXIST %HELIX_CORRELATION_PAYLOAD%\RunnerScripts\xunitrunner-perf\postprocessperfresults.py %HELIX_PYTHONPATH% %HELIX_CORRELATION_PAYLOAD%\RunnerScripts\xunitrunner-perf\postprocessperfresults.py --csvDir $(XunitPerfAnalysisCsvDir)"></TestCommandLines>
+    <PerfTestCommandLines Include="if exist $(AssemblyName).xml (" />
+    <PerfTestCommandLines Include="py $(BenchviewDir)\measurement.py xunit $(AssemblyName).xml --better desc --drop-first-value --append -o $(ProjectDir)measurement.json" />
```
weshaggard: Given that this depends on Python we need to make sure that it is an opt-in and not on by default. I suspect that is already the case as folks need to pass in "Performance=true" or something along those lines, but we need to be careful not to make python a required dependency for corefx builds. We also should update documentations for the performance stuff calling out it requires python.

DrewScoggins: Yeah this is hidden behind setting performance to true.  We can make sure that this is called out in the documentation, but I do not believe that this should affect the main-line build and test scenarios.

PR #1318, opened 2017-02-03
Propagate errors for GenFacades

mellinoe:
Instead of ignoring when `FacadeGenerator.GenerateFacade` returns null, I'm now detecting it and returning `false` from this method. The MSBuild task checks that result and propagates it.

@ericstj This should fix the NRE when GenFacades fails, and give better error messages.

### DIFF
--- a/src/GenFacades/GenFacades.Core/GenFacades.cs
+++ b/src/GenFacades/GenFacades.Core/GenFacades.cs
@@ -18,7 +18,7 @@ namespace GenFacades
     {
         private const uint ReferenceAssemblyFlag = 0x70;
 
-        public static void Execute(
+        public static bool Execute(
             string seeds,
             string contracts,
             string facadePath,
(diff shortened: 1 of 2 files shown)


ericstj:
Looks ok, will those errors logged by the tool be treated as errors in MSBuild (eg: will we get specific errors or will we have to hunt in the log)

mellinoe:
The individual messages that get printed out aren't considered errors, unfortunately, because they're just being printed out using `Trace`, and the `TraceLogger` class just writes to the MSBuild log since i can't distinguish messages/warnings/errors. Am I missing an obvious way to do that?

ericstj:
Attach a listener?  Shouldn't the listener get error vs message?

mellinoe:
`TraceLogger` is a `TraceListener` which is how the errors are being printed out now, I just don't understand how the filtering is supposed to be done; the API for `TraceListener` is a bit unintuitive. It seems like I could create separate TraceListener types with different `Filter` properties?

mellinoe:
I'm going to get this fix in for now. Getting MSBuild to flag the individual errors shouldn't be hard separately.

PR #1320, opened 2017-02-06
Use StringComparer.OrdinalIgnoreCase in LINQ OrderBy calls

akoeplinger:
This makes sure the sort order is consistent even with different culture/locale settings.

I thought about using StringComparer.Ordinal since it feels like the more "correct" in this case but unfortunately this generates a huge diff in e.g. the dotnet/standard repo since APIs are being reordered.

Unfortunately, even StringComparer.OrdinalIgnoreCase doesn't produce the exact same result, but the diff is much smaller, e.g. https://github.com/akoeplinger/standard/commit/dd88153291ed83fd0dadfcb585ab4c2477702346.

With this, I can use GenAPI.exe on OSX and it generates the same file as on Windows.

Fixes https://github.com/dotnet/corefx/issues/15445

/cc @mellinoe @weshaggard

### DIFF
--- a/src/ApiCompat/Program.cs
+++ b/src/ApiCompat/Program.cs
@@ -48,7 +48,7 @@ namespace ApiCompat
 
                 var rules = c.GetExports<IDifferenceRule>();
 
-                foreach (var rule in rules.Select(r => r.GetType().Name).OrderBy(r => r))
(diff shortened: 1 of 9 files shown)


mellinoe:
@akoeplinger Thanks for chasing this down. This was a very annoying problem.

PR #1324, opened 2017-02-07
Add submodule updaters to VersionTools

dagood:
These will be used in build from source to automatically keep the submodules in sync with the individual product repositories.

`IndicatorPackageSubmoduleUpdater` looks for the current indicator package in dotnet/versions, downloads it, extracts the version.txt inside, and compares it to the submodule's current commit. This is good to only update to a commit that has had an official build, but it only works when the submodule actually points at the official repository. (Not the case for any submodules at the moment.)

`LatestCommitSubmoduleUpdater` uses git's `ls-remote` to check the current commit a remote ref is pointing to, and compares it to the submodule's current commit.

@ellismg

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
@@ -6,6 +6,13 @@
     <BuildInfoCacheDir>$(ToolsDir)BuildInfoCache/</BuildInfoCacheDir>
   </PropertyGroup>
 
+  <!-- For backward compatibility, Include XmlUpdateSteps as Xml-type updaters. -->
+  <ItemGroup>
+    <UpdateStep Include="@(XmlUpdateStep)">
+      <UpdaterType>Xml</UpdaterType>
+    </UpdateStep>
+  </ItemGroup>
+
   <!-- Task to update the dotnet/versions repository. -->
   <UsingTask TaskName="UpdatePublishedVersions" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
 
@@ -33,7 +40,7 @@
   <Target Name="UpdateDependencies">
     <UpdateDependencies DependencyBuildInfo="@(DependencyBuildInfo)"
                         ProjectJsonFiles="@(ProjectJsonFiles)"
-                        XmlUpdateStep="@(XmlUpdateStep)"
+                        UpdateStep="@(UpdateStep)"
                         BuildInfoCacheDir="$(BuildInfoCacheDir)" />
   </Target>
 
@@ -46,7 +53,7 @@
 
     <VerifyDependencies DependencyBuildInfo="@(DependencyBuildInfo)"
                         ProjectJsonFiles="@(ProjectJsonFiles)"
-                        XmlUpdateStep="@(XmlUpdateStep)"
+                        UpdateStep="@(UpdateStep)"
                         BuildInfoCacheDir="$(BuildInfoCacheDir)" />
 
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Verifying all auto-upgradeable dependencies... Done." />
@@ -58,7 +65,7 @@
   <Target Name="UpdateDependenciesAndSubmitPullRequest">
     <UpdateDependenciesAndSubmitPullRequest DependencyBuildInfo="@(DependencyBuildInfo)"
                                             ProjectJsonFiles="@(ProjectJsonFiles)"
-                                            XmlUpdateStep="@(XmlUpdateStep)"
+                                            UpdateStep="@(UpdateStep)"
                                             ProjectRepoName="$(ProjectRepoName)"
                                             ProjectRepoOwner="$(ProjectRepoOwner)"
                                             ProjectRepoBranch="$(ProjectRepoBranch)"
--- a/src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
@@ -1,0 +1,96 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using NuGet.Packaging.Core;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.VersionTools.Dependencies.Submodule
+{
+    /// <summary>
+    /// Downloads a package specified from a dependency build info and updates the target git
+    /// submodule to match the version inside.
+    /// </summary>
+    public class IndicatorPackageSubmoduleUpdater : SubmoduleUpdater
+    {
+        /// <summary>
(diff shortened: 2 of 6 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
```diff
@@ -6,6 +6,13 @@
     <BuildInfoCacheDir>$(ToolsDir)BuildInfoCache/</BuildInfoCacheDir>
   </PropertyGroup>
 
+  <!-- For backward compatibility, Include XmlUpdateSteps as Xml-type updaters. -->
```
ellismg: Can we file a tracking issue to do the work to move everyone onto the new plan and then remove this workaround?

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
```diff
@@ -0,0 +1,96 @@
...
+                    .FirstOrDefault(p => p.Id == IndicatorPackageId);
+
+                if (package == null)
+                {
+                    continue;
+                }
+
+                using (ZipArchive archive = DownloadPackageAsync(info, package).Result)
+                {
+                    Stream versionTxt = archive.GetEntry("version.txt")?.Open();
```
ellismg: Wrap this in a using?

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
```diff
@@ -0,0 +1,96 @@
...
+                    if (versionTxt == null)
+                    {
+                        Trace.TraceWarning(
+                            $"Downloaded '{package}' in '{info.BuildInfo.Name}' " +
+                            $"to upgrade '{Path}', but it had no version.txt file. Skipping.");
+                        continue;
+                    }
+
+                    string packageCommitHash = new StreamReader(versionTxt).ReadLine();
```
ellismg: Dispose the streamreader as well?

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                throw new NotSupportedException(
+                    $"A {nameof(PackageDownloadBaseUrl)} must be configured, " +
+                    "as build-infos do not have package feed details.");
+            }
+
+            string downloadUrl = $"{PackageDownloadBaseUrl}/package/{package.Id}/{package.Version}";
+            Trace.TraceInformation($"Downloading '{package}' from '{downloadUrl}'");
+
+            var client = new HttpClient();
```
ellismg: Can we put this in a static so we can re-use them?  It's safe to have multiple operations in flight on the same HttpClient object.

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            Trace.TraceError($"Failed to find '{IndicatorPackageId}' specifying a commit in any build-info.");
+            usedBuildInfos = Enumerable.Empty<DependencyBuildInfo>();
+            return null;
+        }
+
+        protected async Task<ZipArchive> DownloadPackageAsync(DependencyBuildInfo info, PackageIdentity package)
+        {
+            if (PackageDownloadBaseUrl == null)
```
ellismg: Why is this not an argument to the `.ctor` so we can ensure it is always set?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
dagood: Sure. Filed https://github.com/dotnet/core-eng/issues/428 for now.

@dleeapho How should we file BuildTools breaking changes? I suspect this will be more manageable when we get a release model.

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
dagood: There's a plan to eventually have dotnet/versions repo list where the build published its packages, so for any repos that do that, this argument doesn't need to be set. (Then the `NotSupportedException` would be moved to the case where the property is null *and* the build-info doesn't list a download source.)

ellismg: Gotcha, makes sense.

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// The NuGet v2 base url to use to download the indicator package, without a trailing '/'.
+        /// For example, https://dotnet.myget.org/F/dotnet-core/api/v2/package.
+        /// </summary>
+        public string PackageDownloadBaseUrl { get; set; }
+
+        public string IndicatorPackageId { get; }
+
+        public IndicatorPackageSubmoduleUpdater(string indicatorPackageId)
+        {
+            if (indicatorPackageId == null)
```
MichaelSimons: Is an empty package valid?

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
```diff
@@ -0,0 +1,101 @@
...
+            Trace.TraceError($"Failed to find '{IndicatorPackageId}' specifying a commit in any build-info.");
+            usedBuildInfos = Enumerable.Empty<DependencyBuildInfo>();
+            return null;
+        }
+
+        protected async Task<ZipArchive> DownloadPackageAsync(DependencyBuildInfo info, PackageIdentity package)
+        {
+            if (PackageDownloadBaseUrl == null)
+            {
+                throw new NotSupportedException(
```
MichaelSimons: Seem like an InvalidOperationException would be a better choice here.  "DO throw an InvalidOperationException if the object is in an inappropriate state."

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/SubmoduleUpdater.cs
```diff
@@ -0,0 +1,82 @@
...
+            var writer = new StringWriter();
+
+            GitInPath(args)
+                .ForwardStdOut(writer)
+                .Execute();
+
+            return writer.ToString();
+        }
+
+        internal Command GitInPath(params string[] args)
```
MichaelSimons: internal?  Who calls this outside of the class?

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/IndicatorPackageSubmoduleUpdater.cs
dagood: No, I'll tighten that up with `IsNullOrEmpty` and change to `ArgumentException`.

src/Microsoft.DotNet.VersionTools/Dependencies/Submodule/SubmoduleUpdater.cs
dagood: At one point in dev a subclass called it (and I couldn't use `protected` because `Command` is internal) but now it's only used in this class. Changed to `private`. Thanks.

PR #1326, opened 2017-02-09
Initial version of init-docker.sh

MichaelSimons:
The idea behind this script is that we would call it at the beginning of build definitions that utilize Docker images.  This would replace the call to cleanup-docker and the build step to capture the docker version.  The retry logic here will hopefully address some of the flakiness we have been seeing when pulling images.

related to https://github.com/dotnet/core-eng/issues/437

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
@@ -6,7 +6,7 @@ set -e
 set -u
 
 say_err() {
-  printf "%b\n" "Error: $1" >&2
+    printf "%b\n" "Error: $1" >&2
 }
 
 showHelp() {
@@ -49,15 +49,15 @@ image=
 while [ $# -ne 0 ]; do
     name=$1
     case $name in
-        -h|--help|--[Hh]elp)
+        -h|--help)
             showHelp
             exit 0
             ;;
-        -r|--retryCount|--[Rr]etry[Cc]ount)
+        -r|--retryCount)
             shift
             retries=$1
             ;;
-        -w|--waitFactor|--[Ww]ait[Ff]actor)
+        -w|--waitFactor)
             shift
             waitFactor=$1
             ;;
@@ -67,8 +67,8 @@ while [ $# -ne 0 ]; do
             ;;
         *)
             if [ ! -z "$image" ]; then
-              say_err "Unknown argument: \`$name\`"
-              exit 1
+                say_err "Unknown argument: \`$name\`"
+                exit 1
             fi
 
             image="$1"
@@ -87,6 +87,6 @@ echo "Cleaning Docker Artifacts"
 echo
 
 if [ ! -z "$image" ]; then
-  echo "Pulling Docker image $image"
-  execute docker pull $image
+    echo "Pulling Docker image $image"
+    execute docker pull $image
 fi
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
@@ -1,49 +1,92 @@
 #!/usr/bin/env bash
+
+# Stop script on NZEC
 set -e
+# Stop script if unbound variable found (use ${var:-} if intentional)
+set -u
+
+say_err() {
+  printf "%b\n" "Error: $1" >&2
+}
+
+showHelp() {
+    echo "Usage: $scriptName [OPTIONS] [IMAGE_NAME[:TAG|@DIGEST]]"
+    echo
+    echo "Initializes Docker by:"
+    echo "  - Emitting the version of Docker that is being used"
+    echo "  - Removing all containers and images that exist on the machine"
+    echo "  - Ensuring the latest copy of the specified image exists on the machine"
+    echo
+    echo "Options:"
+    echo "  -r, --retryCount    Number of times to retry pulling image on error"
+    echo "  -w, --waitFactor    Time (seconds) to wait between pulls (time is multiplied each iteration)"
+}
+
+# Executes a command and retries if it fails.
+execute() {
+    local count=0
+    until "$@"; do
+        local exit=$?
+        count=$(( $count + 1 ))
+        if [ $count -lt $retries ]; then
+            local wait=$(( waitFactor ** (( count - 1 )) ))
+            echo "Retry $count/$retries exited $exit, retrying in $wait seconds..."
+            sleep $wait
+        else    
+            say_err "Retry $count/$retries exited $exit, no more retries left."
+            return $exit
+        fi
+    done
+
+    return 0
+}
 
 scriptName=$0
+retries=5
+waitFactor=6
+image=
 
-function usage {
-    echo "Usage: $scriptName image"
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+set -e
+
+scriptName=$0
+
+function usage {
```
dagood: We've been using `usage() {` syntax, e.g. in https://github.com/dotnet/buildtools/blob/master/bootstrap/bootstrap.sh. We need to figure out what way is supported by the most shells.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo "Usage: $scriptName image"
+    echo "  image      Name of the Docker image to pull"
+
+    exit 1
+}
+
+# Executes a command and retry up to 5 times until if it fails.
+function execute {
+  local retries=5
```
dagood: 2-space tabs here, but 4 spaces above. I think we use 4-space generally.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+set -e
```
dagood: Consider `set -u` (https://github.com/dotnet/buildtools/blob/e0a047525f44f428db3378c85a08349387288049/bootstrap/bootstrap.sh#L5-L6)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,49 @@
...
+      local exit=$?
+      echo "Retry $count/$retries exited $exit, no more retries left."
+      return $exit
+    fi
+  done
+
+  return 0
+}
+
+if [ -z "$1" ]; then
```
chcosta: '--help' is the common way to get script usage in Linux

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo "Usage: $scriptName image"
+    echo "  image      Name of the Docker image to pull"
+
+    exit 1
+}
+
+# Executes a command and retry up to 5 times until if it fails.
+function execute {
+  local retries=5
+  local waitFactor=3
```
chcosta: can we provide options for changing the retries / wait factor with defaults as provided?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+set -e
+
+scriptName=$0
+
+function usage {
+    echo "Usage: $scriptName image"
+    echo "  image      Name of the Docker image to pull"
```
chcosta: just name or name and tag?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
MichaelSimons: Added

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
MichaelSimons: Changed

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
MichaelSimons: Changed

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
MichaelSimons: Agreed - changed

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
MichaelSimons: Yes - Added.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
MichaelSimons: I changed this to mimic the Docker pull verbage - IMAGE_NAME[:TAG|@DIGEST]

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,92 @@
...
+
+scriptName=$0
+retries=5
+waitFactor=6
+image=
+
+while [ $# -ne 0 ]; do
+    name=$1
+    case $name in
+        -h|--help|--[Hh]elp)
```
dagood: Could you remove the `--help` case?

I think I would actually prefer removing the `--[Hh]elp` case. My interpretation of bootstrap.sh is that the case-insensitive matches are included (with a single `-`) for a little extra parity with bootstrap.ps1, which makes the script less conventional. (I'm not sure I think they're worth it in bootstrap.sh either, unless there's a reason I'm not seeing.)

@crummel Is there more to the case-insensitive options?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
```diff
@@ -0,0 +1,92 @@
...
+            waitFactor=$1
+            ;;
+        -*)
+            say_err "Unknown option: $1"
+            exit 1
+            ;;
+        *)
+            if [ ! -z "$image" ]; then
+              say_err "Unknown argument: \`$name\`"
+              exit 1
```
dagood: Indent only has 2 spaces here and a few other places.

MichaelSimons: Fixed - Having troubles getting my mind to use 4 spaces :)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh
MichaelSimons: Removed case-insensitive matches.

crummel: @dagood You're correct, it's a bootstrap parity thing.  I'm fine with not including it (and probably eliminating from bootstrap too).

PR #1327, opened 2017-02-09
Fix conditional in Dumpling.txt command template

mellinoe:
The current version did not work because of the way we are reading the lines out of this file and injecting them into another script. I tested this by causing a test to fail both:

* normally, through an unhandled System.Exception()
* by forcibly seg-faulting the test

Without this change, the first test case was still trying to call Dumpling.py, but with invalid arguments, because no core file was found.

@sepidehMS


sepidehMS:
Thanks @mellinoe

PR #1328, opened 2017-02-09
Make Dumpling.targets use a proper shell script.

mellinoe:
Instead of reading a text file and inserting raw script lines into RunTests.sh, I'm simply running a Dumpling.sh script from RunTests.sh. The way lines were being copied into RunTests.sh was very fragile and was not working anyways. Instead of fixing it, I've just switched to this, which should be simpler and less error-prone.

@sepidehMS


sepidehMS:
LGTM

PR #1329, opened 2017-02-11
Use either wget or curl, based on what is installed in Dumpling.sh

mellinoe:
Some of our CI machines only have curl. This uses curl if wget is not found, and errors out if neither are found.

@sepidehMS


sepidehMS:
LGTM.
Do we have python installed on all CI machines?

mellinoe:
> Do we have python installed on all CI machines?

Yes, as far as I can tell. See this PR here: https://github.com/dotnet/corefx/pull/16053. I've done a run with Dumpling turned on and am currently working through the issues.

There are several machines which are missing one of the python dependencies ("requests") which will still require some updating of images.

PR #1330, opened 2017-02-11
Update list of APIs legal to call in appcontainer

danmosemsft:
Commit 1 sorts the previous one. Compare insensitively to commit 2 to see the diff.


danmosemsft:
@joperezr @weshaggard

weshaggard:
LGTM - Is this the new allowed set? Is there going to be any need to restrict things to the older set?

joperezr:
> Is this the new allowed set?

Yes
> Is there going to be any need to restrict things to the older set?

That is a good question. I suppose that we might still have some projects building against netcore50 or netcore50aot in corefx, for which we might want to keep the other list around. Does it make sense to have these lists have a $(TargetGroup) in their name so we can keep both?

danmosemsft:
I'm going to commit this as the one that always supports "latest" -- which is what the corefx tree will as far as possible build for. If we want to extend the checker to support "older" then we can grab the older version of this and put it under another name.

ericstj:
Commit descriptions can use a bit more detail next time...

PR #1332, opened 2017-02-14
Added logic to init-tools to ensure correct MSBuild.runtimeconfig.json is used

### LINKED ISSUE #1325: MSBuild.runtimeconfig.json is getting overridden by a file in Microsoft.Build.Runtime
When I updated us to the latest version of MSBuild, we switched to using a new package called Microsoft.Build.Runtime. This includes a runtimeconfig.json file for MSBuild, but it is locked at version 1.0.1 of the shared framework. It overrides the one that we have in this repo. We should explicitly copy out our version in init-tools.sh after we have published the tool-runtime project.

MichaelSimons:
fixes #1325


mellinoe:
LGTM, although the commit author looks like it got a bit messed up.

MichaelSimons:
I fixed the commit author.

PR #1335, opened 2017-02-15
Add local build-info flow for source build

dagood:
Adds `VersionsRepoDir`, which points the tasks to use a versions-repo-like directory on disk.

For dependency update, change the `DependencyBuildInfo` to specify `VersionsRepoDir` metadata rather than `RawVersionsBaseUrl` metadata. E.g. https://github.com/dagood/corefx/commit/4e43203298e1d9c81daa5be91287371e69587716.

The packages created during a source build can be written to the local versions repo directory using `LocalUpdatePublishedVersions`, enabling local flow between two repos.

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
@@ -1,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.VersionTools.Automation
+{
+    public class LocalVersionsRepoUpdater
+    {
+        /// <summary>
+        /// Updates only the Latest_Packages file in the specified on-disk versions repository dir.
+        /// </summary>
+        public void UpdateBuildInfoLatestPackages(
+            IEnumerable<string> packagePaths,
+            string versionsRepoDir,
+            string versionsRepoPath)
+        {
+            if (packagePaths == null)
+            {
+                throw new ArgumentNullException(nameof(packagePaths));
+            }
+            if (string.IsNullOrEmpty(versionsRepoDir))
+            {
+                throw new ArgumentException(nameof(versionsRepoDir));
+            }
+            if (string.IsNullOrEmpty(versionsRepoPath))
+            {
+                throw new ArgumentException(nameof(versionsRepoPath));
+            }
+
+            Dictionary<string, string> packages = packagePaths
+                .Select(path => new NupkgNameInfo(path))
+                // Ignore symbol packages.
+                .Where(t => !t.SymbolPackage)
+                .ToDictionary(t => t.Id, t => t.Version);
+
+            string latestPackagesDir = Path.Combine(
+                versionsRepoDir,
+                versionsRepoPath);
+
+            Directory.CreateDirectory(latestPackagesDir);
+
+            File.WriteAllText(
+                Path.Combine(latestPackagesDir, BuildInfo.LatestPackagesTxtFilename),
+                VersionsRepoUpdater.CreatePackageListFile(packages));
+        }
+    }
+}
--- a/src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
@@ -15,16 +15,16 @@ namespace Microsoft.DotNet.VersionTools.Automation
         /// </summary>
         public void UpdateBuildInfoLatestPackages(
             IEnumerable<string> packagePaths,
-            string versionsRepoDir,
+            string localBaseDir,
             string versionsRepoPath)
         {
             if (packagePaths == null)
             {
                 throw new ArgumentNullException(nameof(packagePaths));
             }
-            if (string.IsNullOrEmpty(versionsRepoDir))
+            if (string.IsNullOrEmpty(localBaseDir))
             {
-                throw new ArgumentException(nameof(versionsRepoDir));
+                throw new ArgumentException(nameof(localBaseDir));
             }
             if (string.IsNullOrEmpty(versionsRepoPath))
             {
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
ellismg: You have a better sense for this than I, but I'll ask anyway.  Did you consider having an abstract base class that housed logic like `CreatePackageListFile` and the gathering of packages with two subclasses, GitHubVersionsRepoUpdater and LocalVersionsRepoUpdater which were just concerned about actually pushing the txt files around?  Is the overhead not worth it for the sharing of the linq clauses and argument checking?

ellismg:
LGTM.  This is really sweet, I'm super excited about the local publishing option as well.  Thanks, Davis.

src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
dagood: I looked at it earlier, briefly--part of the pain was that `NupkgNameInfo` is `internal`. I took a fresh look and decided to share functions as you suggested and just make `NupkgNameInfo` (now `NupkgInfo`) public.

src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
```diff
@@ -2,219 +2,34 @@
...
-
-            using (var reader = new StringReader(latestPackages))
-            {
-                return await BuildInfo.ReadPackageListAsync(reader);
-            }
+            return infos.ToDictionary(i => i.Id, i => i.Version);
         }
 
-        private static string CreatePackageListFile(Dictionary<string, string> packages)
+        protected static string CreatePackageListFile(Dictionary<string, string> packages)
```
MichaelSimons: >CreatePackageListFile [](start = 32, length = 21)

The name of this is a little misleading as it doesn't actually create a file.

src/Microsoft.DotNet.VersionTools/Automation/NupkgInfo.cs
```diff
@@ -21,9 +21,9 @@ public NupkgNameInfo(string path)
...
         }
 
-        public string Id { get; set; }
-        public string Version { get; set; }
-        public string Prerelease { get; set; }
-        public bool SymbolPackage { get; set; }
+        public string Id { get; }
+        public string Version { get; }
+        public string Prerelease { get; }
+        public bool SymbolPackage { get; }
```
MichaelSimons: >SymbolPackage [](start = 20, length = 13)

Consider naming IsSymbolPackage

src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
```diff
@@ -2,219 +2,34 @@
...
-                        }
-                        else
-                        {
-                            Trace.TraceInformation("Encountered exception committing build-info update.");
-                            throw;
-                        }
-                    }
-                }
-            }
+                .Where(t => !t.SymbolPackage);
```
MichaelSimons: Are there a lot of these?  If so you may consider adding a static method that would take in a package path and return whether or not it pointed to a symbols package.  This could then be called in a where condition before the select clause.

src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    public class LocalVersionsRepoUpdater : VersionsRepoUpdater
+    {
+        /// <summary>
+        /// Updates only the Latest_Packages file in the specified on-disk versions repository dir.
+        /// </summary>
+        public void UpdateBuildInfoLatestPackages(
+            IEnumerable<string> packagePaths,
+            string versionsRepoDir,
+            string versionsRepoPath)
```
MichaelSimons: IMO if I saw this signature in intellisense without seeing the implementation I would be confused.

src/Microsoft.DotNet.VersionTools/BuildInfo.cs
```diff
@@ -150,5 +179,16 @@ public static string RawBuildInfoBaseUrl(string rawRepoUrl, string gitRef, strin
...
+
+        private static string FindLatestReleaseFromPackages(IDictionary<string, string> packages)
+        {
+            IEnumerable<NuGetVersion> versions = packages.Values
+                .Select(versionString => new NuGetVersion(versionString));
+
+            return
+                versions.FirstOrDefault(v => v.IsPrerelease)?.Release ??
+                // if there are no prerelease versions, just grab the first version
+                versions.FirstOrDefault()?.ToNormalizedString();
```
MichaelSimons: consider indenting once more.

MichaelSimons:
:shipit:

ellismg:
@dagood Once this is merged I'm going to try update update-dependenices in the CLI to start being able to use to consume local sources.  For now, I hacked the script locally to let me pass on disk sources, but getting to adopt this will be better.

src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
dagood: The packages are always less than half symbol packages, but I think this is worth doing.

src/Microsoft.DotNet.VersionTools/Automation/VersionsRepoUpdater.cs
dagood: Agreed, renaming `CreatePackageListContent`.

src/Microsoft.DotNet.VersionTools/Automation/NupkgInfo.cs
dagood: Will probably remove, instead.

src/Microsoft.DotNet.VersionTools/Automation/LocalVersionsRepoUpdater.cs
dagood: Agreed, I'll make the "base dir"/"relative build info path" clear.

PR #1337, opened 2017-02-16
Move Inbox data to package index

ericstj:
We're growing the inbox concept to accomodate the flat framework packages (eg: NETStandard.Library, MS.NETCore.App, UWP).  As such it no longer makes sense to have these things defined as static framework lists that live in the buildtools package.  I've moved the data into the packageindex and created mechanisms for migration.

/cc @weshaggard @joperezr


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -1208,23 +1201,6 @@
                       Overwrite="true"
                       Condition="'$(_SkipCreatePackage)' != 'true'"/>
   </Target>
-  
-  <!-- Installs locally built package into a packageinstalldir for consuming -->
-  <!-- Overwrites previous installs, supports only one version installed -->
-  <Target Name="InstallLocallyBuiltPackages"
```
weshaggard: 💯

PR #1340, opened 2017-02-17
Create updaters for tooling version files

dagood:
`ToolVersionsUpdater` is used on `.toolversions` files, for bootstrap-acquired dependencies.

`FilePackageUpdater` updates an entire file with a package's version, for `BuildToolsVersion.txt` (and `DotnetCLIVersion.txt`).

This is the first part of getting to this BfS objective:

> 2. We need a way to update the buildtools dependency of everything that uses it to consume the version we just built.

Steps are set up like this: https://github.com/dagood/corefx/commit/1623bdd45fcd13b7b92d03ce776ff8fec1cb4276 (the `DependencyBuildInfo` is for testing).

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Dependencies/ToolVersionsUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Dependencies/ToolVersionsUpdater.cs
@@ -63,11 +63,18 @@ namespace Microsoft.DotNet.VersionTools.Dependencies
 
             public ToolUpdateLineResult(string line, IEnumerable<DependencyBuildInfo> buildInfos)
             {
+                Content = line;
+
                 int separatorIndex = line.IndexOf('=');
-                string name = line.Substring(0, separatorIndex);
+                if (separatorIndex == -1)
+                {
+                    // Ignore lines without a 'name=version' string.
+                    return;
+                }
+
+                string name = line.Substring(0, separatorIndex).Trim();
                 string version = line.Substring(separatorIndex + 1);
 
-                Content = line;
                 OriginalVersion = version;
                 ToolName = name;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.VersionTools/Dependencies/ToolVersionsUpdater.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            public string OriginalVersion { get; }
+
+            public string NewVersion { get; }
+
+            public ToolUpdateLineResult(string line, IEnumerable<DependencyBuildInfo> buildInfos)
+            {
+                int separatorIndex = line.IndexOf('=');
+                string name = line.Substring(0, separatorIndex);
+                string version = line.Substring(separatorIndex + 1);
```
MichaelSimons: Any concerns about a file with an invalid format?  Tracking this down may be painful.

MichaelSimons:
:shipit:

src/Microsoft.DotNet.VersionTools/Dependencies/ToolVersionsUpdater.cs
dagood: I'll add an index check in case there isn't a `=`, but that's the only case I can think of reasonably detecting in this code. Any other thoughts?

MichaelSimons: Yes that is what I was thinking.  The only other thing that comes to mind is if trimming white-space would be of value.  I can go either way on that.

dagood: I believe bootstrap.ps1/sh depend on there not being spaces so it would take some effort to hit that failure, but I'll add a `Trim` on the tool name in case it helps someone fix their .toolversions at some point. (Leaving `version` untrimmed so the upgrade message is accurate.)

PR #1343, opened 2017-02-21
Fix dup keys

ericstj:
/cc @weshaggard

Couple small fixes.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -253,7 +253,7 @@
         <PackageId>$(Id)</PackageId>
         <PackageVersion>$(PackageVersion)</PackageVersion>
         <!-- Some packages support legacy portable profiles where dependencies are provided by targeting pack -->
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -253,7 +253,7 @@
         <PackageId>$(Id)</PackageId>
         <PackageVersion>$(PackageVersion)</PackageVersion>
         <!-- Some packages support legacy portable profiles where dependencies are provided by targeting pack -->
-        <HarvestDependencies Condition="!$([System.String]::new('%(_FilesToPackage.TargetFramework)').StartsWith('portable-'))">true</HarvestDependencies>
+        <HarvestDependencies Condition="!$([System.String]::Copy('%(_FilesToPackage.TargetFramework)').StartsWith('portable-'))">true</HarvestDependencies>
```
weshaggard: Just for my curiosity why is Copy better then new?

ericstj: If you read the commit description you'll see:
>These calls to System.String.ctor were resulting in hundreds of first
chance exceptions in msbuild as it tried to find the correct overload.

>String.Copy doesn't have this problem since there are no overloads
to evaluate.

PR #1344, opened 2017-02-21
Add support for meta-package harvesting

ericstj:
Allow PackageIndex to define package IDs that should be rolled into a
meta-package reference rather than individual package references.

/cc @weshaggard @terrajobst 

This lets us reference the NETStandard.Library meta-package instead of the granular packages.  Consumption side of this change looks like this: https://github.com/ericstj/corefx/commit/38c8073910701ed4934a0c65556c5e52a8433b78

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyMetaPackages.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyMetaPackages.cs
@@ -1,0 +1,80 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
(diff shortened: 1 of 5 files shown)


ericstj:
> do we need versions of for these meta-packages which contain difference sets of packages

The versioning is irrelevant with our current baseline.  If we need versioning in the future we can add it.  I didn't want to add something that complicated this unless we were actually using it.

> Also should we treat them similar to inbox.

Not sure what you mean by this.  The dependency remapping happens after we've done our inbox trimming / dependency group expansion so it will benefit from the inbox logic.

weshaggard:
>Not sure what you mean by this. The dependency remapping happens after we've done our inbox trimming / dependency group expansion so it will benefit from the inbox logic.

I was only meaning should we treat the meta-packages themselves the same way as we do inbox components but I'm fine keeping the separate.

Issue #1338, opened 2017-02-16
IL.targets should call TargetsTriggeredByCompilation after Ilasm


ericstj:
This matches what is done in CSharp.Core.targets and is required for running ApiCompat.

/cc @weshaggard @mellinoe

PR #1346, opened 2017-02-23
Clean IL.targets and make it call TargetsTriggeredByCompilation.

mellinoe:
I did a few things here:

* Cleaned up how the "GetIlasmPath" target worked so that we always use that absolute path on Windows. Previously, if you weren't building from VS, we would just use the name "ilasm" to start the process. That only worked if you were in a developer command prompt. Which is most of the time, admittedly, but this way is cleaner and works just as well.
* Give an error if `IlasmToolPath` is not set, which should only happen outside of Windows. Repos need to set that property, for now at least, until we figure out a better way to restore and deploy native assets along with buildtools.
* Call `TargetsTriggeredByCompilation`, which fixes https://github.com/dotnet/buildtools/issues/1338

@weshaggard, @ericstj

This stuff isn't strictly required for us to upgrade ilasm in corefx (we're just going to override the path property anyways), but it does clean things up and will also cause ApiCompat to run for our IL project.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
@@ -4,35 +4,29 @@
   <!-- Required by Microsoft.Common.targets -->
   <Target Name="CreateManifestResourceNames" Condition="'@(EmbeddedResource)' != ''" />
 
-  <Target Name="GetIlasmPath"
-          Condition="'$(BuildingInsideVisualStudio)' == 'true'">
-    <GetFrameworkPath>
-      <Output TaskParameter="Path" PropertyName="IlasmPath" />
-    </GetFrameworkPath>
-    <PropertyGroup>
-      <IlasmPath>$(IlasmPath)\</IlasmPath>
-    </PropertyGroup>
-  </Target>
+  <PropertyGroup>
+    <IlasmToolPath Condition="'$(IlasmToolPath)' == ''">$(ToolsDir)ilasm</IlasmToolPath>
+  </PropertyGroup>
 
   <Target Name="CoreCompile"
           Inputs="$(MSBuildAllProjects);
                   @(Compile)"
           Outputs="@(IntermediateAssembly);"
           Returns=""
-          DependsOnTargets="GetIlasmPath;$(CoreCompileDependsOn)">
+          DependsOnTargets="$(CoreCompileDependsOn)">
     <PropertyGroup>
       <_OutputTypeArgument Condition="'$(OutputType)' == 'Library'">-DLL</_OutputTypeArgument>
       <_OutputTypeArgument Condition="'$(OutputType)' == 'Exe'">-EXE</_OutputTypeArgument>
 
       <_KeyFileArgument Condition="'$(KeyOriginatorFile)' != ''">-KEY=$(KeyOriginatorFile)</_KeyFileArgument>
-
-      <IlasmExePath Condition="'$(IlasmExePath)' == ''">$(IlasmPath)ilasm</IlasmExePath>
     </PropertyGroup>
 
-    <Exec Command="$(IlasmExePath) -QUIET $(_OutputTypeArgument) $(IlasmFlags) -OUTPUT=@(IntermediateAssembly) $(_KeyFileArgument) @(Compile, ' ')">
+    <Exec Command="$(IlasmToolPath) -QUIET $(_OutputTypeArgument) $(IlasmFlags) -OUTPUT=@(IntermediateAssembly) $(_KeyFileArgument) @(Compile, ' ')">
       <Output TaskParameter="ExitCode" PropertyName="_ILAsmExitCode" />
     </Exec>
     <Error Text="ILAsm failed" Condition="'$(_ILAsmExitCode)' != '0'" />
+
+    <CallTarget Targets="$(TargetsTriggeredByCompilation)" Condition="'$(TargetsTriggeredByCompilation)' != ''"/>
   </Target>
 
   <Import Project="$(MSBuildToolsPath)\Microsoft.Common.targets" />
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
@@ -21,7 +21,10 @@
         "NETStandard.Library": {
           "version": "1.6.1",
           "exclude": "runtime"
-        }
+        },
+        "Microsoft.NETCore.Platforms": "1.1.0",
+        "Microsoft.NETCore.Runtime.CoreCLR": "1.1.0",
+        "Microsoft.NETCore.ILAsm": "2.0.0-beta-25027-03",
       },
       "imports": [
         "dnxcore50",
--- a/dependencies.props
+++ b/dependencies.props
@@ -3,7 +3,7 @@
   <!-- Source of truth for dependency tooling: the commit hash of the dotnet/versions master branch as of the last auto-upgrade. -->
   <PropertyGroup>
     <CoreFxCurrentRef>12abd697857d39b1a07dfae9e01d31f875500ebf</CoreFxCurrentRef>
-    <CoreClrCurrentRef>a8be270baa206a3d87112b7912869661b0c46f0a</CoreClrCurrentRef>
+    <CoreClrCurrentRef>a0ed75550803dad6df36961c168c21f33cd86f6d</CoreClrCurrentRef>
     <ProjectKTfsCurrentRef>b6fe7bf0d879e7a750b15e90743d5191e3b478de</ProjectKTfsCurrentRef>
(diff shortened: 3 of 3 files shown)


ericstj:
>until we figure out a better way to restore and deploy native assets along with buildtools.

You decided not to do something like https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh#L81-L85?

mellinoe:
@ericstj For some reason I remember that stuff being a lot more complicated than it looks now. It probably wouldn't be that hard to hook this in there. But I wonder how we feel about adding more stuff into those kinds of projects and into init-tools.sh like that? The depproj's we have in corefx are quite a bit nicer in that you don't need to list all of the RIDs out and can easily hook them into the actual build process. I'd be happy to go that route, though. @weshaggard thoughts?

ericstj:
I'm fine either way.  Agree that corefx side is a bit easier.

Another idea would be for CLR to decide to ship these tools in shared framework (like they do crossgen), then we could just pull it from there.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
```diff
@@ -5,12 +5,13 @@
   <Target Name="CreateManifestResourceNames" Condition="'@(EmbeddedResource)' != ''" />
 
   <Target Name="GetIlasmPath"
-          Condition="'$(BuildingInsideVisualStudio)' == 'true'">
+          Condition="'$(OS)' == 'Windows_NT' And '$(IlasmToolPath)' == ''">
```
weshaggard: I think I wold prefer we just remove this default ans always set the Path instead of having the OS condition, as I would like to be rid of those conditions.

weshaggard:
I didn't realize we are restoring/publishing for a RID I thought it was just a framework. Given that we are doing it per RID I would suggest we do this package the same way through tool-runtime as it should fit into the same pattern have here. We obviously need to have different support once we change the bootstrapping but so do the other things that are in there.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
mellinoe: Yeah if I hook the restore/publish directly into init-tools, then I can just eliminate this whole target and always use the one under Tools.

mellinoe:
I'm now restoring and published ilasm from within the init-tools comand.

Unfortunately, I'm still hitting an issue when I'm using in on Ubuntu; it seems there might be a problem with paths that are too long: https://github.com/dotnet/coreclr/issues/9757#issuecomment-282174175

That won't necessarily block this going in, but we still wouldn't be able to turn System.Runtime.CompilerServices.Unsafe back on until that is sorted out. So we may just want to hold off on this until there's a new version of the ilasm package that I can point this to.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -56,6 +56,24 @@ if not [%NET46_PUBLISH_ERROR_LEVEL%]==[0] (
 	exit /b %NET46_PUBLISH_ERROR_LEVEL%
 )
 
+set ILASM_PROJECTJSON=%BUILDTOOLS_PACKAGE_DIR%\ilasm\project.json
```
weshaggard: Why can't this just be part of the tool-runtime\project.json?

mellinoe: It feels kind of weird, because this is restored and published alongside that one, and into a different directory.

weshaggard: Perhaps I'm missing something but why does it need to be different? Why can it not be along side everything else? Is it because of the updated CoreCLR package? Is that actually needed for this to function?

mellinoe: Yes, it's because ilasm links against CoreCLR, and against a newer version than we use for the rest of the tools. Publishing ilasm alongside the rest of the tools would not work because it has conflicting dependencies, which will break the rest of the tools.

weshaggard: argh.... do you know how coupled ilasm is the CoreCLR? Could this ilasm run on the CoreCLR that we are currently using for the rest of the tools?

mellinoe: I don't know exactly how coupled they are. I would imagine that there is some set of functionality that ilasm pulls from CoreCLR (and would need to be using the latest version of CoreCLR in order to work properly). Right now, we are able to run the latest ilasm on our 1.1 runtime, but I don't think it's a safe thing to rely on in the future.

weshaggard: >Right now, we are able to run the latest ilasm on our 1.1 runtime, but I don't think it's a safe thing to rely on in the future.

I agree it isn't a great thing to rely on but if it does work right now I think I would prefer we do that as opposed to adding another layer of restore and tool isolation. We have the same problem for all of our tools in BuildTools, as they have to agree on the version of the runtime that they can work on.

mellinoe: Given that the ILAsm package depends on a matching Runtime.CoreCLR package, how should I accomplish that, then? Adding it to the existing tool-runtime/project.json will cause an invalid version of CoreCLR to get pulled in and published with our tools.

weshaggard: I see 2 options:

1) Directly reference the older Runtime package in tool-runtime project.json which will result in a downgrade that we can ignore.
2) Update either manually or via PR in CoreCLR repo to point the ILAsm package at the older version of runtime package (or potentially even remove the dependency).

mellinoe: 1. is an ok hack, although it will result in a bunch of duplicate files being dropped into the root Tools directory (like coreclr, clrjit, S.P.CL, etc.) which is not great (at least IMO). I think we can exclude them from being copied out (with some metadata in the project.json?), but then we need to post-copy ilasm down into the shared folder...

mellinoe:
I included ilasm in tool-runtime/project.json and downgraded the CoreCLR reference. I am going to merge this once my Linux build succeeds.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: Yes we could exclude just the runtime but as you say we would need to then copy ilasm into the shared framework directory. 

Did you file a follow-up issue for this scenario?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.json
```diff
@@ -21,7 +21,10 @@
         "NETStandard.Library": {
           "version": "1.6.1",
           "exclude": "runtime"
-        }
+        },
+        "Microsoft.NETCore.Platforms": "1.1.0",
+        "Microsoft.NETCore.Runtime.CoreCLR": "1.1.0",
+        "Microsoft.NETCore.ILAsm": "2.0.0-beta-25027-03",
```
dagood: This mismatch between ILAsm and CoreCLR version is writing a `publish` warning to stderr, failing the auto-update build: https://devdiv.visualstudio.com/DevDiv/_build?buildId=591416. I can disable fail on stderr, but is this really necessary?

weshaggard: It was the lesser of multiple evils (see the other comments in this PR) but if the warning is causing issues we should see what our options are.

dagood: Ah, expanded an "outdated" thread and see what you mean. I think it's just that one build so far that still had fail on stderr enabled, so I don't think finding another way's a high priority.

weshaggard: It does cause this to be dumped out during init-tools:
```
Restoring BuildTools version 1.0.27-prerelease-01327-01...
Initializing BuildTools...
D:\git\corefx\packages\Microsoft.DotNet.BuildTools\1.0.27-prerelease-01327-01\lib\tool-runtime\project.json(26,52): warn
ing NU1012: Dependency conflict. Microsoft.NETCore.ILAsm 2.0.0-beta-25027-03 expected Microsoft.NETCore.Runtime.CoreCLR
>= 2.0.0-beta-25027-03 but received 1.1.0
```
which I don't really like that noise either so we really should look at other options. However for the auto-update build definition I'm fine with making that not an error.

PR #1348, opened 2017-02-24
[release/1.1.0] Make ValueTuple inbox on Xamarin

ericstj:
/cc @marek-safar @jcouv @weshaggard 

Xamarin needs an update to ValueTuple since they are putting it inbox in their SDKs.  We are planning an update from 1.1.0 branch, which should align with Xamarin timelines.


ericstj:
Issue: https://github.com/dotnet/corefx/issues/14235

jcouv:
Thanks.
Should I add some type forwards in the ValueTuple package relative to mono after that? (is mono a target group in corefx?)

ericstj:
@weshaggard this is for 1.1.0 which is still using the lists.

> Should I add some type forwards in the ValueTuple package relative to mono after that? (is mono a target group in corefx?)

@jcouv, no.  Xamarin is going to ship their own System.ValueTuple which typeforwards to their mscorlib.dll per https://github.com/dotnet/corefx/issues/14235#issuecomment-265909596.  We just put an empty folder in the package so that ours won't conflict with theirs.

ericstj:
This can't be merged until 1.1.0 opens up for check-ins.

ericstj:
@dotnet-bot test this please

PR #1350, opened 2017-02-24
Allow disabling fxcop analyzers

### LINKED ISSUE #1349: Add property to disable analyzers
There is no clean way to suppress analyzers per project because of they are enabled/disabled unconditionally based on project type:

```
    <EnableFxCopAnalyzers Condition="$([System.Text.RegularExpressions.Regex]::IsMatch($(MSBuildProjectDirectory), 'src%24'))">true</EnableFxCopAnalyzers>
    <EnableFxCopAnalyzers Condition="'$(EnableFxCopAnalyzers)' != 'true'">false</EnableFxCopAnalyzers>
```

Analyzers are not applicable to low-level CoreRT components. CoreRT build needs a clean way to disable them.

jkotas:
The fxcopy analyzers do not make sense for CoreRT low-level components

Fixes #1349


jkotas:
@alexperovich PTLA

PR #1353, opened 2017-02-28
Use NuGet.RuntimeModel instead of stale copy

ericstj:
Previously we were writing runtime.json using a stale copy of the NuGet
code (since it wasn't public yet).

Now support is available in the NuGet.RuntimeModel assembly.  Use that
instead.

/cc @Petermarcu @weshaggard


ericstj:
Filed https://github.com/NuGet/Home/issues/4717 about the differences in the format that I worked around.

weshaggard:
For future reference could you add the commit has from nuget that matches the code you are pulling?

ericstj:
> For future reference could you add the commit has from NuGet that matches the code you are pulling?

This isn't a copy, I rewrote it to use public API.  The previous version was a copy because the public API didn't exist at that time.

PR #1355, opened 2017-03-01
Update xunit constants and TargetFrameworkMonikers to be non-versioned

safern:
We are currently testing only against latest version of every target framework. We need to update the `Xunit.Netcore.Extensions` to have a non-version behavior when discovering tests. This will help consolidate the behavior when using `SkipOnTargetFramework` test attribute. This is also needed for netfx vertical test runs in order to filter the tests correctly. 

Currently working on the necessary changes in corefx tests to make them build successfully when we try to ingest the new buildtools version.

/cc @weshaggard @ericstj

### DIFF
--- a/src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
+++ b/src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
@@ -22,8 +22,8 @@ namespace Xunit
         Netcoreapp1_1 = 0x400,
         NetFramework = 0x800,
         Netcoreapp = 0x1000,
-        NetcoreUap = 0x2000,
-        NetcoreUapAot = 0x4000,
+        Uap = 0x2000,
+        UapAot = 0x4000,
         NetcoreCoreRT = 0x8000
     }
 }
--- a/src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
+++ b/src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using Xunit.Sdk;
 
 namespace Xunit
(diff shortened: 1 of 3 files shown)


src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
```diff
@@ -3,33 +3,16 @@
...
 
 using System;
-using Xunit.Sdk;
 
 namespace Xunit
 {
     [Flags]
     public enum TargetFrameworkMonikers
     {
-        Net45 = 0x1,
```
weshaggard: While I agree we only plan to use the versionless ones at this point in time does keeping the versioned ones hurt anything? Just in case we change our mind or do want to add a few specific versioned tests that aren't latest?

safern:
Updated my PR to keep the versioned `TargetFrameworkMonikers` and added the nonversion ones.

src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
```diff
@@ -17,19 +16,14 @@ public enum TargetFrameworkMonikers
...
-        NetFramework = Net45 | Net451 | Net452 | Net46 | Net461 | Net462 | Net463,
-        NetFramework45 = Net45 | Net451 | Net452,
-        NetFramework46 = Net46 | Net461 | Net462,
-
-        Netcoreapp = Netcoreapp1_0 | Netcoreapp1_1,
-
-        NetcoreUwp = Netcore50 | Netcore50aot
+        NetFramework = 0x800,
+        Netcoreapp = 0x1000,
+        NetcoreUap = 0x2000,
```
weshaggard: This should just be uap and uapaot. We don't need netcore in front of those.

PR #1358, opened 2017-03-01
Add support for capturing windows crash dumps to Dumpling.targets

mellinoe:
@sepidehMS Could you take a look?

This adds support for collecting crash dumps on Windows. The differences between this and the Unix script are just a consequence of the different crash dump collection facilities on the operating systems. Unix systems allow us to have the crash dump files placed directly next to the executable file, which makes them very easy to discover. On Windows, we are able to configure a global location for crash dumps via the registry, and our CI machines are configured to do so.

In addition to specifying `EnableDumpling=true`, Windows users must also specify `CrashDumpFolder=<SOMETHING>` in order for this to discover new crash dumps. The process is as follows:

* Ensure dumpling is installed.
* Collect a timestamp and store it in `__TIMESTAMP`.
* Run the tests.
* Call the DumplingHelper.py script with the above `__TIMESTAMP`, the exit code of the tests, `CrashDumpFolder`, and the name of the project
* If the exit code is non-zero, and a crash dump newer than `__TIMESTAMP` is found, then upload it with some metadata (same as Unix).

I've chosen to write the script in Python. In my opinion, we should unify the applicable portions of the Unix script into DumplingHelper.py, but I haven't done that, yet. We still need some parts of the Unix script to be external to the python script, like the part that sets `ulimit -c unlimited`, which needs to affect the calling process rather than the python sub-process. We should be able to merge things like the actual crash dump uploading logic, though.

The Unix path should be unaffected by these changes.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
@@ -3,16 +3,23 @@
     <Content Include="$(MSBuildThisFileDirectory)Dumpling.sh">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="$(MSBuildThisFileDirectory)DumplingHelper.py">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
   </ItemGroup>
   <!-- Setup Dumpling service to collect crash dumps -->
   <Target Name="SetupDumpling"
-          BeforeTargets="GenerateTestExecutionScripts"
-          Condition="'$(TargetOS)'!='Windows_NT'">
-    <ItemGroup>
+          BeforeTargets="GenerateTestExecutionScripts">
+    <ItemGroup Condition="'$(TargetOS)'!='Windows_NT'">
       <TestCommandLines Include="source Dumpling.sh" />
-    </ItemGroup>
-    <ItemGroup>
       <PostExecutionTestCommandLines Include="CollectDumps $%3f $(MSBuildProjectName)" />
+    </ItemGroup>
+    <Error Condition="'$(CrashDumpFolder)' == ''" Text="CrashDumpFolder must be set to use Dumpling on Windows." />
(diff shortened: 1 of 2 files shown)


mellinoe:
@schaabs Here's the logic I'm using to discover the latest crash dump:

https://github.com/dotnet/buildtools/pull/1358/files#diff-3a6b8c3d6499b3af7c2b91e3279a548dR19

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
```diff
@@ -3,16 +3,23 @@
...
-          Condition="'$(TargetOS)'!='Windows_NT'">
-    <ItemGroup>
+          BeforeTargets="GenerateTestExecutionScripts">
+    <ItemGroup Condition="'$(TargetOS)'!='Windows_NT'">
       <TestCommandLines Include="source Dumpling.sh" />
-    </ItemGroup>
-    <ItemGroup>
       <PostExecutionTestCommandLines Include="CollectDumps $%3f $(MSBuildProjectName)" />
     </ItemGroup>
+    <Error Condition="'$(CrashDumpFolder)' == ''" Text="CrashDumpFolder must be set to use Dumpling on Windows." />
```
sepidehMS: Should this be conditioned on Windows?

mellinoe: Yes, it should be, good catch.

sepidehMS:
Overall LGTM. Just out of curiosity, how did you configure CI machines to set the location of crash dumps?

mellinoe:
@sepidehMS Some of them are actually already configured that way. They just set some registry keys:

https://github.com/dotnet/dotnet-ci/blob/master/management/nodes/startupScripts/win2008-20161104.ps1#L49

https://msdn.microsoft.com/en-us/library/windows/desktop/bb787181(v=vs.85).aspx

I still have to double check whether all of our CI machines are configured properly, but it should be simple enough.

sepidehMS:
@mellinoe Thanks.

mellinoe:
@MattGal I also chatted with you about finding windows crash dumps a while back

PR #1364, opened 2017-03-02
Enable windows symbols packages

chcosta:
The PathResolver class used in NuGetPack (https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs#L154), takes into account the OS path directory separator character.  See https://github.com/NuGet/NuGet.Client/blob/c98de8b0635144fb89fd2853b802c03b79f46c2e/src/NuGet.Core/NuGet.Common/PathUtil/PathResolver.cs#L56.  We were previously specifying the manifest file target with '/', which would not get matched by the path resolver on Windows (it wouldn't resolve src content).  This caused us to fail here, https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs#L187, and not produce symbols packages.

/cc @dagood @ericstj @ellismg

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
@@ -197,7 +197,7 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
                     select new ManifestFile()
                     {
                         Source = f.GetMetadata(Metadata.FileSource),
-                        Target = f.GetMetadata(Metadata.FileTarget),
+                        Target = f.GetMetadata(Metadata.FileTarget).Replace('/', Path.DirectorySeparatorChar),
                         Exclude = f.GetMetadata(Metadata.FileExclude)
                     }).OrderBy(f => f.Target, StringComparer.OrdinalIgnoreCase).ToList();
         }
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
@@ -197,7 +197,8 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
                     select new ManifestFile()
                     {
                         Source = f.GetMetadata(Metadata.FileSource),
-                        Target = f.GetMetadata(Metadata.FileTarget).Replace('/', Path.DirectorySeparatorChar),
+                        // Pattern matching in PathResolver requires that we standardize to OS specific directory separator characters
+                        Target = f.GetMetadata(Metadata.FileTarget).Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar),
                         Exclude = f.GetMetadata(Metadata.FileExclude)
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
```diff
@@ -197,7 +197,7 @@ private List<ManifestFile> GetManifestFiles()
                     select new ManifestFile()
                     {
                         Source = f.GetMetadata(Metadata.FileSource),
-                        Target = f.GetMetadata(Metadata.FileTarget),
+                        Target = f.GetMetadata(Metadata.FileTarget).Replace('/', Path.DirectorySeparatorChar),
```
dagood: Is there a problem if this metadata uses `\`, or does that just never happen?

chcosta: I'm not certain I understand your question.  We generate the target path with '/'.  We could do '\\', but that's just flipping the problem.  The issue is that we need to generate a nuspec file with the proper path separator for the platform we're running on or else the PathResolver will fail to match assets.

ericstj:
We've normalized to '/' for targetpath.  Are you sure this is the cause of the problem?

chcosta:
My testing has shown this to be true

ericstj:
How was it working before then?

chcosta:
we aren't currently producing windows symbols packages

ericstj:
I know, but we were in 1.0 and 1.1 without this change.  How did it work then?

chcosta:
i'm not certain what changed specifically, and Davis has more context on the history of symbols packages.  We did only produce packages on one platform during that timeframe, but that doesn't at all address your question because that platform was Windows.  When did we being producing symbols packages?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
dagood: Sorry for the confusion, I just didn't know this:

>  We generate the target path

I agree that `\` and `/` have the same problem, I was just worried that someone might break the convention and end up with mixed slashes. Thanks for the extra info.

ericstj:
We've done so since 1.0.  I'm looking at it to see what changed.

chcosta:
This may be related - https://github.com/dotnet/buildtools/commit/f8450cc87686c9069b5eda4bbac83e02ab72d277

ericstj:
Looks like we were using whatever item path was before:
https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets#L181

But when I added the props binplacing I intentionally normalized to '/' because this is what we do binaries.
https://github.com/dotnet/corefx/blob/master/binplace.targets#L113

I walked through the nuget logic and agree with you that we must standardize to DirectorySeperator character.  Perhaps you can just change this to Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar) and add a comment that says that the pattern matching requires this.

PR #1365, opened 2017-03-03
Fix naming issue with results file

DrewScoggins:
The name of the xml results file created by xunit performance was
different than the file that we were checking andtrying to use.  I have
added the correct ID to resolve the name mismatch.


DrewScoggins:
PTAL @mellinoe

PR #1369, opened 2017-03-03
Enable building packed packages (transport) and enable packing standard

chcosta:
packages with symbols.

/cc @ericstj @ellismg 

Build an additional packed package by specifying `/p:CreatePackedPackage=true`

Add symbols to a standard package by specifying `/p:IncludeSymbolsInPackage=true`

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs
@@ -33,6 +33,10 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
             @"tools\**\*.ps1".Replace('\\', Path.DirectorySeparatorChar)
         };
 
+        private static readonly string _defaultPackedPackagePrefix = "transport";
+        private static readonly string _symbolsPackageExtension = ".symbols.nupkg";
+        private static readonly string _packageExtension = ".nupkg";
+
         [Required]
         public ITaskItem[] Nuspecs
         {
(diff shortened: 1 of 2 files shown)


chcosta:
@weshaggard , are you able to review this or should I wait for Eric?  This is blocking some other work to get symbol packages for Microsoft.NETCore.App, so if I need to wait for Eric then I'll look into privately generating all of the coreclr / corefx runtime / identity packages and using them to validate my changes to Core-Setup.  I'm not opposed to going through that effort if there's any concern with this PR.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs
```diff
@@ -243,11 +301,60 @@ public void Pack(string nuspecPath, Func<string, string> nuspecPropertyProvider,
...
+                }
+                else if (packSymbols)
+                {
+                    packageType = "packed";
+                }
                 Log.LogError($"Error when creating nuget {packageType} package from {nuspecPath}. {e}");
             }
         }
 
+        private Manifest TransformManifestToPackedPackageManifest(Manifest manifest)
```
weshaggard: Why does this need to be a special path in the Pack task? Why can't it just be special mode of the symbols package that is just renamed as "transport"?

weshaggard:
>Build an additional packed package by specifying /p:CreatePackedPackage=true

I assume you plan to use this for Microsoft.NETCore.Runtime.CoreCLR package?

>Add symbols to a standard package by specifying /p:IncludeSymbolsInPackage=true

I assume you plan to use this for the private corefx package?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs
chcosta: We need to change the Id of the package, update the dependencies, and change the runtime graph file which gets included in the identity package to handle the "transport" line of packages.

weshaggard: Sure but we only expect that we will have either a symbols or transport package but not both so you can mostly share that and essentially reduce the logic down to "CreateSymbols" and "Prefix" and if there is a "Prefix" then generate a transport package otherwise just a symbols package.

chcosta: Why wouldn't we want both?  We'd still want symbols to be available for scenarios where the transport package isn't being consumed.  I think we need all three because CoreClr is still producing stand-alone packages, so for them we want standard, symbols, and transport packages.

weshaggard: OK I guess I didn't realize we were going to need both symbol and transport for CoreCLR if that is the case then I'm OK with the general idea.

PR #1384, opened 2017-03-09
Provide a package for CCI + CCI Extensions

terrajobst:
There is currently now way to acquire CCI by itself; one can install the build tools package directly but this also installs a bunch of targets which regular projects most likely cannot use anyway.

@weshaggard

### DIFF
--- a/src/nuget/Microsoft.DotNet.BuildTools.Cci.nuspec
+++ b/src/nuget/Microsoft.DotNet.BuildTools.Cci.nuspec
@@ -1,0 +1,47 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.8.1">
+    <id>Microsoft.DotNet.BuildTools.Cci</id>
+    <version>1.0.0-prerelease</version>
+    <title>Microsoft DotNet BuildTools CCI</title>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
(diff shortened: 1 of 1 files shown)


terrajobst:
Going to merge so I can consume the package from CoreFxTools

src/nuget/Microsoft.DotNet.BuildTools.Cci.nuspec
```diff
@@ -0,0 +1,47 @@
...
+        <dependency id="System.Runtime.Extensions" version="4.3.0" />
+        <dependency id="System.Runtime.InteropServices" version="4.3.0" />
+        <dependency id="System.Text.Encoding" version="4.3.0" />
+        <dependency id="System.Text.RegularExpressions" version="4.3.0" />
+        <dependency id="System.Threading" version="4.3.0" />
+        <dependency id="System.Xml.XDocument" version="4.3.0" />
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="BclRewriter\Microsoft.Cci.dll" target="lib\netstandard13" />
```
akoeplinger: Microsoft.Cci.dll is coming from a nuget package already so that could (should?) just be a `<dependency>`

weshaggard: Agreed should this package just be about Microsoft.CCI.Extensions?

PR #1385, opened 2017-03-10
Add version.txt to every BuildTools package root

dagood:
Putting version.txt (containing the built commit hash) in the packages makes them traceable for automation and people. CoreFX is currently doing this for its packages (using the packaging targets rather than manual nuspecs).

We can use this file to update a checked-in bootstrap script in product repos. An auto-upgrader can take the commit hash from the BuildTools being consumed and download the bootstrap scripts from BuildTools at that hash. It would be better if it were checked into dotnet/versions, but this is an alternative. @MichaelSimons

This also allows the submodule updaters (https://github.com/dotnet/buildtools/pull/1324) to auto-update a BuildTools submodule based on successful builds, by setting BuildTools as an indicator package. @ellismg


ellismg:
LGTM

PR #1386, opened 2017-03-11
Fix syntax for metadata access

jack-pappas:
Minor MSBuild syntax fix for accessing a metadata property. I discovered the issue while trying to get the Microsoft/visualfsharp build working on coreclr; some of the bootstrapping there is currently running via xbuild and mono, which is perhaps more strict about syntax checking.


weshaggard:
Thanks for the fix @jack-pappas

PR #1390, opened 2017-03-15
Explicitly define csharpcoretargetspath and visualbasiccoretargetspath

chcosta:
Re: https://github.com/dotnet/coreclr/issues/10056

@weshaggard @joperezr 

This should unblock building with a VS2017 dev command prompt, but doesn't address the rest of the issues listed in https://github.com/dotnet/corefx/issues/17131


chcosta:

chcosta:
This affects corefx, coreclr, and core-setup.

joperezr:
Correct me if I'm wrong but I believe this would break corefx since we won't be passing in the the override for csc host program

chcosta:
I've built corefx with these changes (https://github.com/dotnet/buildtools/pull/1394) on Windows in a VS2015 and a VS2017 dev command prompt and it built cleanly.  Whether it's actually using the compiler we want it to or not is a different question and that issue is being tracked in https://github.com/dotnet/corefx/issues/17131

PR #1391, opened 2017-03-15
Update certificate for xunit.runner.uwp

ericstj:
This certificate expired and we need a new one.

/cc @crummel @chcosta @weshaggard


chcosta:
Thank  you for fixing this

PR #1396, opened 2017-03-16
Fix meta-package suppression

ericstj:
The condition added previously didn't work because it didn't populate
the trimmed item _ConsolidatedDependencies.

I fixed this by passing the property into the task similar to
ApplyBaseline.

I also added support for an item @(SuppressMetaPackage) that can be used
to suppress a specific meta-package by name.

/cc @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyMetaPackages.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyMetaPackages.cs
@@ -5,6 +5,7 @@
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using NuGet.Frameworks;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/ApplyMetaPackages.cs
```diff
@@ -30,15 +31,33 @@ public class ApplyMetaPackages : PackagingTask
...
+                return true;
+            }
+
             var index = PackageIndex.Load(PackageIndexes.Select(pi => pi.GetMetadata("FullPath")));
             List<ITaskItem> updatedDependencies = new List<ITaskItem>();
 
+            var suppressMetaPackages = new HashSet<string>(SuppressMetaPackages.NullAsEmpty(), StringComparer.OrdinalIgnoreCase);
+
             // We cannot add a dependency to a meta-package from a package that itself is part of the meta-package otherwise we create a cycle
             var metaPackageThisPackageIsIn = index.MetaPackages.GetMetaPackageId(PackageId);
```
weshaggard: You could use the same mechanism for this case and just add metaPackageThisPackageIsIn to your suppressMetaPackages set.

ericstj: Sure, let me do that.

PR #1401, opened 2017-03-20
Remove all global fields and methods in GenFacades.

mellinoe:
It seems that some compilers, like the managed C++ compiler, will emit a bunch of global methods and fields into assemblies. We are removing a lot of other assembly-level metadata from projects being rewritten with GenFacades, but not those. This change causes us to clear out the "<Module>" type, which will wipe out all of those extra global members.

@weshaggard

### DIFF
--- a/src/GenFacades/GenFacades.Core/GenFacades.cs
+++ b/src/GenFacades/GenFacades.Core/GenFacades.cs
@@ -666,8 +666,14 @@ namespace GenFacades
                 // Add reference to seed core assembly up-front so that we keep the same order as the C# compiler.
                 assembly.AssemblyReferences.Add(_seedCoreAssemblyReference);
 
-                // Remove all type definitions except for the "<Module>" type.
+                // Remove all type definitions except for the "<Module>" type. Remove all fields and methods from it.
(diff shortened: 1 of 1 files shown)


src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -666,8 +666,14 @@ public override void RewriteChildren(Assembly assembly)
                 // Add reference to seed core assembly up-front so that we keep the same order as the C# compiler.
                 assembly.AssemblyReferences.Add(_seedCoreAssemblyReference);
 
-                // Remove all type definitions except for the "<Module>" type.
+                // Remove all type definitions except for the "<Module>" type. Remove all fields and methods from it.
                 assembly.AllTypes.RemoveAll(t => t.Name.Value != "<Module>");
+                NamespaceTypeDefinition moduleType = assembly.AllTypes.SingleOrDefault(t => t.Name.Value == "<Module>") as NamespaceTypeDefinition;
```
weshaggard: Depending on the number of types it is probably more efficient to grab the "<Module>" type first then call Clear and then add it back to the AllTypes.

ericstj:
LGTM /cc @erozenfeld

PR #1410, opened 2017-03-30
Add Extensibility point to optional.json so that it can be overwritten in repos

joperezr:
cc: @weshaggard @dagood 
FYI: @AlexGhiondea 

Conditioning the optional.json path so that it can be overwritten in a specific repo. The reason for this is that we want to also restore TestILC from the private feed, but we need the version of the package to match the one from the ProjectN TargetingPack which means that this optional.json needs to live in corefx side.


weshaggard:
>But if you want both (what we already have in it in BuildTools and another thing) you have to duplicate the json file, right?

Yes that is the compromise we are accepting for now.

In fact we should probably go ahead and remove the UAPTools version from the one checked into BuildTools as well and just have on the corefx side.

weshaggard:
>Ideally we would support multiple internal.json files, but then we would have to scan multiple lockfiles to resolve the tools, which I'd say is significantly more complex. A simple override seems reasonable for now.

Yes we will eventually move to PackageReference at which time we can much more easily merge sets of dependencies across buildtools and individual repos.

Issue #1418, opened 2017-04-03
ApiCompat CannotMakeMoreVisible rule shouldn't differentiate between Familiy and FamilyOrAssembly


ViktorHofer:
If implementation is Family and contract is FamilyOrAssembly this shouldn't be marked as an incompatible differnce: https://github.com/dotnet/buildtools/blob/master/src/ApiCompat/Rules/Compat/CannotMakeMoreVisible.cs

ViktorHofer:
Sry I did a pull request from the fork to the fork by mistake 👎

PR #1419, opened 2017-04-03
ApiCompat CannotMakeMoreVisible rule shouldn't differentiate between Familiy and FamilyOrAssembly

ViktorHofer:
fixes https://github.com/dotnet/buildtools/issues/1418

### DIFF
--- a/src/ApiCompat/Rules/Compat/CannotMakeMoreVisible.cs
+++ b/src/ApiCompat/Rules/Compat/CannotMakeMoreVisible.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Cci.Differs.Rules
             // If implementation is protected then contract must be protected as well. 
             if (impl.Visibility == TypeMemberVisibility.Family)
             {
-                if (contract.Visibility != TypeMemberVisibility.Family)
+                if (contract.Visibility != TypeMemberVisibility.Family && contract.Visibility != TypeMemberVisibility.FamilyOrAssembly)
                 {
                     differences.AddIncompatibleDifference(this,
                         "Visibility of member '{0}' is '{1}' in the implementation but '{2}' in the contract.",
@@ -47,7 +47,7 @@ namespace Microsoft.Cci.Differs.Rules
             // If implementation is protected then contract must be protected as well. 
             if (impl.GetVisibility() == TypeMemberVisibility.Family)
             {
-                if (contract.GetVisibility() != TypeMemberVisibility.Family)
+                if (contract.GetVisibility() != TypeMemberVisibility.Family && contract.GetVisibility() != TypeMemberVisibility.FamilyOrAssembly)
                 {
                     differences.AddIncompatibleDifference(this,
                         "Visibility of type '{0}' is '{1}' in the implementation but '{2}' in the contract.",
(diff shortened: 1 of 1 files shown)


src/ApiCompat/Rules/Compat/CannotMakeMoreVisible.cs
```diff
@@ -23,7 +23,7 @@ public override DifferenceType Diff(IDifferences differences, ITypeDefinitionMem
             // If implementation is protected then contract must be protected as well. 
             if (impl.Visibility == TypeMemberVisibility.Family)
```
weshaggard: I think you also want to update the condition to add ` || impl.Visibility == TypeMemberVisibility.FamilyOrAssembly`.

src/ApiCompat/Rules/Compat/CannotMakeMoreVisible.cs
```diff
@@ -47,7 +47,7 @@ public override DifferenceType Diff(IDifferences differences, ITypeDefinition im
             // If implementation is protected then contract must be protected as well. 
             if (impl.GetVisibility() == TypeMemberVisibility.Family)
             {
-                if (contract.GetVisibility() != TypeMemberVisibility.Family)
+                if (contract.GetVisibility() != TypeMemberVisibility.Family && contract.GetVisibility() != TypeMemberVisibility.FamilyOrAssembly)
```
weshaggard: I don't think types can really be `FamilyOrAssembly` (unless we go through this code path for nested types). At any rate it probably doesn't hurt anything to include it here. I would update the condition for the outter if as well like I called in the other comment.

weshaggard:
@ViktorHofer thanks for fixing this. Did you verify that it removes the noise from the baseline files in corefx?

src/ApiCompat/Rules/Compat/CannotMakeMoreVisible.cs
ViktorHofer: Oooh yes, I only cared about from impl to contract but the other way should also be included. Will fix that.

ViktorHofer:
Done. No haven't verified locally yet.

src/ApiCompat/Rules/Compat/CannotMakeMoreVisible.cs
```diff
@@ -7,56 +7,56 @@
...
-                        "Visibility of type '{0}' is '{1}' in the implementation but '{2}' in the contract.",
-                        impl.FullName(), impl.GetVisibility(), contract.GetVisibility());
-                    return DifferenceType.Changed;
-                }
-            }
-
-            return DifferenceType.Unknown;
-        }
-    }
+	// Removed because it appears the *MustExist rules already supercede these.
```
weshaggard: spaces instead of tabs please.

ViktorHofer:
@weshaggard Sorry for the formatting issue. Rebased and dropped last commit and push forced over it. Should be ok now

weshaggard:
Thanks @ViktorHofer for the fix.

ViktorHofer:
@weshaggard: Does it automatically get synced into corefx and if yes how long does it normally take? daily basis?

weshaggard:
BuildTools doesn't automatically get updated, it is done manually on demand. I know @mellinoe is working on updating CoreFx's BuildTools to get his changes over https://github.com/dotnet/buildtools/pull/1422.

PR #1421, opened 2017-04-04
Use PrefixToStrip correctly for better symbol indexing performance

dagood:
This allows `CreateRequest.cmd` to examine the files locally on the build machine for symbol info. The server then replaces `PrefixToStrip` with `UNCPath` and uses that as the location of the files at rest. See https://github.com/dotnet/buildtools/pull/1333 for when these targets were added.

Also:
 * Remove extra Symbols.targets include
 * Fix unpredictable behavior if the extract dir is being reused by deleting the dir before trying to use it. This is useful for manual indexing scenarios using this tooling
 * Add a few timestamped logs to make it easier to track how long each step of indexing takes

---

This significantly improves the speed of the CreateRequest step. I tested it out on a subset of CoreFX packages and got these results from my machine (similar network bottleneck as our build machines):

Before: 4.295 minutes
After: 0.068 minutes (this time is probably smaller in reality due to fixed costs and the short sample)

If that fraction is representative, that would bring the data in https://github.com/dotnet/corefx/pull/16939 to:

 * 4.1 minutes Extracting
 * 9.5 minutes Copying to share
 * **0.3** minutes Submitting symbol server indexing request
 * Total: **13.9** minutes (from 31.6).

This is shorter than the ~16 minute MyGet package push, so it shouldn't delay any aspect of the build if we put it in a parallel build leg.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
@@ -27,7 +27,11 @@
       <RequestCommand Condition="'$(SymbolsRequestDryRun)'!='true'">$(RequestCommand) $(SubmissionArg)</RequestCommand>
     </PropertyGroup>
 
+    <Message Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Running request command..." Importance="High"/>
+
     <Exec Command="$(RequestCommand)" />
+
+    <Message Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Running request command... Done." Importance="High"/>
   </Target>
 
   <!--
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
```diff
@@ -183,10 +207,17 @@
       <Output TaskParameter="Output" ItemName="SymbolPackageFileWithIndex" />
     </AddItemIndices>
 
+    <RemoveDir Directories="$(SymbolPackageExtractDir)"
+               Condition="Exists('$(SymbolPackageExtractDir')" />
```
chcosta: Missing `)`

dagood: Thanks, guess I didn't properly exercise it. Fixing.

dagood:
@chcosta Kind of, but performance isn't mentioned. `PrefixToStrip` is documented as allowing file indexing in a temporary location, so after connecting that indexing happens on the calling machine and that the file share is slow, I realized the perf reason to use `PrefixToStrip` that way. (I used to think it was just a weird utility feature for certain scenarios.) I'll link that (Msft internal) page to you in email--it's a very useful resource on CreateRequest in general.

weshaggard:
LGTM - Thanks for improving this.

PR #1422, opened 2017-04-05
Allow buildtools to be used with a .NET Core 2.0 CLI

mellinoe:
This PR includes a variety of changes that enable the use of buildtools with a .NET Core 2.0 CLI.

* Allow buildtools to be restored and initialized with a 2.0 CLI. This change is backwards-compatible, meaning a 1.1 toolset can still be used, which means nobody will be blocked by the change going in. This is a point-in-time solution, meaning we should eventually remove all of the support for non-2.0 CLI's when our repos have moved off of it.
  * Setting the `BUILDTOOLS_USE_CSPROJ` environment variable causes MSBuild projects targeting .NET Core 2.0 to be used instead of project.json files, which are not supported with the 2.0 toolset.
  * The version numbers in all of our runtimeconfig.json files are automatically updated to the highest shared framework version supported by the CLI being used. This is handled in init-tools.sh and init-tools.ps1, a new initialization script called by init-tools.cmd. The logic for this would be too cumbersome to implement in a batch script.
* PrereleaseResolveNugetPackageAssets correctly handles lower-case package names. There was a breaking change in nuget behavior sometime since our last CLI upgrade. Other than this, there are no other changes needed in our custom tasks.
* ~~Include support for auto-updating MSBuild files with `PackageReference` items in them. This part was surprisingly easy to integrate. I just copied the existing project.json updater and modified the relevant text-editing portions. @dagood~~ I removed this as we don't really need it. Since we can just base PackageReference versions off of other MSBuild properties, we can just rely on the existing XML updater.

Corefx will require a large set of changes in order to consume this. See the diff here:

https://github.com/dotnet/corefx/compare/master...mellinoe:2.0-toolset

Progress towards https://github.com/dotnet/corefx/issues/17260

@weshaggard @joperezr @ericstj @dagood PTAL


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
```diff
@@ -0,0 +1,15 @@
+ param (
+    [Parameter(Mandatory=$true)][string]$ToolRuntimePath,
+    [Parameter(Mandatory=$true)][string]$DotnetCmd
+ )
+
+# Override versions in runtimeconfig.json files with highest available runtime version.
+$mncaFolder = (gi $DotnetCmd).Directory.FullName + "\shared\Microsoft.NETCore.App"
```
dagood: \<nit> `gi`: I usually see style guides saying "don't use aliases in scripts" for powershell stuff, but I don't know if that's generally followed.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
```diff
@@ -0,0 +1,187 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+
+namespace Microsoft.DotNet.VersionTools.Dependencies
+{
+    public class MSBuildPackageReferenceUpdater : IDependencyUpdater
```
dagood: At first glance I'd like this to share with `ProjectJsonUpdater`, but there isn't much value in that if `ProjectJsonUpdater` goes away like it eventually should.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
```diff
@@ -0,0 +1,187 @@
...
+
+                    Action update = FileUtils.GetUpdateFileContentsTask(
+                        filePath,
+                        contents => ReplaceAllDependencyVersions(
+                            contents,
+                            filePath,
+                            dependencyBuildInfos,
+                            out dependencyChanges));
+
+                    // The output json may be different even if there weren't any changes made.
```
dagood: `json` -> `xml`.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            XDocument documentRoot = XDocument.Parse(input);
+
+            dependencyChanges = FindAllDependencyProperties(documentRoot)
+                    .Select(dependencyProperty => ReplaceDependencyVersion(projectJsonFile, dependencyProperty, buildInfos))
+                    .Where(buildInfo => buildInfo != null)
+                    .ToArray();
+
+            StringWriter sw = new StringWriter();
+            documentRoot.Save(sw, SaveOptions.None);
```
dagood: Hmm, `None` vs. `DisableFormatting`? I think `DisableFormatting` with `Parse` using `LoadOptions.PreserveWhitespace` would leave whitespace as similar as possible, but I'm not completely sure.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
```diff
@@ -0,0 +1,187 @@
...
+            return sw.ToString();
+        }
+
+        /// <summary>
+        /// Replaces the single dependency with the updated version, if it matches any of the
+        /// dependencies that need to be updated. Stops on the first updated value found.
+        /// </summary>
+        /// <returns>The BuildInfo used to change the value, or null if there was no change.</returns>
+        private DependencyChange ReplaceDependencyVersion(
+            string projectJsonFile,
```
dagood: `json` -> `xml`, or maybe just change this to `projectFile`.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
```diff
@@ -0,0 +1,187 @@
...
+
+                    string oldVersion;
+                    // Support the new  "inline" Version attribute, or the old-style nested Version element.
+                    XAttribute versionAttribute = null;
+                    XElement versionElement = null;
+
+                    versionAttribute = packageRefElement.Attribute(XName.Get("Version", packageRefElement.GetDefaultNamespace().NamespaceName));
+                    if (versionAttribute == null)
+                    {
+                        versionElement = packageRefElement.Element(XName.Get("Version", packageRefElement.GetDefaultNamespace().NamespaceName));
```
dagood: Repeated long parameter in a long line, hard to read.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
```diff
@@ -0,0 +1,187 @@
...
+                    else
+                    {
+                        if (versionElement == null)
+                        {
+                            Trace.TraceWarning($"Couldn't parse the version information for package '{id}' in '{projectJsonFile}'. Skipping.");
+                            continue;
+                        }
+
+                        oldVersion = versionElement.Value;
+                    }
```
dagood: This structure isn't immediately clear--`if (a!=null) else if (e!=null) else fail` seems better to me.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: That was my thought as well, eventually we will just be left with this one.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: I'll try that out. It might just be that the files I tested were already formatted the same way as this gives me, because I didn't see any extra diffs.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: Fixing.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: Yeah, that is a better pattern, fixing.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: Fixing

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: Fixing

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
mellinoe: Good point. I'll remove all alias usages.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -14,7 +14,11 @@ set BUILDTOOLS_PACKAGE_DIR=%~dp0
...
 set ROSLYNCOMPILERS_VERSION=2.0.0-rc
-set MSBUILD_CONTENT_JSON={"dependencies": { "MicroBuild.Core": "%MICROBUILD_VERSION%", "Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%"},"frameworks": {"netcoreapp1.0": {},"net46": {}}}
+if [%BUILDTOOLS_USE_CSPROJ%]==[] (
+  set MSBUILD_PROJECT_CONTENT={"dependencies": { "MicroBuild.Core": "%MICROBUILD_VERSION%", "Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%"},"frameworks": {"netcoreapp1.0": {},"net46": {}}}
+) ELSE (
+  set MSBUILD_PROJECT_CONTENT=^^^<Project Sdk="Microsoft.NET.Sdk"^^^>^^^<PropertyGroup^^^>^^^<TargetFrameworks^^^>netcoreapp1.0;net46^^^</TargetFrameworks^^^>^^^<DisableImplicitFrameworkReferences^^^>true^^^</DisableImplicitFrameworkReferences^^^>^^^</PropertyGroup^^^>^^^<ItemGroup^^^>^^^<PackageReference Include="MicroBuild.Core" Version="%MICROBUILD_VERSION%" /^^^>^^^<PackageReference Include="Microsoft.Portable.Targets" Version="%PORTABLETARGETS_VERSION%" /^^^>^^^<PackageReference Include="Microsoft.Net.Compilers" Version="%ROSLYNCOMPILERS_VERSION%" /^^^>^^^</ItemGroup^^^>^^^</Project^^^>
```
alexperovich: In the `sh` file you don't even need to escape the newlines, just add newlines inside single quotes. OFC cmd.exe doesn't support anything like that IIRC.

dagood: Good point, only reason to escape them would be if we don't want them to show up in the output, but them being in the output is actually a bonus.

weshaggard: For the msbuild project I wouldn't actually do it the same way. Instead I would check it in as a file and just pass the versions as properties when you do the restore. The only reason we ever generated it here is because we need the version to be in-sync but msbuild has support for doing this by passing in the property.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -14,7 +14,11 @@ set BUILDTOOLS_PACKAGE_DIR=%~dp0
 set MICROBUILD_VERSION=0.2.0
 set PORTABLETARGETS_VERSION=0.1.1-dev
 set ROSLYNCOMPILERS_VERSION=2.0.0-rc
-set MSBUILD_CONTENT_JSON={"dependencies": { "MicroBuild.Core": "%MICROBUILD_VERSION%", "Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%"},"frameworks": {"netcoreapp1.0": {},"net46": {}}}
+if [%BUILDTOOLS_USE_CSPROJ%]==[] (
```
weshaggard: I would like us to do a version check of the CLI and opt-in to the csproj if the version is high enough to support it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -30,51 +34,61 @@ if not exist "%DOTNET_CMD%" (
...
 
 :: Copy Portable Targets Over to ToolRuntime
 if not exist "%PACKAGES_DIR%\generated" mkdir "%PACKAGES_DIR%\generated"
-set PORTABLETARGETS_PROJECTJSON=%PACKAGES_DIR%\generated\project.json
-echo %MSBUILD_CONTENT_JSON% > "%PORTABLETARGETS_PROJECTJSON%"
+set PORTABLETARGETS_PROJECT=%PACKAGES_DIR%\generated\project.json
+echo %MSBUILD_PROJECT_CONTENT% > "%PORTABLETARGETS_PROJECT%"
 @echo on
-call "%DOTNET_CMD%" restore "%PORTABLETARGETS_PROJECTJSON%" %INIT_TOOLS_RESTORE_ARGS% --packages "%PACKAGES_DIR%\."
+call "%DOTNET_CMD%" restore "%PORTABLETARGETS_PROJECT%" %INIT_TOOLS_RESTORE_ARGS% --packages "%PACKAGES_DIR%\."
```
weshaggard: How does this restore work with the msbuild based CLI?

mellinoe: What do you mean? It will work the same as before, except the file will be an MSBuild project instead of a json file now.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
mellinoe: I cleaned up the formatting of both versions so it is much more legible now. Making these into standalone files is doable, but it's separate from the change here. We could have done it before; these files are not actually parameterized, all of the versions are hard-coded in the previous lines, so they could have been hard-coded in a separate file.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: I don't see you creating a csproj for this PORTABLETARGETS_PROJECT or .

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: They were originally generated to put the version number in there without requiring the version number to be duplicated across scripts, that certainly has regressed over time :(

I still think this should be moved out of these scripts but I can wait until @dagood (or someone on core-eng) changes the bootstrapping.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
```diff
@@ -0,0 +1,191 @@
...
+using System.Xml;
+using System.Xml.Linq;
+
+namespace Microsoft.DotNet.VersionTools.Dependencies
+{
+    public class MSBuildPackageReferenceUpdater : IDependencyUpdater
+    {
+        public IEnumerable<string> MSBuildFilePaths { get; }
+
+        public MSBuildPackageReferenceUpdater(IEnumerable<string> msbuildFilePaths)
```
weshaggard: I'm not even sure we need this updater do we? When we start using ProjectReferences I expect us to not need to update the project files most of the time, at least not for the build number changes as I want that to be an msbuild property that we just update in one single file for the entire repo.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
mellinoe: Actually, I missed the fact that we are doing manual copies out of the packages directory later on in the file, and the paths that we copy are based on these version variables, e.g.:
`Robocopy "%PACKAGES_DIR%\MicroBuild.Core\%MICROBUILD_VERSION%\build\." "%TOOLRUNTIME_DIR%\." /E`

So they are used in two places and we'd then have to duplicate them if they were hard-coded in a separate file. I definitely agree we should move them out of the script... Actually I think we should re-evaluate a lot of this crazy stuff we are doing in init-tools, especially if the new MSBuild-based CLI tools makes it obsolete.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: Hm, it's a good point, and I think we discussed this yesterday in person. We would still need some "updater task" that ran on the global properties file, and we'd need to come up with some sort of naming convention, right? I have seen other repos with a pattern like:

```XML
<PackageReference Include="MyPackageName" Version="$(MyPackageNameVersion") />
<PackageReference Include="OtherPackageName" Version="$(OtherPackageNameVersion") />
```

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
mellinoe: It's a bit confusing because the `MSBUILD_PROJECT_CONTENT` (formerly `MSBUILD_CONTENT_JSON`) actually gets put into the `PORTABLETARGETS_PROJECT` file... There's just one generated project.json/csproj here. I'll at least rename the "content" variable so it matches the file name.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
mellinoe: Hm, that should be doable by looking at the output from dotnet --version. I'll try that out.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
weshaggard: I would like to try and just use the xml update and update just the package version in dependencies.props and that should be the only property to be updated.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -30,51 +54,61 @@ if not exist "%DOTNET_CMD%" (
...
 )
 
 :: Microsoft.Build.Runtime dependency is causing the MSBuild.runtimeconfig.json buildtools copy to be overwritten - re-copy the buildtools version.
 Robocopy "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%\." "MSBuild.runtimeconfig.json"
 
 :: Copy Portable Targets Over to ToolRuntime
 if not exist "%PACKAGES_DIR%\generated" mkdir "%PACKAGES_DIR%\generated"
-set PORTABLETARGETS_PROJECTJSON=%PACKAGES_DIR%\generated\project.json
-echo %MSBUILD_CONTENT_JSON% > "%PORTABLETARGETS_PROJECTJSON%"
+set PORTABLETARGETS_PROJECT=%PACKAGES_DIR%\generated\project.json
```
weshaggard: So you are putting different content into the project.json file? Does the tooling care if this is a .json file instead of .csproj?

mellinoe: Ah, I see what you mean. The tooling doesn't care what the file is called, actually. It just assumes it's an MSBuild file.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: I see; we already have these properties in dependencies.props. Let me try using them and make sure everything still works.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: Perhaps we are getting lucky with the restore target but I do know there are targets that make assumptions based on the file extension. However perhaps to make this a little less magical can you just rename the file to neither and just call it something like project_packages or something that doesn't connect to either format.

mellinoe: "Old CLI" will only restore things that are actually called "project.json" as far as I can tell. I can just have this script set the name to "project.csproj" if MSBuild is used.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: One thing we would lose if we removed this would be the validation part, which could kick in if someone accidentally hard-coded a version string into a csproj file that later fell out of sync. If we kept this, we'd still get that validation, even if we didn't plan to use it for actually updating version strings directly.

mellinoe: Using the existing properties in corefx works nicely, so we don't necessarily need to keep this code. Thoughts about the "validation" part, though?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -103,7 +103,11 @@ Robocopy "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%\." "MSBuild.runtimeconf
 
 :: Copy Portable Targets Over to ToolRuntime
 if not exist "%PACKAGES_DIR%\generated" mkdir "%PACKAGES_DIR%\generated"
-set PORTABLETARGETS_PROJECT=%PACKAGES_DIR%\generated\project.json
+if [%BUILDTOOLS_USE_CSPROJ%]==[] (
+  set PORTABLETARGETS_PROJECT=%PACKAGES_DIR%\generated\project.json
```
weshaggard: Probably easy enough to just add another variable for the extension and use it for both projects.

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
weshaggard: We currently don't even have validation enabled in corefx and honestly I don't think it is a big issue, especially given that the properties will not be spread out everywhere. If we see a hard-coded version then it should be a red flag to be eliminated or it should have a comment explaining why it is hard-coded.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/tool-runtime.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
+    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">2.0.0-beta-001776-00</RuntimeFrameworkVersion>
+    <RuntimeIdentifiers>centos.7-x64;debian.8-x64;fedora.23-x64;fedora.24-x64;opensuse.13.2-x64;opensuse.42.1-x64;osx.10.10-x64;rhel.7-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64;win7-x64;win7-x86</RuntimeIdentifiers>
+    <PackageTargetFallback Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">$(PackageTargetFallback);dnxcore50;portable-net45+win8</PackageTargetFallback>
```
weshaggard: Do we actually need these fallbacks? If not I would like to remove them, in particular dnxcore50.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/tool-runtime.csproj
```diff
@@ -0,0 +1,32 @@
...
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
+    <PackageReference Include="Microsoft.Build.Framework" Version="15.1.0-preview-000523-01" />
```
weshaggard: I think it would be good if we use properties for these versions so we can now start to share versions and update them, even potentially update them from individual repos.

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
```diff
@@ -899,6 +899,15 @@ private string GetNuGetPackagePath(string packageId, string packageVersion)
                 {
                     return packagePath;
                 }
+                else
+                {
+                    // Try a lower-case path.
+                    string lowerCasePath = packagePath.ToLowerInvariant();
```
weshaggard: Does this cause us to further diverge from the shipped build task?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/tool-runtime.csproj
mellinoe: I tried to remove them initially, but I believe this package needed them: `Microsoft.Net.Compilers.Targets.NetCore`. I moved us from `Microsoft.Composition` to `System.Composition`, so that one is no longer a problem, at least.

src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
mellinoe: Yes

src/Microsoft.DotNet.VersionTools/Dependencies/MSBuildPackageReferenceUpdater.cs
mellinoe: Ok, I'll remove these changes, then.

dagood:
+cc @ellismg in the unlikely case these changes weren't on your radar.

PR #1425, opened 2017-04-06
Provide a way to ignore missing files during pack

ellismg:
On some platforms (e.g. Unix) we can't build desktop assets. Provide a way
for these projects to signal that when packing nuspec's if files are
missing it's okay and they should just be dropped from the package
(instead of the NuGet logic throwing an exception).

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs
@@ -125,6 +125,15 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
             set;
         }
 
+        /// <summary>
+        /// If true, entries in the manifest which can not be found on disk are ignored.
+        /// </summary>
+        public bool IgnoreMissingFiles
+        {
+            get;
+            set;
+        }
+
         public override bool Execute()
         {
             if (Nuspecs == null || Nuspecs.Length == 0)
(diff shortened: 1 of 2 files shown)


ellismg:
@weshaggard @ericstj I plan to use this technology when enabling building buildtools on *nix.

weshaggard:
While I'm not opposed to this I wonder if a better option would be to switch to pkgproj's and condition those files on the other platforms. Doing it this way potentially allows accidental file removals without realizing it.

ericstj:
@ellismg asked me about this and I suggested it as the quickest way to get things moving.  I do agree that switching to pkgproj and conditioning stuff is better since its intentional but that'd require more effort.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs
```diff
@@ -251,7 +260,27 @@ public void Pack(string nuspecPath, string nupkgPath, Manifest manifest, bool pa
...
+                    if (!IgnoreMissingFiles)
+                    {
+                        return true;
+                    }
+
+                    PackageBuilder temp = new PackageBuilder();
+                    bool didAdd = false;
+
+                    try {
+                        temp.AddFiles(baseDirectoryPath, m.Source, m.Target, m.Exclude);
```
ericstj: Why didn't you just check if the file existed before adding it?

ellismg: ManifestFile is a thin wrapper around the `<File>` entry in the Manifest.  That means that m.Source may not be a flie, but rather a wildcard pattern (including `**` style globs) so there's no good way to test if that exists, that I've found, short of letting NuGet do its resolution.

ericstj: I see.  This is a lot uglier than I was expecting :/.  As we discussed, maybe have a look at pkgproj conversion and if it isn't too much work that might be better.

PR #1426, opened 2017-04-06
Add config file for desktop test runners

safern:
Adding config file for desktop runners (xunit and remote executor).

cc: @weshaggard @joperezr @tarekgh


safern:
> Didn't the name had to match the xunit runer exe? If you are planning to rename it instead when you do the copying, can you add Xunit to the name so it's more descriptive? something like DesktopXUnitRunnerConfigFile.config

It is not just for xunit it is also for RemoteExecutorApp.exe that is why I named it like a "generic" file and I will rename it when copying.

tarekgh:
:+1:

PR #1428, opened 2017-04-08
Get buildtools building on *nix

ellismg:
The gets buildtools building and generating NuGet packages.  The tests don't run for a few reasons:

1. NuGetRuntimeIdentifier is always set to win7-x64 or some other windows like RID.
2. The project.json from the tests assembly does not contain any *NIX rids
3. The tests themselves have some assumptions about filesystem enumeration on *nix which don't hold causing one test to fail.  Talking with Eric we should just update the test, but given that I'm keeping tests disabled for now, I didn't do that.

I don't really understand what the right way to handle NuGetRuntimeIdentifier is.  I assume that we should be detecting it based on the host machine unless someone specifically sets it, but it wasn't 100% clear to me what exactly we need to do.

Once this lands, I would like to update buildtools to consume a newer version of itself.  When testing that locally I ran into some more issues, so I'm holding off on that for now.


bootstrap/bootstrap.sh
```diff
@@ -270,5 +270,9 @@ done <$rootToolVersions
 
 cp $rootToolVersions $bootstrapComplete
 
+if [ -e "$toolsLocalPath/FrameworkTargeting.targets" ]; then
```
ellismg: This works around a casing issue, we have a hack that changes the case of this file in init-tools.sh because the PCL projects expect the "Targets" extension.  Once we can take an updated FrameworkTargeting file (which I'm touching in this PR) we can remove the hack.

config.json
```diff
@@ -175,7 +175,7 @@
           "path": "Tools/msbuild.cmd"
         },
         "unix":{
-          "defaultParameters": "/nologo /verbosity:minimal /clp:Summary /maxcpucount /l:BinClashLogger,Tools/Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log",
+          "defaultParameters": "/nologo /verbosity:minimal /clp:Summary /maxcpucount /l:BinClashLogger,Tools/Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log /p:SkipTests=true",
```
weshaggard: Are you sure you want to put SkipTests=true at this level?

ellismg: Would you rather I set it as property in dir.props?

weshaggard: I think I would prefer it to be somewhere we can add a comment about it, and file an issue on it because we should be able to run tests on unix as well.

ellismg: Moved to the dir.props file, opened https://github.com/dotnet/buildtools/issues/1430 and added a comment.

PR #1433, opened 2017-04-12
Fix casing issue in PrereleaseResolveNuGetPackageAssets.

mellinoe:
Instead of making the entire package path lower-case, we should only lower-case the part that NuGet changes (the package ID).

@ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
@@ -902,7 +902,7 @@ namespace Microsoft.DotNet.Build.Tasks
                 else
                 {
                     // Try a lower-case path.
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs
```diff
@@ -902,7 +902,7 @@ private string GetNuGetPackagePath(string packageId, string packageVersion)
                 else
                 {
                     // Try a lower-case path.
-                    string lowerCasePath = packagePath.ToLowerInvariant();
+                    string lowerCasePath = Path.Combine(packagesFolder, packageId.ToLowerInvariant(), packageVersion);
```
ericstj: Can you double check if NuGet is also ToLower'ing the packageVersion? I believe that can also contain uppercase characters.

mellinoe: It does seem to force that to lower-case as well.

ericstj:
LGTM

Issue #1437, opened 2017-04-14
Error during using GenAPI on linux


WonyoungChoi:
When building a project referencing "Microsoft.DotNet.BuildTools.GenAPI" with dotnet-cli tool on linux, 
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netcoreapp1.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.BuildTools.GenAPI" Version="1.0.0-beta3-01512-01" />
  </ItemGroup>
</Project>
```
Microsoft.DotNet.BuildTools.GenAPI.targets prints following error:
```
/home/zzervb/.nuget/packages/microsoft.dotnet.buildtools.genapi/1.0.0-beta3-01512-01/build/Microsoft.DotNet.BuildTools.GenAPI.targets(27,5): error MSB3073: The command ""/home/zzervb/.nuget/packages/microsoft.dotnet.buildtools.genapi/1.0.0-beta3-01512-01/build/../tools/GenApi.exe" -assembly:"/home/zzervb/test/lib1/bin/Debug/netcoreapp1.0/lib1.dll" -libPath:.....................
```

The file path of GenAPI.exe is wrong and the file doesn't have executable permission. :(

PR #1438, opened 2017-04-14
Fix Microsoft.DotNet.BuildTools.GenAPI.targets for using on linux

WonyoungChoi:
GenAPI.exe file name was wrong in Microsoft.DotNet.BuildTools.GenAPI.targets. 
And use the mono runtime to use .NETFramework 4.6 on linux.

Issue: https://github.com/dotnet/buildtools/issues/1437

### DIFF
--- a/src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
+++ b/src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
@@ -2,9 +2,12 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <GenAPITargetPath Condition="'$(GenAPITargetPath)' == ''">$(TargetDir)$(TargetName).cs</GenAPITargetPath>
-    <_GenAPIPath>$(MSBuildThisFileDirectory)\..\tools\GenApi.exe</_GenAPIPath>
+    <_GenAPIPath>$(MSBuildThisFileDirectory)\..\tools\GenAPI.exe</_GenAPIPath>
 
-    <!-- Sequence ourselves last in PrepareForRun 
+    <_GenAPICommand Condition=" '$(OS)' == 'Windows_NT'">"$(_GenAPIPath)"</_GenAPICommand>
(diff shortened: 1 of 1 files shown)


src/GenAPI.Desktop/Microsoft.DotNet.BuildTools.GenAPI.targets
```diff
@@ -2,9 +2,12 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <GenAPITargetPath Condition="'$(GenAPITargetPath)' == ''">$(TargetDir)$(TargetName).cs</GenAPITargetPath>
-    <_GenAPIPath>$(MSBuildThisFileDirectory)\..\tools\GenApi.exe</_GenAPIPath>
+    <_GenAPIPath>$(MSBuildThisFileDirectory)\..\tools\GenAPI.exe</_GenAPIPath>
 
-    <!-- Sequence ourselves last in PrepareForRun 
+    <_GenAPICommand Condition=" '$(OS)' == 'Windows_NT'">"$(_GenAPIPath)"</_GenAPICommand>
+    <_GenAPICommand Condition=" '$(OS)' != 'Windows_NT'">mono --runtime=v4.0.30319 "$(_GenAPIPath)"</_GenAPICommand>
```
weshaggard: I'm fine introducing the extension point and the casing but I don't really like making the assumption that not windows means mono. In the majority of our tools not windows implies .NET Core and not mono. With the extension point you should be able to set _GenAPICommand in your project or common targets to get this setup for mono.

akoeplinger: @weshaggard this is in GenAPI.Desktop which only targets .NET Framework, so relying on mono as the "Desktop" runtime outside of Windows sounds fine to me.

The `--runtime=v4.0.30319` is unnecessary though, unless you have a very old Mono (and GenAPI probably wouldn't work there anyway).

weshaggard: Fair. I could live with that.

WonyoungChoi:
@weshaggard This PR can be temporary measure. I think ultimately the .netcore runtime (dotnet hostfx) should be used instead of mono and legacy windows .net runtime. We need to consider including the .netcore self-contained GenAPI.exe app in tools directory of this nuget package in the future.
@akoeplinger The "--runtime=v4.0.30319" option is the least restriction action to prevent using very old mono runtime.

PR #1440, opened 2017-04-15
Adding Silverlight public keys signing files

tarekgh:


tarekgh:
https://github.com/dotnet/corefx/issues/18172

tarekgh:
CC @weshaggard 

This change just adding the Silverlight public key files to the buildtools. I didn't change sign.targets here because I am going to apply the changes in corefx repo anyway and also sign.targets in buildtools is very different. 
after getting this merged, I'll apply the needed changes in corefx repo

PR #1441, opened 2017-04-17
Explicit harvesting

ericstj:
/cc @weshaggard


ericstj:
FYI @MattGal and @StephenBonikowsky 

This contains some changes that will impact WCF packaging.  

To retain the old harvesting behavior you can set HarvestIncludeAllPaths=true, but it would be better to make intentional updates to WCF packages around which assets to harvest like I did here: https://github.com/dotnet/corefx/pull/18498/commits/e09ff343e5c243724e222c7cfb368b4313f0c818

This change makes it unnecessary to use SuppressNETStandardInference, so any instance of that can be cleaned up.  

Instead we actually test for NETStandard in the packages to ensure that there is a runnable netstandard impl for any package (eg: it doesn't purely rely on the sum of mutually exclusive frameworks / rids).  If you see failures around this and you don't wish to fix them you can supress them like I've done here: https://github.com/dotnet/corefx/pull/18498/commits/9c0e3b954a1abe9d602ad695052d51c4371e7251#diff-13168d9811b040b8e72d4fa2dc7625f2R25

PR #1447, opened 2017-04-21
Update license for .NET Core NuGet packages

richlander:
We are adopting a plan to change the license and copyright of .NET Core packages. We want to make this change for open source packages. If there are packages that are still closed source, in part of in full, they should not be affected by this change.

Is this the right change to make to achieve the stated goal, above?

I saw that there are [nuspecs in this repo](https://github.com/dotnet/buildtools/search?utf8=%E2%9C%93&q=329770&type=) and know that Roslyn and other components have their own approach. I thought we should start here.

More context: https://github.com/dotnet/corefx/issues/12190

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -54,11 +54,11 @@
     <Authors>Microsoft</Authors>
     <Owners>microsoft,dotnetframework</Owners>
     <Description>TODO</Description>
-    <LicenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</LicenseUrl>
+    <LicenseUrl>https://github.com/dotnet/corefx/blob/master/LICENSE</LicenseUrl>
     <IconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</IconUrl>
-    <Copyright>&#169; Microsoft Corporation.  All rights reserved.</Copyright>
+    <Copyright>.NET Foundation and Contributors</Copyright>
     <Tags></Tags>
-    <RequireLicenseAcceptance>true</RequireLicenseAcceptance>
+    <RequireLicenseAcceptance>false</RequireLicenseAcceptance>
     <Serviceable Condition="'$(Serviceable)' == ''">true</Serviceable>
     <!-- we depend on nuget v2.12 / v3.4 behavior NuGet doesn't support two different min client versions
          so we declare 2.12 and mention in package description that when using 3.x we require 3.4 or later -->


terrajobst:
Love it!

@jaredpar how do you guys generate the `Microsoft.CodeAnalysis.*` packages? You're not using build tools, right? So we probably need to replicate this change there.

@richlander @DamianEdwards who does packaging on the ASP.NET side? We probably want the same change there.

Petermarcu:
@ellismg @ericstj to review the change.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -54,11 +54,11 @@
     <Authors>Microsoft</Authors>
     <Owners>microsoft,dotnetframework</Owners>
     <Description>TODO</Description>
-    <LicenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</LicenseUrl>
+    <LicenseUrl>https://github.com/dotnet/corefx/blob/master/LICENSE</LicenseUrl>
```
bartonjs: I don't know that this should point to master.  If we change the license in 5 years we're (presumably) not retroactively changing the packages; so this should be either a release branch, or a specific commit hash...  Or stay as something not on github.

terrajobst: I think the intent is that we want to express that we'll never change the license.

DamianEdwards:
Adding @eilon for ASP.NET considerations

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
richlander: You are both right, but yes, I agree with @terrajobst that this isn't intended to be a short-term decision. We're not intending on leaving GitHub either ;)

ericstj: If you want to be careful you could just remove the default from the buildtools targets and force repos consuming buildtools to define it.  That way someone won't accidentally ship with the wrong license.

We're already doing that for the file itself, just not the URL apparently.
https://github.com/dotnet/corefx/blob/master/pkg/dotnet_library_license.txt 
https://github.com/dotnet/corefx/blob/master/Packaging.props#L5

Eilon:
I'd like to get an official message from LCA on this. They are rarely happy when I change a license without hearing from them first! 😄

terrajobst:
We've already clarified this with LCA for .NET Core. I don't think we need to treat ASP.NET Core separately because you guys use the same EULA we do.

Eilon:
@terrajobst we use different licenses in GitHub (Apache vs. MIT), though we do use the same license in NuGet.org. For that reason, I'm not certain the conclusion is the same, because ASP.NET Core would end up using a very different license.

terrajobst:
Oh, I didn't know hat. I always assumed you were on MIT as well. In that case let me loop you in on the discussion we have.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
richlander: @ericstj that makes a lot of sense to me.

Is this a matter of going to the other repos and getting them to override the license value?

stephentoub:
@richlander, should this PR from over a year ago be closed?

PR #1448, opened 2017-04-21
Remove Benchview calls when not uploading

DrewScoggins:
We did not have calls to measurement.py behind the LogToBenchview flag
and so when people would do local runs without the Benchview tooling
their runs would fail.  This should be fixed.  I also removed an out of
date comment.


DrewScoggins:
@mellinoe PTAL

mellinoe:
LGTM.

PR #1450, opened 2017-04-21
Enable build of partial reference facades

ericstj:
Adds a mode to GenFacades that starts with an assembly and replaces any
types present in the seeds with typeforwards.  Types not in the seeds
will remain in the assembly.

We'll use this for generating desktop reference facades.

To use this feature set IsPartialFacadeAssembly=true in a reference
project.

After the source is compiled GenFacades will examine all references and
replace any type def in the reference assembly that exists in the
references.

To keep a type in the reference assembly even if it exists in a
reference create an item in the reference project as follows:
```
    <SeedTypePreference Include="namespace.type">
      <Assembly>assemblyName</Assembly>
    </SeedTypePreference>
```

/cc @mellinoe @weshaggard

### DIFF
--- a/src/GenFacades/GenFacades.Core/GenFacades.cs
+++ b/src/GenFacades/GenFacades.Core/GenFacades.cs
@@ -33,7 +33,8 @@ namespace GenFacades
             string[] seedTypePreferencesUnsplit = null,
             bool forceZeroVersionSeeds = false,
             bool producePdb = true,
-            string partialFacadeAssemblyPath = null)
+            string partialFacadeAssemblyPath = null,
+            bool buildPartialReferenceFacade = false)
         {
             if (!Directory.Exists(facadePath))
                 Directory.CreateDirectory(facadePath);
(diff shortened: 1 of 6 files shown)


src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -433,7 +453,27 @@ public Assembly GenerateFacade(IAssembly contractAssembly, IAssemblyReference se
...
+                        }
+                    }
+
                     AddTypeForward(assembly, seedType);
+                    addedForward = true;
+                }
+
+                if (buildPartialReferenceFacade &&  !addedForward)
+                {
+                    Trace.TraceError("Did not find any types in any of the seed assemblies.");
```
weshaggard: I assume in this case it is either an error or you should no longer set IsPartialFacade.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
```diff
@@ -1,5 +1,35 @@
...
+    <FillPartialFacadeDependsOn>
+      EnsureBuildToolsRuntime
+    </FillPartialFacadeDependsOn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IsPartialFacadeAssembly)' == 'true' AND '$(IsReferenceAssembly)' != 'true'">
+    <!-- Tell ResolveMatchingContract to run and resolve contract to project reference -->
+    <ResolveMatchingContract>true</ResolveMatchingContract>
+    <FillPartialFacadeDependsOn>
+      $(FillPartialFacadeDependsOn);ResolveContract
```
weshaggard: Why is this necessary now?

src/GenFacades/GenFacades.Core/GenFacades.cs
ericstj: No this is the case where the type remains in the contract.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.targets
ericstj: It's only used when building a normal partial facade.  That's pre-existing and required in order to resolve the contract.  It's showing as a diff because I refactored the duplicated code into the common targets file.

src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -433,7 +453,27 @@ public Assembly GenerateFacade(IAssembly contractAssembly, IAssemblyReference se
...
                     }
 
+                    if (buildPartialReferenceFacade)
+                    {
+                        // honor preferSeedType for keeping contract type
+                        string preferredSeedAssembly;
+                        bool keepType = _seedTypePreferences.TryGetValue(docId, out preferredSeedAssembly) &&
+                            contractAssemblyName.Equals(preferredSeedAssembly, StringComparison.OrdinalIgnoreCase);
+
+                        if (!keepType)
```
mellinoe: If I'm reading this right, should we have an `else { continue; }` here? If we're keeping the existing type, then we don't need a type-forward, right? It might end up as a no-op later on, I guess.

mellinoe:
Looks good to me.

Issue #1449, opened 2017-04-21
Add GeneratePlatformNotSupportedAssemblyMessage support to GenApis


danmosemsft:
21 CoreFX libraries at least in some configurations set `<GeneratePlatformNotSupportedAssembly>` which contains generic PNSE with messages `Operation is not supported on this platform.`

We should wire through an optional improved message that projects can set like `<GeneratePlatformNotSupportedAssembly_Message>Operations with CNG keys are only supported on Windows platforms.</GeneratePlatformNotSupportedAssembly_Message>`

For now a single message can be used in all PNSE in a generated assembly.

This will help steer devs towards supported API's and explain when they CAN expect the API to work without needing to post questions.

Just pass a parameter to https://github.com/dotnet/buildtools/blob/73d034cddac9ed39af4c3acc4a030cc8598f1006/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs#L204. and put it in notsupported.targets. Then update buildtools in CoreFX and add good messages for all 21 cases.

@sepidehMS could you please give this a shot.

danmosemsft:
https://github.com/dotnet/corefx/issues/18733#issuecomment-296241643

PR #1451, opened 2017-04-21
Add custom PNSE message to GenAPI

sepidehMS:
Resolves https://github.com/dotnet/buildtools/issues/1449. Enables assemblies that set `<GeneratePlatformNotSupportedAssembly>` to pass in an extra optional property to set a custom message for PlatformNotSupportedException like the following:

`<GeneratePlatformNotSupportedAssembly_Message>Operations with CNG keys are only supported on Windows platforms.</GeneratePlatformNotSupportedAssembly_Message>`

cc: @ericstj @weshaggard @danmosemsft

### DIFF
--- a/src/GenAPI/Program.cs
+++ b/src/GenAPI/Program.cs
@@ -151,7 +151,6 @@ namespace GenAPI
                         writer.HighlightBaseMembers = s_hightlightBaseMembers;
                         writer.HighlightInterfaceMembers = s_hightlightInterfaceMembers;
                         writer.PutBraceOnNewLine = true;
-                        writer.ThrowPlatformNotSupportedForCompilation = s_throw;
                         writer.PlatformNotSupportedExceptionMessage = s_exceptionMessage;
                         writer.IncludeGlobalPrefixForCompilation = s_global;
                         return writer;
@@ -238,7 +237,6 @@ namespace GenAPI
         private static bool s_hightlightBaseMembers;
         private static bool s_hightlightInterfaceMembers;
         private static bool s_all;
-        private static bool s_throw;
         private static string s_exceptionMessage;
         private static bool s_global;
 
@@ -269,8 +267,7 @@ namespace GenAPI
                 parser.DefineAliases("hightlightInterfaceMembers", "him");
                 parser.DefineOptionalQualifier("hightlightInterfaceMembers", ref s_hightlightInterfaceMembers, "(-him) [CSDecl] Highlight interface implementation members.");
                 parser.DefineAliases("throw", "t");
-                parser.DefineOptionalQualifier("throw", ref s_throw, "(-t) Method bodies should throw PlatformNotSupportedException.");
-                parser.DefineOptionalQualifier("exceptionMessage", ref s_exceptionMessage, "PlatformNotSupportedException custom message.");
+                parser.DefineOptionalQualifier("throw", ref s_exceptionMessage, "(-t) Method bodies should throw PlatformNotSupportedException with custom message. (Pass Default for PNSE default message)");
                 parser.DefineAliases("global", "g");
                 parser.DefineOptionalQualifier("global", ref s_global, "(-g) Include global prefix for compilation.");
                 parser.DefineQualifier("assembly", ref s_assembly, "Path for an specific assembly or a directory to get all assemblies.");
--- a/src/GenAPI/Program.cs
+++ b/src/GenAPI/Program.cs
@@ -267,7 +267,7 @@ namespace GenAPI
                 parser.DefineAliases("hightlightInterfaceMembers", "him");
                 parser.DefineOptionalQualifier("hightlightInterfaceMembers", ref s_hightlightInterfaceMembers, "(-him) [CSDecl] Highlight interface implementation members.");
                 parser.DefineAliases("throw", "t");
-                parser.DefineOptionalQualifier("throw", ref s_exceptionMessage, "(-t) Method bodies should throw PlatformNotSupportedException with custom message. (Pass Default for PNSE default message)");
+                parser.DefineOptionalQualifier("throw", ref s_exceptionMessage, "(-t) Method bodies should throw PlatformNotSupportedException.");
                 parser.DefineAliases("global", "g");
                 parser.DefineOptionalQualifier("global", ref s_global, "(-g) Include global prefix for compilation.");
                 parser.DefineQualifier("assembly", ref s_assembly, "Path for an specific assembly or a directory to get all assemblies.");
--- a/src/GenAPI/Program.cs
+++ b/src/GenAPI/Program.cs
@@ -152,6 +152,7 @@ namespace GenAPI
                         writer.HighlightInterfaceMembers = s_hightlightInterfaceMembers;
                         writer.PutBraceOnNewLine = true;
                         writer.ThrowPlatformNotSupportedForCompilation = s_throw;
+                        writer.PlatformNotSupportedExceptionMessage = s_exceptionMessage;
                         writer.IncludeGlobalPrefixForCompilation = s_global;
                         return writer;
                     }
@@ -238,6 +239,7 @@ namespace GenAPI
         private static bool s_hightlightInterfaceMembers;
         private static bool s_all;
         private static bool s_throw;
+        private static string s_exceptionMessage;
         private static bool s_global;
 
         private static void ParseCommandLine(string[] args)
@@ -268,6 +270,7 @@ namespace GenAPI
                 parser.DefineOptionalQualifier("hightlightInterfaceMembers", ref s_hightlightInterfaceMembers, "(-him) [CSDecl] Highlight interface implementation members.");
                 parser.DefineAliases("throw", "t");
                 parser.DefineOptionalQualifier("throw", ref s_throw, "(-t) Method bodies should throw PlatformNotSupportedException.");
+                parser.DefineOptionalQualifier("exceptionMessage", ref s_exceptionMessage, "PlatformNotSupportedException custom message.");
                 parser.DefineAliases("global", "g");
                 parser.DefineOptionalQualifier("global", ref s_global, "(-g) Include global prefix for compilation.");
                 parser.DefineQualifier("assembly", ref s_assembly, "Path for an specific assembly or a directory to get all assemblies.");
--- a/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
+++ b/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
@@ -206,7 +206,10 @@ namespace Microsoft.Cci.Writers.CSharp
                 Write("throw new ");
                 if (_forCompilationIncludeGlobalprefix)
                     Write("global::");
-                Write("System.PlatformNotSupportedException(); ");
+                if(_platformNotSupportedExceptionMessage == null)
+                    Write("System.PlatformNotSupportedException();");
+                else
+                    Write($"System.PlatformNotSupportedException(\"{_platformNotSupportedExceptionMessage}\"); ");
             }
             else if (method.ContainingTypeDefinition.IsValueType && method.IsConstructor)
             {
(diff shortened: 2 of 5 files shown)


src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.cs
```diff
@@ -18,6 +18,7 @@ public partial class CSDeclarationWriter : ICciDeclarationWriter
         private bool _forCompilation;
         private bool _forCompilationIncludeGlobalprefix;
         private bool _forCompilationThrowPlatformNotSupported;
```
weshaggard: @ericstj do you think it is worth combining these two? With the presence of the message meaning it is enabled?

danmosemsft:
LGTM. Unless @weshaggard cares there's an underscore.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.cs
danmosemsft: Oh that's a really good idea! Force everyone to add one.

sepidehMS:
@ericstj Could you please take a look?

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.cs
ericstj: Yeah combine them.  Let string.empty indicate no message and null indicate _forCompilationThrowPlatformNotSupported=false.  You don't have to force everyone to add a message, though you could...  Seems like there are some cases where we might want the default message?

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
```diff
@@ -206,7 +206,10 @@ private void WriteMethodBody(IMethodDefinition method)
                 Write("throw new ");
                 if (_forCompilationIncludeGlobalprefix)
                     Write("global::");
-                Write("System.PlatformNotSupportedException(); ");
+                if(_platformNotSupportedExceptionMessage == null)
+                    Write("System.PlatformNotSupportedException();");
+                else
+                    Write($"System.PlatformNotSupportedException(\"{_platformNotSupportedExceptionMessage}\"); ");
```
ericstj: Can you also provide some syntax pattern for folks specify a resource string?

For example, if _platformNotSupportedExceptionMessage starts with some character like `:` or `@` treat the rest of it as code so that folks can do something like `:SR.MyLocalizedResourceMessage`

src/GenAPI/Program.cs
```diff
@@ -266,6 +268,7 @@ private static void ParseCommandLine(string[] args)
                 parser.DefineOptionalQualifier("hightlightInterfaceMembers", ref s_hightlightInterfaceMembers, "(-him) [CSDecl] Highlight interface implementation members.");
                 parser.DefineAliases("throw", "t");
                 parser.DefineOptionalQualifier("throw", ref s_throw, "(-t) Method bodies should throw PlatformNotSupportedException.");
+                parser.DefineOptionalQualifier("exceptionMessage", ref s_exceptionMessage, "PlatformNotSupportedException custom message.");
```
ericstj: Assembly granularity seems OK, but I could imagine we might want a data file that maps doc-id to message in the case a single assembly wanted different messages per type.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
bartonjs: @ericstj Wouldn't we also need to build resource generation into this task?  I assume it currently doesn't bother including SR at all (but I could easily be wrong)

danmosemsft: We are already going to ship with 5% of our exceptoin strings not being in resources. I'm fine with these not being localized and we can have an issue to implement that after we ship.  This is OK for now.

ericstj: > Wouldn't we also need to build resource generation into this task? I assume it currently doesn't bother including SR at all (but I could easily be wrong)

This already works today.  The PNSE infra just generates source that gets passed to compile.  SR generation works the same way.

@danmosemsft it's pretty trivial to implement support for SR now.  Just choose a pattern and use that.  Seems like it'd be better to keep that 5% low rather than add to it, right?

danmosemsft: You're right OK sounds good

ericstj: If you didn't mind hardcoding the "SR." you could just use that as the pattern.  If the string starts with `SR.` then don't quote it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
```diff
@@ -41,6 +41,7 @@
       <GenAPIArgs>$(GenAPIArgs) -libPath:"@(_referencePathDirectories)"</GenAPIArgs>
       <GenAPIArgs>$(GenAPIArgs) -out:"$(NotSupportedSourceFile)"</GenAPIArgs>
       <GenAPIArgs>$(GenAPIArgs) -throw</GenAPIArgs>
+      <GenAPIArgs Condition="'$(GeneratePlatformNotSupportedAssembly_Message)' != ''">$(GenAPIArgs) -exceptionMessage:"$(GeneratePlatformNotSupportedAssembly_Message)"</GenAPIArgs>
```
ericstj: I think you need to update this given your change here: https://github.com/dotnet/buildtools/pull/1451/files#diff-e82cfcbeb709508bae5e486ffa8dcbb6R268

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
weshaggard: I don't really like the idea of hardcoding "SR" here, as this tool is more general. If we want SR we should allow for it to be passed in via the message. Which would mean all we need to do is remove the quoting here and enforce that the message also contain quotes if it needs them.

ericstj: Sure if you can get quotes to stick through the commandline parameter parsing you could do that and just avoid adding quotes in GenAPI

sepidehMS: @ericstj I can't get it working by just passing SR. For instance, I'm trying it for System.Security.Cryptography.OpenSsl and this is the line generated in System.Security.Cryptography.OpenSsl.notsupported.cs:
`public DSAOpenSsl() { throw new System.PlatformNotSupportedException(SR.Cryptography_InvalidDsaParameters_MissingFields); }`

This generates `error CS0103: The
 name 'SR' does not exist in the current context` Where is SR supposed to be defined in this project?

weshaggard: @sepidehMS SR class is generated as part of the build. We need to make sure that isn't disabled for this build configuration.

weshaggard: Looks like we will need to make some adjustments to the notsupport.targets https://github.com/dotnet/buildtools/blob/78260239e4db68ad5c2df774a27d5f9a54fd92e5/src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets#L9 as we omit resources when that is set.

sepidehMS:
cc: @ericstj @weshaggard 

Pushed new commit adding SR support for PNSE custom message and combining `GeneratePlatformNotSupportedAssembly` and `GeneratePlatformNotSupportedAssemblyMessage` properties.

I didn't find a clean way to pass in double quotes to GenAPI. So, I'm checking if the passed-in message starts with "SR." and if it does, I don't include quotes.

Also, since having no string passed to `<GeneratePlatformNotSupportedAssemblyMessage>` and not having this property at all means the same in the csproj and the targets, I'm forcing everyone to pass `Default` to this property if they wish to not customize the PNSE message.

sepidehMS:
@dotnet-bot test this please

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
```diff
@@ -201,12 +201,17 @@ private void WriteMethodBody(IMethodDefinition method)
...
-            if (_forCompilationThrowPlatformNotSupported)
+            if (_platformNotSupportedExceptionMessage != null)
             {
                 Write("throw new ");
                 if (_forCompilationIncludeGlobalprefix)
                     Write("global::");
-                Write("System.PlatformNotSupportedException(); ");
+                if(_platformNotSupportedExceptionMessage.Equals("Default"))
+                    Write("System.PlatformNotSupportedException();");
+                else if(_platformNotSupportedExceptionMessage.StartsWith("SR."))
```
danmosemsft: Another option is to always write it without quotes. I always have to use SR. And why not?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
```diff
@@ -40,7 +39,7 @@
       <GenAPIArgs>-assembly:"%(ResolvedMatchingContract.Identity)"</GenAPIArgs>
       <GenAPIArgs>$(GenAPIArgs) -libPath:"@(_referencePathDirectories)"</GenAPIArgs>
       <GenAPIArgs>$(GenAPIArgs) -out:"$(NotSupportedSourceFile)"</GenAPIArgs>
-      <GenAPIArgs>$(GenAPIArgs) -throw</GenAPIArgs>
+      <GenAPIArgs Condition="'$(GeneratePlatformNotSupportedAssemblyMessage)' != ''">$(GenAPIArgs) -throw:"$(GeneratePlatformNotSupportedAssemblyMessage)"</GenAPIArgs>
```
danmosemsft: I think at this point GeneratePlatformNotSupportedAssemblyMessage is definitely not empty, don't need the condition here, right? It didn't preivously have a condition.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
```diff
@@ -1,15 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <PropertyGroup Condition="'$(GeneratePlatformNotSupportedAssembly)' == 'true'">
```
ericstj: I wouldn't force people to set the string IMHO.  We have plenty of cases where we don't use it.  If I were you I would create a single parameter: -throw.  If a value is specified then use it as the message, if no value is specified then don't pass a message and use the default.  My care level about SR vs quotes is low, but I think you're just creating more work for yourself by changing the behavior here.

ericstj: FYI: behavior of an omitted value is what we'd expect: https://github.com/dotnet/buildtools/blob/3ed0bf65a3312946c959316a779ceb894cebdce7/src/common/CommandLine.cs#L357-L359

So that's more reason not to bother with using "Default"

sepidehMS:
@ericstj @danmosemsft added new commit removing forced Default value. Now, the PlatformNotSupportedAssembly generation can be handled with two different properties in the csproj:

`GeneratePlatformNotSupportedAssembly` (boolean) and `GeneratePlatformNotSupportedAssemblyMessage` (taking some string or SR.message)

If `GeneratePlatformNotSupportedAssemblyMessage` has non-empty value, regardless of `GeneratePlatformNotSupportedAssembly` being set, the PNSE will be generated with custom message. Otherwise, if `'$(GeneratePlatformNotSupportedAssembly)' == 'true'`, default PNSE message will be used.

PR #1453, opened 2017-04-24
Add -global option to GenAPI for including global prefix

WonyoungChoi:
In case of using XXX.System namespace, types in System namespace of BCL can not be referenced at compilation time due to namespace conflicts. 

For example, 
```
namespace MyLib.System {
  public static partial class MyClass {
    public static bool IsChecked { get { throw new Systm.PlatformNotSupportedException(); } }
    ...
  }
}
```
at compilation time, Systm.PlatformNotSupportedException can not be found.

To avoid these conflicts, the -global option is needed to explicitly include the global prefix in every types.

### DIFF
--- a/src/GenAPI/Program.cs
+++ b/src/GenAPI/Program.cs
@@ -53,7 +53,7 @@ namespace GenAPI
                 }
             }
             else
-            { 
+            {
                 using (TextWriter output = GetOutput())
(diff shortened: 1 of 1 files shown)


weshaggard:
Thanks @WonyoungChoi for the feature.

PR #1454, opened 2017-04-24
Rewrite type refs in partial reference facades

ericstj:
When creating a partial reference facade we remove types from the
assembly and replace them with typeforwards.

For the types that are left in the reference facade we need to make sure
they refer to the forwarded types rather than local type defs.

### DIFF
--- a/src/GenFacades/GenFacades.Core/TypeReferenceRewriter.cs
+++ b/src/GenFacades/GenFacades.Core/TypeReferenceRewriter.cs
@@ -1,0 +1,43 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Cci;
+using Microsoft.Cci.MutableCodeModel;
+using System;
+
+namespace GenFacades
+{
+    internal sealed class TypeReferenceRewriter : MetadataRewriter
+    {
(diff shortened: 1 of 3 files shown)


ericstj:
I am doing IL diffs between the libs I previously built through hand-authored CS files fed through CSC vs libs rewritten with this tool to make sure I got everything.  I copied the technique from a tool that @weshaggard wrote in the past.

ericstj:
The diffs look good.  This technique also has the added benefit in that I can remove these workarounds:
https://github.com/dotnet/corefx/commit/45b4c71bff924245d8962035bd751d65c45bf31d#diff-c6b38c104d8ce4d33770fa26bcb7c315R10.  Since we're generating the typeforwards we don't include the internal nested types.

src/GenFacades/GenFacades.Core/TypeReferenceRewriter.cs
```diff
@@ -0,0 +1,43 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Cci;
+using Microsoft.Cci.MutableCodeModel;
+using System;
+
+namespace GenFacades
+{
+    internal sealed class TypeReferenceRewriter : MetadataRewriter
```
weshaggard: Might be interesting to move this to Microsoft.CCI.Extensions at some point.

PR #1461, opened 2017-04-27
Add casting statement on assignment for const Enum type

WonyoungChoi:
In case of using "const enum type field" with initialzation, the constant value
is assigned to the const enum field without casting in the generated code from GenAPI

For example, When using const enum field with initialization as follows:
```
using System;

namespace MyTest
{
  public class MyClass
  {
    public const MyEnum ConstEnum1 = MyEnum.A;
  }

  public enum MyEnum
  {
    A = 1,
    B = 2,
  }
}
```

GenAPI generates following source code:
```
namespace MyTest
{
    public partial class MyClass
    {
        public const MyTest.MyEnum ConstEnum1 = 1;
        public MyClass() { }
    }
    public enum MyEnum
    {
        A = 1,
        B = 2,
    }
}
```

"MyEnum.A" is converted to a constant value "1", but this constant value can not be assigned to "Enum" type variable without casting.

In CSDeclarationWriter::WriteFieldDefinition() of Cci.Extensions, the "casting" should be added if the field type is Enum.

### DIFF
--- a/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Fields.cs
+++ b/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Fields.cs
@@ -52,7 +52,11 @@ namespace Microsoft.Cci.Writers.CSharp
                 {
                     WriteSpace();
                     WriteSymbol("=", true);
-                    WriteMetadataConstant(field.Constant);
+                    if (field.Type.IsEnum) {
+                        WriteFieldDefinitionValue(field);
+                    } else {
+                        WriteMetadataConstant(field.Constant);
+                    }
                 }
 
                 WriteSymbol(";");
@@ -68,6 +72,29 @@ namespace Microsoft.Cci.Writers.CSharp
                 }
                 WriteSymbol(",");
             }
+        }
+
+        private void WriteFieldDefinitionValue(IFieldDefinition field) {
+            var resolvedType = field.Type.ResolvedType;
+
+            if (resolvedType != null) {
+                foreach (var enumField in resolvedType.Fields) {
+                    var enumFieldValue = enumField?.Constant?.Value;
+                    if (enumFieldValue != null && enumFieldValue.Equals(field.Constant.Value))
+                    {
+                        WriteTypeName(field.Type, noSpace: true);
+                        WriteSymbol(".");
+                        WriteIdentifier(enumField.Name);
+                        return;
+                    }
+                }
+            }
+
+            // couldn't find a symbol for enum, just cast it
+            WriteSymbol("(");
+            WriteTypeName(field.Type, noSpace: true);
+            WriteSymbol(")");
+            WriteMetadataConstant(field.Constant);
         }
     }
 }


ericstj:
@WonyoungChoi did you look into if it was possible to use the actual enum?  I think that would produce better looking generated code.

EG: `public const MyTest.MyEnum ConstEnum1 = MyTest.MyEnum.A`

WonyoungChoi:
@ericstj It seems that it's not possible to implement it simply. Because the generated IL doen't have the actual enum name in the field definition. following code is the generated IL code:
```
.namespace MyTest
{
  .class public auto ansi beforefieldinit MyClass
  	extends [System.Runtime]System.Object
  {
    .field public static literal  valuetype MyTest.MyEnum ConstEnum1 = int32(0x00000001)

    // method line 1
    .method public hidebysig specialname rtspecialname 
           instance default void '.ctor' ()  cil managed 
    {
        // Method begins at RVA 0x2050
	// Code size 8 (0x8)
	.maxstack 8
	IL_0000:  ldarg.0 
	IL_0001:  call instance void object::'.ctor'()
	IL_0006:  nop 
	IL_0007:  ret 
    } // end of method MyClass::.ctor

  } // end of class MyTest.MyClass
}

.namespace MyTest
{
  .class public auto ansi sealed MyEnum
  	extends [System.Runtime]System.Enum
  {
    .field  public specialname rtspecialname  int32 value__
    .field public static literal  valuetype MyTest.MyEnum A = int32(0x00000001)
    .field public static literal  valuetype MyTest.MyEnum B = int32(0x00000002)

  } // end of class MyTest.MyEnum
}
```
If want to implement for using the actual enum name, we should find the name in Enum type definition with the compile time constant value of the field definition.

WonyoungChoi:
@ericstj your changes work fine. I've changed this PR with yours. Thank you.

weshaggard:
We have had a long standing bug to convert values to the string Enum name (especially in attribute targets) but it isn't so easy. @ericstj your code works fine for simple cases where you have an exact matching value. However to be complete you need to handle or'ed together flag values as well (or just fallback to casting the value).

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Fields.cs
```diff
@@ -69,5 +73,28 @@ private void WriteFieldDefinition(IFieldDefinition field)
...
+
+        private void WriteFieldDefinitionValue(IFieldDefinition field) {
+            var resolvedType = field.Type.ResolvedType;
+
+            if (resolvedType != null) {
+                foreach (var enumField in resolvedType.Fields) {
+                    var enumFieldValue = enumField?.Constant?.Value;
+                    if (enumFieldValue != null && enumFieldValue.Equals(field.Constant.Value))
+                    {
+                        WriteTypeName(field.Type, true);
```
weshaggard: Can you specify which "true" is as the second parameter? based on the code it looks like it would be for nospace.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Fields.cs
```diff
@@ -69,5 +73,28 @@ private void WriteFieldDefinition(IFieldDefinition field)
...
+                        WriteSymbol(".");
+                        WriteIdentifier(enumField.Name);
+                        return;
+                    }
+                }
+            }
+
+            // couldn't find a symbol for enum, just cast it
+            WriteSymbol("(");
+            WriteTypeName(field.Type, true);
```
weshaggard: same here can you do "nospace: true", so it is clear what the true is for.

ericstj:
> However to be complete you need to handle or'ed together flag values as well (or just fallback to casting the value).

It does fallback to casting the value.

weshaggard:
>It does fallback to casting the value.

I know I was just meaning that we accept that fallback.

WonyoungChoi:
@weshaggard I've fixed. sorry for late fix due to my vacation.

weshaggard:
Thanks @WonyoungChoi for the fix.

PR #1462, opened 2017-04-27
Add xunit filter for method and class

danmosemsft:
`/p:xunitoptions=xxx` blows away other options.

@weshaggard if you don't grab this, I'll merge after you go in. thanks!

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -90,6 +90,9 @@
 
     <XunitOptions Condition="'$(TestTFM)'!=''">$(XunitOptions) -notrait category=non$(TestTFM)tests</XunitOptions>
     <XunitOptions Condition="'$(XunitMaxThreads)'!=''">$(XunitOptions) -maxthreads $(XunitMaxThreads)</XunitOptions>
+    <XUnitOptions Condition="'$(XunitTestClass)' != ''">$(XunitOptions) -class $(XunitTestClass)</XUnitOptions>
+    <XUnitOptions Condition="'$(XunitTestMethod)' != ''">$(XunitOptions) -method $(XunitTestMethod)</XUnitOptions>
+    <XunitOptions>$(XUnitOptions) $(XunitAdditionalOptions)</XunitOptions>
     <XunitTestAssembly Condition="'$(XunitTestAssembly)' == ''">$(TargetFileName)</XunitTestAssembly>
     <XunitArguments>$(XunitTestAssembly) $(XunitOptions)</XunitArguments>
 


stephentoub:
Why is this better than:
```
/p:XunitOptions="-class ..."
```
etc.?  Just wondering why it needs a shortcut that's barely a shortcut.

bartonjs:
> /p:xunitoptions=xxx blows away other optoins.

Does it?

I ran `D:\git\bartonjs\corefx\src\System.Security.Cryptography.X509Certificates\tests>msbuild /t:BuildAndTest /p:XunitOptions="-class System.Security.Cryptography.X509Certificates.Tests.ChainTests"`, and that didn't run any of the `[PlatformSpecific(TestPlatforms.AnyUnix)]` tests in that class.

What options are getting blown away?

danmosemsft:
@bartonjs  if you run this eg `msbuild /t:buildandtest /p:xunitoptions="System.Security.Cryptography.Hashing.Algorithms.Tests.HmacSha384Tests" /p:forceruntests=true  /p:targetgroup=netfx` it will NOT add `-notrait category=nonnetfxtests`. `[PlatformSpecific]` relies on `-notrait category=nonwindowstests` which does not go through XUnitOptions but instead through XunitTraitOptions for some reason.

danmosemsft:
> just wondering why it needs a shortcut that's barely a shortcut.

Seems cleaner, but I don't really care, if you do, I can remove it.

bartonjs:
> -notrait category=nonwindowstests which does not go through XUnitOptions but instead through XunitTraitOptions for some reason.

Dumb question, but: Should we instead be changing the the TFM and perf options off of XunitOptions and onto XunitTraitOptions, so they don't interfere with XunitOptions?  I wouldn't expect wanting to add `-showprogress` to be having those weird behaviours.

safern:
AFAIK, the reason why running `/p:XunitOptions="xxx"` doesn't work is because in msbuild when a property value is assigned as a command line argument it overrides the value and basically ignores all of the new assignments that you do for that variable inside props, targets or proj files. It is basically as a hard set of the variable during that msbuild session. So XunitOptions' value will be the one that was passed from the command line.

I would instead of having an specific property for TestClass and TestMethod I would add a general property named something like `XunitAdditionalOptions` that is appended to `XunitOptions` if it is not empty right before: https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L62

So this way if I want to run a class without parallelization and with showprogress option I don't have to write msbuild /p:XunitTestClass="MyClass" /p:XunitAdditionalOptions="-noparallel -showprogress" instead everything in just one property. 

@danmosemsft do you want me to take over this one as I'm touching buildtools lately?

danmosemsft:
@safern yes, that's why  I made this change, so you could do /p:XunitTestMethod so never having to pass /p:XunitOptions. 

@bartonjs is right though that it would be better to not modify XUnitOptions inside the targets/props at all (since that may be ignored) and instead append $(XunitTestMethod) and everything else to $(XUnitOptions) and pass the lot to xunit. That way you can pass /p:XUnitOptions if you want and/or you can pass /p:XunitTestMethod if you want and eveyrone is happy.

Sure feel free to take it

danmosemsft:
ThankS!

PR #1468, opened 2017-05-02
Enable AssemblyResolver for Full Framework GenFacadesTasks version

safern:
Fixes: https://github.com/dotnet/buildtools/issues/1443

cc: @weshaggard @mellinoe 

FYI: @jcouv


jcouv:
Thanks

src/Microsoft.DotNet.Build.Tasks/GenFacadesTask.cs
```diff
@@ -11,7 +15,7 @@ namespace Microsoft.DotNet.Build.Tasks
     /// <summary>
     /// Runs GenFacades In-Proc.
     /// </summary>
-    public sealed class GenFacadesTask : Task
+    public sealed partial class GenFacadesTask : Task
```
weshaggard: I think you should just inherit from BuildTask instead.

weshaggard: cc @ericstj

safern: That would work and would be cleaner.

safern: Fixed it. Made GenFacadesTask inherit from BuildTask instead.

PR #1469, opened 2017-05-03
Introduce a way to generate the DefaultDllImportSearchPathAttribute w…

AlexGhiondea:
…hen generating the AssemblyInfo

/cc @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -62,6 +62,9 @@
       <AssemblyInfoLines Condition="'$(AssemblyComVisible)'!=''" Include="[assembly:System.Runtime.InteropServices.ComVisible($(AssemblyComVisible))]" />
       <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'"
         Include="[assembly:System.Reflection.AssemblyMetadata(&quot;%(AssemblyMetadata.Identity)&quot;, &quot;%(AssemblyMetadata.Value)&quot;)]" />
+
+      <AssemblyInfoLines Condition="'$(IncludeDllSafeSearchPathAttribute)'=='true'" 
+        Include="[assembly: System.Runtime.InteropServices.DefaultDllImportSearchPathsAttribute(System.Runtime.InteropServices.DllImportSearchPath.AssemblyDirectory | System.Runtime.InteropServices.DllImportSearchPath.System32)]" />
     </ItemGroup>
 
     <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.vbproj'">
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -62,6 +62,9 @@
       <AssemblyInfoLines Condition="'$(AssemblyComVisible)'!=''" Include="[assembly:System.Runtime.InteropServices.ComVisible($(AssemblyComVisible))]" />
       <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'"
         Include="[assembly:System.Reflection.AssemblyMetadata(&quot;%(AssemblyMetadata.Identity)&quot;, &quot;%(AssemblyMetadata.Value)&quot;)]" />
+
+      <AssemblyInfoLines Condition="'$(IncludeDllSafeSearchPathAttribute)'=='true'"
```
weshaggard: Did you want to automatically set this to true for IsDotNetFrameworkAssembly's?

AlexGhiondea: Yes -- that should happen in the CoreFx/CoreCLR repos, right?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -62,6 +62,9 @@
       <AssemblyInfoLines Condition="'$(AssemblyComVisible)'!=''" Include="[assembly:System.Runtime.InteropServices.ComVisible($(AssemblyComVisible))]" />
       <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'"
         Include="[assembly:System.Reflection.AssemblyMetadata(&quot;%(AssemblyMetadata.Identity)&quot;, &quot;%(AssemblyMetadata.Value)&quot;)]" />
+
+      <AssemblyInfoLines Condition="'$(IncludeDllSafeSearchPathAttribute)'=='true'" 
+        Include="[assembly: System.Runtime.InteropServices.DefaultDllImportSearchPathsAttribute(System.Runtime.InteropServices.DllImportSearchPath.AssemblyDirectory | System.Runtime.InteropServices.DllImportSearchPath.System32)]" />
```
joperezr: I assume you tested that the `|` character works just as expected? I ask this because for special characters we had to use the Asccii notation instead, so just making sure.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
weshaggard: You could do that in BuildTools. Somewhere over https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets#L33.

AlexGhiondea: I am not sure if you are saying I can add another assemblymetadata there, or just add a condition on the IsDotNetFrameworkProductAssembly property to set the value. I suspect the second part :)

weshaggard: The later. Just set IncludeDllSafeSearchPathAttribute=true if IsDotNetFrameworkProductAssembly=true.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
AlexGhiondea: `[` seems to be fine -- we use it in the lines above already. I did escape the VB notation which uses `<` and `>`

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
AlexGhiondea: done.

PR #1472, opened 2017-05-04
Add VerifyTypes task to catch duplicate types

ericstj:
Adds a task which accepts a set of files and flags duplicate public
types in that set.

/cc @weshaggard


ericstj:
@dotnet-bot test Innerloop Ubuntu14.04 Debug

ericstj:
That's a bit annoying.  Build succeeds on Windows (because targeting pack is present) but fails on unix because I was missing a "TargetingPackReference" element.  /cc @weshaggard Would be nice if our targeting pack infra behaved the same way as RAR.  Maybe we could do that moving forward: rather than actually setting up references we could just set the properties that RAR uses to locate the framework references.

PR #1473, opened 2017-05-04
Copy S.P.Corelib to S.P.Corelib.ni.dll to ensure the UAP F5 tests con…

AlexGhiondea:
…tinue to work

/cc @gkhanna79 @weshaggard @joperezr


joperezr:
didn't you say that all that was required was for an empty file to be there? if so I think it would be better to have an echo command

joperezr:
Also one more thing, can we log an issue to track the work of reverting this once we fix the runner properly?

AlexGhiondea:
Simply echoing the file is not enough. It is better to just copy this over. We can have an issue tracking the removal of this -- not sure the right repo for it... probably still here.

weshaggard:
Do we have an issue tracking the host update as this will break in shipping scenarios as well now that we have removed the ni.

gkhanna79:
Which shipping scenario will it break? 2.1 based uwp host will have the right change (file issue against me for it).


Thanks,
Gaurav
________________________________

AlexGhiondea:
@gkhanna79 which repo should I file the issue in?

gkhanna79:
@AlexGhiondea Please file it in Core-Setup for 2.1 milestone.

gkhanna79:
@weshaggard Can you please comment on what shipping scenarios would break?

weshaggard:
>@weshaggard Can you please comment on what shipping scenarios would break?

Right now just the F5 debug scenarios in UAP but we aren't supporting that in 2.0 so I think it is fine to support it in 2.1.

gkhanna79:
Yes, those would be supported for 2.1 and https://github.com/dotnet/core-setup/pull/2115 would account for the host update for the same.

PR #1480, opened 2017-05-09
Morediskusagetelemetry

smile21prc:
Enable disk usage reporting for TEMP and Nuget Cache。


smile21prc:
@chcosta , can you take a look? In this PR I:
1. Only enabled disk usage reporting and didn't include clean up code.
2. Renamed some APIs to be more general as they'll be reused by my clean up code as well (clean up TEMP/Nuget Dirs in additional to agent working dirs).

PR #1481, opened 2017-05-09
Make license url property overridable 

tarekgh:


tarekgh:
@weshaggard @ericstj could you please have a look? this is the first step to get the build tools allow override the properties. then I'll make the different repos override it.

cc @richlander @terrajobst

tarekgh:
https://github.com/dotnet/corefx/issues/12190

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -1162,6 +1162,8 @@
     <PropertyGroup  Condition="'$(AllowPartialPackages)' == 'true' AND '@(_missingPackageFile)' != ''">
       <Description>WARNING: This package is missing files @(_missingPackageFile->'%(TargetPath)/%(Filename)%(Extension)') %0A$(Description)</Description>
     </PropertyGroup>
+
+    <Error Condition="'$(LicenseUrl)' == ''" Text="LicenseUrl property is not defined." />
```
tarekgh: This Error element will be useful when we remove LicenseUrl from this targets file later.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -54,11 +54,11 @@
...
     <Owners>microsoft,dotnetframework</Owners>
     <Description>TODO</Description>
-    <LicenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</LicenseUrl>
+    <LicenseUrl Condition="'$(LicenseUrl)' == ''">http://go.microsoft.com/fwlink/?LinkId=329770</LicenseUrl>
     <IconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</IconUrl>
-    <Copyright>&#169; Microsoft Corporation.  All rights reserved.</Copyright>
+    <Copyright Condition="'$(Copyright)' == ''">.NET Foundation and Contributors</Copyright>
     <Tags></Tags>
-    <RequireLicenseAcceptance>true</RequireLicenseAcceptance>
+    <RequireLicenseAcceptance>false</RequireLicenseAcceptance>
```
ericstj: Why was this changed?  You should probably condition the property just in case folks don't want the change.

tarekgh:
@ericstj I have conditioned RequireLicenseAcceptance property too. let me know if you have any other comments. Thanks.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
weshaggard: If we don't plan to remove the default LicenseUrl then I would just remove this error as it is just dead code.

tarekgh: The plan so far is to remove the property from the build tools. but I'll go ahead and remove this Error now and then we can add it later when we remove the property

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -54,11 +54,11 @@
     <Authors>Microsoft</Authors>
     <Owners>microsoft,dotnetframework</Owners>
     <Description>TODO</Description>
-    <LicenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</LicenseUrl>
+    <LicenseUrl Condition="'$(LicenseUrl)' == ''">http://go.microsoft.com/fwlink/?LinkId=329770</LicenseUrl>
     <IconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</IconUrl>
-    <Copyright>&#169; Microsoft Corporation.  All rights reserved.</Copyright>
+    <Copyright Condition="'$(Copyright)' == ''">.NET Foundation and Contributors</Copyright>
```
weshaggard: @tarekgh sorry to ping this old PR but do you recall why you change the text in the Copyright? It appears that we never end up using this text as we also override this property in https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props#L7, so I want to make sure we aren't messing something up.

PR #1482, opened 2017-05-09
Add MSBuild task that will get the hash out from an assembly

joperezr:
PTAL: @weshaggard 

This adds a task that we will use in core-setup in order to get the commit hash out from an assembly.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GetVersionFromAssembly.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GetVersionFromAssembly.cs
@@ -1,0 +1,48 @@
+﻿using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Newtonsoft.Json;
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetVersionFromAssembly : Task
+    {
+        [Required]
+        public string PathToAssembly { get; set; }
+
+        [Output]
+        public string AssemblyVersion { get; set; }
+
+        public override bool Execute()
+        {
+            if (!File.Exists(PathToAssembly))
+            {
+                Log.LogError($"{PathToAssembly} does not exist.");
+                return false;
+            }
+            try
+            {
+                Regex regex = new Regex(@"Commit Hash: ([A-Za-z0-9].*)");
+                string productVersion = FileVersionInfo.GetVersionInfo(PathToAssembly).ProductVersion;
+                Match match = regex.Match(productVersion);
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/GetVersionFromAssembly.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetVersionFromAssembly : Task
+    {
+        [Required]
+        public string PathToAssembly { get; set; }
+
+        [Output]
+        public string AssemblyVersion { get; set; }
```
weshaggard: This shouldn't be AssemblyVersion it should be AssemblySHA or something.

src/Microsoft.DotNet.Build.Tasks/GetVersionFromAssembly.cs
```diff
@@ -0,0 +1,48 @@
...
+                if (match.Success && match.Groups.Count > 1)
+                {
+                    AssemblyVersion = match.Groups[1].Value;
+                }
+                else
+                {
+                    Log.LogError($"{PathToAssembly} has an invalid ProductVersion or does not contain a commit hash.");
+                    return false;
+                }
+            } catch (Exception)
```
weshaggard: new line needed after brace.

src/Microsoft.DotNet.Build.Tasks/GetVersionFromAssembly.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                Log.LogError($"{PathToAssembly} does not exist.");
+                return false;
+            }
+            try
+            {
+                Regex regex = new Regex(@"Commit Hash: ([A-Za-z0-9].*)");
+                string productVersion = FileVersionInfo.GetVersionInfo(PathToAssembly).ProductVersion;
+                Match match = regex.Match(productVersion);
+                if (match.Success && match.Groups.Count > 1)
```
weshaggard: @ericstj should we go ahead and have this task expose the other version information from this assembly?

src/Microsoft.DotNet.Build.Tasks/GetVersionFromAssembly.cs
joperezr: good point, will fix.

src/Microsoft.DotNet.Build.Tasks/GetVersionFromAssembly.cs
joperezr: Will add.

joperezr:
Closing this since it doesn't make sense anymore since we will get this info on a different way.

PR #1483, opened 2017-05-09
Add System namespace to PNSE Message in GenAPI

sepidehMS:
This should fix errors in https://github.com/dotnet/corefx/pull/19506. When using resource strings as the message for auto-generated PNSE, few projects like the following hit `The name 'SR' does not exist in the current context` errors, as the `System` namespace is missing:

- Microsoft.Win32.Registry
- Microsoft.Win32.Registry.AccessControl
- System.Security.Cryptography.Cng (In some configurations contains APIs from Microsoft.Win32.SafeHandles)

cc: @ericstj @danmosemsft


sepidehMS:
cc: @weshaggard

PR #1484, opened 2017-05-10
Pass ProjectReferencesCreatingPackages to resolve task

mellinoe:
If a csproj has P2P references, this property needs to be passed for the PrereleaseResolveNugetPackageAssets to work. It just happens that none of our projects in corefx/standard have P2P references, so we've never hit it. core-setup has a project where we need this.

@weshaggard


weshaggard:
cc @ericstj

ericstj:
Do we really use package/project replacement in one of our repos?  I'd really prefer we just do that explicitly with a ProjectReference.

mellinoe:
> Do we really use package/project replacement in one of our repos? I'd really prefer we just do that explicitly with a ProjectReference.

We're just using a ProjectReference in core-setup; this is still needed.

PR #1490, opened 2017-05-11
Automatically detect shared fx version in crossgen.sh

mellinoe:
Instead of hard-coding the shared framework version in crossgen.sh, we will now automatically detect it based on the folders present under shared/Microsoft.NETCore.App/. The CoreCLR version is still hard-coded, but I don't see an easy way to fix that with just the information present in the shared framework.


mellinoe:
I'm going to merge this now to unblock myself in core-setup.

ellismg:
LGTM

PR #1491, opened 2017-05-11
Passing path when saving build definitions in VSTS.

salvmd:
@markwilkie @chcosta @dagood 

Creating this to discuss the fix for dotnet/core-eng#163.

Some further questions:

- The method RemoveIdentifiableInformation, which removes instance identifiable fields, is called on the local checked in definition, but the examples I've seen of local definitions never seem to have these fields; only VSTS seems to know about them. Is this just a validation mechanism to ensure local definitions are not written with these fields by accident?
- It looks like the path was not being sent to VSTS because the logic considered it an instance identifiable field. If we remove it from that list of identifiable fields, the path can be sent and retrieved from VSTS normally; are there any other consequences I'm missing for removing this field from that list?
- These changes fix the issue when using the default config for VstsDefinitionClientConfig; however, if using a custom config that adds the path again to the list of identifiable fields, this would break again in the same way. Is this (using a different config) something common that we should defend against?
- Regarding the extra logic in IsMatching to remove the trailing separators, added this because the Build API does this on their end, so saving /ABC/ the first time would break on the second time since we would be comparing local: /ABC/ against retrieved: /ABC . If there is another tool that validates definitions before checking them in, would it be better to move this check there to ensure no checked-in definition ever have trailing separators?

Added some comments in VstsDefinitionHttpClient.cs because I was a little confused about why both of those methods were called before updating (the reason being only RetrieveDefinitionByIdAsync gets the full definition data) :)

Tested these cases:

1. Saving a definition with the root path. Running again updates the same definition.
2. Saving the same definition but with \SomePath\. It creates a new definition under the specified path. Running again updates the same definition.
3. Saving the same definition but with \SometPath. It updates the same definition as the previous test.

### DIFF
--- a/src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
+++ b/src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
@@ -5,6 +5,7 @@
 using Microsoft.DotNet.Build.VstsBuildsApi.Configuration;
 using Newtonsoft.Json.Linq;
 using System;
+using System.IO;
 
 namespace Microsoft.DotNet.Build.VstsBuildsApi
 {
@@ -19,8 +20,17 @@ namespace Microsoft.DotNet.Build.VstsBuildsApi
 
         protected override bool IsMatching(JObject localDefinition, JObject retrievedDefinition)
         {
+            string localDefinitionPath = localDefinition["path"].ToString();
+
+            // If it's not the root directory.
+            if (Path.GetDirectoryName(localDefinitionPath) != null) {
+                // Remove trailing path separators since VSTS does the same when creating definitions.
+                localDefinitionPath = localDefinitionPath.TrimEnd(
+                    Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            }
+
             return localDefinition["quality"].ToString() == "definition" &&
-                localDefinition["path"].ToString() == retrievedDefinition["path"].ToString();
+                localDefinitionPath == retrievedDefinition["path"].ToString();
         }
 
         protected override string GetDefinitionProject(JObject definition)
--- a/src/Microsoft.DotNet.Build.VstsBuildsApi/Configuration/VstsDefinitionClientConfig.cs
+++ b/src/Microsoft.DotNet.Build.VstsBuildsApi/Configuration/VstsDefinitionClientConfig.cs
@@ -2,6 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using System.Linq;
+
 namespace Microsoft.DotNet.Build.VstsBuildsApi.Configuration
 {
     /// <summary>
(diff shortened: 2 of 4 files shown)


dagood:
Hi @salvmd, I'll see if I can help:

> The method RemoveIdentifiableInformation, which removes instance identifiable fields, is called on the local checked in definition [...] Is this just a validation mechanism to ensure local definitions are not written with these fields by accident?

I don't know the original intent, but I agree that it does this and it seems worthwhile to keep.

> It looks like the path was not being sent to VSTS because the logic considered it an instance identifiable field. If we remove it from that list of identifiable fields, the path can be sent and retrieved from VSTS normally; are there any other consequences I'm missing for removing this field from that list?

There's another positive consequence (don't know if you missed it, but thought I'd point it out): removing it from that list also means that during the update flow, `path` won't be flowed from the old definition to the new definition in [VstsDefinitionClient.cs#L102](https://github.com/salvmd/buildtools/blob/7fad8591e2b49274bd013794c851ee4ca42443ce/src/Microsoft.DotNet.Build.VstsBuildsApi/VstsDefinitionClient.cs#L102). This means that after your change, it always uses the value from the JSON, which is good.

> Is this (using a different config) something common that we should defend against?

Right now, this library is used in only one place, and it uses the default config. In principle, I think it makes sense to ensure the behavior will stay the same if someone theoretically doesn't change their custom config.

> If there is another tool that validates definitions before checking them in, would it be better to move this check there to ensure no checked-in definition ever have trailing separators?

There isn't, but we have considered making a way for devs to validate their local changes like this. It would be nice to catch "style" issues like this one during a validation step, even when they are correctable by this library.

Hope that helps clarify some things. Thanks for the comment changes. :smile:

salvmd:
Thanks @dagood!

> In principle, I think it makes sense to ensure the behavior will stay the same if someone theoretically doesn't change their custom config.

I agree. I'll add some validation for this when creating the config.

dagood:
I think I said that a bit ambiguously... what I meant was that if someone doesn't change their custom config and upgrades `Microsoft.DotNet.Build.VstsBuildsApi`, the behavior probably shouldn't change due to the upgrade. I don't *think* adding `path` support needs to be a breaking change, but rather a new feature that users can opt into.

salvmd:
Ah, I see your point. Sounds good to me; I'll remove the validation.
Do you have any other comments/suggestions? :)

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
```diff
@@ -19,8 +20,17 @@ public VstsBuildHttpClient(JObject definition, VstsApiEndpointConfig config)
 
         protected override bool IsMatching(JObject localDefinition, JObject retrievedDefinition)
         {
+            string localDefinitionPath = localDefinition["path"].ToString();
+
+            // If it's not the root directory.
+            if (Path.GetDirectoryName(localDefinitionPath) != null) {
```
dagood: I can't find VSTS docs for the `path` parameter, but I'm not comfortable using `System.IO.Path` to do this since it's a server path rather than on a file system. In my experience it's safe to assume the separator will always be `\`, and I don't know what the behavior is with `/`. Really wish the API docs were complete.

\<nit>Also, the brace should be on the next line.\</nit>

src/Microsoft.DotNet.Build.VstsBuildsApi/VstsBuildHttpClient.cs
```diff
@@ -19,8 +20,17 @@ public VstsBuildHttpClient(JObject definition, VstsApiEndpointConfig config)
 
         protected override bool IsMatching(JObject localDefinition, JObject retrievedDefinition)
         {
+            string localDefinitionPath = localDefinition["path"].ToString();
```
dagood: I prefer `.Value<string>()` rather than `.ToString()` because it's a more exact match, but I see `ToString` is used in this file already so I'm ok with it.

PR #1497, opened 2017-05-14
Adding task that will read netcoreapp versions file to get out repo's hashes

joperezr:
PTAL: @weshaggard 

Adding a simple task that will read the netcoreapp versions file to parse out repo's hashes.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GetNetCoreAppVersionsFromFile.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GetNetCoreAppVersionsFromFile.cs
@@ -1,0 +1,58 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetNetCoreAppVersionsFromFile : Task
+    {
+        [Required]
+        public string PathToVersionsFile { get; set; }
+
+        [Output]
+        public string CorefxHash { get; set; }
+
+        [Output]
+        public string DotnetVersionsCorefxHash { get; set; }
+
+        [Output]
+        public string CoreclrHash { get; set; }
+
+        [Output]
+        public string DotnetVersionsCoreclrHash { get; set; }
+
+        [Output]
+        public string CoresetupHash { get; set; }
+
+        [Output]
+        public bool HasCoreclrVersion { get; set; }
+
+        public override bool Execute()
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/GetNetCoreAppVersionsFromFile.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetNetCoreAppVersionsFromFile : Task
+    {
+        [Required]
+        public string PathToVersionsFile { get; set; }
+
+        [Output]
+        public string CorefxHash { get; set; }
```
weshaggard: I think I would make this a little more generic so that we can easily extend it to things like standard version or others as well. Instead of returning hard-coded properties, return an ITaskItem[] and add set of items which have identity of the key and metadata of the sha.

joperezr: Apart from the standard hash, do we expect anything else going into that versions.txt file? If not, I wonder if it would be better to just add one more hardcoded one for standard. I'm happy to do it generic as well, I just want to understand a bit more what do we expect to use this task for.

weshaggard: I expect we will want things like WCF as well when we do this for UAP package as an example and just making it more generic feels like the easier thing to do. 

Do you have a commit for how you are consuming this? Perhaps that will help inform me if it will be more difficult to consume in the generic fashion.

joperezr: I see, it shouldn't be more difficult to consume, so I'll change it to be generic. I don't have a commit yet with how to consume it, as I'm working on a change that will calculate the path to this versions file as since it is an SDK project that restores it, the package won't be restored to the `PackagesDir`. Once this is merged, I'll send out the corefx PR.

joperezr:
@weshaggard, I've made the requested changes, let me know if that is more or less what you expected.

joperezr:
Merging now. Will address any subsequent feedback later.

src/Microsoft.DotNet.Build.Tasks/GetNetCoreAppVersionsFromFile.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class GetNetCoreAppVersionsFromFile : Task
+    {
+        [Required]
+        public string PathToVersionsFile { get; set; }
+
+        [Required]
```
joperezr: I'll remove the parameter since you are right, it doesn't really make sense and I can just return all.

PR #1498, opened 2017-05-14
Enable to pass a TargetFrameworkMoniker to ActiveIssueAttribute

safern:
With the desktop test effort we started disabling tests with an active issue on github by doing: 
```cs
[SkipOnTargetFramework(TagetFrameworkMonikers.NetFramework, "dotnet/corefx #issue")]
```
When that is not the intended usage of this attribute. We want to use this attribute when the test is specific to another framework or an intended difference on behavior. 

With this PR we allow to use `[ActiveIssue(1234, TargetFrameworkMonikers.NetFramework)]` in example to disable a test that has an issue open tracking the failure only on desktop, or if we find a failure only on uap or netcoreapp disable the test through ActiveIssueAttribute only for that target framework.

cc: @weshaggard @danmosemsft 

FYI: @stephentoub

### DIFF
--- a/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
@@ -18,7 +18,7 @@ namespace Xunit
         public ActiveIssueAttribute(string issue, TestPlatforms platforms) { }
         public ActiveIssueAttribute(int issueNumber, TargetFrameworkMonikers framework) { }
         public ActiveIssueAttribute(string issue, TargetFrameworkMonikers framework) { }
-        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)~0) { }
-        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)~0) { }
+        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)0) { }
+        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)0) { }
     }
 }
--- a/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
@@ -18,7 +18,7 @@ namespace Xunit
         public ActiveIssueAttribute(string issue, TestPlatforms platforms) { }
         public ActiveIssueAttribute(int issueNumber, TargetFrameworkMonikers framework) { }
         public ActiveIssueAttribute(string issue, TargetFrameworkMonikers framework) { }
-        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = TargetFrameworkMonikers.None) { }
-        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = TargetFrameworkMonikers.None) { }
+        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = TargetFrameworkMonikers.All) { }
+        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = TargetFrameworkMonikers.All) { }
     }
 }
--- a/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
@@ -18,7 +18,7 @@ namespace Xunit
         public ActiveIssueAttribute(string issue, TestPlatforms platforms) { }
         public ActiveIssueAttribute(int issueNumber, TargetFrameworkMonikers framework) { }
         public ActiveIssueAttribute(string issue, TargetFrameworkMonikers framework) { }
-        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = TargetFrameworkMonikers.All) { }
-        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = TargetFrameworkMonikers.All) { }
+        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)~0) { }
+        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)~0) { }
     }
 }
--- a/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
@@ -16,9 +16,7 @@ namespace Xunit
     {
         public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any) { }
         public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any) { }
-
         public ActiveIssueAttribute(int issueNumber, TargetFrameworkMonikers framework) { }
-
         public ActiveIssueAttribute(string issueNumber, TargetFrameworkMonikers framework) { }
     }
 }
--- a/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
@@ -14,9 +14,11 @@ namespace Xunit
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
     public class ActiveIssueAttribute : Attribute, ITraitAttribute
     {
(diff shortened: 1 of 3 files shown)


weshaggard:
For my information is the goal that SkipOnFramework is a permanent disable for the platform and Active issue is that it is just a bug we plan to fix?

safern:
> For my information is the goal that SkipOnFramework is a permanent disable for the platform and Active issue is that it is just a bug we plan to fix?

Correct. So when we where trying to get a clean Desktop run we disabled a lot of tests with SkipOnFramework pointing to an issue that is still open, because the test failure needs investigation. We did this in order to enable CI legs so that we don't go backwards in the progress we made the past month. 

Here is an example: https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestsWrite.cs#L412

This test was marked as skip when #18806 is still open and needs further investigation. So this will allow us to add the temporary attribute ActiveIssue and the test will only be disabled for a platform in specific. Instead of using an attribute that is intended to be "permanent"

src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
```diff
@@ -16,5 +16,7 @@ public class ActiveIssueAttribute : Attribute, ITraitAttribute
     {
         public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any) { }
         public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any) { }
+        public ActiveIssueAttribute(int issueNumber, TargetFrameworkMonikers framework) { }
```
weshaggard: Why not all TestPlatforms and TargetFrameworkMonikers to both be passed? I would just add TargetFrameworkMonikers as another optional parameter and default it to 0 (which is an no value).

src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
```diff
@@ -28,17 +28,46 @@ public class ActiveIssueDiscoverer : ITraitDiscoverer
             Debug.Assert(ctorArgs.Count() >= 2);
 
             string issue = ctorArgs.First().ToString();
-            TestPlatforms platforms = (TestPlatforms)ctorArgs.Last();
+            TestPlatforms platforms = TestPlatforms.Any;
+            TargetFrameworkMonikers frameworks = TargetFrameworkMonikers.All;
+            
+            if (ctorArgs.Count() > 2)
+            {
+                platforms = (TestPlatforms)ctorArgs.ElementAt(1);
```
weshaggard: Why make this assumption? Why not just loop over the arguments and they try casting each argument to both types and do the work based on which type it is?

src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
```diff
@@ -24,6 +24,7 @@ public enum TargetFrameworkMonikers
         Netcoreapp = 0x1000,
         Uap = 0x2000,
         UapAot = 0x4000,
-        NetcoreCoreRT = 0x8000
+        NetcoreCoreRT = 0x8000,
+        All = ~0
```
weshaggard: While All make some sense here it doesn't make a lot of sense in the SkipOnFramework context. I wonder if we really need to expose anything more here. You can just blindly use 0.

safern: Ok. Makes sense

src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
safern: Fixed

src/xunit.netcore.extensions/TargetFrameworkMonikers.cs
safern: Fixed

src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
```diff
@@ -28,17 +28,40 @@ public class ActiveIssueDiscoverer : ITraitDiscoverer
...
+            
+            foreach (object arg in ctorArgs.Skip(1)) // First argument is the issue number.
+            {
+                if (arg is TestPlatforms)
+                {
+                    platforms = (TestPlatforms)arg;
+                }
+                else
+                {
+                    frameworks = (TargetFrameworkMonikers)arg;
```
weshaggard: Just to be safe I would also check the type of the arg in this case as well and only cast if the type matches.

src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
```diff
@@ -28,17 +28,40 @@ public class ActiveIssueDiscoverer : ITraitDiscoverer
...
+                }
+            }
+        
             if ((platforms.HasFlag(TestPlatforms.FreeBSD) && RuntimeInformation.IsOSPlatform(OSPlatform.Create("FREEBSD"))) ||
                 (platforms.HasFlag(TestPlatforms.Linux) && RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) ||
                 (platforms.HasFlag(TestPlatforms.NetBSD) && RuntimeInformation.IsOSPlatform(OSPlatform.Create("NETBSD"))) ||
                 (platforms.HasFlag(TestPlatforms.OSX) && RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) ||
                 (platforms.HasFlag(TestPlatforms.Windows) && RuntimeInformation.IsOSPlatform(OSPlatform.Windows)))
             {
-                yield return new KeyValuePair<string, string>(XunitConstants.Category, XunitConstants.Failing);
```
weshaggard: Looks like you removed the failing category.

weshaggard: I would expect that to be returned if frameworks is 0 as an example.

safern: I'm using ~0 as a default so that it has all of the frameworks' flag, so when doing if(framework.HasFlag) for all of the TargetFrameworkMonikers is going to be true and will return to skip the tests for that platform. Either way works fine, I've tested it with both. Which one you prefer? With 0 as might be more readable?

src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
safern: Sounds good

src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
weshaggard: Sorry missed the `~`. However it looks like you are still never returning category failing anywhere any longer. Isn't that going to cause filtered to break.

safern: No, as it will return category nonnetcoreapptests, category nonnetfxtests, etc. For all the TargetFrameworkMonikers that are included in the if statements when a TargetFrameworkMoniker is not passed in. If we want to include category failing would be when framework == 0 and would change framework to 0 so no other flag is included. 

If we always return category failing it will always when the attribute is present and the platform matches, will skip it for all the frameworks, that is why I rather did it through the framework flags. 

I think using the 0 approach would be more readable for people in the future that get to this file to update whatever. So I will update it to follow that and return category failing when == 0 only.

weshaggard: sounds good. I just want to make sure we have the exact same behavior if no one passed or cared about the TargetFrameworkMoniker.

safern: Yes. I tested it locally for all the cases and we have the same behavior as now on either approach :)

weshaggard: I guess I'm missing something but I don't know how this case https://github.com/dotnet/buildtools/blob/57b45254428999e5770743369c21778386bb4321/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L363 would work if we never return "failing" as a category.

safern: Because we would be returning category non(targetgroup)tests for all of the frameworks and we pass that notrait by default when running the tests. 

https://github.com/dotnet/buildtools/blob/57b45254428999e5770743369c21778386bb4321/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L59

weshaggard: I guess it is hard to determine with all the various flags. However do keep in mind that it isn't only corefx that uses these and so other repo's might be passing a different set of flags with the simplest being "nonfailing", so I would like to make sure they continue to work as the do.

safern: So lets say TargetGroup  = netfx and a test has [ActiveIssue(1232)] the default values would be passed, `TestPlatforms.Any` and `(TargetFrameworkMonikers)~0`. Then the first if will be true as platform.HasFlag(TestPlatforms.Windows) will be true, so then since framework = ~0 all the flags are present for it. So it will return:
```
category nonnetcoreapptests
category nonnetfxtests
category nonuaptests
category nonuapaottests
category noncorerttests
```

And in tests.targets since `_bc_TargetGroup = netfx` it will call xunit with argument `-notrait category=nonnetfxtests` so it will skip the test as the discoverer returned a `KeyValuePair`with that category. 

Still moving to 0 as it is more reliable and better to follow the category failing when non targetframework and testplatform are passed.

safern: > I guess it is hard to determine with all the various flags. However do keep in mind that it isn't only corefx that uses these and so other repo's might be passing a different set of flags with the simplest being "nonfailing", so I would like to make sure they continue to work as the do.

Commented at the same time :) 

Didn't know this was used in other repos. I will move to the other approach.

safern: Fixed.

safern:
Thanks for the multiple iteration reviews @weshaggard

PR #1501, opened 2017-05-15
Fixenvvar

smile21prc:
Use HOME path from env var instead of hard coding it.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/CleanupVSTSAgent.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/CleanupVSTSAgent.cs
@@ -414,6 +414,8 @@ namespace Microsoft.DotNet.Build.Tasks
                     return Environment.GetEnvironmentVariable("TMPDIR");
                 else if (DirExists(Environment.GetEnvironmentVariable("TMP")))
                     return Environment.GetEnvironmentVariable("TMP");
+                else if (DirExists(Path.Combine(Environment.GetEnvironmentVariable("HOME"), "myagent/_work/_temp")))
+                    return Path.Combine(Environment.GetEnvironmentVariable("HOME"), "myagent/_work/_temp");
                 else
                 {
                     Log.LogMessage("No TEMP dir found.");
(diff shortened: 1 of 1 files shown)


smile21prc:
@MattGal @chcosta , could you please take a look at this PR? Thanks.

smile21prc:
@MattGal, Thank Matt for the review. This whole clean-up step isn't added for OSX, so not need to test this change for OSX now:
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=6386

And I agree with you, the whole script needs to be tested for OSX when it's enabled.

src/Microsoft.DotNet.Build.Tasks/CleanupVSTSAgent.cs
```diff
@@ -414,8 +414,8 @@ private string GetTEMPDirectory()
                     return Environment.GetEnvironmentVariable("TMPDIR");
                 else if (DirExists(Environment.GetEnvironmentVariable("TMP")))
                     return Environment.GetEnvironmentVariable("TMP");
-                else if (DirExists("/home/DDITAdministrator/myagent/_work/_temp"))
-                    return "/home/DDITAdministrator/myagent/_work/_temp";
+                else if (DirExists(Path.Combine(Environment.GetEnvironmentVariable("HOME"), "myagent/_work/_temp")))
```
chcosta: If all you're doing is checking for the existence of a directory as your "temp" directory, why does it matter what OS you're on?

chcosta:
```C#
private void AddDirToListIfExist(List<string> dirs, string directory)
```
`dirs` could be a HashSet

chcosta: Why not follow the `AddDirToListIfExist` pattern which is used in GetNugetCacheDirectories?

chcosta: Method name should be `AddDirToListIfExists`, missing trailing 's'

PR #1504, opened 2017-05-17
crossgen.sh improvements

ellismg:
This PR has two changes:

1) Don't crossgen a file if the ni is already present.  This will address the segmentation faults we see when we try to crossgen System.Private.Corelib.dll (the .ni.dll already exists, it's in the package).
2) If `NUGET_PACKAGES` is set in the environment, respect that variable as the location to restore packages to.

I'd like some sort of knob to allow me to restore packages to a shared location as part of the build from source work, and `NUGET_PACKAGES` seemed like the obvious choice, since the tooling understands that by default.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -54,13 +54,17 @@ crossgen_single()
 {
     __file=$1
     if [[ $__file != *.ni.dll && $__file != *.ni.exe ]]; then
-        $__crossgen /Platform_Assemblies_Paths $__sharedFxDir:$__toolsDir /nologo /MissingDependenciesOK /ReadyToRun $__file > /dev/null
-        if [ $? -eq 0 ]; then
-            __outname="${__file/.dll/.ni.dll}"
-            __outname="${__outname/.exe/.ni.exe}"
-            echo "$__file -> $__outname"
+        if [[ ($__file == *.dll && -e ${__file/.dll/.ni.dll}) || ($__file == *.exe && -e ${__file/.exe/.ni.exe}) ]]; then
```
weshaggard: This isn't going to help System.Private.CoreLib case because we no longer name the file ni.dll it is just .dll

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -78,7 +82,7 @@ __targetDir=$1
 __scriptpath=$(cd "$(dirname "$0")"; pwd -P)
 __toolsDir=$__scriptpath/../Tools
 __dotnet=$__toolsDir/dotnetcli/dotnet
-__packagesDir=$__scriptpath/../packages
+__packagesDir="${NUGET_PACKAGES:-${__scriptpath}/../packages}"
```
weshaggard: Did we ever end up putting crossgen into the shared framework? That would make this script a lot simpler.

ellismg: It is in Microsoft.NETCore.App now, in the tools folder.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
ellismg: It helped locally with the latest buildtools.  Do you know when that change went in?

weshaggard: https://github.com/dotnet/coreclr/pull/11359

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
weshaggard: In that case does it make since to stop hardcoding the coreclr package version and instead use the shared framework version to restore Microsoft.NETCore.App instead?

ellismg: I can fix that after 2.0

ellismg:
@weshaggard I added an explicit check to not crossgen System.Private.Corelib.dll.

I also added a commit that makes it possible to specify CommitHash and VersionSeedDate properties explicitly instead of always shelling out to `git` (which is useful for the build for source stuff).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -285,7 +291,12 @@
       <LatestDateCommand Condition="'$(RunningOnUnix)'=='true'">git show -s --format=%25cd --date=short HEAD</LatestDateCommand>
     </PropertyGroup>
 
-    <Exec Command="$(LatestDateCommand)" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+    <PropertyGroup Condition="'$(VersionSeedDate)' != ''">
+      <LatestDateCommandExitCode>0</LatestDateCommandExitCode>
+    </PropertyGroup>
+
+
```
weshaggard: nit: extra blank line

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
weshaggard: sounds good. Might be worth filing an issue.

PR #1507, opened 2017-05-20
Add xunit.netcore.extensions full framework project to be able to use xunit extensibility features

safern:
We where consuming xunit's extensibility netstandard package, and the desktop and netstandard have different behaviors. So in order to get full extensibility and behavior we should add this in the Xunit.Netcore.Extensions package. 


tarekgh:
LGTM

tarekgh:
Looks Ubuntu build failed because it was trying to build net425 of the extension. we need to skip building that for Linux

safern:
> Looks Ubuntu build failed because it was trying to build net425 of the extension. we need to skip building that for Linux

Yes. I will update that. Thanks!

safern:
Thanks for reviewing :)

PR #1510, opened 2017-05-22
Mark SkippedTestCase as serializable for Full Framework

safern:
This is needed because xunit creates different AppDomains, so if it not serializable when the condition in ConditionalFact/Theory attributes is not met the test will be marked as not discovered instead of skipped. So the test summary will not be correct and we will never catch when the tests are skipped for desktop.

cc: @weshaggard

### DIFF
--- a/src/xunit.netcore.extensions/SkippedTestCase.cs
+++ b/src/xunit.netcore.extensions/SkippedTestCase.cs
@@ -6,16 +6,14 @@ using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Xunit;
 using Xunit.Abstractions;
 using Xunit.Sdk;
 
 namespace Xunit.NetCore.Extensions
 {
-#if FULL_FRAMEWORK
-    [Serializable]
-#endif
     /// <summary>Wraps another test case that should be skipped.</summary>
-    internal sealed class SkippedTestCase : IXunitTestCase
+    internal sealed class SkippedTestCase : LongLivedMarshalByRefObject, IXunitTestCase
     {
         private readonly IXunitTestCase _testCase;
         private readonly string _skippedReason;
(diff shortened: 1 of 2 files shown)


src/xunit.netcore.extensions/SkippedTestCase.cs
```diff
@@ -2,16 +2,18 @@
...
 using System.Threading.Tasks;
+using Xunit;
 using Xunit.Abstractions;
 using Xunit.Sdk;
 
 namespace Xunit.NetCore.Extensions
 {
     /// <summary>Wraps another test case that should be skipped.</summary>
-    internal sealed class SkippedTestCase : IXunitTestCase
+    internal sealed class SkippedTestCase : LongLivedMarshalByRefObject, IXunitTestCase
```
weshaggard: LongLivedMarshalByRefObject exists in both the netstandard and netfx xunit?

safern:
I've changed the approach to instead of using #if-def on our side for the Desktop version, to make `SkippedTestCase` inherit from `Xunit.LongLivedMarshalByRefObject` which handles the #if-defs already. For netfx it will inherit from `MarshalByRefObject` for netstandard it is an empty class.

See: https://github.com/xunit/xunit/blob/master/src/xunit.runner.utility/Utility/LongLivedMarshalByRefObject.cs

src/xunit.netcore.extensions/SkippedTestCase.cs
safern: I missed your comment while I was publishing mine on the main thread. But yes, it exists in both.

https://github.com/dotnet/buildtools/pull/1510#issuecomment-303228025

We just had the same timing haha :)

safern: I think that was the goal of creating a custom abstract class that inherits from MarshalByRefObject so that they didn't have to #if-def all of their types and they could just make them all inherit from this one.

PR #1511, opened 2017-05-23
Remove dead files

danmosemsft:
We use [2 pinvoke analyzer baselines, ](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.CodeAnalysis/PackageFiles/codeAnalysis.targets#L15) one for OneCore and one for UWP.

We use PinvokeAnalyzer_Win32Apis.txt -- the only unique entries in PinvokeAnalyzer_OneCoreApis.txt are apisets and we don't use those anymore so deleting it.
For UWP we use PinvokeAnalyzer_Win32UWPApis.txt -- I don't know the purpose of PinvokeAnalyzer_UWPApis.txt but it's not accurate and not touched in 18 months so deleting it.


weshaggard:
> I don't know the purpose of PinvokeAnalyzer_UWPApis.txt but it's not accurate and not touched in 18 months so deleting it.

The purpose of it is make sure we use only APIs that are part of the modern sdk, i.e. this is the policy part of the store ingestion. We should actually update and use that, assuming that isn't what your new PinvokeAnalyzer_Win32UWPApis.txt is about.

danmosemsft:
> assuming that isn't what your new PinvokeAnalyzer_Win32UWPApis.txt is about.

That file existed already and it contains the latest data. I don't know how PinvokeAnalyzer_UWPApis.txt was meant to be different, but it's out of date and all in terms of api sets so it seems to have no use/value.

PR #1512, opened 2017-05-23
Return exit code from dumpling collect script

weshaggard:
Address https://github.com/dotnet/corefx/issues/20172


weshaggard:
cc @danmosemsft @mellinoe

mellinoe:
I'm fine with this as long as the "truncation" problem with the python API doesn't break us down the line. Thanks for fixing this.

PR #1513, opened 2017-05-23
Statically evaulate TargetOSTrait

weshaggard:
If you don't statically evaluate TargetOSTrait then we don't correctly
filter Xunit tests in VS via the default test runs parameters we have.

Should address https://github.com/dotnet/corefx/issues/18469.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -58,6 +58,14 @@
 
     <XunitOptions>$(XunitOptions) -notrait category=non$(_bc_TargetGroup)tests</XunitOptions>
 
+    <TargetOS Condition="'$(TargetOS)' == ''">$(DefaultOSGroup)</TargetOS>
+    <TargetOSTrait Condition="'$(TargetOS)'=='Windows_NT'">nonwindowstests</TargetOSTrait>
+    <TargetOSTrait Condition="'$(TargetOS)'=='Linux'">nonlinuxtests</TargetOSTrait>
+    <TargetOSTrait Condition="'$(TargetOS)'=='OSX'">nonosxtests</TargetOSTrait>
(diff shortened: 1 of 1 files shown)


weshaggard:
cc @mellinoe

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -58,6 +58,14 @@
 
     <XunitOptions>$(XunitOptions) -notrait category=non$(_bc_TargetGroup)tests</XunitOptions>
 
+    <TargetOS Condition="'$(TargetOS)' == ''">$(DefaultOSGroup)</TargetOS>
+    <TargetOSTrait Condition="'$(TargetOS)'=='Windows_NT'">nonwindowstests</TargetOSTrait>
+    <TargetOSTrait Condition="'$(TargetOS)'=='Linux'">nonlinuxtests</TargetOSTrait>
+    <TargetOSTrait Condition="'$(TargetOS)'=='OSX'">nonosxtests</TargetOSTrait>
+    <TargetOSTrait Condition="'$(TargetOS)'=='FreeBSD'">nonfreebsdtests</TargetOSTrait>
+    <TargetOSTrait Condition="'$(TargetOS)'=='NetBSD'">nonnetbsdtests</TargetOSTrait>
+    <XunitOptions Condition="'$(TargetOSTrait)' != ''">$(XunitOptions) -notrait category=$(TargetOSTrait)</XunitOptions>
```
safern: Should we instead of appending it to `XunitOptions` add it to `XunitTraitOptions` that is added to TestCommandLine through the placeholder replacement here: https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L191 ?

weshaggard: We cannot use XunitTraitOptions because that isn't setup statically and we need it to be static for VS.

safern: Oh I see. Sounds reasonable.

Issue #1508, opened 2017-05-22
ActiveIssueAttribute should be able to be used in test classes.


safern:
ActiveIssueAttribute is currently marked only for test methods, we should allow them to be used to disable a whole test class.

cc: @weshaggard

Issue #1446, opened 2017-04-18
Set AllowMultipleUse to true on SkipOnTargetFrameworkAttribute


AtsushiKan:
If [ActiveIssue] can be used multiple times on the same test, [SkipOnTargetFramework] should be able to do so as well.

I've been adding hundreds of [SkipOnTargetFramework(UapAot, "BlahBlahBlah is not supported on UapAot"] and it's a PITA when there's a [SkipOnTargetFramework(NetFramework)] already there for a completely different reason.

This can be considered part of https://github.com/dotnet/buildtools/issues/1395

safern:
@danmosemsft taking this one as I've been touching this side lately.

PR #1515, opened 2017-05-24
Fix SkipOnTargetFrameworkAttribute, ActiveIssueAttribute and add Desktop tests custom trace listener 

safern:
This PR includes three different fixes:

* Change SkipOnTargetFrameworkAttribute to allow multiple usages: https://github.com/dotnet/buildtools/issues/1446
* Change ActiveIssueAttribute to allow usage to disable a whole test class.
* Add DesktopTestTraceListener to solve issues when a Debug.Assert fails inside a test execution. With this behavior we will not have assert pop-ups anymore so tests will not hang when running in CI/Helix and will throw an exception so that the test is detected as failed. In order to complete this we will need to follow up with: https://github.com/dotnet/corefx/issues/20109#issuecomment-303774134

Fixes: https://github.com/dotnet/buildtools/issues/1446
Fixes: https://github.com/dotnet/buildtools/issues/1508

cc: @weshaggard


src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
```diff
@@ -11,7 +11,7 @@ namespace Xunit
     /// Apply this attribute to your test method to specify an active issue.
     /// </summary>
     [TraitDiscoverer("Xunit.NetCore.Extensions.ActiveIssueDiscoverer", "Xunit.NetCore.Extensions")]
-    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true)]
```
weshaggard: While this allows it to be applied does it correctly cause the entire class to be skipped?

src/xunit.netcore.extensions.46/DesktopTestTraceListener.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// </summary>
+    public class DesktopTestTraceListener : DefaultTraceListener
+    {
+        /// <summary>
+        /// Override of <see cref="DefaultTraceListener.Fail" /> to handle Assert failures with custom behavior.
+        /// When an Assert failure happens during test execution we will rather throw a DebugAssertException so that the test fails and we have a full StackTrace.
+        /// </summary>
+        public override void Fail(string message, string detailMessage)
+        {
+            StackTrace stack = new StackTrace(true);
```
weshaggard: Why are you manually creating the stack trace to pass? When you throw the exception it should create the stack.

src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
safern: Yes. I've tested locally. Since this is handled by Xunit it gets all the attributes in a member (either class or method) and it calls the discoverer or GetTraits method for each of them and add those traits to each guy. So yes xunit will call out GetTraits multiple time with each AttributeInfo and then we will return the corresponding traits for each Attribute.

src/xunit.netcore.extensions.46/DesktopTestTraceListener.cs
safern: I added it as I grabbed the code from what DefaultListener is doing but without the UI dialog. It makes sense to leave the stack trace to the expection. Good thought. Will fix that

Issue #1275, opened 2016-12-20
Add property for specifying XUnit method/class to test


naamunds:
To specify a method or class to run when running tests for a project using the BuildTools `Test` target, one needs to specify `"/p:XunitOptions=-method <method_name>"` or `"/p:XunitOptions=-class <class_name>"`.  As mentioned in https://github.com/dotnet/buildtools/pull/1272#discussion_r93084987, this is unintuitive.

In addition, specifying it using `XunitOptions` overwrites some of the options used by the `Test` target code, most notably `-xml $(XunitResultsFileName)`, meaning that no XML results file is produced when running a single method or class in this way.

Using explicit properties such as `XunitMethodName` and `XunitClassName` would make the usage clearer and solve the overwrite issue.

safern:
Fixed in: #1518

PR #1518, opened 2017-05-25
Xunit options

safern:
This is to allow people to pass additional options to xunit through msbuild.

cc: @weshaggard @danmosemsft

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -57,7 +57,7 @@
     <CoverageOptions>-oldStyle -filter:"{CoverageFilter}" -excludebyfile:"*\Common\src\System\SR.*" -nodefaultfilters -excludebyattribute:*.ExcludeFromCodeCoverage* -skipautoprops -hideskipped:All -threshold:1</CoverageOptions>
     <CoverageCommandLine>$(CoverageOptions) -returntargetcode -register:user -target:$(TestProgram) -output:$(CoverageOutputFilePath)</CoverageCommandLine>
     <TestHost>$(CoverageHost)</TestHost>
-    <XunitOptions>$(XunitOptions) -parallel none</XunitOptions>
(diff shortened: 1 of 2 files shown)


danmosemsft:
https://github.com/dotnet/buildtools/issues/1275
https://github.com/dotnet/buildtools/pull/1462

danmosemsft:
I don't think this is the right fix. The goal is to allow /p:XUnitOptions to work (as well as it would be nice to add the two custom properties I had in my PR for class and method in case folks want those). The problem is if you do this /pXunitOptions, all these lines appending to it, don't work because the command line value wins. This is by (bad, original) design of MSBuild.

To fix that, XUnitOptions needs to be dumped into some other property up front, say `<_XUnitOptions>`. From that point down all appending etc is on _XUnitOptions.

safern:
Could you take a look at my latest commit? With that way we would have XunitAdditionalOptions, XunitMethodName and XunitClassName so that is easier for everyone so that they don't have to learn the xunit switches such as -method or -class. We could leave XunitOptions as proposed to be the one people uses through msbuild and rename the actual property as you propose. Which is the same I'm doing here but without renaming, so whichever you prefer.

bartonjs:
I'd prefer to keep my motor memory of setting XunitOptions, and thus the property inside msbuild should be renamed.  Needing to learn a new way of doing the same task is always frustrating.

weshaggard:
>I'd prefer to keep my motor memory of setting XunitOptions, and thus the property inside msbuild should be renamed. Needing to learn a new way of doing the same task is always frustrating.

I tend to agree. Just rename the one we use internally to _XunitOptions.

safern:
Well I guess it is decided, the 3 of you agreed :) so I will update the PR.

safern:
Updated the PR to have 3 options:
* XunitOptions
* XunitMethodName
* XunitMethodClass

Could you PTAL?

safern:
> You might want to do a search for XunitOptions in other targets to make sure there aren't any other places.

Didn't actually know there where more targets that involved Xunit. Will verify before merging. Thanks for the heads up.

safern:
There was an instance of XunitOptions in CodeCoverage.targets where it disables parallelization. I updated it to be _XunitOptions as well.

PR #1520, opened 2017-05-26
Add symbol signing targets

dagood:
Adds targets and new utility tasks to create signed catalog files and inject them into symbol packages for upload to MyGet. Uses `makecat` to create the catalogs and MicroBuild to sign them.

 1. Reads the entries in the symbol package zips
 2. Finds ones that have an extension we want to sign
 3. Extracts those entries into an intermediate directory
 4. Creates one catalog file per package
 5. Signs the catalog files
 6. Injects the signed catalog files
 7. Injects a list of cataloged files into each symbol package (for EmbedIndex)

The first commit is from when we were going to have one catalog file per signed file for simplicity, but that ended up being extremely slow: ~1 hour (which would be part of the package publish leg). A single catalog per package takes ~5 minutes total.

I don't like the `MSBuild` tasks, but so far they seem like the clearest way to make the per-package ops work.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
@@ -2,6 +2,8 @@
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <UsingTask TaskName="AddItemIndices" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="ZipFileExtractToDirectory" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileGetEntries" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask TaskName="ZipFileInjectFile" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
     <SymbolsRequestIntermediateDir Condition="'$(SymbolsRequestIntermediateDir)'==''">$(BaseIntermediateOutputPath)SymbolsRequest\</SymbolsRequestIntermediateDir>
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks/ZipFileExtractToDirectory.cs
```diff
@@ -13,22 +13,28 @@ namespace Microsoft.DotNet.Build.Tasks
...
         public string DestinationDirectory { get; set; }
 
         /// <summary>
-        /// Indicates if the destination archive should be overwritten if it already exists.
+        /// Indicates if the destination directory should be overwritten if it already exists.
         /// </summary>
         public bool OverwriteDestination { get; set; }
 
+        /// <summary>
+        /// The paths to include in the extraction, relative to root. If null,
```
dagood: It's only file (entry) paths that are supported--I'll clarify.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
```diff
@@ -224,4 +226,174 @@
...
+             Properties="SymbolPackageFilePath=%(SymbolPackageFileWithIndex.Identity);
+                         SymbolPackageIndex=%(SymbolPackageFileWithIndex.Index);
+                         ExtensionsToCatalog=$(ExtensionsToCatalog);
+                         SymbolCatalogIntermediateDir=$(SymbolCatalogIntermediateDir);
+                         MakeCatCommand=$(MakeCatCommand)" />
+
+    <!-- Find the catalogs to sign them. -->
+    <ItemGroup>
+      <FilesToSign Include="$(SymbolCatalogIntermediateDir)**\*.cat">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
```
crummel: Is `InternalCertificateId` the variable that already exists and is used for signing regular binaries?  If so, I think we'd be better off using a different variable like `SymbolCatalogCertificateId` in case a repo wants to sign binaries with a different certificate (or the file types a certificate supports changes).  We could default the new variable to `$(InternalCertificateId)` for now in the build definition.

dagood: I'm setting it manually in my test build with `/p` (I was unable to find `InternalCertificateId` referenced anywhere), so changing the name to distinguish sounds fine to me.

src/Microsoft.DotNet.Build.Tasks/ZipFileExtractToDirectory.cs
```diff
@@ -50,7 +56,24 @@ public override bool Execute()
                 if (!Directory.Exists(Path.GetDirectoryName(DestinationDirectory)))
```
dagood: Removed the redundant check entirely, instead.

PR #1522, opened 2017-05-30
Add Remote Execution to UAP runner

tarekgh:


tarekgh:
@AlexGhiondea could you please have a look? 

CC @joperezr

tarekgh:
I'll look at the failure as I am not able to produce it locally. but will investigate

src/xunit.runner.uap/RunnerRemoteExecutionServiceUAP/RunnerRemoteExecutionServiceUAP.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                argList.Add(arg);
+                i++;
+            }
+            additionalArgs = argList.ToArray();
+
+            // Load the specified assembly, type, and method, then invoke the method.
+            // The program's exit code is the return value of the invoked method.
+            object instance = null;
+            string log = $"RemoteExecuter: {assemblyName} {methodName} {string.Join(", ", additionalArgs)}{Environment.NewLine}";
```
AlexGhiondea: Should we change this to a StringBuilder and use it to log more info from this code? Would that help debug this in the future?

src/xunit.runner.uap/RunnerRemoteExecutionServiceUAP/RunnerRemoteExecutionServiceUAP.cs
```diff
@@ -0,0 +1,146 @@
...
+                Assembly a = Assembly.Load(new AssemblyName(assemblyName));
+                Type t = a.GetType(typeName);
+                MethodInfo mi = t.GetTypeInfo().GetDeclaredMethod(methodName);
+                if (!mi.IsStatic)
+                {
+                    instance = Activator.CreateInstance(t);
+                }
+
+                // Invoke the test
+                object result = mi.Invoke(instance, additionalArgs);
```
AlexGhiondea: This makes the assumption that test methods only take in arguments as objects, right?

While not necessary for this change, we might have to 'cast' each parameter to the type expected by the method...

src/xunit.runner.uap/RunnerRemoteExecutionServiceUAP/RunnerRemoteExecutionServiceUAP.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            returnData["Log"] = log;
+        }
+
+        private bool GetStringValueFromValueSet(ValueSet message, string key, out string val)
+        {
+            val = null;
+            
+            object valObject;
+            if (!message.TryGetValue(key, out valObject))
```
AlexGhiondea: Will this do a case insensitive search for the key?

src/xunit.runner.uap/RunnerRemoteExecutionServiceUAP/RunnerRemoteExecutionServiceUAP.cs
tarekgh: We always require a strings as the method parameters. you may look at what we already doing today here:

https://github.com/dotnet/corefx/blob/master/src/Common/tests/System/Diagnostics/RemoteExecutorConsoleApp/RemoteExecutorConsoleApp.cs#L55

src/xunit.runner.uap/RunnerRemoteExecutionServiceUAP/RunnerRemoteExecutionServiceUAP.cs
tarekgh: This will be case sensitive. I don't think we need to complicate it to support case insensitive as we'll have limited cases communicating between the service and the client.

src/xunit.runner.uap/RunnerRemoteExecutionServiceUAP/RunnerRemoteExecutionServiceUAP.cs
tarekgh: I'll change this to SB. thanks.

joperezr:
Why do you need the pfx key file?

tarekgh:
@joperezr 

> Why do you need the pfx key file?

without it, I cannot compile the runner. I don't know how this compile today or we just skip it

src/dirs.proj
```diff
@@ -4,6 +4,7 @@
   <ItemGroup>
     <!-- Exclude the UAP runner from the build as it depends on Win10 SDKs -->
     <ProjectsToExclude Include="xunit.runner.uap\XUnit.Runner.Uap.csproj" />
+    <ProjectsToExclude Include="xunit.runner.uap\RunnerRemoteExecutionServiceUAP\RunnerRemoteExecutionServiceUAP.csproj" />
```
weshaggard: Is the expectation that this always gets built manually and we upload it somewhere?

tarekgh: Yes. The new added project is referenced from xunit.runner.uap which is already excluded. the way we update the runner xunit.runner.uap today is by manually compiling it and packaging it. I think @AlexGhiondea used buildtools repo as a source control only and not to compile it.

src/xunit.runner.uap/RunnerRemoteExecutionServiceUAP/RunnerRemoteExecutionServiceUAP.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                // Invoke the test
+                object result = mi.Invoke(instance, additionalArgs);
+                var task = result as Task<int>;
+                returnData["Results"] = task != null ? task.GetAwaiter().GetResult() : (int) result;
+            }
+            catch (Exception exc)
+            {
+                returnData["Results"] = -4;
+                log.Append(exc);
```
danmosemsft: exc.ToSTring() maybe?

tarekgh: it has the same effect.

PR #1524, opened 2017-05-30
GenAPI: Write security attributes in property accessors too

akoeplinger:
Like it's done e.g. for methods: https://github.com/mono/buildtools/blob/14a12e45dd954bfd6e98aecc366e8742aeaa6590/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs#L32


weshaggard:
@alexperovich thanks for all the fixes. What are you using genapi for that ran into all these issues?

akoeplinger:
@weshaggard just some tooling in Mono that we use for producing API diffs :)

PR #1526, opened 2017-05-30
GenAPI: Implement "volatile" modifier

akoeplinger:


weshaggard:
Looks like you got a compilation issue to work out here.

akoeplinger:
Fixed the build error 👍

PR #1530, opened 2017-05-31
Fix the success code returned from remote execution

tarekgh:
Currently the remote execution returning 0 as success code but in corefx repo we use the value 42 as the success code. changing the success code to 42 so corefx tests work seamlessly with the uap runner


tarekgh:
@AlexGhiondea could you please have a look at this minor fix. 

cc @danmosemsft

tarekgh:
thanks @stephentoub for your review!

PR #1532, opened 2017-06-01
GenAPI: Add global prefix to explicit interface method and property

WonyoungChoi:
When use "-global" option, the global prefix "global::" should be added to name of explicit interface methods and properties as follows.

```
object global::System.Collections.IEnumerator.Current { get { throw new global::System.PlatformNotSupportedException("Not Supported Feature"); } }
```


stephentoub:
This PR is a year and a half old.  Can it be closed?

PR #1533, opened 2017-06-01
When GenFacades fails it should return an error code that does not me…

AlexGhiondea:
…an success

/cc @weshaggard


mellinoe:
https://github.com/dotnet/buildtools/pull/1533/files#diff-d966ef3e16dd5d4a70644b66bd4f0bfeR50

needs to return 1 as well.

AlexGhiondea:
@mellinoe ah -- yes! I changed that in a different version of that file and missed it here. Will fix shortly.

PR #1534, opened 2017-06-01
Fix duplicate file keys when creating catalogs.

crummel:
MakeCat apparently does not handle duplicate keys well even though we are supposed to be using the `<HASH>` as the key.  This change makes us use the full file path for the key so there are no duplicates.


crummel:
Yes, it wasn't putting all the file hashes in the catalog so verification failed.  Looks like it only uses the first one it saw if there's multiple files with the same tag.

PR #1535, opened 2017-06-01
Move core-setup tasks into BuildTools

chcosta:
This change is moving the tasks from https://github.com/dotnet/core-setup/tree/master/tools-local/Microsoft.DotNet.Build.Tasks.Local into BuildTools.  

The files which are modified (not new additions), *should* be non-breaking in repos which utilize these tasks though I will have to make some small modifications to core-setup when I update that repo to consume these changes.

https://github.com/dotnet/core-setup/blob/master/tools-local/Microsoft.DotNet.Build.Tasks.Local/DownloadBlobFromAzure.cs and https://github.com/dotnet/core-setup/blob/master/tools-local/Microsoft.DotNet.Build.Tasks.Local/DownloadBlobsFromAzure.cs have been collapsed into the DownloadFromAzure task.  I've validated previous behavior for DownloadFromAzure task in CoreFx repo.

I'm not porting https://github.com/dotnet/core-setup/blob/master/tools-local/Microsoft.DotNet.Build.Tasks.Local/GetHostInformation.cs because I think we can replace that task with BuildTools pre-existing https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/GetTargetMachineInfo.cs task.

All tasks are moving into CloudTestTasks except for GenerateChecksums.cs which is moving into Microsoft.DotNet.Build.Tools.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
@@ -31,6 +31,8 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
 
         public string BlobNamePrefix { get; set; }
 
+        public ITaskItem[] BlobNames { get; set; }
+        
         public override bool Execute()
         {
             return ExecuteAsync().GetAwaiter().GetResult();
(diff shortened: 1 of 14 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
```diff
@@ -48,96 +50,64 @@ public override bool Execute()
...
-                            responseFile.LoadXml(response.Content.ReadAsStringAsync().GetAwaiter().GetResult());
-                            XmlNodeList elemList = responseFile.GetElementsByTagName("Name");
-
-                            blobsNames.AddRange(elemList.Cast<XmlNode>()
-                                                        .Select(x => x.InnerText)
-                                                        .ToList());
-
-                            nextMarker = responseFile.GetElementsByTagName("NextMarker").Cast<XmlNode>().FirstOrDefault()?.InnerText;
-                        }
+                        blobNames = blobNames.Where(b => b.StartsWith(BlobNamePrefix)).ToList<string>();
```
weshaggard: Getting the blobs this way is going to be much less efficient. The reason we added the BlobNamePrefix is because we could directly query the blobs that way and it would limit to just the blobs that start with that instead of getting all the blobs which overtime will be a ton which will be a noticeable performance issue.

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
```diff
@@ -31,6 +31,8 @@ public sealed class DownloadFromAzure : AzureConnectionStringBuildTask
 
         public string BlobNamePrefix { get; set; }
 
+        public ITaskItem[] BlobNames { get; set; }
```
weshaggard: Is there a scenario were we want to get an arbitrary set of blobs that don't share a common root?

chcosta: In Core-Setup, we only download the blobs we intend to publish - https://github.com/dotnet/core-setup/blob/master/publish/publish.proj#L136

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
chcosta: This is the scenario where a list of `BlobNames` has been explicitly specified, and we're ensuring that they match a certain prefix.  I wouldn't this to affect performance in any way (if we support passing an explicit itemgroup blob list).

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
```diff
@@ -48,96 +50,64 @@ public override bool Execute()
...
-                        responseFile.LoadXml(await response.Content.ReadAsStringAsync());
-                        XmlNodeList elemList = responseFile.GetElementsByTagName("Name");
-
-                        blobsNames.AddRange(elemList.Cast<XmlNode>()
-                                                    .Select(x => x.InnerText)
-                                                    .ToList());
-                        nextMarker = responseFile.GetElementsByTagName("NextMarker").Cast<XmlNode>().FirstOrDefault()?.InnerText;
-                    }
-                    while (!string.IsNullOrEmpty(nextMarker))
+                        AccountName = AccountName,
```
weshaggard: You dropped the BlobNamePrefix here.

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
weshaggard: The performance hit comes from the fact that you are downloading a list of all the blob names which will become huge overtime. The storage APIs are optimized based on blob name prefix which is why if we can use a common prefix we won't need to download all the names but instead just the subset under the prefix.

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
chcosta: line 63, below

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
chcosta: This code path is skipping the list retrieval, the list is already specified by the user in `BlobNames`

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
weshaggard: Sorry your correct that is accounted for.

src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
weshaggard: Got it. That makes sense. I'm OK with that.

src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
```diff
@@ -26,22 +26,7 @@
...
-    <Compile Include="FilterForUpload.cs" />
-    <Compile Include="GetPerfTestAssemblies.cs" />
-    <Compile Include="ListAzureContainers.cs" />
-    <Compile Include="SendToHelix.cs" />
-    <Compile Include="UploadClient.cs" />
-    <Compile Include="UploadToAzure.cs" />
-    <Compile Include="WriteBuildStatsJson.cs" />
-    <Compile Include="RemoveItemMetadata.cs" />
-    <Compile Include="WriteItemsToJson.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)*.cs" />
```
weshaggard: But why? I realize wildcards seem like a good idea but they usually do nothing but cause trouble in some way.

chcosta: I kept moving files over to buildtools then forgetting to add them to the csproj, so I changed this to a wildcard with the thought that msbuild based cli project systems do it this way anyways.  I can move this back to being explicit since there are pros-cons to both methods and I don't have a super strong need to make this change.

PR #1536, opened 2017-06-02
Tool runtime improvements

ellismg:
Two major changes here:
- Stop publishing the tool runtime standalone.
- Clean up crossgen.sh to pull crossgen from the Microsoft.NETCore.App package, using the version that matches whatever is in the shared folder of the dotnet cli we're using.

I hit these issues while locally tying to move the version of the CLI that CoreFX uses forward to a preview2 build, which lead to these changes.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
@@ -1,10 +1,7 @@
 #!/usr/bin/env bash
+set -euo pipefail
 
 # Restores crossgen and runs it on all tools components.
-
-__CoreClrVersion=2.0.0-preview1-25204-02
-__MyGetFeed=https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
-
 usage()
 {
     echo "crossgen.sh <directory>"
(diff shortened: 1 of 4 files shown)


ellismg:
Wes or Eric, do you understand the logic in init-tools.cmd here:

https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd#L83-L97

I'm assuming we want to change this to remove the runtimes but keep the two publishes for different TFMs to two different locations.

I can dig in but if either of you knew "oh yeah, this is exactly what's going on" that would be helpful.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -54,45 +49,46 @@ crossgen_single()
...
     usage
 fi
 
+__MyGetFeed=${BUILDTOOLS_CROSSGEN_FEED:-https://dotnet.myget.org/F/dotnet-core/api/v3/index.json}
 __targetDir=$1
 __scriptpath=$(cd "$(dirname "$0")"; pwd -P)
 __toolsDir=$__scriptpath/../Tools
 __dotnet=$__toolsDir/dotnetcli/dotnet
-__packagesDir="${NUGET_PACKAGES:-${__scriptpath}/../packages}"
+__packagesDir=${NUGET_PACKAGE:-$__scriptpath/../packages}
```
ellismg: I will undo this one change (i was using a version of init-tools.sh that didn't have and added it back but slightly differently)!

weshaggard:
>I'm assuming we want to change this to remove the runtimes but keep the two publishes for different TFMs to two different locations.

We needed to publish for netcoreapp for running CLI applications and also publish for net46 so we get the correct dependencies when using the full framework msbuild. Yeah I think we can get away with simply removing the runtime for those.

ellismg:
> Yeah I think we can get away with simply removing the runtime for those.

Thanks, I'll do that.

ellismg:
I think this is ready now.  I did build locally on both Ubuntu and windows and stuff is working.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -54,30 +49,29 @@ crossgen_single()
 {
     __file=$1
     if [[ $__file != *.ni.dll && $__file != *.ni.exe ]]; then
-        if [[ ($__file == *.dll && -e ${__file/.dll/.ni.dll}) || ($__file == *.exe && -e ${__file/.exe/.ni.exe}) ]]; then
```
weshaggard: Is there any harm in keeping this condition to not ngen something that already has an ni?

ellismg: No, I will add it back (the removal of if is an artifact of how I developed this change where I started from what was in the CoreFX version of build tools, instead of what was in HEAD).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
```diff
@@ -12,4 +13,7 @@ foreach ($file in Get-ChildItem $ToolRuntimePath *.runtimeconfig.json)
     Write-Host "Correcting runtime version of" $file.FullName
     $text = (Get-Content $file.FullName) -replace "1.1.0","$highestVersion"
     Set-Content $file.FullName $text
-}
\ No newline at end of file
+}
+
+# Make a directory in the root of the tools folder that matches the buildtools version
```
weshaggard: What is the purpose of making this directory?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -80,22 +78,25 @@ if not [%RESTORE_ERROR_LEVEL%]==[0] (
...
+call "%DOTNET_CMD%" publish "%TOOLRUNTIME_PROJECT%" -f net46 -o "%TOOLRUNTIME_DIR%\net46"
 set NET46_PUBLISH_ERROR_LEVEL=%ERRORLEVEL%
 @echo off
 if not [%NET46_PUBLISH_ERROR_LEVEL%]==[0] (
   echo ERROR: An error ocurred when running: '"%DOTNET_CMD%" publish "%TOOLRUNTIME_PROJECT%" -f net46'. Please check above for more details.
   exit /b %NET46_PUBLISH_ERROR_LEVEL%
 )
 
+:: Copy some roslyn files which are published into runtimes\any\native to the root
+Robocopy "%TOOLRUNTIME_DIR%\runtimes\any\native" "%TOOLRUNTIME_DIR%\."
```
weshaggard: I assume this was the only thing that was left out when you stopped publish for the RID?

ellismg: > I assume this was the only thing that was left out when you stopped publish for the RID?

As far as I could tell, yes.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
ellismg: The directory did not exist after changing how publishing works, but init-tools.cmd in each repository expects it to exist so it can write the semaphore into it.  It seemed cleaner to just create it here instead of requiring init-tools.cmd to build it.

Apparently CoreCLR was hitting this even without my changes: https://github.com/dotnet/coreclr/pull/11912 and their workaround was to hork with the outer init-tools.cmd.

weshaggard: Sounds good. might be worth adding the semaphore note to the comment.

PR #1544, opened 2017-06-06
init-tools.sh: fall back to portable Linux when Linux host has unknown rid

tmds:
CC @ellismg


ellismg:
@tmds In 1a841480272ed46b459bf928881d3f3d0dd85e9d I moved off of publishing standalone, which sidesteps the entire problem.  Do you think that is sufficient?

tmds:
@ellismg even better! Thanks

PR #1553, opened 2017-06-08
Add -Force to init-tools.ps1

karajas:
This is to unblock Maestro updates failing with directory already exists.
https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=792972&_a=summary

/cc @ellismg @dagood


karajas:
Added https://github.com/dotnet/buildtools/issues/1554 to track fixing the actual issue

PR #1561, opened 2017-06-09
Change Microsoft.xunit.extensibility.execution.netcore redist package links to BuildTools ones

dagood:
Copy license and project urls from `Microsoft.xunit.netcore.extensions.nuspec` to `Microsoft.xunit.extensibility.execution.netcore.nuspec`. This should avoid some confusion when seeing the package without context.

@brianrob @karajas


brianrob:
Thank you @dagood!

PR #1563, opened 2017-06-12
Enabling build of UAP Launcher and Runner on Windows

joperezr:
cc: @weshaggard @tarekgh 

We want to start building both the launcher and the Runner for our F5 uap tests on our official builds so that the binaries are signed. These changes will enable the build for both of them on Windows. I added the NO MERGE label for now since I suspect Jenkins machines won't be able to build the Runner, but if it does succeed then this would be ready for merging.


tarekgh:
That will be great if we can get it build here.

joperezr:
Closing this since we won't be able to build the runner for now.

PR #1569, opened 2017-06-14
Add CodeGeneration capability which is required for some of our MemoryMappedFile tests

joperezr:
cc: @JeremyKuhne @AlexGhiondea @tarekgh 

Already did a private Helix run with these changes and have verified that it fixes the issues and it doesn't cause any regressions.


tarekgh:
LGTM

PR #1570, opened 2017-06-14
Port the IBC merge infrastructure changes from master

AlexGhiondea:


dagood:
Looks reasonable to me. Have you tried it out in the context of CoreFX 1.1.0?

AlexGhiondea:
@dagood not yet -- could you do a test run (not in corefx necessarily) to make sure that the code will work as intended?

dagood:
I don’t have time (this is imo where most of the work is) but I can point to https://github.com/dotnet/corefx/pull/17053 as an example of what commands use this in the CoreFX official build and https://github.com/dotnet/core-eng/blob/master/Documentation/Project-Docs/fetch-internal-tooling.md for a bit more detail on the properties. The OptionalToolSourcePassword is a PAT you can generate yourself on https://devdiv.visualstudio.com/_details/security/tokens/Edit with permission to read package feeds. There is a specific PAT we use for the official builds I can help with when it's time for that.

AlexGhiondea:
@dagood is there a local test that can be done just to trigger your code? I want to avoid checking in this to buildtools only to find out that it does not actually work? Is there a simple test that I/you can do to make sure the code inside the target works?

dagood:
If you build the package on your machine, in CoreFX you can [`set BUILDTOOLS_SOURCE=C:\foo\buildtools\bin\packages`](https://github.com/dotnet/corefx/blob/7d9476a04c0198661fece6f93fae0f750c2675fc/init-tools.cmd#L9), modify BuildToolsVersion.txt to be the version of your built package, run init-tools to get the package set up in context, then run `sync`/`msbuild`/etc. commands. I agree it makes sense not to merge this until it's validated.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
```diff
@@ -0,0 +1,135 @@
...
+
+    <Error Text="Optional tooling has not been restored. Use the 'RestoreOptionalToolingPackages' target first."
+           Condition="!Exists('$(OptionalToolingProjectLockJsonPath)')" />
+
+    <ItemGroup>
+      <OptionalToolingTargetMoniker Include=".NETFramework,Version=v4.5" />
+    </ItemGroup>
+
+    <PrereleaseResolveNuGetPackageAssets ProjectLockFile="$(OptionalToolingProjectLockJsonPath)"
+                                         NuGetPackagesDirectory="$(PackagesDir)"
```
dagood: I think this is a good fix. Pretty sure it works in master because newer versions of nuget put the package dir in the lockfile, which means https://github.com/dotnet/buildtools/blob/4f918b5f338f0cac04b29cd4bfe17acfd5d8e1b9/src/Microsoft.DotNet.Build.Tasks/PrereleaseResolveNuGetPackageAssets.cs#L267 can read it. If you see that entry in the 1.1.0 optional tooling lockfile you could try to port that, but the fix you have here seems like a much safer bet.

AlexGhiondea: Yup -- that is exactly why master is fine :). Thanks for checking!

PR #1574, opened 2017-06-19
Enable configuration value to have any case

danmosemsft:
Fix this
```
"C:\git\corefx\src\System.Console\tests\System.Console.Tests.csproj" (default target) (1) ->
(AnnotateProjectReference target) ->
  C:\git\corefx\buildvertical.targets(155,5): error MSB4018: The "FindBestConfigurations" task failed unexpectedly.\r [C:\git\corefx\src\System.Console\tests\System.Console.Tests.csproj]
C:\git\corefx\buildvertical.targets(155,5): error MSB4018: System.ArgumentException: Unknown value 'release' found in configuration 'netstandard-Windows_NT-release'.  Expected property 'ConfigurationGroup' with one of values Debug, Release.\r [C:\git\corefx\src\System.Console\tests\System.Console.Tests.csproj]
C:\git\corefx\buildvertical.targets(155,5): error MSB4018:    at Microsoft.DotNet.Build.Tasks.ConfigurationFactory.ParseConfiguration(String configurationString, Boolean permitUnknownValues)\r 
```

There are other .ToDictionary() 's in there I didn't modify: this works.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Configuration/ConfigurationFactory.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/Configuration/ConfigurationFactory.cs
@@ -44,7 +44,7 @@ namespace Microsoft.DotNet.Build.Tasks
             }
 
             AllPropertyValues = propertyValueGrouping
-                .ToDictionary(g => g.Key, g => g.Single());
+                .ToDictionary(g => g.Key, g => g.Single(), StringComparer.OrdinalIgnoreCase);
 
             PropertyValues = AllPropertyValues.Values
                 .GroupBy(v => v.Property)


danmosemsft:
@weshaggard @ericstj good point -- what do you suggest? That doesn't sound trivial to solve, unless the name of the configuration used exclusively originates in this task, in which case we can have it normalize it.

weshaggard:
>That doesn't sound trivial to solve, unless the name of the configuration used exclusively originates in this task, in which case we can have it normalize it.

I suggest we probably should try and normalize the paths in the bin placing targets so that we guarantee they are always consistent.

danmosemsft:
I may not have time to look at this for a while, but I guess the message could at least indicate that the casing is the problem..

PR #1575, opened 2017-06-19
Reduce spew when running UAP tests

danmosemsft:
Minor changes to reduce the console output when I run UAP tests.

Note that the app launcher looks for a file `\AC\temp\stdout.txt` but our xunit wrapper never writes such a file as it does not redirect console output there. (Now we have a real Console class not a stub, that may be possible)

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -236,7 +236,7 @@
       <TargetExecutableNames Include="xunit.console.netcore.exe"/>
       <TargetExecutableNames Condition="'%(ProjectReference.Filename)' == 'RemoteExecutorConsoleApp'" Include="RemoteExecutorConsoleApp.exe"/>
       <IlcInputFolderContents Include="$(ILCFXInputFolder)/*" />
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -236,7 +236,7 @@
       <TargetExecutableNames Include="xunit.console.netcore.exe"/>
       <TargetExecutableNames Condition="'%(ProjectReference.Filename)' == 'RemoteExecutorConsoleApp'" Include="RemoteExecutorConsoleApp.exe"/>
       <IlcInputFolderContents Include="$(ILCFXInputFolder)/*" />
-      <TestCommandLines Include="mklink /H %(IlcInputFolderContents.Filename)%(IlcInputFolderContents.Extension) $(_Runtime_Path)%(IlcInputFolderContents.Filename)%(IlcInputFolderContents.Extension)" />
+      <TestCommandLines Include="if not exist %(IlcInputFolderContents.Filename)%(IlcInputFolderContents.Extension) mklink /H %(IlcInputFolderContents.Filename)%(IlcInputFolderContents.Extension) $(_Runtime_Path)%(IlcInputFolderContents.Filename)%(IlcInputFolderContents.Extension)" />
```
danmosemsft: Good point..

danmosemsft:
@safern I'm abandoning this: it sounds like you'll be looking at the problem.

PR #1576, opened 2017-06-19
Disable UI Logging in Xunit.Runner.Uap to avoid crashes when the log is super long

safern:
When there is a lot of failures and we try to dump everything to the UI it slows the test execution and it crashes it so we get no testResults.xml

With this change we will see the output in the console at the end of the execution which is usually fast without the UI logging. 

cc: @danmosemsft @tarekgh @joperezr


tarekgh:
cc @AlexGhiondea

tarekgh:
LGTM

PR #1578, opened 2017-06-20
Enable MultiFile ILC runs for Release-x64

joperezr:
cc: @danmosemsft @tijoytom @joshfree @yizhang82 @MattGal @botaberg 

Once this changes get ingested by corefx, we will have multi-file ILC test runs on Release-x64 runs. I have already verified that this works and sent a sample job to Helix  to make sure results are expected.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -224,6 +224,8 @@
             ====================================================== -->
     <PropertyGroup Condition="'$(UseDotNetNativeToolchain)' == 'true'">
       <ILCBuildType Condition="'$(ILCBuildType)' == ''">ret</ILCBuildType>
+      <EnableMultifileTesting Condition="'$(ConfigurationGroup)' == 'Release' And '$(ArchGroup)' == 'x64'">true</EnableMultifileTesting>
+      <_UseSharedAssemblies Condition="'$(EnableMultifileTesting)' == 'true'">-useSharedAssemblies</_UseSharedAssemblies>
       <_TestILCFolder>%RUNTIME_PATH%\TestILC</_TestILCFolder>
       <_Runtime_Path>%RUNTIME_PATH%\ILCInputFolder\</_Runtime_Path>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -240,8 +242,9 @@
       <TestCommandLines Include="copy /y $(_TestILCFolder)\default.rd.xml  %EXECUTION_DIR%" />
       <TestCommandLines Include="rmdir /S /Q %EXECUTION_DIR%int" />
       <TestCommandLines Include="rmdir /S /Q %EXECUTION_DIR%native" />
+      <TestCommandLines Condition="'$(EnableMultifileTesting)' == 'true'" Include="call %RUNTIME_PATH%\TestILC\ilc.exe -buildSharedAssemblies -buildtype $(ILCBuildType) -v diag"/>
```
tijoytom: Are you sure you want to do this of every test ? Ideally  this need to be done after the package is pulled from ILC . We need this happen only once , buildSharedAssemblies can take 2 - 4  minutes depending on the machine.

tijoytom: Also remove the verbose logging , it will mess up the console output

yizhang82: if we only do this once, it should be fine to have verbose logging there.

joperezr: >Are you sure you want to do this of every test 

It's not about whether we want it or not, when we pull the ILC toolchain we don't yet have the corefx runtime built so we can't do it at that time, and there  is no good central place to put it in order to run it once the runtime is there. It is definitelly something we can do eventually with some infrastructure changes, but this should be fine for now IMO.

joperezr: I chatted with @weshaggard offline and there is a better way of doing this and only doing it once, so I'll close this and send the new PR soon.

PR #1580, opened 2017-06-21
Adding option to enable  using shared assemblies for multi-file tests

joperezr:
cc: @danmosemsft @safern @tijoytom


joperezr:
After this is in and ingested in corefx, I will set `EnableMultiFileILCTests` on that side for Release-x64 configurations.

joperezr:
This PR replaces #1578

tijoytom:
@joperezr I might be missing something here , where is the buildSharedAssembly part ?

joperezr:
that will be in corefx so that we only do it once as you suggested in the other PR. But I need this in first, so that I can consume it in corefx PR.

tijoytom:
@joperezr thanks , makes sense . Thanks for doing this ... @botaberg We should do this for x86 too ( leave out ARM for now since we there are lot for failure in ARM single file now)

PR #1583, opened 2017-06-24
add UapNotUapAot to TFM enum

danmosemsft:
three problems to solve:

1) Making this work ` [SkipOnTargetFramework(~TargetFrameworkMonikers.UapAot)]`

- `SkipOnTargetFrameworkDiscoverer` determines the test has traits `nonXXXXtests` where XXX is every TFM for which `(~TargetFrameworkMonikers.UapAot).HasFlag(TFM)` is true.
- Because `TFM.Uap = TFM.UapAot | 0x2000`, this is NOT true for TFM.Uap (the negated mask of UapAot does not HasFlag of TFM.Uap because it's missing the UapAot bit out of TFM.Uap). So this test is determined to have all nonXXXtests traits except `nonUapaottests` and `nonUaptests`.
- When running tests, `tests.targets` passes `-notrait nonTARGETGROUPtests` to xunit. In the case of a UAP test run, this would be `-notrait nonuaptests`. This does not exclude the test marked with the attribute above, because the test is NOT considered to have trait `nonuaptests`.

2) Making tihs work `[SkipOnTargetFramework(TargetFrameworkMonikers.UapNotUapAot)]` .. should skip specifically Uap and not Uapaot.

3) Making this work  `[SkipOnTargetFramework(~TargetFrameworkMonikers.Uap)]` .. should skip everything except uap, including skipping uapaot.

Proposal: Add `UapNotUapAot = 0x2000,` and special case that bit so it yields `nonUapTests`. This should fix both problems (I think? my head hurts.)

This is also a lesson in why one should never use negated constants or flags in any situation ever.

### DIFF
--- a/src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
+++ b/src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
@@ -53,7 +53,7 @@ namespace Xunit.NetCore.Extensions
                     yield return new KeyValuePair<string, string>(XunitConstants.Category, XunitConstants.NonNetfxTest);
                 if (frameworks.HasFlag(TargetFrameworkMonikers.Netcoreapp))
                     yield return new KeyValuePair<string, string>(XunitConstants.Category, XunitConstants.NonNetcoreappTest);
-                if (frameworks.HasFlag(TargetFrameworkMonikers.Uap))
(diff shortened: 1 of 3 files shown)


danmosemsft:
@safern @stephentoub I'm inclined to think it's not worth making a change here..

danmosemsft:
@hughbe

safern:
I'll update the xunit extensions once the build that was triggered with this merge is done

PR #1592, opened 2017-07-04
Look for type in Corefx.Private.TestUtilities in Conditional tests

hughbe:
Previously, we'd only look in the current test assembly. This blocks moving PlatformDetection to CoreFx.Private.TestUtilities.

### DIFF
--- a/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
+++ b/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using Xunit.Abstractions;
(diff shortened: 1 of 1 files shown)


src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
```diff
@@ -54,11 +55,30 @@ internal class ConditionalTestDiscoverer
...
+                    ITypeInfo type = GetTypeFromSymbol(testMethod.TestClass.Class.Assembly, symbols[0]);
                     if (type != null)
                     {
                         declaringType = type.ToRuntimeType();
                     }
+                    else
+                    {
+                        try
+                        {
+                            var assemblyName = new AssemblyName("CoreFx.Private.TestUtilities");
```
weshaggard: I don't like us hardcoding the name of this assembly here. We need a more general mechanism for finding these methods. Perhaps we can look at all assemblies referenced or we add another options that allows folks to put the assembly name as part of the string name.

tarekgh:
@hughbe  I already have a PR https://github.com/dotnet/buildtools/pull/1629 please have a look and let me know if we can close this one

hughbe:
that's a lot better. thanks!

PR #1596, opened 2017-07-06
Update pinvoke list

danmosemsft:
Refreshed from the source to pick up some additions.
I don't know why there's removes: I don't think we care about any of those.


danmosemsft:
I'm just going to go ahead and merge: @pjanotti may update it again.

PR #1599, opened 2017-07-07
Add MakePri to buildtools to use it for creating resources.pri for Uap tests

safern:
Contributes: https://github.com/dotnet/corefx/issues/21692

cc: @tarekgh @danmosemsft @weshaggard @joperezr


tarekgh:
:+1:

tarekgh:
Is it x86 or amd64 version? and from where did you pick the file?

joperezr:
I would instead have it in a subfolder called "net46". Reason is that all tools in buildtools package's root are .NetCore exes and desktop exes are located in the net46 subfolder.

joperezr:
@tarekgh it's the x64 version

tarekgh:
> it's the x64 version

I would prefer to use x86 version even if we are building on x64. this will make our life easier if we decided at any point to build on x86 platform

safern:
> Is it x86 or amd64 version? and from where did you pick the file?

It is the x86 version. I got it from my installed Windows SDK

tarekgh:
>  I got it from my installed Windows SDK

just ensure this is latest SDK so we'll have the most updated tool version. thanks for getting this in.

safern:
Ok. I will ensure it is the latest and change the folder name to "net46" :)

safern:
Updated it to be latest version -> 10.0.15063.0 and also renamed to net46 it's folder

weshaggard:
Is there any reason we needed to check-in the binary? We generally avoid checking them in and instead put them into a nuget package we can restore from.

Also for future reference it would be good to put the version number and where you got them either in the commit message or a readme that lives next to the file.

safern:
> We generally avoid checking them in and instead put them into a nuget package we can restore from.

But I think we talked about didn't wanting to put it into a nuget because it was a Windows SDK binary from what I remember. If we rather have it in a nuget package, I can go ahead and do the change.

weshaggard:
>But I think we talked about didn't wanting to put it into a nuget because it was a Windows SDK binary 

I don't see any difference between that and checking it into a public repo.

safern:
> I don't see any difference between that and checking it into a public repo.

Makes sense. Do you rather have it in a nuget package to restore from?

weshaggard:
Yes that is the approach we have been generally taking to avoid checking in binaries into our git repos.

tarekgh:
@safern please log future issue to track moving this file to a nuget package.

PR #1601, opened 2017-07-10
Copy resx files in resources.targets to AssemblyName.resw file

safern:
Contributes to: https://github.com/dotnet/corefx/issues/21692

This is needed as MakePri.exe doesn't support .resx files. Also they have to be called as the assembly so that when we create the .pri file the resources are mapped correctly.

cc: @tarekgh @weshaggard @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
@@ -11,7 +11,6 @@
     
     <!-- For UAP, we need to bin place the resources as resw files to create the runner's pri file. -->
     <ResourcesFolderPath Condition="'$(ResourcesFolderPath)' == ''">$(RuntimePath)resw</ResourcesFolderPath>
-    <CommonPriFile Condition="'$(CommonPriFile)' == ''">$(ResourcesFolderPath)/resources.pri</CommonPriFile>
 
     <!-- A number of the imports below depend on these default properties -->
     <AssemblyVersion Condition="'$(AssemblyVersion)'==''">999.999.999.999</AssemblyVersion>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
@@ -8,6 +8,9 @@
 
   <PropertyGroup>
     <BuildToolsTaskDir Condition="'$(BuildToolsTaskDir)' == ''">$(ToolsDir)</BuildToolsTaskDir>
+    
+    <!-- For UAP, we need to bin place the resources as resw files to create the runner's pri file. -->
+    <ResourcesFolderPath Condition="'$(ResourcesFolderPath)' == ''">$(RuntimePath)resw</ResourcesFolderPath>
 
     <!-- A number of the imports below depend on these default properties -->
     <AssemblyVersion Condition="'$(AssemblyVersion)'==''">999.999.999.999</AssemblyVersion>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
@@ -11,7 +11,7 @@
     
     <!-- For UAP, we need to bin place the resources as resw files to create the runner's pri file. -->
     <ResourcesFolderPath Condition="'$(ResourcesFolderPath)' == ''">$(RuntimePath)resw</ResourcesFolderPath>
-    <CommonPriFile Condition="'$(CommonPriFile)' == ''">$(RuntimePath)resources.pri</CommonPriFile>
+    <CommonPriFile Condition="'$(CommonPriFile)' == ''">$(ResourcesFolderPath)/resources.pri</CommonPriFile>
 
     <!-- A number of the imports below depend on these default properties -->
     <AssemblyVersion Condition="'$(AssemblyVersion)'==''">999.999.999.999</AssemblyVersion>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -20,20 +20,25 @@
           GenerateResourcesSource;
           $(CompileDependsOn);
       </CompileDependsOn>
+      <ResourcesFolderPath Condition="'$(BuildingUAPVertical)' == 'true'">$(RuntimePath)resw</ResourcesFolderPath>
```
joperezr: I thought you said that you would need this property in a different targets file. If so, this should probably be in Build.Common.props instead

safern: I tried adding it to Build.Common.props but by the time resources.targets is called it is not defined yet.

joperezr: We should then find a different place but I don't think this is the right one. Also, if we want to put it here, we should at least make this overridable by checking if it's set already.

safern: We could add it to dir.props then, right?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
```diff
@@ -8,6 +8,9 @@
 
   <PropertyGroup>
     <BuildToolsTaskDir Condition="'$(BuildToolsTaskDir)' == ''">$(ToolsDir)</BuildToolsTaskDir>
+    
+    <!-- For UAP, we need to bin place the resources as resw files to create the runner's pri file. -->
+    <ResourcesFolderPath Condition="'$(ResourcesFolderPath)' == ''">$(RuntimePath)resw</ResourcesFolderPath>
```
weshaggard: This looks like it will cause them to be bin-placed for all configurations. Is that correct?

safern: No, this is only the property definition. They are being bin-placed here: https://github.com/dotnet/buildtools/pull/1601/files#diff-dccabf83e08d3281efb7ebb5798da47aR35

And that is conditoned to UAPVertical.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -26,14 +26,18 @@
...
-    
 
     <GenerateResourcesCode
         ResxFilePath="$(StringResourcesPath)" 
         OutputSourceFilePath="$(IntermediateResOutputFileFullPath)" 
         AssemblyName="$(AssemblyName)" />
 
+    <Copy Condition="'$(BuildingUAPVertical)' == 'true'"
+          SourceFiles="$(StringResourcesPath)"
+          DestinationFiles="$(ResourcesFolderPath)\$(AssemblyName).resw"
```
weshaggard: We are assuming that resx files are exactly the same format as resw files is that true for our usages?

tarekgh: @safern already tried all resx files we have in corefx and worked after renaming it to resw. so far this is good enough till we find a case break this assumption.

safern: I just tried this locally with makepri and used the produced resources.pri inside the runner and ran a library that loads a string from ResourceManager and it works perfectly.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
```diff
@@ -8,6 +8,10 @@
...
+    <CommonPriFile Condition="'$(CommonPriFile)' == ''">$(RuntimePath)resources.pri</CommonPriFile>
```
tarekgh: can we store this file under sub-folder like "PRI" or even with the rest of the rew files?

safern: yes we can do that. I think the resw folder would be ok

safern: Fixed

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -26,14 +26,21 @@
...
         ResxFilePath="$(StringResourcesPath)" 
         OutputSourceFilePath="$(IntermediateResOutputFileFullPath)" 
         AssemblyName="$(AssemblyName)" />
 
+    <Copy Condition="'$(BuildingUAPVertical)' == 'true'"
+          SourceFiles="$(StringResourcesPath)"
+          DestinationFiles="$(ResourcesFolderPath)\FxResources.$(AssemblyName).SR.resw"
+          SkipUnchangedFiles="true" />
+
+    <Delete Condition="'$(BuildingUAPVertical)' == 'true' AND Exists('$(CommonPriFile)')"
```
weshaggard: I'm not sure it is a good idea to do this for every project given the common file is shared and we build in parallel

tarekgh: what you suggest @weshaggard?

weshaggard: Where are we generating this file? I would suggest we always regenerate if we see any new resw files that are newer. So a target that has inputs of *.SR.resw files.

safern: Fixed. I didn't remove the CommonPriFile property definition as we will be using it later in test.targets. Should I move it to dir.props instead?

safern: At the end I deleted the property definition as it is only needed in test.targets so it can be defined there

tarekgh:
LGTM

weshaggard:
Are you going to make the changes to test.targets in another PR?

safern:
> Are you going to make the changes to test.targets in another PR?

Yes as I'm writing a task that does all the required work to create the common.pri file. So I will submit a PR with that task and the changes to tests.targets

safern:
Thanks for the reviews :)

PR #1605, opened 2017-07-11
Convert Portable to Windows PDBs during symbol archive targets

dagood:
Adds a task and target that are used in the archive process to generate and then publish Windows PDBs for each Portable PDB. I've verified this locally, and I'm working on verifying it with a fake full symbol publish build def now.

I've already mirrored `Microsoft.DiaSymReader.Converter` and `Microsoft.DiaSymReader.Native` on the dotnet-buildtools feed.

There's a workaround in `Microsoft.DotNet.Build.Tasks.net45.csproj` to use `PrereleaseResolveNuGetPackageAssets` to grab the x86 natives. Without it, only the x64 natives end up in the buildtools package, but we run with x86 msbuild. I didn't want to just switch the project to x86 in case there's a compatibility reason for it being x64 now.

Currently I chose not to support this task on .NET Core. The task does compile and ship on Core, and when the dependencies are present, it works fine. But there was no clear way to gather and pack up all the dependencies needed, and full desktop framework is the only scenario. There isn't a whole lot to figure out (or work around) if it's useful to support Core in the future.

https://github.com/dotnet/core-eng/issues/698

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
@@ -39,4 +39,25 @@
     <TargetingPackReference Include="System.Xml" />
     <TargetingPackReference Include="System.Xml.Linq" />
   </ItemGroup>
+  <Target Name="AfterBuild">
+    <!-- Manually binplace DiaSymReader runtimes for win7-x86, which the official builds use. -->
+    <PrereleaseResolveNuGetPackageAssets AllowFallbackOnTargetSelection="true"
+                                         IncludeFrameworkReferences="false"
+                                         NuGetPackagesDirectory="$(PackagesDir)"
+                                         RuntimeIdentifier="win7-x86"
+                                         ProjectLanguage="$(Language)"
+                                         ProjectLockFile="project.lock.json"
+                                         TargetMonikers="$(NuGetTargetMoniker)">
+      <Output TaskParameter="ResolvedCopyLocalItems"
+              ItemName="_ReferenceCopyLocalPathsFromDependencies" />
+    </PrereleaseResolveNuGetPackageAssets>
+
+    <ItemGroup>
+      <_DiaSymReaderNativeCopyLocalItems Include='@(_ReferenceCopyLocalPathsFromDependencies)'
+                                         Condition="'%(NuGetPackageId)' == 'Microsoft.DiaSymReader.Native'" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(_DiaSymReaderNativeCopyLocalItems)"
+          DestinationFolder="$(OutputPath)" />
+  </Target>
 </Project>
--- a/src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
@@ -40,7 +40,10 @@
     <TargetingPackReference Include="System.Xml.Linq" />
   </ItemGroup>
   <Target Name="AfterBuild">
-    <!-- Manually binplace DiaSymReader runtimes for win7-x86, which the official builds use. -->
+    <!--
+      Manually binplace DiaSymReader runtimes for win7-x86, which the official builds use.
+      This workaround should be removed, see https://github.com/dotnet/buildtools/issues/1608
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
```diff
@@ -39,4 +39,25 @@
     <TargetingPackReference Include="System.Xml" />
     <TargetingPackReference Include="System.Xml.Linq" />
   </ItemGroup>
+  <Target Name="AfterBuild">
+    <!-- Manually binplace DiaSymReader runtimes for win7-x86, which the official builds use. -->
```
crummel: Is this worth preemptively logging an issue for?  Then the comment can refer back to it and we have an idea of the work that has to happen to remove the hack.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
```diff
@@ -403,4 +448,24 @@
...
+    @(PortableFileToConvert)
+      * ItemSpec: The DLL file associated with a Portable PDB to convert.
+      * PdbPath: The path to the Portable PDB file to convert.
+      * TargetPath: The output path for the generated Windows PDB. Full filename.
+    @(ConversionOptions) [optional]
+      * ItemSpec: An entry in the Microsoft.DiaSymReader.Tools.PdbConversionOptions flags enum to
+        use for all conversions performed.
+  -->
+  <Target Name="CreateWindowsPdbsFromPortablePdbs">
+    <Error Text="BuildTools does not support Portable PDB conversion to Windows PDB in .NET Core. Run msbuild using the desktop framework."
```
crummel: MSBuildRuntimeType is `core` for Linux, I'm guessing?  Does this break there or does `SubmitSymbolsRequest` never get called?  I'd be fine with skip and warn instead of error out if we're unsure about effects on client repos.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
dagood: Good idea, also opens it up for discussion about worth.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
dagood: It's `core` when running on .NET Core, even on Windows. Running the task fails to do some interop stuff because the DLLs it needs aren't present.

Here's the error:

```
The "ConvertPdbsToPortablePdbs" task failed unexpectedly.\r [C:\git\corefx\build.proj]
System.InvalidCastException: Unable to cast COM object of type 'System.__ComObject' to interface type 'Microsoft.DiaSymReader.ISymUnmanagedWriter8'. This operation failed because the QueryInterface call on the COM component for the interface with IID '{5BA52F3B-6BF8-40FC-B476-D39C529B331E}' failed due to the following error: No such interface supported (Exception from HRESULT: 0x80004002 (E_NOINTERFACE)).\r [C:\git\corefx\build.proj]
at Microsoft.DiaSymReader.SymReaderFactory.CreateWindowsPdbWriter(Object pdbStream, Object metadataProvider) in E:\A\_work\7\s\src\Microsoft.DiaSymReader.Converter\SymReader\SymReaderFactory.cs:line 89\r [C:\git\corefx\build.proj]
...
```

I'm not really a fan of a warning or silent continue: I feel like I'd rather make sure people confirm for certain that they don't want to produce Windows PDBs. A concern would be someone trying to change symbol publishing to run on Core without realizing that this is one of the effects.

I'll add a property for skipping this process entirely, though. I suppose you could override the target, but a property would be clearer.

crummel: I like the property for skipping it, that sounds good.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
dagood: Oh, I mistook what this was a comment about. I'll be adding issues for this hack and Core support.

dagood:
Tested it out in VSTS on a CoreFX package set. We still build Windows PDBs on Windows, which it successfully skipped. It converted the Portable PDBs for Linux and OSX to Windows PDBs. Merging.

PR #1606, opened 2017-07-11
Add Full GIT URL to version info.

vancem:
Previously we had the commit ID in the file version information.  However we have
several Repos so this is needlessly ambigous.  Add the full url.   Now when you do a
filever -v you get something like ths

ProductVersion   4.6.25511.0 built by: vancem-VANCEM1. SrcCode: https://github.com/dotnet/coreclr/tree/de5f92822dbac17160da9e796e21918d2b812e38

Which unambiguously identifies vi the source code.  

@ericstj @dagood  

I have tested this by getting the coreclr repo to use the new build tools and seen that it 
produces the above URL.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -292,7 +292,7 @@
     
     <!-- ************************* -->
 
-    <!-- SrcCodeRef is something that identifies the source code along with at tag (e.g. SrcCode: http//.....) -->
+    <!-- SrcCodeRef is something that identifies the source code along with at tag (e.g. SrcCode: https://.....) -->
     <!-- We do the best we can, and fall back as necessary -->
     <PropertyGroup Condition="'$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A'">
       <SrcCodeRef>SrcCode%3A $(GitHubRepositoryUrl)/tree/$(LatestCommit)</SrcCodeRef>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -286,7 +286,7 @@
     </PropertyGroup>
 
     <!-- The GitHubRepositoryUrl is formed out of the simple GitHubRepositoryName (e.g. coreclr or corefx) -->
-    <PropertyGroup Condition="'$(GitHubOwnerName)' != '' AND '$(GitHubRepositoryName)' != ''" AND '$(GitHubRepositoryUrl)' == ''">
+    <PropertyGroup Condition="'$(GitHubOwnerName)' != '' AND '$(GitHubRepositoryName)' != '' AND '$(GitHubRepositoryUrl)' == ''">
       <GitHubRepositoryUrl>https:/github.com/$(GitHubOwnerName)/$(GitHubRepositoryName)</GitHubRepositoryUrl>
     </PropertyGroup>
     
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -294,11 +294,11 @@
 
     <!-- SrcCodeRef is something that identifies the source code along with at tag (e.g. SrcCode: http//.....) -->
     <!-- We do the best we can, and fall back as necessary -->
-    <PropertyGroup Condition="'$(GitHubRepositoryUrl)' != ''">
+    <PropertyGroup Condition="'$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A'">
       <SrcCodeRef>SrcCode%3A $(GitHubRepositoryUrl)/tree/$(LatestCommit)</SrcCodeRef>
     </PropertyGroup>
     <!-- If we dont have the GitHubRepositoryUrl at least put the commit ID (which is N/A if that does not exist) -->
-    <PropertyGroup Condition="'$(GitHubRepositoryUrl)' == ''">
+    <PropertyGroup Condition="'$(GitHubRepositoryUrl)' == '' OR '$(LatestCommit)' == 'N/A'">
       <SrcCodeRef>Commit%3A $(LatestCommit)</SrcCodeRef>
     </PropertyGroup>
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -167,11 +167,11 @@
       <NativeVersionLines Include="#undef VER_PRODUCTVERSION" />
       <NativeVersionLines Include="#define VER_PRODUCTVERSION          $(MajorVersion),$(MinorVersion),$(BuildNumberMajor),$(BuildNumberMinor)" />
       <NativeVersionLines Include="#undef VER_PRODUCTVERSION_STR" />
-      <NativeVersionLines Include="#define VER_PRODUCTVERSION_STR      &quot;$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)$(BuiltByString). Commit Hash%3A $(LatestCommit)&quot;" />
+      <NativeVersionLines Include="#define VER_PRODUCTVERSION_STR      &quot;$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)$(BuiltByString). $(SrcCodeRef)&quot;" />
       <NativeVersionLines Include="#undef VER_FILEVERSION" />
       <NativeVersionLines Include="#define VER_FILEVERSION             $(MajorVersion),$(MinorVersion),$(BuildNumberMajor),$(BuildNumberMinor)" />
       <NativeVersionLines Include="#undef VER_FILEVERSION_STR" />
-      <NativeVersionLines Include="#define VER_FILEVERSION_STR         &quot;$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)$(BuiltByString). Commit Hash%3A $(LatestCommit)&quot;" />
+      <NativeVersionLines Include="#define VER_FILEVERSION_STR         &quot;$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)$(BuiltByString). $(SrcCodeRef)&quot;" />
       <NativeVersionLines Include="#ifndef VER_LEGALCOPYRIGHT_STR" />
       <NativeVersionLines Include="#define VER_LEGALCOPYRIGHT_STR      &quot;\xa9 Microsoft Corporation.  All rights reserved.&quot;" />
       <NativeVersionLines Include="#endif" />
@@ -207,7 +207,7 @@
 
     <ItemGroup>
       <SourceFileLines />
-      <SourceFileLines Include="static char sccsid%5B%5D %5F%5Fattribute%5F%5F%28%28used%29%29 %3D %22%40%28%23%29Version $(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)$(BuiltByString). Commit Hash%3A $(LatestCommit)%22%3B" />
+      <SourceFileLines Include="static char sccsid%5B%5D %5F%5Fattribute%5F%5F%28%28used%29%29 %3D %22%40%28%23%29Version $(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)$(BuiltByString). $(SrcCodeRef)%22%3B" />
       <!-- Since this is a source file, compiler will complain if there is no new line at end of file, so adding one bellow. -->
       <SourceFileLines Include=" " />
     </ItemGroup>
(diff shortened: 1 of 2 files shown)


dagood:
This won't work in some scenarios:

 * Official builds clone from VSTS, not GitHub, to build. As written, this would actually leak the PAT.
 * Locally, I don't have a `origin` remote: I've named mine `dotnet` and `dagood`.

(This would also probably require a change in BfS to keep out the git dependency, /cc @ellismg.)

dagood:
Instead, we could set this as a pipebuild parameter and consume it during the build, or generate a GitHub URL based on the project name.

vancem:
What is your suggestion on the best way to fix this?   Note that it does not have to work in all cases (I will fix it so that it fails gracefully to what it did before if it can't get a URL prefix).  It only has to work for our actual build machines (local builds it will just fall back).    

Note that we will have a similar problem when we want to embed source code URLs into the PDBS.   I was assuming that all our repos (that do official builds) were set up the same way to fetch, but I did not anticipate that we fetched from a cloned VSTS URL. 

I may be best if we simply have canonical file (e.g. git-remote.txt) file that lives in the repo base which holds this canonical URL.   If the file does not exist, we give up and just give the commit ID.  

However I would prefer something where we don't have to change every repo we use.   I am assuming there is an obvious mapping from the VSTS url to the GitHub url.     If FOR BUILD SERVERS we use a consistent name for the origin (again we can bail on the ULR if don't find the origin name), and then map that to the GIT url it would solve the problem. 

I think the git dependency is OK (since it is no worse than before...).  

What do you suggest?   We will need some solution for automatic source code fetching.  

Vance

dagood:
We already have a `PB_GitHubRepositoryName` VSTS variable that we pass through our build definitions for other steps--I'd say it makes sense to use that and generate a GitHub url. We could pass that into the build as a msbuild property called `GitHubRepositoryName`. The value of this variable is defined per product, but it's in infrastructure outside the git repo itself (and it's already been there for a while). The strings are like `corefx`, `coreclr`, `wcf`.

When we source-link to GitHub in our PDBs, I think the msbuild property explicit approach will work just as well there.

I agree that falling back to commit hash is fine in any nonstandard case.

For security reasons, I really don't want to go down the path of trying to map the git remote. I'm not confident that there is a centralized point of control for all of the builds this code may end up applying to, and I don't want to surprise anyone.

A "soft" git dependency is fine--I was just noting that BfS may need to react.

vancem:
> We already have a PB_GitHubRepositoryName VSTS variable that we pass through our build definitions for other steps--I'd say it makes sense to use that and generate a GitHub url. We could pass that into the build as a msbuild property called GitHubRepositoryName

When you say VSTS variable, do you mean that it is an environment variable (that is I can say $(PB_GitHubRepositoryName) in the versioning.targets file and if it is not set just give up?   Also it sounds like the repository name is just a word like 'coreclr' which means I need to assume https://github.com/dotnet/ prefix (do we anticipate use this infrastructure for things outside the 'dotnet' user?).    If all that sounds OK, I will update PR).  

Note that it is not clear how to test this (except by explicitly setting PB_GitHubRepositoryName as an environment variable).

ellismg:
From a build from source perspective, as long as there is a way to set some property to give the same information (as Vance has in his PR) we should be okay.  We'll just set that property in the git-less build.

dagood:
> When you say VSTS variable, do you mean that it is an environment variable (that is I can say $(PB_GitHubRepositoryName) in the versioning.targets file and if it is not set just give up?

I mean a variable [in the VSTS build definitions](https://github.com/dotnet/corefx/blob/9eafffc6054c8f039d788465621bcfd49a379ca1/buildpipeline/DotNet-CoreFx-Trusted-Windows-NoTest.json#L302). Technically VSTS does inject these variables into the environment, but we avoid making dependencies on environment variables in the repos. Instead we prefer to link them up explicitly in the msbuild context by calling msbuild with `/p:GitHubRepositoryName=$(PB_GitHubRepositoryName)`, which is also done in that same VSTS definition json file. (Then you use it here as `GitHubRepositoryName`.)

Falling back to commit hash sounds fine to me.

> Also it sounds like the repository name is just a word like 'coreclr' which means I need to assume https://github.com/dotnet/ prefix (do we anticipate use this infrastructure for things outside the 'dotnet' user?).

That sounds like a good default to me, and it seems like it would be easy to configure.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -264,6 +264,14 @@
       Condition="'$(SkipVersionGeneration)' != 'true' AND '$(VersionPropsImported)' != 'true'">
     <!-- ############################### -->
 
+    <!-- Get the Url of the Git Repository -->
+    <Exec Command="git config --get remote.origin.url 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true" Condition="'$(GitRepoUrlRaw)' == ''">
```
weshaggard: You should consider using `git ls-remote --get-url` instead which will give you the url for the first remote setup. That at least eliminates your dependency on the name "origin".

weshaggard:
I think it is reasonable to make a best case effort to read the url if the property isn't set but that does mean we do have the potential security risk of leaking our PAT. @dagood are you sure it is in the url itself?

dagood:
> dagood are you sure it is in the url itself?

Yep, confirmed locally that the way we clone in CoreFX builds then running that git command returns the VSTS url with the PAT.

vancem:
First, let me say that I don't understand much of anything about the build pipeline works (I am slowly learning).   In particular I don't know what PAT stands for, but presumably it is a URL that has some information that we don't wish to be completely public.  

Second, ideally I wanted something that 'just worked' on any of our repositories where the build tools existed (that is you don't need to make modifications to each repo to 'turn it on'.     That is what made using GIT commands appealing (since the existing code already had a GIT dependency).  However it seems that the GIT repository that the build machines use is not the public GIT repository (not surprising) and thus this simple does not give us what we want.   

Because if we want something SIMPLE, it is easy enough 'RepoURL.txt' file and simply have MSBuild fetch the prefix from this.  This is easy and is probably OK that it is obscure because it should not change (ever frankly), but it does require a change to be made to every repository to 'opt in', which is unfortunate.  

Davis suggested using the PB_GitHubRepositoryName property  (which is just the last component of the url (like coreclr or corfx but we can simply concatenated a http://github.com/dotnet in front)  but having poked around a bit I am concerned that this may be fragile (for CoreCLR this seems to be named GitHubRepositoryName not PB_GitHubRepositoryName).  It is hard for me to tell if this is uniformly set for all repos where we use the buildtools, and I don't know how to test this stuff.   (and it sounds you also need to plumb t into msbuild).  However if someone to knows the system better is confident and gives me guidance I am happy to do that.  

I also note that there are variables like [PB_VsoCorefxGitUrl](https://github.com/dotnet/corefx/blob/9eafffc6054c8f039d788465621bcfd49a379ca1/buildpipeline/DotNet-CoreFx-Trusted-Windows-NoTest.json#L346) that are exactly what we want, but this variable is not uniform (there is a similar value in coreCLR but it is called GitHubRepo).  

At this point, I think there are two basic strategies:

1. Variable based:  fix things so we have a uniform variable that has been plumed through that holds the simple name of repository (e.g PB_GitHubRepositoryName) and use that.  
2. Use git ls-remote --get-url, to get the URL from GIT, but only use it to fetch the simple respository name (e.g. get the corefx name and then form the url using http://github/dotnet/corefx).   This avoids issues with VSO and information leakage. 

What I like about 2, is that I believe I can implement it right now and I believe it will be at least as robust as anything else (I am happy only do only if PB_GitHubRepositoryName is not defined, and use that to define it).   

I will pursue option (2) for the moment.  If people think it is overkill (PB_GitHubRepositoryName should already be uniformly set), then I will modify the PR.

weshaggard:
>Use git ls-remote --get-url, to get the URL from GIT, but only use it to fetch the simple respository name

How do you plan to fetch just the simple from that command? It gives the full url which will be formatted differently between official and not builds.

> (e.g. get the corefx name and then form the url using http://github/dotnet/corefx).

If we do end up with a solution that appends the github URL please make sure it is https so we don't introduce a potential security hole given we don't know how folks might use this information. 

>I will pursue option (2) for the moment. If people think it is overkill (PB_GitHubRepositoryName should already be uniformly set), then I will modify the PR.

Regardless of the approach we need to have a property folks can set to override as we are not always going to be in the context of git (i.e. the source tar-balls). Whatever property we choose should not begin with PB_ as that PB stands for PipeBuild and indicates that property is private to the build pipeline and shouldn't be used directly in our BuildTools. So if we do use one it should be something like GitHubRepositoryName.

vancem:
> How do you plan to fetch just the simple name from that command? It gives the full url which will be formatted differently between official and not builds.

I will use MSBuild functions to extract the string after the last / in the URL.  

I am happy to call it GitHubRepositoryName, which seems to already have this meaning as the 'simple' name of the repository (not the full URL)

weshaggard:
The official url looks like something like:
`https://$(PB_VsoAccountName):$(VsoPassword)@devdiv.visualstudio.com/DevDiv/_git/$(PB_VsoRepositoryName)/` so it will not have the name corefx anywhere in that url and your parsing will essentially be empty for that url. So perhaps the full solution sill requires the property to be set. 

To answer your other question PAT is personally access token which is essentially the VsoPassword in that url.

vancem:
I would have expected $(PB_VsoRepositoryName) to expand to something that had 'corefx' as its last component.    Is that not true?

However as a first step, I will NOT introduce the complexity of using GIT to try to fall back and simply assume that GitHubRepositoryName is defined (and it is what we want.   We seem to already use it, so I suspect it has already been plumbed so lets see how far that gets us.

weshaggard:
>I would have expected $(PB_VsoRepositoryName) to expand to something that had 'corefx' as its last component. Is that not true?

That expands to DotNet-CoreFX-Trusted for our official builds and it looks like it has a trailing slash (not sure if that is needed for anything). 

>However as a first step, I will NOT introduce the complexity of using GIT to try to fall back and simply assume that GitHubRepositoryName is defined (and it is what we want. We seem to already use it, so I suspect it has already been plumbed so lets see how far that gets us.

I suspect that is probably the best and if we find places it isn't plumbed we can correctly plumb it.

vancem:
OK I have changed things so that if you if GitHubRepositoryName is defined then it creates the URL https:/github.com/dotnet/$(GitHubRepositoryName)/tree/$(CommitID)  for the source code.    It will fall back gracefully to what it did before if it can't do this. 

Note that normal dev builds do NOT have GitHubRepositoryName.   We really only need this for official builds and I suspect (but do not know) that it might already be defined for official builds (although it may not be uniform across repositories).  

Can someone offer guidance about how to insure that this variable is defined?

Thanks

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -278,6 +278,30 @@
...
+
+    <!-- The GitHubOwnerName is the name of the entity that 'owns' a particular github repository (e.g the 'dotnet in https://github/dotnet/coreclr) -->
+    <!-- we assume by default that these build tools are used by the 'dotnet' group.  This can be overridden. -->
+    <PropertyGroup Condition="'$(GitHubOwnerName)' == ''">
+      <GitHubOwnerName>dotnet</GitHubOwnerName>
+    </PropertyGroup>
+
+    <!-- The GitHubRepositoryUrl is formed out of the simple GitHubRepositoryName (e.g. coreclr or corefx) -->
+    <PropertyGroup Condition="'$(GitHubOwnerName)' != '' AND '$(GitHubRepositoryName)' != '' AND '$(GitHubRepositoryUrl)' == ''">
+      <GitHubRepositoryUrl>https:/github.com/$(GitHubOwnerName)/$(GitHubRepositoryName)</GitHubRepositoryUrl>
```
dagood: `https:/` is missing a slash.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -278,6 +278,30 @@
...
+    </PropertyGroup>
+    
+    <!-- ************************* -->
+
+    <!-- SrcCodeRef is something that identifies the source code along with at tag (e.g. SrcCode: http//.....) -->
```
dagood: `http//` should probably be `https://`

dagood:
> Can someone offer guidance about how to insure that this variable is defined?

First, make sure `/p:GitHubRepositoryName=$(PB_GitHubRepositoryName)` is passed to the step that will call this target. (If the repo style avoids depending on the environment from inside msbuild.)

Then, make sure it's passed into the build leg. One way is `pipelines.json`: currently in CoreFX/CoreCLR it is defined in the publish pipeline, which means it's only available in the publish build legs. Unfortunately there's no way to define a "all pipelines" global variable in `pipelines.json`, so you probably want to edit the VSTS definition for that repo to pass it in as a global pipeline variable there.

---

You could also define the `GitHubRepositoryName` msbuild variable in dir.props, come to think of it. This would be similar to your standalone file suggestion, but it's a more typical place and you can put in a comment about it.

vancem:
> You could also define the GitHubRepositoryName msbuild property in dir.props, come to think of it. This would be similar to your standalone file suggestion, but it's a more typical place and you can put in a comment about it.

I think this is probably for the best.   The less 'build system magic' we have the better and it seems that deriving this name from existing variables is not going to work.    Thus here is the plan

1. Check in what we have here.  Basically if someone sets GitHubRepositoryName.
2. I will be making a set of changes to the repos to set this variable in the dir.props files.

So are we happy with the changes here?   I have tested this I will need an official build of this to do the necessary modification to use the new build tools.

PR #1611, opened 2017-07-14
Add task to merge the framework resources with the UAP runner resources

safern:
Contributes: https://github.com/dotnet/corefx/issues/21692

This is the work required to get all the resources from the framework and create a resources.pri that merges the framework resources and the test resources. This handles String.resx files and EmbeddedResources that are not called String.resx.

Also in order for this to be safe in parallel there is a change that needs to happen in corefx but I will submit the PR of that changes after this is merged. The change is in tests.builds by adding:

```xml
<PropertyGroup>
    <TraversalBuildDependsOn>
      MakeCommonResourcesPriFile;
      $(TraversalBuildDependsOn)
    </TraversalBuildDependsOn>
  </PropertyGroup>
```

Also I've submitted a private helix run and it is green with no test failures with the required change in coreclr to actually load the resources from pri file: https://mc.dot.net/#/user/safern/pr~2Fcorefx~2Fmaster~2F/test~2Ffunctional~2Fuap~2F/20170713.3526

cc: @tarekgh @weshaggard @joperezr


morganbr:
@tarekgh , would it be possible to avoid the resource merging part by having CoreClr resolve framework resources from embedded the same way UAPAOT does? Makepri is slow and merging has failures sometimes, so it would be preferable to not need to merge framework resources into the test resources (the same is true for actual user applications).

tarekgh:
> would it be possible to avoid the resource merging part by having CoreClr resolve framework resources from embedded the same way UAPAOT does? Makepri is slow and merging has failures sometimes, so it would be preferable to not need to merge framework resources into the test resources (the same is true for actual user applications).

I would like to keep the merging to the PRI because this is the customers scenario. also it will give us some idea if we run into any issue with MakePri. performance wise, it is not that slow especially most of the tests doen't have resources so we'll not merge and instead we will copy the pre-created common PRI

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
```diff
@@ -0,0 +1,154 @@
...
+            File.WriteAllText(_modifiedConfigFile, text);
+        }
+
+        private void MergePriFiles()
+        {
+            string projectRoot = Path.GetDirectoryName(RunnersPriFilePath);
+            ProcessStartInfo startInfo = new ProcessStartInfo()
+            {
+                FileName = MakePriExecutablePath,
+                Arguments = $"versioned /o /pr {projectRoot} /cf {_modifiedConfigFile} /of {NewPriFilePath} /if {RunnersPriFilePath}",
```
tarekgh: >Arguments = $"versioned /o /pr {projectRoot} /cf {_modifiedConfigFile} /of {NewPriFilePath} /if {RunnersPriFilePath}", [](start = 16, length = 118)

do we need to add quotations around the different parameters? this will help avoid problem if we have spaces in the file paths

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -9,11 +9,32 @@
...
+    <StringResourcesPath Condition="'$(StringResourcesPath)'=='' And Exists('$(ResourcesSourceOutputDirectory)Strings.resx')">$(ResourcesSourceOutputDirectory)Strings.resx</StringResourcesPath>
     <IntermediateResOutputFileFullPath Condition="'$(MSBuildProjectExtension)' == '.csproj'">$(IntermediateOutputPath)SR.cs</IntermediateResOutputFileFullPath>
     <IntermediateResOutputFileFullPath Condition="'$(MSBuildProjectExtension)' == '.vbproj'">$(IntermediateOutputPath)SR.vb</IntermediateResOutputFileFullPath>
     <DefineConstants Condition="'$(ConfigurationGroup)' == 'Debug'">$(DefineConstants);DEBUGRESOURCES</DefineConstants>
   </PropertyGroup>
+
+  <Target Name="CopyResxFilesToReswFiles" Condition="'$(BuildingUAPVertical)' == 'true'" BeforeTargets="CoreCompile">
+
+    <ItemGroup Condition="'$(BuildingUAPVertical)' == 'true'">
+      <AllResxFiles Include="@(EmbeddedResource)" Condition="'%(Extension)' == '.resx'">
```
tarekgh: >AllResxFiles [](start = 7, length = 12)

could you prefix the internal property names with underscore?

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
safern: Yes, we can add that. Will fix.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
safern: Yes, will fix.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -159,13 +160,63 @@
           />
   </Target>
 
+  <PropertyGroup Condition="'$(BuildingUAPVertical)' == 'true'">
+    <CommonPriFile Condition="'$(CommonPriFile)' == ''">$(RuntimePath)/resw/resources.pri</CommonPriFile>
```
tarekgh: >CommonPriFile [](start = 5, length = 13)

prefix with underscore please

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -159,13 +160,63 @@
...
+    <TestSpecificPriFile>$(TestPath)\resources.pri</TestSpecificPriFile>
+  </PropertyGroup>
+
+
+
+  <Target Name="CalculateResWFiles" Condition="'$(BuildingUAPVertical)' == 'true'">
+    <ItemGroup>
+      <TestResWFiles Include="$(TestResourcesFolderPath)\*.resw" />
+      <CommonResWFiles Include="$(ResourcesFolderPath)\*.resw" />
+    </ItemGroup>
```
tarekgh: is it possible these can be moved to MakeCommonResourcesPriFile target?

tarekgh:
other than my comments, LGTM.

@weshaggard please have a look. thanks

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -159,13 +160,63 @@
...
+  </Target>
+
+  <Target Name="MakeCommonResourcesPriFile"
```
safern: Sure. Will add them.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: Will fix.

morganbr:
@tarekgh , I was sorta wondering if we should change that for customers too. The framework assemblies are already carrying their resources, so copying them to PRI makes debug builds and deployments slower.

tarekgh:
This will need to have a changes in the VS targets to do that. I am not objecting the idea if you can get VS team commit doing that. the change from their side can be simple by enabling the following property to be true by default

```XML
<SkipMergingFrameworkResources>true</SkipMergingFrameworkResources>
```

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                throw new ArgumentException($"App runner resources.pri file: {RunnersPriFilePath} was not found.");
+            }
+
+            if (!File.Exists(MakePriExecutablePath))
+            {
+                throw new ArgumentException($"MakePri.exe was not found in path: {MakePriExecutablePath}");
+            }
+
+            _tempFolder = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName(), "makeprifiles");
```
weshaggard: Any reason to use the temp directory as opposed to a configured directly like the obj directory in a repo?

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
weshaggard: Whenever I see use directly calling ProcessStart in a Task I ask myself can we just use the Exec task instead or can we wrap it? Otherwise we are in the world of needing the maintain the process interactions manually ourselves which isn't great.

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
tarekgh: I think using the intermediate folder would be better here so we don't have to worry deleting it ourselves

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -9,11 +9,38 @@
...
+
+    <PropertyGroup>
+      <ProjectHasResources Condition="'@(AllResxFiles)' != ''">true</ProjectHasResources>
+      <TestProjectNeedsModifiedPriFile Condition="'$(IsTestProject)' == 'true' AND '$(ProjectHasResources)' == 'true' AND '$(ResourcesShouldGoToCommonPri)' != 'true'">true</TestProjectNeedsModifiedPriFile>
+      <TestResourcesFolderPath Condition="'$(TestProjectNeedsModifiedPriFile)' == 'true'">$(RuntimePath)$(AssemblyName).Resw</TestResourcesFolderPath>
+      <ResWDestinationPath Condition="'$(TestResourcesFolderPath)' != ''">$(TestResourcesFolderPath)</ResWDestinationPath>
+      <ResWDestinationPath Condition="'$(TestResourcesFolderPath)' == ''">$(ResourcesFolderPath)</ResWDestinationPath>
+    </PropertyGroup>
+
+    <Copy SourceFiles="@(AllResxFiles->'%(FullPath)')"
```
weshaggard: Why do we need to copy the resx files to resw files here? I thought we were already outputing resw files from your other changes.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -9,11 +9,38 @@
...
+          DestinationFiles="@(AllResxFiles->'$(ResWDestinationPath)/%(ReswName).resw')"
+          SkipUnchangedFiles="true">
+          <Output TaskParameter="CopiedFiles" ItemName="FilesCreated" />
+    </Copy>
+
+    <!-- We need to touch the copied files so that the target that uses them can track the inputs and outputs with the copied timestamp -->
+    <Touch Files="@(FilesCreated)" />
```
weshaggard: I'm not sure I understand why you need to touch the files here. You just copied them and if they were unchanged then we don't want them to invalidate the inputs of the other target.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
weshaggard: Nevermind I see you removed that below.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -159,13 +160,63 @@
           />
   </Target>
 
+  <PropertyGroup Condition="'$(BuildingUAPVertical)' == 'true'">
+    <CommonPriFile Condition="'$(CommonPriFile)' == ''">$(RuntimePath)/resw/resources.pri</CommonPriFile>
+    <TestSpecificPriFile>$(TestPath)\resources.pri</TestSpecificPriFile>
+  </PropertyGroup>
+
+
```
weshaggard: nit 3 empty lines

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -159,13 +160,63 @@
...
+
+  <Target Name="CalculateResWFiles" Condition="'$(BuildingUAPVertical)' == 'true'">
```
weshaggard: Is there a reason you put this in a target instead of just in the global scope?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -159,13 +160,63 @@
...
+  </Target>
+
+  <Target Name="MakeCommonResourcesPriFile"
+          DependsOnTargets="CalculateResWFiles"
+          Condition="'$(BuildingUAPVertical)' == 'true'"
+          Inputs="@(CommonResWFiles)"
+          Outputs="$(CommonPriFile)">
+
+    <MakeResourcesPriFile ResWFiles="@(CommonResWFiles)"
+                          RunnersPriFilePath="$(TestHostRootPath.Replace('/', '\'))\Runner\resources.pri"
```
weshaggard: Why do we have to renormalize the path slashes in all these places?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -197,6 +248,12 @@
       <OutputFolderForTestDependencies>$(BinDir)/TestDependencies</OutputFolderForTestDependencies>
     </PropertyGroup>
 
+    <!-- For UAP we need to copy the generated resources.pri containing the framework resources to the test folder -->
+    <Copy Condition="'$(BuildingUAPVertical)' == 'true' AND '$(TestProjectNeedsModifiedPriFile)' != 'true'"
+          SourceFiles="$(CommonPriFile)"
```
weshaggard: Should we eliminate this special case and handle it directly in MakeTestSpecificResourcesPriFile so after that target we only have a resources.pri file. Either a copy of the common one or the common plus the test pri.

weshaggard: We generally want to avoid extra custom copies and get the full set of files to copy into the RunnerFolderContents item which gets used for our script generation.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -283,7 +340,7 @@ robocopy /S /NP %EXECUTION_DIR%int\%(Identity)\ %EXECUTION_DIR%native\
       <TestCommandLines Include="mklink /H %(RunnerFolderContents.RecursiveDir)%(RunnerFolderContents.Filename)%(RunnerFolderContents.Extension) %RUNTIME_PATH%\Runner\%(RunnerFolderContents.RecursiveDir)%(RunnerFolderContents.Filename)%(RunnerFolderContents.Extension)" />
 
       <!-- Copy the runtime binaries over -->
-      <RuntimePathContents Include="$(RuntimePath)\**\*" />
+      <RuntimePathContents Include="$(TestHostRootPath)\UAPLayout\*"/>
```
weshaggard: Do we have all the necessary runtime files in this directory?

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
safern: Ok, will look at the Exec task to try to go that route.

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
safern: No reason in particular. I just wanted to use some directory that wasn't in the common path where I would be adding files that are used just once by makepri to the repo. Is there a way to get the intermediate directory inside the task or should I pass it as a parameter?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
safern: The problem that we where having here is that the Copy task preserves the timestamp of the original copied file so when building only one src assembly we would copy over that resx file if it exists with and old timestamp. If we run the tests then the common pri file will be created with a recent timestamp. However if we after that build another src assembly that has a resx file a new resw file will be created but with a timestamp older than the common pri file that was created previously, so if we build any other tests the common pri will not be updated because the MakeCommonPriFile target's inputs  are older than the output so the task would be skipped. So the only good solution that we found was touching the resw files so that their timestamp is the time when we copied them and if we add a new one after we create the common pri file it will have a newer timestamp than the common pri file and it will be updated.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: When building a test this target is loaded at the same time as resources.targets, so we could hit the case where this test have an embedded resource but since the global scope is calculated on load time and no in execution time by the targets the files wouldn't be updated and will not contain that new resw that was just created by resources.targets. So we need to put it inside a target so that this target is executed right before creating the pri file which means resources.targets was already executed and it will have all the necessary files.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: I did the normalization to avoid errors within makepri because I didn't know if that could cause any exceptions, but I can keep them with no normalization if it doesn't hurt makepri and you rather have them without normalization.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: I can move that to that target and have the target executed every time and just condition it to either copy the common file or call the task to create a new one containing the tests resources.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: Yes, I've already submitted a helix private run and I had no issues, all the tests passed.

src/Microsoft.DotNet.Build.Tasks/MakeResourcesPriFile.cs
weshaggard: You should just add a parameter for the path.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
weshaggard: >However if we after that build another src assembly that has a resx file a new resw file will be created but with a timestamp older than the common pri file that was created previously

I'm not sure I understand. As I would expect it to already have a resw as well in this context.  In what context would you have a resx and not a resw and then later build again and get the resw? 

Keeping the touch here will essentially eliminate the incremental build for the common pri which is what we want. We will need to find another solution.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
weshaggard: Fair I can see an argument that more resw files will be created during this build invocation.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
weshaggard: Yes I would like to avoid all these extra string copying to fix the slashes if they aren't necessary. It really makes the code more error prone and less performant.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
weshaggard: Yes I think that would be better.

safern:
I've addressed PR feedback. Deleted the custom task that created the pri file and move to use existing exec task and everything is now created inside targets.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -9,11 +9,56 @@
...
+          Outputs="@(_ReswOutputFiles)" >
+
+    <Copy SourceFiles="@(_AllResxFiles->'%(FullPath)')"
+          DestinationFiles="@(_ReswOutputFiles)"
+          SkipUnchangedFiles="true">
+          <Output TaskParameter="CopiedFiles" ItemName="FilesCreated" />
+    </Copy>
+
+    <!-- We need to touch the copied files so that the target that uses them can track the inputs and outputs with the copied timestamp -->
+    <Touch Files="@(FilesCreated)" />
```
tarekgh: >  <Touch Files="@(FilesCreated)" /> [](start = 2, length = 35)

do we still need to use touch?

tarekgh:
LGTM

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -9,11 +9,56 @@
...
+    <ItemGroup>
+      <_ReswOutputFiles Include="@(_AllResxFiles->'$(_ResWDestinationPath)/%(ReswName).resw')" />
+    </ItemGroup>
+
+  </Target>
+
+  <Target Name="CopyResxFilesToReswFiles" 
+          Condition="'$(BuildingUAPVertical)' == 'true'" 
+          DependsOnTargets="CalculateResourcesFileNames"
+          Inputs="@(_AllResxFiles->'%(FullPath)')"
```
weshaggard: Instead of doing this item transform multiple times I would suggest capturing it in the target above.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
weshaggard: Yes the touch is still needed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -9,11 +9,56 @@
...
+      <_ResWDestinationPath Condition="'$(TestResourcesFolderPath)' == ''">$(ResourcesFolderPath)</_ResWDestinationPath>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <_ReswOutputFiles Include="@(_AllResxFiles->'$(_ResWDestinationPath)/%(ReswName).resw')" />
+    </ItemGroup>
+
+  </Target>
+
+  <Target Name="CopyResxFilesToReswFiles"
```
weshaggard: While the general logic looks correct are you able to verify that it correctly works incrementally?

safern: Yes, I did verify it locally and it correctly works incrementally.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
safern: Correct, it is still needed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
safern: Ok. Will fix this.

safern: I've fixed this.

safern:
Thanks for your reviews @tarekgh and @weshaggard

PR #1612, opened 2017-07-16
Update ReportGenerator version to 2.5.9

stephentoub:


src/common/test-runtime/project.json
```diff
@@ -9,7 +9,7 @@
     "System.Runtime.InteropServices.RuntimeInformation": "4.4.0-beta-24813-03",
     "System.Linq.Parallel": "4.0.1",
     "OpenCover": "4.6.519",
-    "ReportGenerator": "2.5.0",
+    "ReportGenerator": "2.5.9",
```
weshaggard: While it is probably find to update this version we don't really use this in corefx any longer. Once you integrate this change into corefx you will need to update. https://github.com/dotnet/corefx/blob/master/external/test-runtime/XUnit.Runtime.depproj#L62, ideally you should change it to use the msbuild property $(ReportGeneratorVersion) now.

stephentoub: > we don't really use this in corefx any longer

So I can just update corefx and not build tools? Can all this be deleted from build tools then? I'm not clear on why it's in both places.

weshaggard: We will at least need an update the CodeCoverage.targets file. Instead of updating the version number though I think I would go about deleting the version numbers. Then add the properties to dependencies.props in corefx so they are defined and can be picked up by both CodeCoverage.targets as well as the Xunit.Runtime.depproj file.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -6,7 +6,7 @@
   -->
   <PropertyGroup>
     <OpenCoverVersion>4.6.519</OpenCoverVersion>
-    <ReportGeneratorVersion>2.5.0</ReportGeneratorVersion>
+    <ReportGeneratorVersion>2.5.9</ReportGeneratorVersion>
```
weshaggard: Just to be safe instead of deleting these it might be worth add a condition to set them to these values if they aren't already set.

PR #1615, opened 2017-07-21
Adding SSPI and LSA apis to UWP whitelist baseline

pjanotti:


src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer_Win32UWPApis.txt
```diff
@@ -4791,9 +4796,32 @@ shcore.dll!CreateRandomAccessStreamOnFile
 shcore.dll!CreateRandomAccessStreamOverStream
 shcore.dll!CreateStreamOverRandomAccessStream
 shcore.dll!IsProcessInIsolatedContainer
+shell32.dll!SHGetKnownFolderPath
```
danmosemsft: I checked on SHGetKnownFolderPath. They are not adding it to WACK. We will have a new API. So please leave this out of this file, and keep the baseline file in CoreFX until we stop calling it.

pjanotti: ack

PR #1628, opened 2017-08-08
Initial set of changes to use msbuild that comes with the CLI on Windows

joperezr:
cc: @weshaggard 

This are only the changes required to build on Windows, we still need to do some work for Linux side. Also, in here we will always be using .NET Core MSBuild on Windows, the right approach to me would be to be able to choose which one, but throwing this out just to get initial feedback.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -10,6 +10,7 @@ if [%PACKAGES_DIR%] == [] set PACKAGES_DIR=%TOOLRUNTIME_DIR%
 set PACKAGES_DIR=%PACKAGES_DIR:"=%
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
 set MICROBUILD_VERSION=0.2.0
+set ROSLYNCOMPILERS_VERSION=2.0.0-rc
 
 :: Determine if the CLI supports MSBuild projects. This controls whether csproj files are used for initialization and package restore.
 set CLI_VERSION=
@@ -26,7 +27,8 @@ if [%BUILDTOOLS_USE_CSPROJ%]==[] (
 { ^
   "dependencies": ^
     { ^
-      "MicroBuild.Core": "%MICROBUILD_VERSION%" ^
+      "MicroBuild.Core": "%MICROBUILD_VERSION%", ^
+      "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%" ^
     }, ^
   "frameworks": {"netcoreapp1.0": {},"net46": {}} ^
 }
@@ -41,6 +43,7 @@ if [%BUILDTOOLS_USE_CSPROJ%]==[] (
   ^^^</PropertyGroup^^^> ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
+    ^^^<PackageReference Include=^"Microsoft.Net.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
   ^^^</ItemGroup^^^> ^
  ^^^</Project^^^>
   set PROJECT_EXTENSION=csproj
@@ -88,6 +91,8 @@ if not [%NET46_PUBLISH_ERROR_LEVEL%]==[0] (
   exit /b %NET46_PUBLISH_ERROR_LEVEL%
 )
 
+Robocopy "%TOOLRUNTIME_DIR%\runtimes\any\native" "%TOOLRUNTIME_DIR%\."
+
 :: Copy Portable Targets Over to ToolRuntime
 if not exist "%PACKAGES_DIR%\generated" mkdir "%PACKAGES_DIR%\generated"
 set PORTABLETARGETS_PROJECT=%PACKAGES_DIR%\generated\project.%PROJECT_EXTENSION%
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -28,9 +26,7 @@ if [%BUILDTOOLS_USE_CSPROJ%]==[] (
 { ^
   "dependencies": ^
     { ^
-      "MicroBuild.Core": "%MICROBUILD_VERSION%", ^
-      "Microsoft.Portable.Targets": "%PORTABLETARGETS_VERSION%", ^
-      "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%" ^
+      "MicroBuild.Core": "%MICROBUILD_VERSION%" ^
```
weshaggard: Why do we need to keep MicroBuild here?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -94,12 +88,6 @@ if not [%NET46_PUBLISH_ERROR_LEVEL%]==[0] (
...
 )
 
-:: Copy some roslyn files which are published into runtimes\any\native to the root
-Robocopy "%TOOLRUNTIME_DIR%\runtimes\any\native" "%TOOLRUNTIME_DIR%\."
-
-:: Microsoft.Build.Runtime dependency is causing the MSBuild.runtimeconfig.json buildtools copy to be overwritten - re-copy the buildtools version.
-Robocopy "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%\." "MSBuild.runtimeconfig.json"
-
 :: Copy Portable Targets Over to ToolRuntime
 if not exist "%PACKAGES_DIR%\generated" mkdir "%PACKAGES_DIR%\generated"
```
weshaggard: I thought we didn't need to portable targets?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: I'd like to clean out this entire project generation.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.cmd
```diff
@@ -1,2 +1,2 @@
-@call msbuild.exe %*
+@call %~dp0dotnetcli\dotnet.exe msbuild %*
```
weshaggard: We probably want to add some logic to see if msbuild is on the path and if so then use it otherwise fallback to dotnet.exe msbuild. With some way to force that if needed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
```diff
@@ -4,7 +4,6 @@
     <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
     <EnableDefaultItems>false</EnableDefaultItems>
     <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">2.0.0-beta-001776-00</RuntimeFrameworkVersion>
```
weshaggard: While it probably isn't necessary we should probably start passing this version in to avoid restoring it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: I don't need it to build myself, but I thought this was required for official builds which is why I didn't touch that.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
joperezr: We don't, I kept this for the microbuild stuff, but if that is not required I'll go ahead and nuke this as well

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.cmd
joperezr: Yup that is the plan, this is just temporary. I also plan to do msbuild specific parameters in here( like nodeReuse) cc @maririos

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
joperezr: Sounds good

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: We still need the microbuild stuff but we should move it to the toolruntime project instead.

joperezr: I'm not sure if its possible as I don't know if the stuff will be published correctly, but I'll try this out.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -26,7 +27,8 @@ if [%BUILDTOOLS_USE_CSPROJ%]==[] (
 { ^
   "dependencies": ^
     { ^
-      "MicroBuild.Core": "%MICROBUILD_VERSION%" ^
+      "MicroBuild.Core": "%MICROBUILD_VERSION%", ^
+      "Microsoft.Net.Compilers": "%ROSLYNCOMPILERS_VERSION%" ^
```
weshaggard: Why add here? Can't we just put them into the toolruntime.csproj?

joperezr: yeah, I haven't done those changes yet, this PR only got updated because I've been making changes on my local buildtools and keeping track of the changes by pushing but I'm mainly getting corefx to build with the changes first, After getting that, I will update all of this logic to be msbuild based instead.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/msbuild.cmd
eerhardt: I don't think keying off the `%PATH%` is a good idea.  If I happen to use a VS developer command prompt and build this repo, by default I think it should build the same as if I opened a blank `cmd` prompt.

Instead, I think the `force` option should be the other way around.  By default, use the `dotnet.exe msbuild`, and keep it consistent between Windows and Linux.  If someone wants to force the build to use a full framework MSBuild, then they can pass in an arg to override the MSBuild to use.

joperezr: I believe our only concern on using the .NETCore MSBuild that comes with the CLI as the default was perf, since the desktop version was usually faster. Once we are ready to push this through, we can try and measure perf again with each case, and then decide on which one to take based on both consistency across platforms and build time.

eerhardt: If we have concerns about perf - then our customers will too.  We need to push on getting these things into our product.  That's the point about dogfooding.

joperezr:
Closing up old PR. I will eventually finish this in a new PR once we are ready to onboard the new msbuild inside the CLI for our repos.

PR #1629, opened 2017-08-09
Enable XUnit conditional filtering uses other assemblies

tarekgh:
Conditional Fact and Theory in our xunit filtering is always used types from the test assembly that currently running. We are working on moving PlatformDetection code to CoreFx.Private.TestUtilities assembly as it is naturally fit there and avoid including the cs files in every test project need to use it. We have many tests using Conditional Fact/Theory which calls PlatformDetection methods. The way we apply this filtering is we create the PlatformDetection type from the running test assembly. After we move PlatformDetection to CoreFx.Private.TestUtilities, this mechanism will not work. To make it work, we adding the way to allow specify the assembly containing the type through providing the qualified type name which include the assembly name.

### DIFF
--- a/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
+++ b/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
@@ -48,16 +48,36 @@ namespace Xunit.NetCore.Extensions
                     continue;
                 }
 
-                string[] symbols = conditionMemberName.Split('.');
                 Type declaringType = testMethodDeclaringType;
 
-                if (symbols.Length == 2)
+                // We have qualified type name with the assembly name, something like
+                // [ConditionalFact("System.PlatformDetection, CoreFx.Private.TestUtilities!" + nameof(PlatformDetection.IsNonZeroLowerBoundArraySupported))]
+                // We don't use '.' as separator in such case and use '!' because the qualified type name can have '.' to include the type namespace.
+                if (conditionMemberName.IndexOf('!') > 0)
                 {
-                    conditionMemberName = symbols[1];
-                    ITypeInfo type = testMethod.TestClass.Class.Assembly.GetTypes(false).Where(t => t.Name.Contains(symbols[0])).FirstOrDefault();
-                    if (type != null)
+                    // get the method name
+                    string[] symbols = conditionMemberName.Split('!');
+                    if (symbols.Length == 2)
                     {
-                        declaringType = type.ToRuntimeType();
+                        conditionMemberName = symbols[1];
+                        Type requestedType = Type.GetType(symbols[0]);
+                        if (requestedType != null)
+                        {
+                            declaringType = requestedType;
+                        }
+                    }
+                }
+                else
+                {
+                    string[] symbols = conditionMemberName.Split('.');
+                    if (symbols.Length == 2)
+                    {
+                        conditionMemberName = symbols[1];
+                        ITypeInfo type = testMethod.TestClass.Class.Assembly.GetTypes(false).Where(t => t.Name.Contains(symbols[0])).FirstOrDefault();
+                        if (type != null)
+                        {
+                            declaringType = type.ToRuntimeType();
+                        }
                     }
                 }
 


tarekgh:
CC @stephentoub @weshaggard @safern @danmosemsft

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
```diff
@@ -48,16 +48,36 @@ internal class ConditionalTestDiscoverer
                     continue;
                 }
 
-                string[] symbols = conditionMemberName.Split('.');
                 Type declaringType = testMethodDeclaringType;
 
-                if (symbols.Length == 2)
+                // We have qualified type name with the assembly name, something like
+                // [ConditionalFact("System.PlatformDetection, CoreFx.Private.TestUtilities!" + nameof(PlatformDetection.IsNonZeroLowerBoundArraySupported))]
```
safern: This example I think is not correct. Because there you are passing System.PlatformDetection as a property to be a condition and it is a class. ConditionalFact and ConditionalTheory attributes receive either one or more conditional properties that will be evaluated and they all have to be true in order to execute the test. Could we update the comment?

tarekgh: the whole string passed as the attribute parameter would be 

```
System.PlatformDetection, CoreFx.Private.TestUtilities!IsNonZeroLowerBoundArraySupported`
```

which is correct. the part before the '!'is the type name and the part after '!' is the method name. I am not sure what is wrong?

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
```diff
@@ -48,16 +48,36 @@ internal class ConditionalTestDiscoverer
...
+                // We don't use '.' as separator in such case and use '!' because the qualified type name can have '.' to include the type namespace.
+                if (conditionMemberName.IndexOf('!') > 0)
                 {
-                    conditionMemberName = symbols[1];
-                    ITypeInfo type = testMethod.TestClass.Class.Assembly.GetTypes(false).Where(t => t.Name.Contains(symbols[0])).FirstOrDefault();
-                    if (type != null)
+                    // get the method name
+                    string[] symbols = conditionMemberName.Split('!');
+                    if (symbols.Length == 2)
```
danmosemsft: if symbols.Length != 2 then throw?

tarekgh: we didn't do that before so I am not sure what would be the effect of throwing here. look at the original code and you'll see we don't throw but produce some error

safern: Actually it wouldn't discover the test instead of causing a failure which perhaps is better?

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
safern: That you are passing multiple arguments and for ConditionalFact and Theory every argument is a condition to evaluate, so in this case it would look for two different properties: 

* System.PlatformDetection
* CoreFx.Private.TestUtilities!IsNonZeroLowerBoundArraySupported

So that would be two different entries. 

https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Attributes/ConditionalFactAttribute.cs#L16

https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs#L41

So if we want the full name perhaps you need a little more parsing? Maybe the assembly name move it at the beginning, something like: 

`AssemblyName!ClassName.PropertyName`

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
danmosemsft: Eh, probably the LookupConditionalMethod would then fail anyway so you get your failure.

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
danmosemsft: I agree `AssemblyName!ClassName.PropertyName` is much clearer and more standard. I jsut realized that's what I assumed it was and misunderstood.

danmosemsft:
@tarekgh I guess you want to close https://github.com/dotnet/buildtools/pull/1592

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
safern: Right now if  it couldn't find the method it will return an ExecutionErrorTestCase which means that the test will not be discovered and will print an error saying that the property didn't exist. 

https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs#L65-L75

So we don't really throw exceptions we print an error message for that specific test through xunit (which I think is better) 

This is the error message that would be displayed: https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs#L105-L108

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
tarekgh: Just to clarify, the qualified type name is "System.PlatformDetection, CoreFx.Private.TestUtilitie" which we can just pass it to Type.GetType without any manipulation. so going with "AssemblyName!ClassName.PropertyName" I'll have to reconstruct the qualified type name from it before I can use it. so I am not seeing any benefit doing that.

safern: I just missed that the comma was inside quotes :( so that is why I thought it was wrong, I thought you where passing 2 parameters in the example. Sorry about that. 

If the qualified name is expected as you suggest and it is easier I'm fine with sticking with that, however I think the other one is much clearer. It is unfortunate that it the qualified name structure.

Thanks for clarifying @tarekgh

tarekgh: Thanks @safern 

@danmosemsft please let me know if you have any more comments

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
bartonjs: Well, if the length is 1 it should keep doing what it was doing... there's no reason we should change the attributes that are targeting local, functionality-named members.

```C#

public static bool PlatformSupportsRsaPss { get; } = PlatformDetection.IsWindows10;

[ConditionalFact(nameof(PlatformSupportsRsaPss)]
public static void SomePssTest()
{
    ...
}
```

That's my preferred style, since it says why it's conditional, not where it's conditional.  (So, to me, the need for change here doesn't really exist)

safern: I think if the length is 1 it didn't change the actual behavior, since the default type is the test class (https://github.com/tarekgh/buildtools/blob/663eada3fea68d68e4a1768bf56afaea2e48c54b/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs#L51) 

Also at the beginning it is assuming that the ConditionalName is the one found as a parameter (https://github.com/tarekgh/buildtools/blob/663eada3fea68d68e4a1768bf56afaea2e48c54b/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs#L43)

Afterwards it will try to construct the type and method name if it finds a `!` or a `.` and look it up after the calculation: https://github.com/tarekgh/buildtools/blob/663eada3fea68d68e4a1768bf56afaea2e48c54b/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs#L85

So I don't think this will hurt the existing behavior, but we will gain another way to filter out tests without having to wrap a PlatformDetection property inside a local property.

bartonjs: @safern Ah, fair enough.  I hadn't looked at the full code, just what popped through in mail and the context view here.  So I thought `!` was a mandatory value now, not that we were down a path where it had already been detected.

Don't mind me, nothing to see here...

tarekgh:
thanks @safern @danmosemsft for your review

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
weshaggard: I agree with others here that the the assembly name should be part of the type and not attached to the member. Even if we cannot use "." I still think it would be better to have "<Assembly>!Type, Member" as the format as that seems more logical to me.

PR #1630, opened 2017-08-10
Add task to get the resources and merge them into the common pri for external dependencies

safern:
The WCF team encountered an issue where the resources for their dependencies (in this case .NET Core) were not present in the pri file. The reason why this happened is because we assumed that the other repos where the same as coreFX but we didn't count on coreFX as an external dependency. So we need to get those external dependencies and generate their resw files in order to create the pri file through mergepri.exe.

FYI: @StephenBonikowsky

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.net45/CreateExternalAssembliesResources.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/CreateExternalAssembliesResources.cs
@@ -1,0 +1,78 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Collections;
+using System.IO;
+using System.Reflection;
+using System.Resources;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class CreateExternalAssembliesResources : Task
+    {
+        [Required]
+        public ITaskItem[] InputAssemblies { get; set; }
+
+        [Required]
+        public string OutputPath { get; set; }
+
+        public override bool Execute()
+        {
+            try
+            {
+                if (!Directory.Exists(OutputPath))
+                {
+                    Directory.CreateDirectory(OutputPath);
+                }
+
+                CreateReswFiles();
+            }
+            catch (Exception e)
+            {
+                Log.LogErrorFromException(e, showStackTrace: false);
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+
+        public void CreateReswFiles()
+        {
+            foreach (ITaskItem assemblySpec in InputAssemblies)
+            {
+                string assemblyPath = assemblySpec.ItemSpec;
+                try
+                {
+                    Assembly assembly = Assembly.LoadFrom(assemblyPath);
+                    foreach (string resourceName in assembly.GetManifestResourceNames())
+                    {
+                        if (!resourceName.EndsWith(".resources"))
+                        {
+                            continue; // we only need to get the resources strings to produce the resw files.
+                        }
+
+                        string reswName = Path.GetFileNameWithoutExtension(resourceName);
+                        string reswPath = Path.Combine(OutputPath, $"{reswName}.resw");
+                        using (FileStream stream = File.Create(reswPath))
+                        using (ResourceReader resourceReader = new ResourceReader(assembly.GetManifestResourceStream(resourceName)))
+                        using (ResXResourceWriter resourceWriter = new ResXResourceWriter(stream))
+                        {
+                            IDictionaryEnumerator enumerator = resourceReader.GetEnumerator();
+                            while (enumerator.MoveNext())
+                            {
+                                resourceWriter.AddResource(enumerator.Key.ToString(), enumerator.Value.ToString());
+                            }
+                        }
+                    }
+                }
+                catch (BadImageFormatException)
+                {
+                    continue; // native assemblies can't be loaded.
+                }
+            }
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
@@ -56,15 +56,13 @@ namespace Microsoft.DotNet.Build.Tasks
                         string reswName = Path.GetFileNameWithoutExtension(resourceName);
                         string reswPath = Path.Combine(OutputPath, $"{reswName}.resw");
                         using (ResourceReader resourceReader = new ResourceReader(assembly.GetManifestResourceStream(resourceName)))
+                        using (ReswResourceWriter resourceWriter = new ReswResourceWriter(reswPath))
                         {
-                            ReswResourceWriter resourceWriter = new ReswResourceWriter(reswPath);
                             IDictionaryEnumerator enumerator = resourceReader.GetEnumerator();
                             while (enumerator.MoveNext())
                             {
                                 resourceWriter.AddResource(enumerator.Key.ToString(), enumerator.Value.ToString());
                             }
-
-                            resourceWriter.Close();
                         }
                     }
                 }
@@ -76,7 +74,7 @@ namespace Microsoft.DotNet.Build.Tasks
         }
     }
 
-    internal class ReswResourceWriter
+    internal class ReswResourceWriter : IDisposable
     {
         private readonly XElement _root;
 
@@ -102,7 +100,7 @@ namespace Microsoft.DotNet.Build.Tasks
             _root.Add(newElement);
         }
 
-        internal void Close()
+        public void Dispose()
         {
             _document.Save(_filePath);
         }
--- a/src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
@@ -12,7 +12,7 @@ using System.Resources;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
-    public class CreateExternalAssembliesResources : Task
+    public class ExtractResWResourcesFromAssemblies : Task
     {
         [Required]
         public ITaskItem[] InputAssemblies { get; set; }
@@ -24,11 +24,7 @@ namespace Microsoft.DotNet.Build.Tasks
         {
             try
             {
-                if (!Directory.Exists(OutputPath))
-                {
-                    Directory.CreateDirectory(OutputPath);
-                }
-
+                Directory.CreateDirectory(OutputPath);
                 CreateReswFiles();
             }
             catch (Exception e)
@@ -56,9 +52,8 @@ namespace Microsoft.DotNet.Build.Tasks
 
                         string reswName = Path.GetFileNameWithoutExtension(resourceName);
                         string reswPath = Path.Combine(OutputPath, $"{reswName}.resw");
-                        using (FileStream stream = File.Create(reswPath))
                         using (ResourceReader resourceReader = new ResourceReader(assembly.GetManifestResourceStream(resourceName)))
-                        using (ResXResourceWriter resourceWriter = new ResXResourceWriter(stream))
+                        using (ResXResourceWriter resourceWriter = new ResXResourceWriter(reswPath))
                         {
                             IDictionaryEnumerator enumerator = resourceReader.GetEnumerator();
                             while (enumerator.MoveNext())
--- a/src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
@@ -6,12 +6,10 @@ using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
 using System.Resources;
 using System.Xml.Linq;
-using System.Linq;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
@@ -22,6 +20,9 @@ namespace Microsoft.DotNet.Build.Tasks
 
         [Required]
         public string OutputPath { get; set; }
+
+        [Required]
+        public string InternalReswDirectory { get; set; }
 
         public override bool Execute()
         {
@@ -43,6 +44,12 @@ namespace Microsoft.DotNet.Build.Tasks
             foreach (ITaskItem assemblySpec in InputAssemblies)
             {
                 string assemblyPath = assemblySpec.ItemSpec;
+
+                if (!ShouldExtractResources($"FxResources.{Path.GetFileNameWithoutExtension(assemblyPath)}.SR.resw", assemblyPath))
+                {
+                    continue; // we skip framework assemblies that resources already exist and don't need to be extracted to avoid reading dll metadata.
+                }
+                
                 try
                 {
                     Assembly assembly = Assembly.LoadFrom(assemblyPath);
@@ -53,8 +60,14 @@ namespace Microsoft.DotNet.Build.Tasks
                             continue; // we only need to get the resources strings to produce the resw files.
                         }
 
-                        string reswName = Path.GetFileNameWithoutExtension(resourceName);
-                        string reswPath = Path.Combine(OutputPath, $"{reswName}.resw");
+                        string reswName = $"{Path.GetFileNameWithoutExtension(resourceName)}.resw";
+
+                        if (!reswName.StartsWith("FxResources") && !ShouldExtractResources(reswName, assemblyPath)) // already checked for FxResources previously
+                        {
+                            continue; // resw output file already exists and is up to date, so we should skip this resource file.
+                        }
+
+                        string reswPath = Path.Combine(OutputPath, reswName);
                         using (ResourceReader resourceReader = new ResourceReader(assembly.GetManifestResourceStream(resourceName)))
                         using (ReswResourceWriter resourceWriter = new ReswResourceWriter(reswPath))
                         {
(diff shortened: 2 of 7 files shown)


src/Microsoft.DotNet.Build.Tasks.net45/CreateExternalAssembliesResources.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+            catch (Exception e)
+            {
+                Log.LogErrorFromException(e, showStackTrace: false);
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+
+        public void CreateReswFiles()
```
weshaggard: Doesn't this task already exist in our shipping tools? If so can we please use that one instead of creating our own.

tarekgh: I am not aware of any task doing this exact function (extracting resw from assemblies). The only place has a code like that is the implementation of GenerateResource task and this part is kind of internal which we cannot call it.
We used to do the same in .NetNative transforms but I think this is not happening anymore.

safern: Yes, @tarekgh and I looked at the VS tasks that we had before doing the first step a few weeks ago and we didn't find anything easy to achieve through that direction, so that is why we decided doing it on our end. This is actually "similar" on what we do in the .NetNative transforms but using a ResourceReader and a ResXResourceWriter which from what Tarek told me we didn't have available by the time that task was written.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -6,6 +6,7 @@
   <Import Project="$(ToolsDir)PerfTesting.targets" Condition="Exists('$(ToolsDir)PerfTesting.targets') and '$(Performance)' == 'true'"/>
   <UsingTask TaskName="GenerateTestExecutionScripts" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="GetTargetMachineInfo" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+  <UsingTask Condition="'$(BuildingUAPVertical)' == 'true'" TaskName="CreateExternalAssembliesResources" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
```
tarekgh: >CreateExternalAssembliesResources [](start = 70, length = 33)

I would call this ExtractResWResourcesFromAssemblies

src/Microsoft.DotNet.Build.Tasks.net45/CreateExternalAssembliesResources.cs
weshaggard: I was under the impression that any PCL library that gets consumed in a UWP project would have its resources stripped out and put into a resw just the same way we are doing here which is why I figured there must already be a task to do it. If the task is too different or not available to our builds then putting this here is fine I just want to avoid duplication if we can.

safern: Yes we actually tried to use that task before doing it all ourselves but it was kind of different and not available :(

tarekgh: @weshaggard, as I mentioned extracting the resources was done internally inside GenerateResource task when building ResW. I don't think we can call GenerateResource just to extract the resources

http://index/?query=Microsoft.AppxPackage.Targets&rightProject=CBA(Repo)&file=src%5Ctools%5CBuildProcess%5CMSBuild%5CMicrosoft%5CVisualStudio%5Cv14.0%5CAppxPackage%5CMicrosoft.AppxPackage.Targets&line=1047

src/Microsoft.DotNet.Build.Tasks.net45/CreateExternalAssembliesResources.cs
```diff
@@ -0,0 +1,78 @@
...
+        public ITaskItem[] InputAssemblies { get; set; }
+
+        [Required]
+        public string OutputPath { get; set; }
+
+        public override bool Execute()
+        {
+            try
+            {
+                if (!Directory.Exists(OutputPath))
```
danmosemsft: you don't need the directory.exists check. it will just do nothing if it already exists

safern: Oh, you're right. Thanks.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: Will fix.

src/Microsoft.DotNet.Build.Tasks.net45/CreateExternalAssembliesResources.cs
```diff
@@ -0,0 +1,78 @@
...
+                        if (!resourceName.EndsWith(".resources"))
+                        {
+                            continue; // we only need to get the resources strings to produce the resw files.
+                        }
+
+                        string reswName = Path.GetFileNameWithoutExtension(resourceName);
+                        string reswPath = Path.Combine(OutputPath, $"{reswName}.resw");
+                        using (FileStream stream = File.Create(reswPath))
+                        using (ResourceReader resourceReader = new ResourceReader(assembly.GetManifestResourceStream(resourceName)))
+                        using (ResXResourceWriter resourceWriter = new ResXResourceWriter(stream))
```
tarekgh: >ResXResourceWriter(stream) [](start = 71, length = 26)

ResXResourceWriter has a constructor which takes a string file path instead of stream. so you don't have to create the stream manually and pass it here.

safern: Good call.

tarekgh:
other than minor comments, LGTM

safern:
Addressed PR feedback, @weshaggard are you also OK with the changes?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -170,6 +171,23 @@
...
   </PropertyGroup>
+  
+  <ItemGroup Condition="'$(BuildingUAPVertical)' == 'true'">
+    <_AllRuntimeDllFiles Include="$(RuntimePath)\*.dll" />
+  </ItemGroup>
+
+  <Target Name="CreateReswFilesForExternalDependencies" 
+          Condition="'$(BuildingUAPVertical)' == 'true'"
+          Inputs="@(_AllRuntimeDllFiles)"
+          Outputs="$(_CommonPriFile)">
```
weshaggard: You aren't outputting _CommonPriFile here.

weshaggard: This should probably be the list of resw files.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
```diff
@@ -38,6 +39,7 @@
     <TargetingPackReference Include="System.Threading.Tasks" />
     <TargetingPackReference Include="System.Xml" />
     <TargetingPackReference Include="System.Xml.Linq" />
+    <TargetingPackReference Include="System.Windows.Forms" />
```
weshaggard: This task is only going to work on .NET Framework? We have to be careful that this also works on core as we will be doing builds, including windows builds on core as well.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: Ok, sounds good.

src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj
safern: Yes this is only going to work on .NET Framework but it is supported when building UAP right? I tested locally for building the whole test tree in UAP and in .NET Core and it worked fine, however I can move it to be netstandard1.6 and use the metadata reader and output the resources manually without the ResxResourceWriter. 

Why I would need to use the metadata reader is because Assembly.LoadFrom is not available until NS2.0 and we are currently building against NS1.6; are we planning on moving to the new toolchain and 2.0? 

Do you want me to go with the NS approach?

tarekgh: I think Wes's point is using System.Windows.Forms would not exist if msbuild will use net core and not the full framework. This means you just want to avoid using ResXResourceWriter and you may write the generated resw manually using Xml writer.

safern: Ok sounds good I will update to generate the resw file using Xml writer, if he would like to make it full NS, I found that this can be done with MetadataReader rather easier than what we talk the other day, so I could change that as well.

safern: Ok, sounds good. Will update that.

src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
```diff
@@ -0,0 +1,231 @@
...
+                }
+                catch (BadImageFormatException)
+                {
+                    continue; // native assemblies can't be loaded.
+                }
+            }
+        }
+    }
+
+    internal class ReswResourceWriter
```
tarekgh: would be nice if you implement the IDisoposable interface and then use "using" block when calling this class

safern: Fixed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -169,7 +170,32 @@
...
+    <_ExternalReswOutputPath>$(ResourcesFolderPath)/external/</_ExternalReswOutputPath>
   </PropertyGroup>
+  
+  <ItemGroup Condition="'$(BuildingUAPVertical)' == 'true'">
+    <_AllRuntimeDllFiles Include="$(RuntimePath)\*.dll" />
+    <_ExternalReswFiles Include="$(_ExternalReswOutputPath)*.resw" />
+    
+    <!-- The first time the CreateReswFilesForExternalDependencies target is executed the itemgroup _ExternalReswFiles will be empty
+    and that will avoid the target from executing, so we add a dummy file if they are empty so that the target is executed the first time. -->
+    <_ExternalReswFiles Condition="@(_ExternalReswFiles) == ''" Include="$(_ExternalReswOutputPath)dummy.resw" />
```
weshaggard: I guess I was expecting a transform from the *.dll to *.resw here so we know we always have all the matching files.

safern: We have no way to know what the resw file is going to be as that information comes from the dll's metadata, it can be a custom name, i.e we add an embedded resource for testing purposes that it's resources file is called SantiResources.resx and I can override the resource manifest name to be `Santi.This.IsMyResourceManager` in the csproj like in [here](https://github.com/dotnet/corefx/blob/master/src/System.Composition.Hosting/src/System.Composition.Hosting.csproj#L72) and that would mean that the resw file needs to be called `Santi.This.IsMyResourceManager.resw`

Since that is embedded into the dll's metadata there is no way to know what resw files the task will output and that is why the output resw files I create them [here](https://github.com/dotnet/buildtools/pull/1630/files#diff-4dba46151dfcc864a9186f4ae3de64c1R56) which comes from the ResourceName. That is why in the beginning I made the Outputs to be the common pri file, because that one is static, and if it is newer than the dlls then we don't need to update it.

weshaggard: OK talking this over with @tarekgh I think what you have will work.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -169,7 +170,32 @@
...
+  </ItemGroup>
+
+  <Target Name="CreateReswFilesForExternalDependencies" 
+          Condition="'$(BuildingUAPVertical)' == 'true'"
+          Inputs="@(_AllRuntimeDllFiles)"
+          Outputs="@(_ExternalReswFiles)">
+
+    <ItemGroup>
+      <_RuntimeDllFilesWithoutResw Include="@(_AllRuntimeDllFiles)"
```
weshaggard: What about cases where we rebuild the dll and now the resw is outdated? Is that something to worry about here? Do we need to take into account if the timestamp of the resw is older then the dll re-run it?

We could probably do this filter in the task itself.

safern: It will definitely be re ran if a dll is updated. That is what we want right? Since they could change the resources?

safern: @weshaggard you okay with this changes? This is the behavior we want right?

safern:
Fixed the filtering of dlls against resw files to be done inside the task itself.

tarekgh:
LGTM

src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
```diff
@@ -72,6 +85,25 @@ public void CreateReswFiles()
                 }
             }
         }
+
+        private bool ShouldExtractResources(string expectedReswFileName, string assemblyPath)
```
weshaggard: Do you also need to handle the case where it isn't a managed assembly?

src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
```diff
@@ -43,6 +44,12 @@ public void CreateReswFiles()
...
                 string assemblyPath = assemblySpec.ItemSpec;
+
+                if (!ShouldExtractResources($"FxResources.{Path.GetFileNameWithoutExtension(assemblyPath)}.SR.resw", assemblyPath))
+                {
+                    continue; // we skip framework assemblies that resources already exist and don't need to be extracted to avoid reading dll metadata.
+                }
+                
                 try
                 {
                     Assembly assembly = Assembly.LoadFrom(assemblyPath);
```
weshaggard: What happens if this is a native assembly?

src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
```diff
@@ -0,0 +1,261 @@
...
+                string assemblyPath = assemblySpec.ItemSpec;
+
+                if (!ShouldExtractResources($"FxResources.{Path.GetFileNameWithoutExtension(assemblyPath)}.SR.resw", assemblyPath))
+                {
+                    continue; // we skip framework assemblies that resources already exist and don't need to be extracted to avoid reading dll metadata.
+                }
+                
+                try
+                {
+                    Assembly assembly = Assembly.LoadFrom(assemblyPath);
```
weshaggard: Are we every going to hit an assembly load failure due to missing dependencies? For example if we load a .NET Core assembly on .NET Framework then it might not have all the dependencies resolved.

src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
safern: `Assembly.LoadFrom` will throw a `BadImageFormatException` which I'm catching and ignoring and going to the next assembly:

https://github.com/dotnet/buildtools/pull/1630/files#diff-4dba46151dfcc864a9186f4ae3de64c1R84

weshaggard:
FYI - https://github.com/Microsoft/msbuild/pull/2254 that is a link to a PR that is doing similar resource extraction in msbuild.

src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
safern: It will just not find the file and return true for the first check: https://github.com/dotnet/buildtools/pull/1630/files#diff-4dba46151dfcc864a9186f4ae3de64c1R48

But then when trying to load the assembly it will fall into the `catch (BadImageFormatException)` and will ignore it and continue to the next assembly.

src/Microsoft.DotNet.Build.Tasks.net45/ExtractResWResourcesFromAssemblies.cs
safern: So then we should probably use `Assembly.ReflectionOnlyLoadFrom` which doesn't load its dependencies, right?

safern:
So I updated it to use the MetadataReader and tested it locally and works perfectly

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -251,11 +276,16 @@
     <!-- We call MakePri.exe to create common resources.pri file only if this test project has its own resources. -->
     <Exec Command="$(_MakePriCommand)" StandardOutputImportance="Low" StdErrEncoding="Unicode" Condition="'$(TestProjectNeedsModifiedPriFile)' == 'true'" />
 
-    <Copy Condition="'$(TestProjectNeedsModifiedPriFile)' != 'true'"
+    <Copy Condition="'$(TestProjectNeedsModifiedPriFile)' != 'true' AND Exists('$(_CommonPriFile)')"
           SourceFiles="$(_CommonPriFile)"
           DestinationFiles="$(TestPath)\resources.pri"
           SkipUnchangedFiles="true" />
 
+    <Copy Condition="'$(TestProjectNeedsModifiedPriFile)' != 'true' AND !Exists('$(_CommonPriFile)')"
```
weshaggard: Should this be conditioned on the existence of Runner\resources.pri instead?

safern: The Runner\resources.pri file should always exist as it comes from the Runner's package and that is the base PRI that we use to merge from to create the CommonPriFile. The reason why I added this is in the scenario where a repo doesn't have any resources embedded into their projects and the external dependencies doesn't have resources either (this case is not likely to happen, but just thinking of the worst scenario) the list of resw files will be empty so MakePri will not create the CommonPriFile because it had no inputs. 

So I did this to protect the build, so that when this happens and there is no CommonPriFile, we use the base file which comes from the runner itself, the build succeeds and the tests wont fail.

safern:
Thanks for your reviews @tarekgh @weshaggard and @danmosemsft :)

PR #1634, opened 2017-08-10
Support passing Type to XUnit ConditionalFact/Theory

tarekgh:
[ConditionalFact(typeof(PlatformDetection), "nameof(PlatformDetection.IsNonZeroLowerBoundArraySupported)")]

### DIFF
--- a/src/xunit.netcore.extensions/Attributes/ConditionalFactAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/ConditionalFactAttribute.cs
@@ -11,7 +11,14 @@ namespace Xunit
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
     public sealed class ConditionalFactAttribute : FactAttribute
     {
+        public Type     CalleeType { get; private set; }
         public string[] ConditionMemberNames { get; private set; }
+
+        public ConditionalFactAttribute(Type calleeType, params string[] conditionMemberNames)
+        {
+            CalleeType = calleeType;
+            ConditionMemberNames = conditionMemberNames;
+        }
 
         public ConditionalFactAttribute(params string[] conditionMemberNames)
         {
--- a/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
+++ b/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
@@ -23,11 +23,37 @@ namespace Xunit.NetCore.Extensions
                                                         IMessageSink diagnosticMessageSink,
                                                         ITestMethod testMethod,
                                                         IEnumerable<IXunitTestCase> testCases,
-                                                        IEnumerable<string> conditionMemberNames)
+                                                        object[] conditionArguments)
         {
             // A null or empty list of conditionMemberNames is treated as "no conditions".
             // and the test cases will not be skipped.
-            // Example: [ConditionalFact()] or [ConditionalFact((string[]) null)]
+            // Example: [ConditionalFact()]
+            if (conditionArguments == null || conditionArguments.Length == 0)
+            {
+                return testCases;
+            }
+
+            string [] conditionMemberNames;
+
+            Type calleeType = conditionArguments[0] as Type;
+            if (calleeType != null)
+            {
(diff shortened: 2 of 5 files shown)


tarekgh:
@weshaggard @safern @danmosemsft

src/xunit.netcore.extensions/Attributes/ConditionalFactAttribute.cs
```diff
@@ -11,8 +11,15 @@ namespace Xunit
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
     public sealed class ConditionalFactAttribute : FactAttribute
     {
+        public Type     CalleeType { get; private set; }
         public string[] ConditionMemberNames { get; private set; }
 
+        public ConditionalFactAttribute(Type calleeType, params string[] conditionMemberNames)
```
weshaggard: Why params?

tarekgh: you can specify multiple methods

tarekgh: [ConditionalFact(typeof(PlatformDetection), "IsWinRT", "IsNotNanoServer")]

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
```diff
@@ -48,35 +74,44 @@ internal class ConditionalTestDiscoverer
...
                 {
-                    // get the method name
-                    string[] symbols = conditionMemberName.Split('!');
-                    if (symbols.Length == 2)
+                    declaringType = testMethodDeclaringType;
+
+                    // We have qualified type name with the assembly name, something like
+                    // [ConditionalFact("System.PlatformDetection, CoreFx.Private.TestUtilities!" + nameof(PlatformDetection.IsNonZeroLowerBoundArraySupported))]
+                    // We don't use '.' as separator in such case and use '!' because the qualified type name can have '.' to include the type namespace.
+                    if (conditionMemberName.IndexOf('!') > 0)
```
weshaggard: I would say we don't support this odd format now with the Type approach. Lets eliminate some of the complexity.

tarekgh: I thought it wouldn't hurt. I'll remove it.

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
```diff
@@ -48,35 +74,44 @@ internal class ConditionalTestDiscoverer
...
                     {
-                        conditionMemberName = symbols[1];
-                        Type requestedType = Type.GetType(symbols[0]);
-                        if (requestedType != null)
+                        // get the method name
+                        string[] symbols = conditionMemberName.Split('!');
```
safern: So we are still going to support the qualified name scenario? I imagine this is to cover the case where we could have multiple conditions that live under different assemblies? For example one that is under PlatformDetection and other that is under the test class itself?

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
safern: What about we hit this scenario? https://github.com/dotnet/buildtools/pull/1634#discussion_r132597294

I'm OK with removing it and forcing people to instead wrap it under a property inside the test class.

src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs
tarekgh: No I'll remove this as Wes suggested in his comment

safern: Yes I commented that and didn't see Wes's comment since I was already reviewing :)

I think is better to remove it as well.

tarekgh:
Thanks, @weshaggard and @safern for your quick review.

PR #1636, opened 2017-08-13
Enable CoreFX build to tag DLLs with GITHub URL

vancem:
In PR #1606 we added logic so that the version resource had not just the GIT commit ID
but also the full Github URL.   This worked well for native code but it turns out that
It did not work for managed code.

This was mostly because for managed code the variables needed to be 'exported' explicitly to a *.props
file so that another instance of MSBUILD would pick it up.   There was already a variable that had this
behavior called BuiltByString, and so this fix is to simply put the additional information for the GIT commit
ID into this variable (which would then be attached appropriately).

While I was at it I change the syntax a bit to make it more friendly to maching parsing.   key-value pairs
now always begin with an @ and are separted by spaces (e.g.  @BuiltBy: vancem @SrcCode: http//github.

@dagood

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -353,11 +353,11 @@
     <!-- and has the general form @Name1: Value1 @Name2: Value2  ... which makes it easy to parse -->
     <PropertyGroup Condition="'$(BuiltByString)' == ''">
       <!-- Set the @BuiltBy key-value pair -->
-      <BuiltByString Condition="'$(VersionUserName)' != '' AND '$(VersionHostName)' != ''">$(BuiltByString)%20%40BuiltBy: $(VersionUserName)-$(VersionHostName)</BuiltByString>
+      <BuiltByString Condition="'$(VersionUserName)' != '' AND '$(VersionHostName)' != ''">$(BuiltByString) %40BuiltBy: $(VersionUserName)-$(VersionHostName)</BuiltByString>
       <!-- Set the @SrcCode key-value pair -->
-      <BuiltByString Condition="'$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A'">$(BuiltByString)%20%40SrcCode%3A $(GitHubRepositoryUrl)/tree/$(LatestCommit)</BuiltByString>
+      <BuiltByString Condition="'$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A'">$(BuiltByString) %40SrcCode: $(GitHubRepositoryUrl)/tree/$(LatestCommit)</BuiltByString>
       <!-- If that does not work, try setting the @Commit key-value pair -->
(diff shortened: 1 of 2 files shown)


vancem:
I have tested this update on CoreCLR and CoreFX repos and it works as desired.

vancem:
I should mention that most of the actual delta is the removal of the (recently added) SrcCodeRef variable.  This variable was the key-value pair for the source code, but it had been 'folded' into the BuiltByString variable (which was already a key-value pair indicating the user who built the dll).   

I did this because there is 'plumbing' associated with this variable (actions that save this variable to a .props file and read it back at other places), and I want to add to it.   If desired we can rename this variable, but that will make the change significantly bigger, and something like 'BuildInfoString' would be better, BuiltByString is really not that bad either.  

@dotnet-bot  Test Pre-Merge signing tool

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -360,8 +348,16 @@
...
     </Exec>
 
-    <PropertyGroup>
-      <BuiltByString Condition="'$(BuiltByString)' == '' AND '$(VersionUserName)' != '' AND '$(VersionHostName)' != ''">%20built by: $(VersionUserName)-$(VersionHostName)</BuiltByString>
+    <!-- Set BuiltByString (which is added to the end of the informational version) -->
+    <!-- This string has additional information as well (like the GIT commit URL) -->
+    <!-- and has the general form @Name1: Value1 @Name2: Value2  ... which makes it easy to parse -->
+    <PropertyGroup Condition="'$(BuiltByString)' == ''">
+      <!-- Set the @BuiltBy key-value pair -->
+      <BuiltByString Condition="'$(VersionUserName)' != '' AND '$(VersionHostName)' != ''">$(BuiltByString)%20%40BuiltBy: $(VersionUserName)-$(VersionHostName)</BuiltByString>
```
dagood: Why `%20` rather than a space?

dagood:
`Pre-Merge signing tool` is a security process step, not a test: to get it green, you need to click on the Details link and go through the process to sign the commit. (Some people just ignore it.)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
vancem: My guess is that it would not matter either way, I was simply avoiding changing anything that I did not need to change.  The original text I was fiddling with had a %20built by: prefix (that is they used %20 for the space), so I did the same.  

Given that there are other spaces in the string, it will almost certainly work with a leading space, so if you want, I can change it (below as well).  Let me know.  

Note that I used %40 for the @ out of an abundance of caution.   Since these strings get used in MSBuild statements and @ is a special character for MSBuild it is likely to cause grief.     We could pick something else, I just want something that is 'letter like' (it has been used in identifiers for some language or another), but is very unlikely to ever be needed in the 'value' part of the key-value pairs (so we don't have the complexity of escape sequences).

dagood: Ah, I didn't notice it was already there. It seems like the original thought was that MSBuild removes leading non-encoded spaces, but I've never seen that happening. I think changing the `%20` to ` ` now would be good so this is more readable in the future.

The `%40` for `@` makes sense, thanks.

PR #1637, opened 2017-08-14
Add netcoreapp2.1 as a valid option for SupportedFrameworks

ericstj:
/cc @weshaggard @mellinoe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -941,6 +941,9 @@
     <DefaultValidateFramework Include="netcoreapp2.0">
       <RuntimeIDs>win7-x86;win7-x64;win10-arm64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.23-x64;fedora.24-x64;linuxmint.17-x64;opensuse.13.2-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64</RuntimeIDs>
     </DefaultValidateFramework>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -941,6 +941,9 @@
     <DefaultValidateFramework Include="netcoreapp2.0">
       <RuntimeIDs>win7-x86;win7-x64;win10-arm64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.23-x64;fedora.24-x64;linuxmint.17-x64;opensuse.13.2-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64</RuntimeIDs>
     </DefaultValidateFramework>
+    <DefaultValidateFramework Include="netcoreapp2.1">
+      <RuntimeIDs>win7-x86;win7-x64;win10-arm64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.23-x64;fedora.24-x64;linuxmint.17-x64;opensuse.13.2-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64</RuntimeIDs>
```
weshaggard: Not sure how much it matters but this RID list looks a bit dated. For example: we no longer support ubuntu16.10 we now support osx10.12 and 10.13.

ericstj: It was just a copy/paste from 2.0.  If you have information that would make this a better list, please file an issue and we can address that separately.

This list is just a sampling, so its not meant to be complete, but for whatever is listed in there we will do a unique applicability evaluation to ensure support.

PR #1642, opened 2017-08-19
Allow generate resources with assembly name has dashes

tarekgh:
As we generate cs code for accessing the embedded resources. The generated type will be named as FxResources.AsemblyName.SR. if the AssemblyName has characters not allowed in the namespace names (e.g. dashes), this will cause compilation errors. The fix here is we normalize the assembly name before we use it.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
@@ -75,7 +75,21 @@
     <GenerateResourcesCode
         ResxFilePath="$(StringResourcesPath)" 
         OutputSourceFilePath="$(IntermediateResOutputFileFullPath)" 
-        AssemblyName="$(AssemblyName)" />
+        AssemblyName="$(AssemblyName)" >
+
+        <Output TaskParameter="NormalizedAssemblyName" PropertyName="_NormalizedAssemblyName" />
+    </GenerateResourcesCode>
+
+    <ItemGroup>
+      <!-- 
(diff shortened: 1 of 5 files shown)


tarekgh:
CC @weshaggard @safern could you please have a look? Thanks.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -75,7 +75,21 @@
     <GenerateResourcesCode
         ResxFilePath="$(StringResourcesPath)" 
         OutputSourceFilePath="$(IntermediateResOutputFileFullPath)" 
-        AssemblyName="$(AssemblyName)" />
+        AssemblyName="$(AssemblyName)" >
+
+        <Output TaskParameter="NormalizedAssemblyName" PropertyName="_NormalizedAssemblyName" />
```
weshaggard: While the normalization might work wouldn't it be easier to just allow projects to set the name via some other property so we can control it for a given project as necessary?

tarekgh: That is a good idea too. we may consider have both, the override property and the normalization too.

tarekgh: @weshaggard I'll not go with your idea now because we'll need to have a way to handle such cases when executing the resources extraction that @safern did as we'll not have the needed information (e.g. extracting the resources of the runtime assemblies on WCF repro). I'll still think about the idea to see if we can have a reliable way to do it.

PR #1644, opened 2017-08-23
Add Microsoft.Dotnet.Build.Tasks.Feed toolset

karajas:
Initial PR for collecting feedback.
This introduces a set of tasks to control the input and output feeds of a repo. It currently only supports the creation of Transport Feeds.

/cc @weshaggard @chcosta @jaredpar

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -83,7 +83,6 @@
       <Version>$(MicrosoftDiaSymReaderConverterPackageVersion)</Version>
     </StaticDependency>
 
-    <NuGetDependency Include="NuGet.Client"/>
     <NuGetDependency Include="NuGet.Commands"/>
     <NuGetDependency Include="NuGet.Packaging"/>
     <NuGetDependency Include="NuGet.ProjectModel"/>
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
@@ -6,8 +6,6 @@ using Microsoft.Build.Framework;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Globalization;
-using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Threading;
@@ -32,7 +30,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
         private string _leaseId;
         private string _leaseUrl;
 
-        public AzureBlobLease(string accountName, string accountKey, string connectionString, string containerName, string blobName, Microsoft.Build.Utilities.TaskLoggingHelper log, string maxWait = null, string delay=null)
+        public AzureBlobLease(string accountName, string accountKey, string connectionString, string containerName, string blobName, Microsoft.Build.Utilities.TaskLoggingHelper log, string maxWait = null, string delay = null)
         {
             _accountName = accountName;
             _accountKey = accountKey;
@@ -45,7 +43,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             _leaseUrl = $"{AzureHelper.GetBlobRestUrl(_accountName, _containerName, _blobName)}?comp=lease";
         }
 
-        public void Acquire()
+        public string Acquire()
         {
             Stopwatch stopWatch = new Stopwatch();
             stopWatch.Start();
@@ -60,7 +58,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                     { AutoRenewLeaseOnBlob(this, _accountName, _accountKey, _containerName, _blobName, leaseId, _leaseUrl, _log); },
                       _cancellationTokenSource.Token);
                     _leaseId = leaseId;
-                    return;
+                    return _leaseId;
                 }
                 catch (Exception e)
                 {
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
@@ -1,0 +1,292 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using MSBuild = Microsoft.Build.Utilities;
+using Newtonsoft.Json;
+using NuGet.Versioning;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using System.IO;
+using Microsoft.DotNet.Build.CloudTestTasks;
+using System;
+using NuGet.Packaging;
+using System.Text.RegularExpressions;
+using System.Threading;
+
+namespace Microsoft.DotNet.Build.Tasks.Feed
+{
+    public sealed class BlobFeed
+    {
+        private MSBuild.TaskLoggingHelper Log;
+        public string AccountName { get; set; }
+        public string AccountKey { get; set; }
+        public string ContainerName { get; set; }
+        public string IndexDirectory { get; set; }
+
+        private static readonly CancellationTokenSource TokenSource = new CancellationTokenSource();
+        private static readonly CancellationToken CancellationToken = TokenSource.Token;
+
+        public BlobFeed(string accountName, string accountKey, string containerName, string indexDirectory, MSBuild.TaskLoggingHelper loggingHelper)
+        {
+            AccountName = accountName;
+            AccountKey = accountKey;
+            ContainerName = containerName;
+            IndexDirectory = indexDirectory;
+            Log = loggingHelper;
+        }
+
+        public string FeedUrl
+        {
+            get
+            {
+                return AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
+            }
+        }
+
+        public async Task<bool> CheckIfFeedExists()
+        {
+            string url = $"{FeedUrl}?restype=container";
+            using (HttpClient client = new HttpClient())
+            {
+                client.DefaultRequestHeaders.Clear();
+                var request = AzureHelper.RequestMessage("GET", url, AccountName, AccountKey).Invoke();
+                using (HttpResponseMessage response = await client.SendAsync(request))
+                {
+                    if (response.IsSuccessStatusCode)
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Container {ContainerName} exists for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    else
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Container {ContainerName} does not exist for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    return response.IsSuccessStatusCode;
+                }
+            }
+        }
+
+        public async Task<bool> CheckIfBlobExists(string blobPath)
+        {
+            string url = $"{FeedUrl}/{blobPath}?comp=metadata";
+            using (HttpClient client = new HttpClient())
+            {
+                client.DefaultRequestHeaders.Clear();
+                var request = AzureHelper.RequestMessage("GET", url, AccountName, AccountKey).Invoke();
+                using (HttpResponseMessage response = await client.SendAsync(request))
+                {
+                    if (response.IsSuccessStatusCode)
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Container {ContainerName} exists for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    else
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Container {ContainerName} does not exist for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    return response.IsSuccessStatusCode;
+                }
+            }
+        }
+
+        public string GenerateIndexes(IEnumerable<string> items, string relativePath)
+        {
+            Log.LogMessage(MessageImportance.Low, $"START generating indexes for {relativePath}");
+            foreach (var package in items)
+            {
+                Tuple<string, NuGetVersion> blobPath = GetNugetPackageVersion(package);
+                GeneratePackageServiceIndex(Path.Combine(relativePath, blobPath.Item1), new List<string>() { blobPath.Item2.ToFullString() });
+            }
+            Log.LogMessage(MessageImportance.Low, $"DONE generating indexes for {relativePath}");
+            return IndexDirectory;
+        }
+
+        public Tuple<string, NuGetVersion> GetNugetPackageVersion(string package)
+        {
+            using (var reader = new PackageArchiveReader(package))
+            using (var nuspecStream = reader.GetNuspec())
+            {
+                NuspecReader nuspec = new NuspecReader(nuspecStream);
+                return new Tuple<string, NuGetVersion>(nuspec.GetId(), nuspec.GetVersion());
+            }
+        }
+        public string CalculateRelativeUploadPath(string item, string relativePath)
+        {
+            string relativeBlobPath = CalculateBlobPath(item, relativePath);
+            if (Path.GetExtension(item) == ".nupkg")
+            {
+                Tuple<string, NuGetVersion> nugetPackage = GetNugetPackageVersion(item);
+                return relativeBlobPath.Substring(0, relativeBlobPath.IndexOf(nugetPackage.Item2.ToFullString()) - 1);
+            }
+            else
+            {
+                return relativeBlobPath.Substring(0, relativeBlobPath.IndexOf(Path.GetFileName(item)) - 1);
+            }
+        }
+
+        public string CalculateBlobPath(string item, string relativePath)
+        {
+            string blobPath = string.Empty;
+            if (Path.GetExtension(item) == ".nupkg")
+            {
+                Tuple<string, NuGetVersion> nugetPackage = GetNugetPackageVersion(item);
+                blobPath = Path.Combine(Path.Combine(relativePath, Path.Combine(nugetPackage.Item1, nugetPackage.Item2.ToFullString())), Path.GetFileName(item));
+            }
+            else
+            {
+                blobPath = item.Substring(item.IndexOf(relativePath));
+            }
+            return blobPath.ToLowerInvariant().Replace("\\", "/");
+        }
+
+        public string GeneratePackageServiceIndex(string redirectUrl, IEnumerable<string> values)
+        {
+            string pathToTempFolder = Path.Combine(IndexDirectory, redirectUrl);
+            if (Directory.Exists(pathToTempFolder))
+            {
+                Directory.Delete(pathToTempFolder, true);
+            }
+            Directory.CreateDirectory(pathToTempFolder);
+            string packageIndexJsonLocation = Path.Combine(pathToTempFolder, "index.json");
+            using (var streamWriter = new StreamWriter(File.OpenWrite(packageIndexJsonLocation)))
+            using (JsonTextWriter writer = new JsonTextWriter(streamWriter))
+            {
+                writer.WriteStartObject();
+                writer.WritePropertyName("versions");
+                writer.WriteStartArray();
+                foreach (string version in values)
+                {
+                    writer.WriteRawValue($"\"{version}\"");
+                }
+                writer.WriteEndArray();
+                writer.WriteEndObject();
+            }
+            return packageIndexJsonLocation;
+        }
+
+        private string GenerateRootServiceIndex(string redirectUrl)
+        {
+            string pathToTempFolder = Path.Combine(IndexDirectory, redirectUrl);
+            if (Directory.Exists(pathToTempFolder))
+            {
+                Directory.Delete(pathToTempFolder, true);
+            }
+            Directory.CreateDirectory(pathToTempFolder);
+            string rootIndexJsonLocation = Path.Combine(pathToTempFolder, "index.json");
+            using (FileStream fs = File.OpenWrite(rootIndexJsonLocation))
+            using (var streamWriter = new StreamWriter(fs))
+            using (JsonTextWriter writer = new JsonTextWriter(streamWriter))
+            {
+                writer.WriteStartObject();
+                writer.WritePropertyName("version");
+                writer.WriteRawValue("\"3.0.0\"");
+                writer.WritePropertyName("resources");
+                writer.WriteStartArray();
+                writer.WriteStartObject();
+                writer.WritePropertyName("@id");
+                writer.WriteRawValue($"\"{FeedUrl}/{redirectUrl}/\"");
+                writer.WritePropertyName("@type");
+                writer.WriteRawValue("\"PackageBaseAddress/3.0.0\"");
+                writer.WritePropertyName("comment");
+                writer.WriteRawValue("\"Base URL of Azure storage where NuGet package registration info for intermediaries is stored.\"");
+                writer.WriteEndObject();
+                writer.WriteEndArray();
+                writer.WritePropertyName("@context");
+                writer.WriteStartObject();
+                writer.WritePropertyName("@vocab");
+                writer.WriteRawValue("\"http://schema.nuget.org/schema#\"");
+                writer.WritePropertyName("comment");
+                writer.WriteRawValue("\"http://www.w3.org/2000/01/rdf-schema#comment\"");
+                writer.WriteEndObject();
+                writer.WriteEndObject();
+            }
+            return rootIndexJsonLocation;
+        }
+
+        public bool IsSanityChecked(IEnumerable<string> items)
+        {
+            Log.LogMessage(MessageImportance.Low, $"START checking sanitized items for feed");
+            foreach (var item in items)
+            {
+                if (items.Any(s => Path.GetExtension(item) != ".nupkg"))
+                {
+                    Log.LogError($"{item} is not a nupkg");
+                    return false;
+                }
+            }
+            List<string> duplicates = items.GroupBy(x => x)
+                    .Where(group => group.Count() > 1)
+                    .Select(group => group.Key).ToList();
+            if (duplicates.Count > 0)
+            {
+                Log.LogError($"Duplicates found: {duplicates}");
+                return false;
+            }
+            Log.LogMessage(MessageImportance.Low, $"DONE checking for sanitized items for feed");
+            return true;
+        }
+
+        public async Task<bool> CreateFeedContainer(string relativePath)
+        {
+            ValidateContainerName(ContainerName);
+            string url = $"{FeedUrl}?restype=container";
+            using (HttpClient client = new HttpClient())
+            {
+                Tuple<string, string> headerBlobType = new Tuple<string, string>("x-ms-blob-public-access", "blob");
+                List<Tuple<string, string>> additionalHeaders = new List<Tuple<string, string>>() { headerBlobType };
+                var createRequest = AzureHelper.RequestMessage("PUT", url, AccountName, AccountKey, additionalHeaders);
+
+                using (HttpResponseMessage response = await AzureHelper.RequestWithRetry(Log, client, createRequest))
+                {
+                    try
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            "Received response to create Container {0}: Status Code: {1} {2}",
+                            ContainerName, response.StatusCode, response.Content.ToString());
+                    }
+                    catch (Exception e)
+                    {
+                        Log.LogErrorFromException(e, true);
+                    }
+                }
+            }
+            string item = GenerateRootServiceIndex(relativePath);
+            string uploadPath = CalculateBlobPath(item, relativePath);
+            Log.LogMessage($"Uploading root index.json to {uploadPath}.");
+            try
+            {
+                UploadClient uploadClient = new UploadClient(Log);
+                await
+                    uploadClient.UploadBlockBlobAsync(
+                        CancellationToken,
+                        AccountName,
+                        AccountKey,
+                        ContainerName,
+                        item,
+                        uploadPath);
+            }
+            catch (Exception e)
+            {
+                Log.LogErrorFromException(e);
+            }
+            return !Log.HasLoggedErrors;
+        }
+
+        private void ValidateContainerName(string container)
+        {
+            if (container.Length < 3 || container.Length > 63 || !Regex.IsMatch(container, @"^[a-z0-9]+(-[a-z0-9]+)*$"))
+                throw new Exception("Container Name is invalid");
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -1,0 +1,164 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using MSBuild = Microsoft.Build.Utilities;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Threading;
+using System.IO;
+using Microsoft.DotNet.Build.CloudTestTasks;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.Build.Tasks.Feed
+{
+    sealed class BlobFeedAction
+    {
+        private MSBuild.TaskLoggingHelper Log;
+        private static readonly CancellationTokenSource TokenSource = new CancellationTokenSource();
+        private static readonly CancellationToken CancellationToken = TokenSource.Token;
+
+        public BlobFeed feed;
+        public int MaxClients { get; set; } = 8;
+
+        public BlobFeedAction(string accountName, string accountKey, string containerName, string indexDirectory, MSBuild.TaskLoggingHelper Log)
+        {
+            this.feed = new BlobFeed(accountName, accountKey, containerName, string.IsNullOrWhiteSpace(indexDirectory) ? Path.GetTempPath() : indexDirectory, Log);
+            this.Log = Log;
+        }
+
+        public async Task<bool> PushToFeed(IEnumerable<string> items, string relativePath, bool allowOverwrite = false)
+        {
+            if (feed.IsSanityChecked(items))
+            {
+                if (CancellationToken.IsCancellationRequested)
+                {
+                    Log.LogError("Task PushToFeed cancelled");
+                    CancellationToken.ThrowIfCancellationRequested();
+                }
+
+                using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
+                {
+                    await Task.WhenAll(items.Select(item => PushItemToFeed(item, relativePath, clientThrottle, allowOverwrite)));
+                }
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+
+        public async Task<bool> PushToFeedFlat(IEnumerable<string> items, string relativePath, bool allowOverwrite = false)
+        {
+            if (CancellationToken.IsCancellationRequested)
+            {
+                Log.LogError("Task PushToFeedFlat cancelled");
+                CancellationToken.ThrowIfCancellationRequested();
+            }
+
+            using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
+            {
+                await Task.WhenAll(items.Select(item => PushItemToFeed(item, relativePath, clientThrottle, allowOverwrite)));
+            }
+            return !Log.HasLoggedErrors;
+        }
+
+        public async Task<bool> PushItemToFeed(string item, string relativePath, SemaphoreSlim clientThrottle, bool allowOverwrite)
+        {
+            try
+            {
+                string uploadPath = feed.CalculateBlobPath(item, relativePath);
+                string packageDirectory = feed.CalculateRelativeUploadPath(item, relativePath);
+
+                if (await feed.CheckIfBlobExists(uploadPath) && !allowOverwrite)
+                {
+                    throw new Exception($"Item {uploadPath} already exists and Overwrite is false.");
+                }
+                await UploadAsync(CancellationToken, item, uploadPath, clientThrottle, allowOverwrite);
+
+                List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, packageDirectory);
+                if (!listAzureBlobs.Any(x => x.Contains(uploadPath)))
+                {
+                    throw new Exception($"Uploaded package {uploadPath} is not present on feed. Cannot update index.json.");
+                }
+
+                await UploadIndexJson(clientThrottle, true, packageDirectory, listAzureBlobs);
+            }
+            catch (Exception e)
+            {
+                Log.LogErrorFromException(e);
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+
+        private async Task UploadIndexJson(SemaphoreSlim clientThrottle, bool allowOverwrite, string packageDirectory, List<string> listAzureBlobs)
+        {
+            listAzureBlobs.Remove(listAzureBlobs.Find(x => x.Contains("index.json")));
+            List<string> updatedVersions = new List<string>();
+            foreach (var version in listAzureBlobs)
+            {
+                string versionToCheck = version.Substring(packageDirectory.Length + 1).Split('/')[0];
+                NuGetVersion nugetVersion = null;
+                if (NuGetVersion.TryParse(versionToCheck, out nugetVersion))
+                {
+                    updatedVersions.Add(versionToCheck);
+                }
+            }
+            string packageIndexJsonLocation = feed.GeneratePackageServiceIndex(packageDirectory, updatedVersions);
+            if (!await feed.CheckIfBlobExists(packageIndexJsonLocation))
+            {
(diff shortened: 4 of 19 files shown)


karajas:
TODO:

- [ ] 	Formatting changes
- [ ] 	Add the lease api call
- [ ] Add the option to push to a flat container (the code is all there, need to add the actual flag)

karajas:
PTAL.

karajas:
Error during restore.
```
17:26:23   Restoring all packages...
17:30:30 
EXEC : error : Failed to download package from 'https://api.nuget.org/v3-flatcontainer/runtime.aot.system.globalization/4.0.11/runtime.aot.system.globalization.4.0.11.nupkg'. [D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj]
17:30:30 EXEC : error : Response status code does not indicate success: 504 (Gateway Timeout). [D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj]
17:30:30 EXEC : error : Unable to load package 'runtime.aot.System.Globalization'. [D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj]
17:30:31 
D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj(41,5): error MSB3073: The command ""D:\j\workspace\Windows_NT_pr---6cb64db8\Tools/dotnetcli/dotnet.exe" restore --packages "D:\j\workspace\Windows_NT_pr---6cb64db8\packages" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://dotnet.myget.org/F/nuget-build/api/v3/index.json --source https://api.nuget.org/v3/index.json "D:\j\workspace\Windows_NT_pr---6cb64db8\src"" exited with code 1.
17:30:31 
17:30:31 
EXEC : error : Failed to download package from 'https://api.nuget.org/v3-flatcontainer/runtime.aot.system.globalization/4.0.11/runtime.aot.system.globalization.4.0.11.nupkg'. [D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj]
17:30:31 EXEC : error : Response status code does not indicate success: 504 (Gateway Timeout). [D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj]
17:30:31 EXEC : error : Unable to load package 'runtime.aot.System.Globalization'. [D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj]
17:30:31 
D:\j\workspace\Windows_NT_pr---6cb64db8\build.proj(41,5): error MSB3073: The command ""D:\j\workspace\Windows_NT_pr---6cb64db8\Tools/dotnetcli/dotnet.exe" restore --packages "D:\j\workspace\Windows_NT_pr---6cb64db8\packages" --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://dotnet.myget.org/F/nuget-build/api/v3/index.json --source https://api.nuget.org/v3/index.json "D:\j\workspace\Windows_NT_pr---6cb64db8\src"" exited with code 1.
```

dependencies.props
```diff
@@ -82,15 +82,17 @@
...
     <NuGetDependency Include="NuGet.Commands"/>
+    <NuGetDependency Include="NuGet.Common"/>
+    <NuGetDependency Include="NuGet.ContentModel"/>
     <NuGetDependency Include="NuGet.Packaging"/>
     <NuGetDependency Include="NuGet.ProjectModel"/>
     <NuGetDependency Include="NuGet.RuntimeModel"/>
     <NuGetDependency Include="NuGet.Versioning"/>
     <StaticDependency Include="@(NuGetDependency)">
-      <Version>4.3.0-preview2-4095</Version>
+      <Version>4.3.0-preview1-3937</Version>
```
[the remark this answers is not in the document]
karajas: NuGet.Client doesn't have version 4.3.0-preview2-4095(https://dotnet.myget.org/feed/nuget-build/package/nuget/NuGet.Client) and it requires that it's dependencies to be of the same version (4.3.0).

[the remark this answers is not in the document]
karajas: Actually, we may not require this dependency now though, will remove and try again.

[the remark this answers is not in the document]
karajas: Yup resolved, sorry conflated the two issues

karajas:
PTAL @weshaggard @chcosta

karajas:
An example for publishing: https://github.com/dotnet/corefx/compare/master...karajas:TransportFeed

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        public BlobFeedAction(string accountName, string accountKey, string containerName, MSBuild.TaskLoggingHelper Log)
+        {
+            this.feed = new BlobFeed(accountName, accountKey, containerName, Log);
+            this.Log = Log;
+        }
+
+        public Task<bool> DeleteFromFeed(string filterString = "")
+        {
+            throw new NotImplementedException();
```
weshaggard: Why are all these throwing NotImplementedException?

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
```diff
@@ -0,0 +1,27 @@
...
+    <id>Microsoft.DotNet.Build.Tasks.Feed</id>
+    <version>1.0.0-prerelease</version>
+    <title>Microsoft DotNet Build Feed tools </title>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
+    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=518631</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <summary>.NET Helix Cloud Test Execution Tools</summary>
```
weshaggard: This isn't the correct summary.

src/Microsoft.DotNet.Build.Tasks.Feed/ConfigureInputFeed.cs
```diff
@@ -0,0 +1,57 @@
...
+            nugetConfigBody += $"<fallbackPackageFolders>{Environment.NewLine}<clear />{Environment.NewLine}</fallbackPackageFolders>{Environment.NewLine}<packageSources>{Environment.NewLine}<clear />{Environment.NewLine}";
+            for(int i=0;i<EnableFeeds.Length;i++)
+            {
+                nugetConfigBody += $"<add key=\"inputFeed{i}\" value=\"{ EnableFeeds[i].ItemSpec }\" />{Environment.NewLine}";
+            }
+            nugetConfigBody += $"</packageSources>{Environment.NewLine}";
+            nugetConfigBody += $"<disabledPackageSources>{Environment.NewLine}";
+            for (int i = 0; i < DisableFeeds.Length; i++)
+            {
+                nugetConfigBody += $"<add key=\"disableFeed{i}\" value=\"{ DisableFeeds[i].ItemSpec }\" value=\"true\" />{Environment.NewLine}";
```
weshaggard: What cases do we care about DisabledFeeds? 

Unless I'm miss-understanding the spec https://docs.microsoft.com/en-us/nuget/schema/nuget-config-file#disabledpackagesources I think the feed needs to be listed in the package source list and the key be listed here. You cannot have two values. 

I personally would just removed the DisabledFeed stuff unless we have a scenario for it right now.

src/Microsoft.DotNet.Build.Tasks.Feed/IFeed.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Build.Tasks.Feed
+{
+    interface IFeed<T>
```
weshaggard: I'm not sure I fully understand the point of these interfaces. Who do you expect to use them? What other implementations do you expect?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -0,0 +1,249 @@
...
+                        writer.WritePropertyName("comment");
+                        writer.WriteRawValue("\"http://www.w3.org/2000/01/rdf-schema#comment\"");
+                        writer.WriteEndObject();
+                        writer.WriteEndObject();
+                    }
+                }
+            }
+        }
+
+        public bool IsSanityChecked(IEnumerable<string> items)
```
weshaggard: I don't see any users of this? What is its purpose exactly?

src/Microsoft.DotNet.Build.Tasks.Feed/FeedActionTask.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.DotNet.Build.Tasks.Feed
+{
+
+    public class FeedActionTask : MSBuild.Task
+    {
+
+        [Required]
+        public string Task { get; set; }
```
weshaggard: Task is overloaded term I think I would give it a better description. Is there any reason you don't just have two Tasks? One for Push and one for Configure?

src/Microsoft.DotNet.Build.Tasks.Feed/FeedActionTask.cs
```diff
@@ -0,0 +1,136 @@
...
+                        {
+                            Log.LogError($"No items to push. Please check ItemGroup ItemsToPublish.");
+                        }
+                        if (string.IsNullOrEmpty(RelativePath))
+                        {
+                            Log.LogWarning($"No relative path. Items are pushed to root of container.");
+                        }
+                        if (Type == "Blob") {
+                            //FeedAccountName is of the form - https://account.blob.core.windows.net/container/
+                            string removeHeader = FeedAccountName.Substring(FeedAccountName.IndexOf("https://")+8);
```
weshaggard: This parsing seems a little too magical and making heavy assumptions about the FeedAccountName input string.

src/Microsoft.DotNet.Build.Tasks.Feed/Models/IndexJson.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Build.Tasks.Feed
+{
+    public class IndexJson
+    {
+        public List<string> versions { get; set; }
```
weshaggard: nit: Should be capital V for public Properties.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+            else
+            {
+                throw new Exception("Push failed to due overwrite being false.");
+            }
+            return !Log.HasLoggedErrors;
+        }
+
+        public async Task<List<string>> ListAzureBlobs(string AccountName, string AccountKey, string ContainerName, string FilterBlobNames = "")
+        {
+            List<string> blobsNames = new List<string>();
```
weshaggard: We already have a number code paths that do things like list, upload, download, etc. We should be calling and sharing that code instead of duplicating it.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+                    }
+                }
+            }
+        }
+
+        private async Task MergeIndexes(CancellationToken ct, string item, string relativePath, List<string> indexJsonList)
+        {
+            //pull equivalent index.json
+            Tuple<string, NuGetVersion> blobPath = feed.CalculateBlobPath(item, relativePath);
+            string relativeBlobPath = Path.Combine(relativePath, blobPath.Item1).ToLowerInvariant().Replace("\\", "/");
```
weshaggard: I don't understand why we need to do this contains and pass in the index.json as opposed to just assuming the index.json exists in the directory and fetching it.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+            if (containerExists)
+            {
+                if (Directory.Exists(DownloadDirectory))
+                {
+                    Directory.Delete(DownloadDirectory, true);
+                }
+                Directory.CreateDirectory(DownloadDirectory);
+                using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
+                {
+                    List<string> listAzureBlobs = await ListAzureBlobs(feed.AccountName, feed.AccountKey, feed.ContainerName);
```
weshaggard: Why do we need to gather the list of index.json files? The convention is that it lives in the root of a package name directory correct? We should just use that and grab it instead of listing all blobs in the container and then parsing the strings.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+                        foreach(var version in generated)
+                        {
+                            if (!downloaded.Contains(version))
+                            {
+                                downloaded.Add(version);
+                            }
+                        }
+                        downloaded.Sort();
+
+                        JsonSerializer jsonSerializer = new JsonSerializer();
```
weshaggard: Why are we doing all this work in the scope of the HttpClient using block?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+            //pull equivalent index.json
+            Tuple<string, NuGetVersion> blobPath = feed.CalculateBlobPath(item, relativePath);
+            string relativeBlobPath = Path.Combine(relativePath, blobPath.Item1).ToLowerInvariant().Replace("\\", "/");
+            try
+            {
+                if (indexJsonList.Any(x => x.Contains(relativeBlobPath)))
+                {
+                    //download index.json to download folder
+                    string blob = indexJsonList.Find(x => x.Contains(relativeBlobPath));
+                    using (HttpClient client = new HttpClient())
```
weshaggard: We need to put this DownloadFromAzure into a helper type and use it instead of duplicating all this code.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+                    await Task.WhenAll(items.Select(item => MergeIndexes(CancellationToken, item, relativePath, indexJsonBlobs)));
+                }
+            }
+            if (CancellationToken.IsCancellationRequested)
+            {
+                Log.LogError("Task UploadToAzure cancelled");
+                CancellationToken.ThrowIfCancellationRequested();
+            }
+
+            List<string> indexList = new List<string>(Directory.EnumerateFiles(Path.Combine(IndexDirectory, relativePath), "index.json", SearchOption.AllDirectories));
```
weshaggard: You are duplicating the file path here and below that should be shared to prevent getting out of sync.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,389 @@
...
+                    {
+                        return item.Substring(item.IndexOf(relativePath)).Replace("\\", "/");
+                    };
+                    //upload all items
+                    await Task.WhenAll(items.Select(item => UploadAsync(CancellationToken, item, calculatePath(item), clientThrottle)));
+                }
+            }
+            else
+            {
+                throw new Exception("Push failed to due overwrite being false.");
```
weshaggard: Don't you need to know that it exists already before failing?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: I wanted to implement them at some point, should I comment them out or delete them and update the doc to reflect the actual apis needed?

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
karajas: fixed.

src/Microsoft.DotNet.Build.Tasks.Feed/IFeed.cs
karajas: My thinking was we might use different backing sources eg. MyGet, VSTS, Azure. I can get rid of this if needed.

weshaggard: It feels like a little bit of over-engineering at this point.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
karajas: I wanted a quick check for files that aren't nupkgs and are duplicated. Will invoke in the right place.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: Each package has an index.json.

karajas:
PTAL @weshaggard

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: fixed

src/Microsoft.DotNet.Build.Tasks.Feed/FeedActionTask.cs
karajas: fixed.

src/Microsoft.DotNet.Build.Tasks.Feed/FeedActionTask.cs
karajas: fixed.

src/Microsoft.DotNet.Build.Tasks.Feed/Models/IndexJson.cs
karajas: removed.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: fixed.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: fixed.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: removed.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: fixed.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: fixed.

karajas:
PTAL when you get a chance.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
```diff
@@ -29,7 +29,7 @@ public partial class AzureBlobLease
         private string _accountName;
         private string _accountKey;
         private Microsoft.Build.Utilities.TaskLoggingHelper _log;
-        private string _leaseId;
+        public string _leaseId { get; private set; }
```
chcosta: Properties are generally capitalized, and no '_'.  Why turn this into a property? I don't see it referenced anywhere.

karajas: I want to use the same instance of AzureBlobLease to acquire and renew the leaseid, and be able to pass it to the header of subsequent requests.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        public BlobFeed(string accountName, string accountKey, string containerName, MSBuild.TaskLoggingHelper loggingHelper)
+        {
+            AccountName = accountName;
+            AccountKey = accountKey;
+            ContainerName = containerName;
+            log = loggingHelper;
+        }
+
+        public string GetFeedUrl
```
chcosta: Given that this is a property and not a method, I would just name it `FeedUrl`

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            get
+            {
+                return $"https://{ AccountName }.blob.core.windows.net/";
+            }
+        }
+
+        public async Task<bool> CheckIfFeedExists()
+        {
+            string url = $"{ GetFeedUrl }{ContainerName}?restype=container";
```
chcosta: nit: inconsistent spacing

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -0,0 +1,217 @@
...
+            using (var nuspecStream = reader.GetNuspec())
+            {
+                NuspecReader nuspec = new NuspecReader(nuspecStream);
+                return new Tuple<string, NuGetVersion>(nuspec.GetId(), nuspec.GetVersion());
+            }
+        }
+
+        private void GeneratePackageServiceIndex(string accountName, string redirectUrl, string value)
+        {
+            string pathToTempFolder = Path.Combine(Directory.GetCurrentDirectory(), "tmp", redirectUrl);
```
weshaggard: Using the tmp directory a lot causes a bunch of garbage. What is the strategy for cleaning up the garbage?

karajas: We could add a flag that deletes after publish in this task, else we can have the automated cleanup in official builds handle this.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,179 @@
...
+            if (feed.IsSanityChecked(items))
+            {
+                if (CancellationToken.IsCancellationRequested)
+                {
+                    Log.LogError("Task PushToFeed cancelled");
+                    CancellationToken.ThrowIfCancellationRequested();
+                }
+
+                List<string> indexList = new List<string>(Directory.EnumerateFiles(Path.Combine(IndexDirectory, relativePath), "index.json", SearchOption.AllDirectories));
+                List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, string.Empty);
```
weshaggard: Looks like you are still listing all the blobs in the entire container. I don't understand why that is necessary. You should scope your list to just the blob for each item.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+
+        public async Task<bool> PushToFeedFlat(IEnumerable<string> items, string relativePath)
+        {
+            using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
+            {
+                List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, string.Empty);
```
weshaggard: Another place were you are listing all blobs in the container.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        private async Task UploadAsync(CancellationToken ct, string item, string uploadPath, SemaphoreSlim clientThrottle, List<string> listAzureBlobs)
+        {
+            if (!File.Exists(item))
+                throw new Exception(string.Format("The file '{0}' does not exist.", item));
+
+            await clientThrottle.WaitAsync();
+            string leaseId = string.Empty;
+            bool isLeaseRequired = listAzureBlobs.Contains(uploadPath);
```
weshaggard: Do we need a lease for all things or just the "index.json" files we are updating?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,179 @@
...
+            finally
+            {
+                clientThrottle.Release();
+            }
+
+            try
+            {
+                if (isLeaseRequired)
+                {
+                    blobLease.Release();
```
weshaggard: This release should be in the finally block above. We need to ensure that we release the lease if anything bad happens in the Upload as well.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: I think it would be better to check for all, in case we do have conflicting uploads on blobs during parallel builds

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,179 @@
...
+                    CancellationToken.ThrowIfCancellationRequested();
+                }
+
+                List<string> indexList = new List<string>(Directory.EnumerateFiles(Path.Combine(IndexDirectory, relativePath), "index.json", SearchOption.AllDirectories));
+                List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, string.Empty);
+                using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
+                {
+                    Func<string, string> calculatePath = (string item) =>
+                    {
+                        Tuple<string, NuGetVersion> blobPath = feed.CalculateBlobPath(item);
```
weshaggard: I think I would move this entire function into the feed.CalculateBlobPath and just add the relativePath as a parameter. That eliminates an inline function and extra closure/delegate calls here.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,179 @@
...
+        public BlobFeed feed;
+        public int MaxClients { get; set; } = 8;
+
+        public BlobFeedAction(string accountName, string accountKey, string containerName, MSBuild.TaskLoggingHelper Log)
+        {
+            this.feed = new BlobFeed(accountName, accountKey, containerName, Log);
+            this.Log = Log;
+        }
+
+        public async Task<bool> PushToFeed(IEnumerable<string> items, string relativePath, bool containerExists = false)
```
weshaggard: I think I would split this function to only work on a single item and the do the parallelism based on that new split function instead how you are currently splitting things.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,179 @@
...
+                        Tuple<string, NuGetVersion> blobPath = feed.CalculateBlobPath(item);
+                        string relativeBlobPath = Path.Combine(relativePath, blobPath.Item1, blobPath.Item2.ToFullString());
+                        return Path.Combine(relativeBlobPath, Path.GetFileName(item)).ToLowerInvariant().Replace("\\", "/");
+                    };
+                    await Task.WhenAll(items.Select(item => UploadAsync(CancellationToken, item, calculatePath(item), clientThrottle, listAzureBlobs)));
+
+                    if (containerExists)
+                    {
+                        listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, string.Empty);
+                        items.All(item => UpdateIndexes(item, relativePath, listAzureBlobs));
```
weshaggard: I would probably just add your new packages to the list instead of listing them from azure again.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
weshaggard: I think that is fine as long as it doesn't have a large performance penalty.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
weshaggard: You are currently putting that tmp directory under the current working directory which may or may not ever be cleaned up. I think it is probably best to have someone pass you in a directory to work with or to use the real temp directory.

karajas:
Addressed PR feedback, @weshaggard PTAL.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: Will fix.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -71,102 +59,89 @@ public BlobFeedAction(string accountName, string accountKey, string containerNam
...
+                await Task.WhenAll(items.Select(item => PushItemToFeed(item, relativePath, listAzureBlobs)));
             }
             return !Log.HasLoggedErrors;
         }
 
-        private async Task UploadAsync(CancellationToken ct, string item, string uploadPath, SemaphoreSlim clientThrottle, List<string> listAzureBlobs)
+        public async Task<bool> PushItemToFeed(string item, string relativePath, IEnumerable<string> listAzureBlobs)
+        {
+            string uploadPath = feed.CalculateBlobPath(item, relativePath);
+            using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
```
weshaggard: What is the point in the throttle here given you are uploading only one item?

karajas:
Moved the throttle out and made the upload check on a subset of the blobs based on prefix.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,154 @@
...
+                }
+                catch (Exception)
+                {
+                    Log.LogError($"Unable to obtain lease on {uploadPath}");
+                }
+            }
+            try
+            {
+                Log.LogMessage($"Uploading {item} to {uploadPath}.");
+                    UploadClient uploadClient = new UploadClient(Log);
```
weshaggard: nit indention.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -0,0 +1,244 @@
...
+                    return response.IsSuccessStatusCode;
+                }
+            }
+        }
+
+        public string GenerateIndexes(IEnumerable<string> items, string relativePath)
+        {
+            log.LogMessage(MessageImportance.Low, $"START generating indexes for {relativePath}");
+            GenerateRootServiceIndex(relativePath);
+            foreach (var package in items)
```
weshaggard: Only generate the root index here.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            if (feed.IsSanityChecked(items))
+            {
+                if (CancellationToken.IsCancellationRequested)
+                {
+                    Log.LogError("Task PushToFeed cancelled");
+                    CancellationToken.ThrowIfCancellationRequested();
+                }
+
+                List<string> indexList = new List<string>(Directory.EnumerateFiles(Path.Combine(feed.IndexDirectory, relativePath), "index.json", SearchOption.AllDirectories));
```
weshaggard: Lets not traverse here just use the items list. We should have a separate step to upload the root index json.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,154 @@
...
+                clientThrottle.Release();
+            }
+        }
+
+        private async Task<bool> UpdateIndexes(string item, string relativePath)
+        {
+            try
+            {
+                string searchPath = feed.CalculateRelativeUploadPath(item, relativePath);
+                List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, searchPath);
```
weshaggard: Should already have the blobs and we can pass them in.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,154 @@
...
+                if (listAzureBlobs.Any(x => x.Contains(searchPath)))
+                {
+                    List<string> versions = listAzureBlobs.FindAll(x => x.Contains(searchPath));
+                    versions.Remove(versions.Find(x => x.Contains("index.json")));
+                    versions = versions.Select(x =>
+                    {
+                        string version = x.Substring(searchPath.Length + 1);
+                        version = version.Substring(0, version.IndexOf("/"));
+                        return version;
```
weshaggard: Use Version.TryParse on the substring and if that succeeds add that version.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -0,0 +1,244 @@
...
+                            MessageImportance.Low,
+                            "Received response to create Container {0}: Status Code: {1} {2}",
+                            ContainerName, response.StatusCode, response.Content.ToString());
+                    }
+                    catch (Exception e)
+                    {
+                        log.LogErrorFromException(e, true);
+                    }
+                }
+            }
```
weshaggard: Should we add the root index.json file here?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -0,0 +1,93 @@
...
+                }
+                BlobFeedAction blobFeedAction = new BlobFeedAction(AccountName, AccountKey, ContainerName, IndexDirectory, Log);
+                bool containerExists = await blobFeedAction.feed.CheckIfFeedExists();
+                if (!containerExists)
+                {
+                    await blobFeedAction.feed.CreateFeedContainer();
+                }
+                if (!PublishFlatContainer)
+                {
+                    if (!containerExists)
```
weshaggard: plumb and allow overwriting of packages option to the task.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,154 @@
...
+                await Task.WhenAll(items.Select(item => PushItemToFeed(item, relativePath, clientThrottle)));
+            }
+            return !Log.HasLoggedErrors;
+        }
+
+        public async Task<bool> PushItemToFeed(string item, string relativePath, SemaphoreSlim clientThrottle)
+        {
+            string uploadPath = feed.CalculateBlobPath(item, relativePath);
+            List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, feed.CalculateRelativeUploadPath(item, relativePath));
+            await UploadAsync(CancellationToken, item, uploadPath, clientThrottle, listAzureBlobs.Contains(uploadPath));
```
weshaggard: Move the list after you upload the package and add an assertion that the package you just uploaded is in that list.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -0,0 +1,154 @@
...
+            return !Log.HasLoggedErrors;
```
weshaggard: Move the UpdateIndex and Uploading of the index to this function.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -64,16 +58,53 @@ public BlobFeedAction(string accountName, string accountKey, string containerNam
...
-            List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, feed.CalculateRelativeUploadPath(item, relativePath));
-            await UploadAsync(CancellationToken, item, uploadPath, clientThrottle, listAzureBlobs.Contains(uploadPath));
+            try
+            {
+                Tuple<string, NuGetVersion> nugetPackage = feed.GetNugetPackageVersion(item);
+                string uploadPath = feed.CalculateBlobPath(item, relativePath);
+                string searchPath = feed.CalculateRelativeUploadPath(item, relativePath);
+
+                List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, uploadPath);
+                if (listAzureBlobs.Contains(uploadPath) && !allowOverwrite)
```
weshaggard: Only do the list if allowOverwrite is false. Should just do a file exists instead of list.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -64,16 +58,53 @@ public BlobFeedAction(string accountName, string accountKey, string containerNam
...
+                {
+                    throw new Exception($"Item {uploadPath} already exists and Overwrite is false.");
+                }
+                await UploadAsync(CancellationToken, item, uploadPath, clientThrottle, allowOverwrite);
+
+                listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, searchPath);
+                if (!listAzureBlobs.Any(x => x.Contains(nugetPackage.Item2.ToFullString())))
```
weshaggard: Shouldn't this just be Contains(uploadpath)?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -64,16 +58,53 @@ public BlobFeedAction(string accountName, string accountKey, string containerNam
...
+        public async Task<bool> PushItemToFeed(string item, string relativePath, SemaphoreSlim clientThrottle, bool allowOverwrite)
         {
-            string uploadPath = feed.CalculateBlobPath(item, relativePath);
-            List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, feed.CalculateRelativeUploadPath(item, relativePath));
-            await UploadAsync(CancellationToken, item, uploadPath, clientThrottle, listAzureBlobs.Contains(uploadPath));
+            try
+            {
+                Tuple<string, NuGetVersion> nugetPackage = feed.GetNugetPackageVersion(item);
+                string uploadPath = feed.CalculateBlobPath(item, relativePath);
+                string searchPath = feed.CalculateRelativeUploadPath(item, relativePath);
```
weshaggard: searchPath = packageDirectory

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -64,16 +58,53 @@ public BlobFeedAction(string accountName, string accountKey, string containerNam
...
+                if (!listAzureBlobs.Any(x => x.Contains(nugetPackage.Item2.ToFullString())))
+                {
+                    throw new Exception($"Uploaded package {uploadPath} is not present on feed. Cannot update index.json.");
+                }
+
+                listAzureBlobs.Remove(listAzureBlobs.Find(x => x.Contains("index.json")));
+                List<string> updatedVersions = new List<string>();
+                foreach (var version in listAzureBlobs)
+                {
+                    string versionToCheck = version.Substring(searchPath.Length + 1);
```
weshaggard: do the substrings in one line.

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
```diff
@@ -0,0 +1,30 @@
...
+    <copyright>Copyright © Microsoft Corporation</copyright>
+    <tags></tags>
+    <dependencies>
+     <!-- Do not add any dependencies to this package as we want to install it versionless (i.e. -ExcludeVersion) and we don't want the dependencies to be installed that way -->
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll" target="lib" />
+    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib" Exclude="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll"/>
+    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib\desktop" />
```
weshaggard: Why is this "lib\desktop" it should be "lib\net45".

PR #1645, opened 2017-08-23
Allow for files to be copied to a new path while harvesting

ericstj:
/cc @weshaggard 

I ran into this when updating corefx to net471.

I need to copy a harvested asset into a runtime-specific folder since latest package added RID-split cross-compile.


weshaggard:
I thought we already supported that option via https://github.com/dotnet/corefx/blob/32f3e82596c8c35691b7cdbf1a27e27784acc652/src/System.Reflection.TypeExtensions/pkg/System.Reflection.TypeExtensions.pkgproj#L15.

ericstj:
That's a move not a copy.

PR #1647, opened 2017-08-24
Tool-runtime should use MS.NETCore.App version built into CLI

ericstj:
/cc @mellinoe @weshaggard 

See https://github.com/dotnet/corefx/pull/23513#issuecomment-324726971


ericstj:
Confirmed this resolved the problem for me

PR #1649, opened 2017-08-25
Make ExtractResWResourcesFromAssemblies derive from BuildTask

ericstj:
This is necessary so that the AssemblyResolver unifies dependencies.

/cc @safern @weshaggard


ericstj:
See https://github.com/dotnet/corefx/pull/23513#issuecomment-325027350

PR #1650, opened 2017-08-28
Complete roslyn update to 2.0.0-rc4

ericstj:
Previously f99e5ef88ac3e6b909676619a309147283f7c241 updated the roslyn package restored
by the tool-runtime to 2.0.0-rc4 but missed updating these two other places.

The result was that the new compiler was used on core but not desktop-MSBuild.

This change completes the update.

/cc @weshaggard @benaadams

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <RoslynVersion>2.0.0-rc</RoslynVersion>
+    <RoslynVersion>2.0.0-rc4</RoslynVersion>
     <RoslynPackageName>Microsoft.Net.Compilers</RoslynPackageName>
     <RoslynTargetsPath>$(ToolRuntimePath)</RoslynTargetsPath>
   </PropertyGroup>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -11,7 +11,7 @@ set PACKAGES_DIR=%PACKAGES_DIR:"=%
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
 set MICROBUILD_VERSION=0.2.0
 set PORTABLETARGETS_VERSION=0.1.1-dev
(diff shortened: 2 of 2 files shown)


ericstj:
FYI, I've tested picking up this update on the latest CoreFx and it now builds clean after https://github.com/dotnet/corefx/pull/23559

jkotas:
When I have tried to pick this up for CoreRT, I am getting errors about mismatched `System.Collections.Immutable` version like:

```
  D:\corert\Tools\sign.targets(110,5): error : D:\corert\bin/obj/Windows_NT.x64.Debug/ILCompiler.DependencyAnalysisFramework\ILCompiler.DependencyAnalysisFramework.dll: I/O error reading or writing PE file: Could not load file or assembly 'System.Collections.Immutable, Version=1.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. [D:\corert\src\ILCompiler.DependencyAnalysisFramework\src\ILCompiler.DependencyAnalysisFramework.csproj]
```

ericstj:
That doesn't look like mismatched: it looks like it didn't find the file.  Did you restore the tool-runtime?

Can you confirm that S.C.I exists in the tools/net46 folder?

If so, then the likely cause is that the new roslyn lifted the dependency and this task isn't unifying its dependencies and it is the first to load that assembly in the process.  Similar to this: https://github.com/dotnet/buildtools/commit/56cd202cf1dd9b885489593905fdf34396c476b7. I would have expected a ref-def mismatch rather than a missing file.

ericstj:
Also, you were likely broken by https://github.com/dotnet/buildtools/commit/f99e5ef88ac3e6b909676619a309147283f7c241 and not this change, since this change doesn't effect the tool-runtime binaries, only the targets and the roslyn subfolder.

jkotas:
> Can you confirm that S.C.I exists in the tools/net46 folder?

Yes, both S.C.I and S.R.M exist in the tools/net46 folder. They are the exact same versions (1.2.1 and 1.4.1) as in the older buildtools that work fine.

> the likely cause is that the new roslyn lifted the dependency and this task isn't unifying its dependencies

Yes, this looks to be the case. What would be the best way to fix it?

ericstj:
I'm putting in a fix into BuildTools to make all the tasks use the AssemblyResolver.  That should avoid these types of issues in the future.  Do you have a repro of the corert failure?  Would you be able to test a private build?

jkotas:
The repro is to point CoreRT repo to latest buildtools via BuildToolsVersion.txt and then run build.cmd.  Yes, I can test a private build.

ericstj:
Ok, 
```
set BUILDTOOLS_SOURCE=https://www.myget.org/F/ericstj-staging/api/v3/index.json
```

Then update BuildToolsVersion.txt to 2.0.0-prerelease-01928-0

Delete bin and tools, then build.

jkotas:
Works fine for CoreRT. Thanks!

PR #1651, opened 2017-08-28
Allow projects to disable the Roslyn in BuildTools.

eerhardt:
When using BuildTools in a repo like core-setup, where we want to use the dotnet-cli to build managed projects, we need to be able to not use the Roslyn properties that come in BuildTools.

For example, setting `<RoslynTargetsPath>$(ToolRuntimePath)</RoslynTargetsPath>` breaks any SDK project using the dotnet-cli.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -52,7 +52,7 @@
 
   <Import Project="$(MSBuildThisFileDirectory)BuildVersion.targets" />
 
-  <Import Project="$(MSBuildThisFileDirectory)Roslyn.Common.props" />
+  <Import Project="$(MSBuildThisFileDirectory)Roslyn.Common.props" Condition="'$(DisableBuildToolsRoslynVersion)' != 'true'" />
 
   <!-- Restore commands -->
   <PropertyGroup>


weshaggard:
@eerhardt This hook is present because we need to be able to push all our repo's to a newer csc which likely isn't in the SDK yet. While I don't mind the extensibility hook we need to be sure we are using the same compiler throughout our product.

Is there a better fix to allow the SDK based projects to use the downloaded Roslyn instead?

eerhardt:
The issue is that the BuildTools copy of `Microsoft.CSharp.Core.targets` is hacked up and incompatible with the real product that we ship.

In the call to the `<Csc` task, it uses a task parameter that doesn't exist:

`OverrideToolHost="$(OverrideToolHost)"`

Thus, when using BuildTools + an SDK project, you get the error:

```
F:\git\core-setup\Tools\Microsoft.CSharp.Core.targets(106,11): error MSB4064: The "OverrideToolHost" parameter is not supported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property. [F:\git\core-setup\src\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj]
F:\git\core-setup\Tools\Microsoft.CSharp.Core.targets(67,5): error MSB4063: The "Csc" task could not be initialized with its input parameters.  [F:\git\core-setup\src\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj]
```

Looking, I can't even find the checked-in code for this task.  https://github.com/search?utf8=%E2%9C%93&q=OverrideToolHost+&type=Code

At the end of the day, even if we had support for using a specific Roslyn version, or the BuildTools downloaded version, something here needs to change.  This `OverrideToolHost` issue blocks any SDK-based project from using BuildTools.  And instead of bringing in more and more MSBuild .targets files from BuildTools to try to hack it further, it is better to dogfood the actual shipping product.  Thus, using the `Microsoft.CSharp.Core.targets` that ships with the .NET SDK.

The change here allows SDK-based projects to compile.  It doesn't affect any of our existing (non-SDK based) projects, since they will never set this property.

##

> we need to be able to push all our repo's to a newer csc which likely isn't in the SDK yet.

BuildTools is on `2.0.0-rc4` - from 7 months ago

https://github.com/dotnet/buildtools/blob/fe78c74748285023c47910a768177a7845755872/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L4

The dotnet/cli is currently on `2.3.2-beta1-61921-05`. 
https://github.com/dotnet/cli/blob/9af126356bf662020374e693388d4c16c26f7362/build/DependencyVersions.props#L6.

Getting new Roslyn versions into the .NET SDK isn't that hard.

> we need to be sure we are using the same compiler throughout our product.

I agree that is a goal to strive for, but we are not doing that today.  The two managed libraries in core-setup are built with a pre-release version of the .NET SDK, which contains a pre-release Roslyn.  This change doesn't make it any worse.

> Is there a better fix to allow the SDK based projects to use the downloaded Roslyn instead?

Instead of trying to find a way to get the .NET SDK to use the downloaded Roslyn that BuildTools has, we should push a feature into the product that is the real way to get the .NET SDK to use a specific version of Roslyn.  That way other projects/users can benefit from this feature.

I think I remember seeing support for this, but I can't remember where.  @jaredpar - Do you know of the official way we tell our customers how to use a specific version of Roslyn?

jaredpar:
> I think I remember seeing support for this, but I can't remember where. @jaredpar - Do you know of the official way we tell our customers how to use a specific version of Roslyn?

From the perspective of BuildTools I don't know a way. Roslyn can insert custom compiler builds into all of our core products: CLI, MSBuild, CoreXT, etc ...  But I'm unfamiliar with how BuildTools packages and overrides the compiler here hence can't provide an answer. 

I'd more ask the opposite way: why isn't BuildTools just using one of the supported ways of distributing Roslyn? Those we take a lot of time to support and would be able to provide good answers too.

eerhardt:
@jaredpar - that's what I was asking about - how do normal users of "dotnet build" use a custom/new Roslyn version?

Do they PackageReference some Roslyn NuGet package that does it?  Is there doc on it somewhere?

jaredpar:
> @jaredpar - that's what I was asking about - how do normal users of "dotnet build" use a custom/new Roslyn version?

This is how we do it today:

- PackageReference to Microsoft.NetCore.Compilers. This NuGet package is setup such that CLI can use our binaries. 
- Import the Microsoft.NetCore.Compilers.props file from the build directory before any common build props files. 

There is a bit more complication if you want to flip between desktop and CLI builds this way. The logic is mostly captured here:

https://github.com/dotnet/roslyn/blob/master/build/Targets/Settings.props#L47

weshaggard:
For reference we have issue https://github.com/dotnet/corefx/issues/17131 to address this problem (should be in buildtools and not corefx). We just haven't funded the work yet. I definitely want to do this in a more supported way, probably very similar to what Roslyn is doing.

eerhardt:
Can I merge this change to unblock my core-setup work?

weshaggard:
>Can I merge this change to unblock my core-setup work?

I'm fine with this as a temporary workaround but I would like to make sure we fix core-setup to use the same Roslyn targets once we fix the other issue https://github.com/dotnet/corefx/issues/17131.

eerhardt:
Agreed.

Issue #1587, opened 2017-06-27
Test path doesn't include build configuration


danmosemsft:
After running S.R.E.Tests for netcoreapp, I tried to run them for ILC. Unfortunately they share a single 'netstandard' output folder which will contain the wrong corelib:

`System.Private.CoreLib : Gatekeeper error GK0028 : GK0028 Assembly 'System.Private.CoreLib' is a mixed mode assembly containing managed and native code.  Only pure managed assemblies are compatible with .NET Native.`

To fix this I had to delete c:\git\corefx\bin\Windows_NT.AnyCPU.Debug\System.Runtime.Extensions.Tests\netstandard\ and rebuild the test folder. 

The problem here is that the test path doesn't include the build configuration, it only has the target group used to build and this could be the same for more than one vertical. 

@joperezr

joperezr:
cc: @weshaggard 

Now that we are actually messing with the contents in the test folder (in uap and uapaot we call mklink for several things) and also since the actual RunTests.cmd produced there is different across different verticals, we should add build configuration to the testpath, such that we still support building and running tests across different verticals on the same enlistment without the need to be cleaning out folders.

weshaggard:
We should have two outputs here one is the build output which is `c:\git\corefx\bin\Windows_NT.AnyCPU.Debug\System.Runtime.Extensions.Tests\netstandard\` and the other is where we execute the tests from which should be based on the configuration we are running the tests from. We used to have that split but we may have lost it with the engineering work but that is what we should do to segregate these assets.

joperezr:
that sounds fair, and yes, we did lose that when we moved to dev/eng work.

safern:
Buildtools changes handled in: https://github.com/dotnet/buildtools/pull/1652

CoreFX changes handled in: https://github.com/dotnet/corefx/pull/23633

PR #1652, opened 2017-08-29
Add necessary configurations in order to binplace the tests per buildconfiguration

safern:
This needs some changes in dir.targets in CoreFX to enable it. 

Contributes to: https://github.com/dotnet/buildtools/issues/1587

## Why is this necessary?

This is necessary in order to be able to have a test working directory different than the output directory. The output directory is shared across target groups if the test assemblies target netstandard or uap, but when running in different runtimes (uap, netcoreapp, etc.) dependencies in order to be able to run the tests change and we fill the directory with a bunch of items that are specific to that target framework. This causes to have some clashes in the bin when trying to run the same test assembly against different target frameworks without having a clean repo/output test directory. (i.e without this changes if I want to run a test project against both uap and uapaot, assuming I already built vertical for both, I would have to run tests against one TFM and then delete the test's output directory and rerun them against the other one).

This change will avoid that, and will allow to have a specific test working directory different than the output directory. This can be per tfm, pero build configuration or however the consuming repo sets this up. 

## New Behavior with this change

With this change we will have the next behavior when building/running tests:

- When building the tests we will still bin place the test assets to the `OutputDirectory` (`corefx\bin\OSGroup.Configuration\TestProject\TargetGroup/`)
- When running the tests we will bin place the tests assets to this new [`TestPath`](https://github.com/dotnet/corefx/pull/23633/files#diff-4de71e7a0fd952be821f8af2768281dcR34) which will be the `TestPath`. This new path will be the test's working directory where we will bin place the RunTests.cmd and necessary dependencies in tests.targets. Also the `testResults.xml` will be found here after the test execution.
- `TestPath` is set to be `bin/tests/<TestProject>/<BuildConfiguration>/` (i.e `corefx\bin\tests\System.Collections.Immutable.Tests`) in `dir.props`
- This `TestPath` will be added to the `BinPlaceDir` in `FrameworkTargeting.targets` when `BinPlaceTest` is set to true. 
- Then in the [`BinPlaceFiles`](https://github.com/safern/buildtools/blob/4d39247ff2c152ac50ade6128a44afc6f9993ca2/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets#L243) target the test assets will be hardlinked to the `TestPath`.

This will isolate every test execution per build configuration and will allow people to run the same test project for all of the supported configurations on the same repo without having to clean the repo.

cc: @weshaggard @ericstj

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -345,7 +345,7 @@
     <ItemGroup>
       <_currentBinPlaceConfigurations Include="@(_bestBinlaceConfigurations)" Condition="'%(Identity)' == '$(Configuration)' OR '%(Identity)-$(ConfigurationGroup)' == '$(Configuration)'" />
 
-      <BinPlaceDir Condition="'$(BinPlaceTest)' == 'true'" Include="@(_currentBinPlaceConfigurations->'%(TestWorkingPath)')" />
+      <BinPlaceDir Condition="'$(BinPlaceTest)' == 'true'" Include="@(_currentBinPlaceConfigurations->'%(TestPath)')" />
       <BinPlaceDir Condition="'$(BinPlaceRuntime)' == 'true'" Include="@(_currentBinPlaceConfigurations->'%(RuntimePath)')" />
       <BinPlaceDir Condition="'$(BinPlaceRef)' == 'true'" Include="@(_currentBinPlaceConfigurations->'%(RefPath)')" />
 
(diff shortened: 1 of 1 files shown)


ericstj:
Seems reasonable.  Can you explain the end-to-end solution in the corefx issue?  I didn't get that by reading the thread (Wes had a suggestion but I didn't see how you intend to implement it).  Given that you'll be merging things in pieces it would help to understand the full end-to-end before you commit the parts.  We've had a big problem in the past where folks grow the test infra organically and it becomes a messier with every change.

Also, this commit message misuses the term *configurations.*  You're adding *support* to binplace tests.  Avoid terms like *necessary* since that doesn't really help explain why its necessary.

safern:
> Can you explain the end-to-end solution in the corefx issue?

@ericstj I've added the description on my corefx PR. Could you please review both PRs? 

https://github.com/dotnet/corefx/pull/23633#issue-253522995

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -376,6 +377,7 @@
     <AdditionalCleanDirectories Include="@(BinPlaceConfiguration->'%(RuntimePath)')" />
     <AdditionalCleanDirectories Include="@(BinPlaceConfiguration->'%(PackageFileRefPath)')" />
     <AdditionalCleanDirectories Include="@(BinPlaceConfiguration->'%(PackageFileRuntimePath)')" />
+    <AdditionalCleanDirectories Include="@(BinPlaceConfiguration->'%(TestWorkingPath)')" />
```
ericstj: Consider naming this TestPath for parallelism with RefPath and RuntimePath.

ericstj:
nit: the commit description is still missing detail and uses the term *necessary* without ever describing why its necessary.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
safern: I will. I just named it different to differentiate it from the <TestPath> property we use in tests.targets

safern:
I've updated the description to this PR. I will add that description to the merge commit when merging the PR after I address your PR feedback.

safern:
I've updated the comment and added the details into the commit message.

PR #1653, opened 2017-08-29
Fix race condition because of assembly name normalization

tarekgh:
We used to generate the normalized assembly name in the target GenerateResourcesSource and we override the resw name of the resource file there too. In the incremental build the target GenerateResourcesSource will get skipped but overriding the resw item name still be done as it is used outside the target in other places. This will cause the normalized name property not initialized but it is used in setting the resw name. and that will cause a race condition because many projects will try to access the exact same resw file name while building.
The fix is to move the assembly name normalization outside GenerateResourcesSource so we’ll be sure we always initialize the normalized assembly name property.

Fixes #23618

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
@@ -31,17 +31,13 @@ namespace Microsoft.DotNet.Build.Tasks
         [Required]
         public string AssemblyName { get; set; }
 
-        [Output]
-        public string NormalizedAssemblyName { get; set; }
-
         public bool DebugOnly { get; set; }
 
         public override bool Execute()
         {
             try
             {
-                NormalizedAssemblyName = Helper.NormalizeAssemblyName(AssemblyName);
-                _resourcesName = "FxResources." + NormalizedAssemblyName;
+                _resourcesName = "FxResources." + AssemblyName;
 
                 using (_targetStream = File.CreateText(OutputSourceFilePath))
                 {
--- a/src/Microsoft.DotNet.Build.Tasks/NormalizeAssemblyName.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/NormalizeAssemblyName.cs
@@ -8,7 +8,7 @@ using System;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
-    public class NormalizeAssemblyName : Task
+    public class NormalizeAssemblyName : BuildTask
     {
         [Required]
         public string AssemblyName { get; set; }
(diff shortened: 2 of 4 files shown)


tarekgh:
CC @ericstj @weshaggard @safern

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -31,17 +31,13 @@ public class GenerateResourcesCode : Task
...
-
         public bool DebugOnly { get; set; }
 
         public override bool Execute()
         {
             try
             {
-                NormalizedAssemblyName = Helper.NormalizeAssemblyName(AssemblyName);
-                _resourcesName = "FxResources." + NormalizedAssemblyName;
+                _resourcesName = "FxResources." + AssemblyName;
```
ericstj: Doesn't this have other side-effects?

ericstj: I see you pass in the normalized value.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets
```diff
@@ -62,34 +63,40 @@
...
 
-  <Target Name="GenerateResourcesSource"
-          Condition="'$(StringResourcesPath)'!='' AND '$(OmitResources)'!='true'"
-          Inputs="$(StringResourcesPath)"
-          Outputs="$(IntermediateResOutputFileFullPath)">
-
-    <GenerateResourcesCode
-        ResxFilePath="$(StringResourcesPath)" 
-        OutputSourceFilePath="$(IntermediateResOutputFileFullPath)" 
+  <Target Name="AssemblyNameNormalization" Condition="'$(StringResourcesPath)'!='' AND '$(OmitResources)'!='true'">
```
ericstj: Naming is bad.  Can you follow the same pattern we use elsewhere: name the target after the one task it calls?

src/Microsoft.DotNet.Build.Tasks/NormalizeAssemblyName.cs
```diff
@@ -0,0 +1,34 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class NormalizeAssemblyName : Task
```
safern: It should inherit from BuildTask so that when we cross compile it for netfx it has the AssemblyResolver magic.

tarekgh: although we don't need this one here as I am not adding any new dependencies but I'll do it anyway.

tarekgh:
@ericstj @safern let me know if you have any more comments. Thanks.

ericstj:
Please validate the fix by building a src project twice and ensuring that a new resw isn't written to the bin folder.

tarekgh:
> Please validate the fix by building a src project twice and ensuring that a new resw isn't written to the bin folder.

Yes I already did that with @safern

safern:
> Yes I already did that with @safern

Yes we did a repro without the fix and got a repro and then followed the same steps with the fix and works perfectly.

PR #1654, opened 2017-08-29
Make all BuildTools tasks derive from BuildTask

ericstj:
BuildTask provides our common desktop support in a base class by
hooking AssemblyResolve and providing runtime unification for assemblies
referenced by the task.

/cc @jkotas @weshaggard


ericstj:
I tested this by consuming in corefx and it was successful.  Asked @jkotas to test on CoreRT which is currently broken per https://github.com/dotnet/buildtools/pull/1650#issuecomment-325522726

ericstj:
Did so for me on CoreRT as well.  Merging

jkotas:
@ericstj Thanks!

PR #1656, opened 2017-08-29
Clean up fixes for Microsoft.Dotnet.Build.Tasks.Feed

karajas:
@weshaggard @chcosta


karajas:
Merging as these are fixes that were asked for in a previous pr.

PR #1659, opened 2017-08-31
Add support for arch-specific `ActiveIssue`s.

pgavlin:
This allows the user to disable a particular test when targeting a
specific platform without disabling the test for other platforms. We
intend to use this in CoreCLR to disable tests that e.g. fail on x86 or
ARM but not on other platforms.

### DIFF
--- a/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
@@ -18,7 +18,9 @@ namespace Xunit
         public ActiveIssueAttribute(string issue, TestPlatforms platforms) { }
         public ActiveIssueAttribute(int issueNumber, TargetFrameworkMonikers framework) { }
         public ActiveIssueAttribute(string issue, TargetFrameworkMonikers framework) { }
-        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)0) { }
-        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)0) { }
+        public ActiveIssueAttribute(int issueNumber, TestArchitectures architectures) { }
+        public ActiveIssueAttribute(string issueNumber, TestArchitectures architectures) { }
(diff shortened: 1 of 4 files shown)


pgavlin:
@stephentoub @weshaggard PTAL

pgavlin:
Is there any way to test the changes that I've made? I didn't notice anything at-a-glance.

weshaggard:
>Is there any way to test the changes that I've made? I didn't notice anything at-a-glance.

There are no tests for this in BuildTools unfortunately. The way I usually test it is to build a local copy and then copy over the local tools version my corefx build.

pgavlin:
>There are no tests for this in BuildTools unfortunately. The way I usually test it is to build a local copy and then copy over the local tools version my corefx build.

Okay, I'll give that a shot before merging.

src/xunit.netcore.extensions/Attributes/ActiveIssueAttribute.cs
```diff
@@ -18,7 +18,9 @@ public class ActiveIssueAttribute : Attribute, ITraitAttribute
         public ActiveIssueAttribute(string issue, TestPlatforms platforms) { }
         public ActiveIssueAttribute(int issueNumber, TargetFrameworkMonikers framework) { }
         public ActiveIssueAttribute(string issue, TargetFrameworkMonikers framework) { }
-        public ActiveIssueAttribute(int issueNumber, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)0) { }
-        public ActiveIssueAttribute(string issue, TestPlatforms platforms = TestPlatforms.Any, TargetFrameworkMonikers framework = (TargetFrameworkMonikers)0) { }
+        public ActiveIssueAttribute(int issueNumber, TestArchitectures architectures) { }
+        public ActiveIssueAttribute(string issueNumber, TestArchitectures architectures) { }
```
danmosemsft: nit: `string issue`

danmosemsft:
@pgavlin can we merge this - did you test it? if so you can just hit merge.

danmosemsft:
@pgavlin ?

krwq:
ping here since it would be useful to use it in #32710

safern:
In order for this to be enabled in corefx and other repos that are moving to arcade, this now needs to be added into arcade. If I have some spare time I can grab it -- but no promises 😄

ViktorHofer:
Closing as xunit isn't present in buildtools anymore.

PR #1661, opened 2017-09-05
Updating the version of the roslyn compilers we use for building

joperezr:
cc: @weshaggard @jaredpar @VSadov 

Updating the version of roslyn compilers we use for building the repos. With these set of changes I'm able to build corefx in Windows and non-Windows succesfully.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -40,9 +40,14 @@
     <OverrideToolHost Condition="'$(RunningOnUnix)' != 'true'">$(DotnetCliPath)dotnet.exe</OverrideToolHost>
     <BuildToolsTaskDir Condition="'$(RunningOnCore)'!='true'">$(ToolsDir)net46/</BuildToolsTaskDir>
     <BuildToolsTaskDir Condition="'$(RunningOnCore)'=='true'">$(ToolsDir)</BuildToolsTaskDir>
+    <RoslynIncompatibleMsbuildVersion Condition="'$(MSBuildToolsVersion)' == '2.0' OR
+                                                 '$(MSBuildToolsVersion)' == '3.5' OR
+                                                 '$(MSBuildToolsVersion)' == '4.0' OR
+                                                 '$(MSBuildToolsVersion)' == '12.0' OR
+                                                 '$(MSBuildToolsVersion)' == '14.0'">true</RoslynIncompatibleMsbuildVersion>
     <UseRoslynCompilers Condition="'$(UseRoslynCompilers)'=='' and '$(RunningOnUnix)'=='true'">false</UseRoslynCompilers>
     <!-- UseSharedCompilation is not supported in MSBuild 14.0 with the new version of roslyn, so make sure we are not using that for building -->
-    <UseSharedCompilation Condition="'$(UseSharedCompilation)' == '' and '$(RunningOnUnix)' != 'true' and '$(UseRoslynCompilers)' != 'false' and '$(VisualStudioVersion)' != '14.0'">true</UseSharedCompilation>
+    <UseSharedCompilation Condition="'$(UseSharedCompilation)' == '' and '$(RunningOnUnix)' != 'true' and '$(UseRoslynCompilers)' != 'false' and '$(RoslynIncompatibleMsbuildVersion)' != 'true'">true</UseSharedCompilation>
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -41,7 +41,7 @@
     <BuildToolsTaskDir Condition="'$(RunningOnCore)'!='true'">$(ToolsDir)net46/</BuildToolsTaskDir>
     <BuildToolsTaskDir Condition="'$(RunningOnCore)'=='true'">$(ToolsDir)</BuildToolsTaskDir>
     <UseRoslynCompilers Condition="'$(UseRoslynCompilers)'=='' and '$(RunningOnUnix)'=='true'">false</UseRoslynCompilers>
-    <UseSharedCompilation Condition="'$(UseSharedCompilation)' == '' and '$(RunningOnUnix)' != 'true' and '$(UseRoslynCompilers)' != 'false'">true</UseSharedCompilation>
+    <UseSharedCompilation>false</UseSharedCompilation>
```
jaredpar: That's really going to hurt build throughput. Why is this happening?

joperezr: yeah, I've seen builds are much slower than before, but this is required since this is a very recent build of roslyn which doesn't support msbuild 14.0 anymore. since we don't want to drop support from that version of msbuild just yet, we have to add this workaround and manually set csc path or else we would get an error like: `Microsoft.Net.Compilers is only supported on MSBuild v15.0 and above`

joperezr: Would you suggest to have something like this instead:
```xml
<UseSharedCompilation Condition="'$(UseSharedCompilation)' == '' and '$(RunningOnUnix)' != 'true' and '$(UseRoslynCompilers)' != 'false' and '$(VisualStudioVersion)' != '14.0'">true</UseSharedCompilation>
```

Where I check the VisualStudioVersion to see if its 14.0 or not?

eerhardt: > since we don't want to drop support from that version of msbuild just yet

When do we plan on dropping support for it?  VS 2017 has been out for 6 months.

joperezr: just chatted with @weshaggard about that, the problem is that these changes are in buildtools and we don't want to assume that it's ok to drop support of VS2015 on all of the repos that consume buildtools. Being a general repo that gets consumed by a lot of repos, we rather keep these changes compatible with msbuild 14.0 for the moment.

jaredpar: Thought the point of BuildTools was a light weight way to provide tooling to repos. But it can't support the latest C# compiler? That seems bad.

joperezr: That's why I just changed the condition, what we want is to support all versions of C# compiler and let the repo decide which one to use, as opposed to only supporting the newer version or the older version.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -105,6 +105,10 @@ cp -R "${__PACKAGES_DIR}"/[Mm]icro[Bb]uild.[Cc]ore/"${__MICROBUILD_VERSION}/buil
 # Copy some roslyn files over
 cp $__TOOLRUNTIME_DIR/runtimes/any/native/* $__TOOLRUNTIME_DIR/
 
+#Temporarily rename roslyn compilers to have exe extension
```
weshaggard: Please file a follow-up issue for this.

VSadov:
@joperezr - There is a newer compiler package available - 2.6.0-rdonly-ref-62106-01 , but that adds some more checks and could potentially require source changes.
I'd prefer to get build working and then update to the newer version. - Just to do one thing at a time.

Once this change is in, we should move to a newer version.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: #1669

joperezr:
temporarily rolling this back to unblock buildtools update in corefx, will re -add it once I have the changes ready in corefx for the update.

joperezr:
cc: @ericstj

ericstj:
Thanks @joperezr

PR #1663, opened 2017-09-06
Remove reference assembly attribute from partial ref facade when it is a full facade

ericstj:
Not all desktop tooling handles reference assemblies.  This helps reduce the number of reference assemblies we produce: if an assembly contains entirely type-forwards we'll strip the reference assembly attribute.

/cc @weshaggard

### DIFF
--- a/src/GenFacades/GenFacades.Core/GenFacades.cs
+++ b/src/GenFacades/GenFacades.Core/GenFacades.cs
@@ -489,6 +489,15 @@ namespace GenFacades
                             return forwardedTypes.TryGetValue(oldType.DocId(), out newType) ? newType : oldType;
                         });
 
+                        var remainingTypes = assembly.AllTypes.Where(t => t.Name.Value != "<Module>");
+
+                        if (!remainingTypes.Any())
+                        {
+                            Trace.TraceInformation($"Removed all types from {contractAssembly.Name} thus will remove ReferenceAssemblyAttribute.");
+                            assembly.AssemblyAttributes.RemoveAll(ca => ca.FullName() == "System.Runtime.CompilerServices.ReferenceAssemblyAttribute");
+                            assembly.Flags &= ~ReferenceAssemblyFlag;
+                        }
+
                         typeRefRewriter.Rewrite(assembly);
                     }
                 }


ericstj:
With this change, the following no longer get ReferenceAssemblyAttribute:
```
System.Threading.Tasks.Extensions
Microsoft.Win32.Registry
System.CodeDom
System.ComponentModel.Annotations
System.Configuration.ConfigurationManager
System.Data.Odbc
System.Data.SqlClient
System.IO.Packaging
System.IO.Ports
System.Security.AccessControl
System.Security.Cryptography.Pkcs
System.Security.Cryptography.ProtectedData
System.Security.Cryptography.Xml
System.Security.Principal.Windows
System.ServiceProcess.ServiceController
System.IO.Packaging
System.Security.Cryptography.Cng
```

We can decide if we want to reship those in servicing.

src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -489,6 +489,14 @@ public Assembly GenerateFacade(IAssembly contractAssembly, IAssemblyReference se
                             return forwardedTypes.TryGetValue(oldType.DocId(), out newType) ? newType : oldType;
                         });
 
+                        var remainingTypes = assembly.AllTypes.Where(t => t.Name.Value != "<Module>");
+
+                        if (!remainingTypes.Any())
+                        {
+                            Trace.TraceInformation($"Removed all types from {contractAssembly.Name} thus will remove ReferenceAssemblyAttribute.");
+                            assembly.AssemblyAttributes.RemoveAll(ca => ca.FullName() == "System.Runtime.CompilerServices.ReferenceAssemblyAttribute");
```
nguerrera: IIRC, this is informational, and the real bit lives in assemblyflags. Am I misremembering or did the compiler/runtime change?

weshaggard: correct the assemblyflags is the real bit.

weshaggard: but we should also remove the ReferenceAssemblyAttribute.

ericstj: I see, let me strip that as well.

ericstj: We're defining that as 0x70, are all those bits part of the reference assembly flags?  I can't find this in the ecma spec.

ericstj: It uses both the attribute and the flags:
https://github.com/dotnet/coreclr/blob/13e7c4368da664a8b50228b1a5ef01a660fbb2dd/src/vm/pefile.inl#L128-L159

https://github.com/dotnet/coreclr/blob/cb8cfba3b61f18f81787322f0a2563d118c26b8a/src/inc/corhdr.h#L736-L779

Those flags actually correspond to the processor architecture.  0x70 means "no platform".  Clearing it means unspecified.

PR #1664, opened 2017-09-07
Update release/2.0.0 to current 

karajas:
Between corefx, coreclr and core-setup the latest commit is https://github.com/dotnet/buildtools/commit/80da51aa59ae5fe14fbce1894183bf8d5c6d7b42. 

Also included in this changeset is the change to the versioning of BuildTools for servicing out of this branch.

/cc @ericstj @weshaggard


karajas:
Will also update with cherry picked fixes that are required.

ericstj:
Do you plan on going through these with Shiproom?  Have you reviewed them for anything that looks concerning for servicing?

weshaggard:
To be fair the only commits that need to be looked at are between https://github.com/dotnet/buildtools/commit/3162466ab961d8bc851a60d24c3a80649ddc631c (version 2.0.0-prerelease-01721-01) and https://github.com/dotnet/buildtools/commit/80da51aa59ae5fe14fbce1894183bf8d5c6d7b42 (version 2.0.0-prerelease-01903-01)

There are a lot less changes between those 2 commits. I don't know why the release/2.0.0 branch is currently on the older Apr 24th commit.

ericstj:
Agreed.

It's on the older April 24th commit because I was the only one to update it, which I did during one of the preview releases when we needed to make fixes after forking from master.  I raised the issue during the stabilization of RTM (when I was out of office) and other folks were going to do it but never got around to it.  Thus the current state.

karajas:
1. Updated shiproom. 
2. Going through the commits it doesn't seem like anything major that would potentially break the build.
3. Validated the build on my machine.

I also moved the other packages to servicing naming... Most of them should be consumed by the buildtools package but we may have separate consumers for CloudTest, VSTSRepoAPI, VersionTools, etc. This may help with servicing down the line.

karajas:
PTAL.

weshaggard:
LGTM

PR #1665, opened 2017-09-08
Adding SourceLink Support (source code from GitHub) for dotnet builds

vancem:
You have to set the variable UseSourceLink to activate it.  Thus this change has no effect unless you opt in.   

Also older C# compilers this will fail unless you generate portable PDBs (later C# compilers support MSPDB).  This may limit when you can turn this on.  

There are additional issues getting this working end-to-end but I think this is all we need in build-tools.

I have been testing this on the corefx and coreclr repo.  I need a build of build-tools however before those repos can be updated.   

@dagood, @tmat, @mikem8361

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
@@ -8,5 +8,13 @@
     <InformationalVersion Condition="'$(InformationalVersion)' == ''">$(AssemblyFileVersion)$(BuiltByString). $(SrcCodeRef)</InformationalVersion>
     <Product Condition="'$(Product)' == ''">Microsoft%AE .NET Framework</Product>
     <AssemblyTitle Condition="'$(AssemblyTitle)' == ''">$(AssemblyName)</AssemblyTitle>
+	<!-- File passed to C# compiler that defines the mapping from source to GIT URL.  Created in versioning.targets --> 
+    <SourceLinkFilePath>$(BaseIntermediateOutputPath)source_link.json</SourceLinkFilePath>
   </PropertyGroup>
+
+  <PropertyGroup Condition="Exists('$(SourceLinkFilePath)')">
+    <!-- Defining this variable tells the C# compiler to use this as the /sourcelink option value. -->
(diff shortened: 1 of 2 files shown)


tmat:
Looks good to me, but I'm not familiar with the build tools.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
```diff
@@ -8,5 +8,13 @@
     <InformationalVersion Condition="'$(InformationalVersion)' == ''">$(AssemblyFileVersion)$(BuiltByString). $(SrcCodeRef)</InformationalVersion>
     <Product Condition="'$(Product)' == ''">Microsoft%AE .NET Framework</Product>
     <AssemblyTitle Condition="'$(AssemblyTitle)' == ''">$(AssemblyName)</AssemblyTitle>
+	<!-- File passed to C# compiler that defines the mapping from source to GIT URL.  Created in versioning.targets --> 
+    <SourceLinkFilePath>$(BaseIntermediateOutputPath)source_link.json</SourceLinkFilePath>
   </PropertyGroup>
+
+  <PropertyGroup Condition="Exists('$(SourceLinkFilePath)')">
```
dagood: I think it makes sense to check `UseSourceLink` here too, just in case someone wants to turn off source linking but doesn't know to delete the file.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
```diff
@@ -8,5 +8,13 @@
     <InformationalVersion Condition="'$(InformationalVersion)' == ''">$(AssemblyFileVersion)$(BuiltByString). $(SrcCodeRef)</InformationalVersion>
     <Product Condition="'$(Product)' == ''">Microsoft%AE .NET Framework</Product>
     <AssemblyTitle Condition="'$(AssemblyTitle)' == ''">$(AssemblyName)</AssemblyTitle>
+	<!-- File passed to C# compiler that defines the mapping from source to GIT URL.  Created in versioning.targets -->
```
dagood: Looks like a tab character got mixed in rather than four spaces.

PR #1668, opened 2017-09-11
Add TargetFrameworkMonikers.IsInAppContainer

danmosemsft:
Currently tests that require appcontainer filter with TFM.IsUapNotUapAot even though they may not care whether the configuration is Aot. Add TFM.IsInAppContainer so they can express this intent. If we later add support for AppContainer in Aot configuration runs, we don't need to update the tests, only the filtering.

relates to https://github.com/dotnet/corefx/issues/19931


danmosemsft:
Hmm maybe this isn't the right approach as IsInAppContainer is not really a TFM. Instead, tests that specifically want "in app container" should filter on PlatformDetection.IsWinRT  (aka IsInAppContainer)

PR #1671, opened 2017-09-12
Ensure deps file contains correct extension

ericstj:
Previously the GenerateDepsJson task was assuming all assemblies were
dlls, however that is not the case; xunit.console.netcore.exe is an exe.

This didn't seem to cause an issue if the file existed in the
application folder.  If xunit.console.netcore.exe existed in the app folder, the host would
ignore the incorrect entry in the framework deps file.

If, however, the file was missing from the app folder and had incorrect
extension, the host would fail at startup.

This change fixes the deps file to have the correct extension.

The only diff after making this fix can be seen here: https://gist.github.com/ericstj/abdd9ba519dff947c3ac8910c9f37324/revisions#diff-4ed1ec087ae487fff3bdda792419a18e


ericstj:
@dotnet-bot test this please

karajas:
Thanks for picking this up, LGTM

PR #1682, opened 2017-09-19
Update the include list for dumpling

DrewScoggins:
This adds saving off the full runtime that we unpack as part of running
the tests as well as the system libs that live in lib and usr/lib.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
@@ -5,6 +5,10 @@
     </Content>
     <DumplingIncPaths Include="$(RuntimePath)" />
     <DumplingIncPaths Include="$(TestPath)" />
+    <DumplingIncPaths Condition="'$(TargetOS)'!='Windows_NT'" Include="/lib/x86_64-linux-gnu/" />
+    <DumplingIncPaths Condition="'$(TargetOS)'!='Windows_NT'" Include="/usr/lib/x86_64-linux-gnu/" />
+    <DumplingIncPaths Condition="'$(TargetOS)'!='Windows_NT'" Include="$HELIX_CORRELATION_PAYLOAD" />
+    <DumplingIncPaths Condition="'$(TargetOS)'!='Windows_NT'" Include="/lib64/" />
   </ItemGroup>
   <Target Name="PreinstallDumpling"
           BeforeTargets="TestAllProjects">


DrewScoggins:
@mellinoe @danmosemsft PTAL

danmosemsft:
@janvorli @mikem8361 @safern anything look missing?

danmosemsft:
How large are these on typical test machines? On my test ubuntu box usr/lib/x86_64-linux-gnu is 1.1GB but it has some enormous libs that maybe test machines wont have. Does dumpling trigger when tests crash on my own boxes? Perhaps there's a way to  be more specific, or have a size cap? Is there any file type we can filter for maybe?

mikem8361:
I’m not sure copying everything in this list of directories in the right approach.  I thought at some point the scripts Scott created would load the core dump in lldb and upload only the modules referenced in the core dump.  It would certainly be a lot smaller footprint and work across all our supported distros which have slightly different directory structure sometimes.

mikem

mellinoe:
Perhaps we should just start with `$HELIX_CORRELATION_PAYLOAD`? It wasn't clear to me from the metting whether the rest of it was needed to debug in regular circumstances. I guess I need clarification on that part.

danmosemsft:
What is `$HELIX_CORRELATION_PAYLOAD` ?

From the meeting, my understanding is that if we don't have the binaries matching the stack frames, we will not be able to unwind the stack, possibly even including other frames. Thus today we require the matching distro, which is so painful that nobody debugs test dumps. 

So if that understanding is right, we need to solve this properly.

@DrewScoggins can dumpling enumerate the loaded modules, from the dump? Did @schaabs have that working?

safern:
> What is $HELIX_CORRELATION_PAYLOAD

This is the test runtime folder that contains the shared framework and all its managed symbols. It also contains all the necessary lib*.so files and SOS

DrewScoggins:
Yeah, we will need the $HELIX_CORRELATION_PAYLOAD. For the other directories, I just looked at all of the static dependencies that libcoreclr.so had, and included the directories that the dependencies lived in. When I looked on my local machine the size of all the directories zipped was about 240MB. That does not seem prohibitive, but if we think that is too much to be saving per dump we can try and walk it back.

As for dumpling loading the dump and finding the dependencies, I am not sure if that ever worked, but I am fairly certain that it does not work right now. Just looking at the generated zips with the dump files we do not have everything that we need.  I could believe that we could add this feature or get it working, but that seems like something that we could do going forward.

mikem8361:
The problem with stack unwinding is if a different version of the runtime module is on the machine the core dump is loaded.  If the module doesn’t exist, lldb will use the in-memory/core image to get the unwind info.

mikem

DrewScoggins:
Adding @janvorli. From talking with him he made it seem that we needed to include all of the system libraries (stuff in lib/x86_64-linux-gnu) in order to properly load the core dumps across machines. Is this not right?

DrewScoggins:
Also after looking at the dumpling source code it does look like we used to grab dependencies based on the files that were needed from loading the dump, but this feature was removed when dumpling moved to async dump uploads as a perf requirement.  We should certainly add this feature back, but this seems like a good stop-gap until then.

mellinoe:
How big is the entire folder we would be uploading on our CI machines?

DrewScoggins:
I just looked on one of our lab machines they have a size of 288MB uncompressed. This is about what the size was on my local VM.  Adding these folders would about double the size of the dumps that we are uploading, just to offer some context.  Although we could likely remove some of the stuff that we are saving as we may have some duplicate stuff.

mellinoe:
That amount seems fine to me.

PR #1685, opened 2017-09-20
Improve UAP Runner logging and enable showprogress flag

safern:
So re-opening the PR as I've fixed the timeout issue. The issue was a deadlock in the UI thread when initializing the app since I was blocking it when trying to get the StreamWriter for the log in the same thread as the UI. 

The fix was to move the log to the `XunitTestRunner` class since we don't need to log anything from the `MainPage`

I've ran private runs to confirm this.

cc: @tarekgh @danmosemsft @joperezr


safern:
It is easier to review by reviewing the last commit, since the other 3 didn't change and they were already reviewed in my previews PR.

tarekgh:
LGTM

safern:
Thanks @tarekgh I will publish a new package and will merge once that is done to update it in corefx.

safern:
The package is in myget so I will go ahead and merge, and update corefx side.

PR #1687, opened 2017-09-20
Changes required in order to move corefx to 2.0.0 CLI

joperezr:
cc: @weshaggard @karajas @ericstj 

Changes required to move corefx to the 2.0.0 CLI version and build succesfully.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
@@ -110,52 +110,56 @@ namespace Microsoft.DotNet.Build.Tasks
                 }
             }
 
-            targetsSection.Remove(runtimeTarget);
-            newDepsJson.Remove("libraries");
+            string platformPackageName = "";
+            foreach (var package in targetsSection[runtimeTarget].Children())
+            {
+                JProperty packageProperty = package as JProperty;
+                if (packageProperty != null && packageProperty.Name.Contains($"runtime.{rid}.Microsoft.NETCore.App"))
+                {
+                    platformPackageName = packageProperty.Name;
+                    break;
+                }
+            }
+
+            if (string.IsNullOrEmpty(platformPackageName))
+                Log.LogError($"Unable to find platform package that matches runtime.{rid}.Microsoft.NETCore.App when generating deps.json file.");
+
+            JObject platformPackage = (JObject)targetsSection[runtimeTarget][platformPackageName];
+            platformPackage.Remove("runtime");
+            platformPackage.Remove("native");
+            JObject platformRuntime = new JObject();
+            JObject platformNative = new JObject();
+
             foreach (var assembly in assemblyNames)
             {
-                JObject runtimes = new JObject();
-                JObject runtimeLocation = new JObject();
-                string key = $"{assembly.AssemblyName.Name}/{assembly.AssemblyName.Version.Major}.{assembly.AssemblyName.Version.Minor}.{assembly.AssemblyName.Version.Build}";
-                runtimeLocation.Add(assembly.FileName, new JObject());
-                runtimes.Add("runtime", runtimeLocation);
                 try
                 {
-                    targetValue.Add(key, runtimes);
-                }
-                catch (System.ArgumentException)
-                {
-
-                }
-
-                try
-                {
-                    libraryValue.Add(key, ConstructLibraryNode());
+                    platformRuntime.Add($"runtimes/{rid}/lib/netcoreapp2.0/{assembly.FileName}", new JObject());
                 }
                 catch (System.ArgumentException)
                 {
 
                 }
             }
-            targetsSection.Add(runtimeTarget, targetValue);
-            newDepsJson.Add("libraries", libraryValue);
+            // Add mscorlib and System facades to platformRuntime
+            platformRuntime.Add($"runtimes/{rid}/lib/netcoreapp2.0/mscorlib.dll", new JObject());
+            platformRuntime.Add($"runtimes/{rid}/lib/netcoreapp2.0/System.dll", new JObject());
 
-            // Delete mscorlib.dll references comming from CoreCLR package. They do not exist anymore.
-            var mscorlibProperties = new List<JProperty>();
-            foreach (var item in newDepsJson.Descendants())
+            foreach (var nativeAssembly in nonManagedAssemblyFilePaths)
             {
-                var property = item as JProperty;
-                if (property == null)
-                    continue;
+                try
+                {
+                    if (nativeAssembly.EndsWith(".dll"))
+                        platformNative.Add($"runtimes/{rid}/native/{nativeAssembly}", new JObject());
+                }
+                catch (System.ArgumentException)
+                {
 
-                var name = property.Name;
-                if (name.EndsWith("/lib/netstandard1.0/mscorlib.dll") || name.EndsWith("/native/mscorlib.ni.dll"))
-                    mscorlibProperties.Add(property);
+                }
             }
-            foreach (var item in mscorlibProperties)
-            {
-                item.Remove();
-            }
+
+            platformPackage.Add("runtime", platformRuntime);
+            platformPackage.Add("native", platformNative);
 
             if (!string.IsNullOrEmpty(OutputPath))
             {
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
@@ -149,7 +149,7 @@ namespace Microsoft.DotNet.Build.Tasks
             {
                 try
                 {
-                    if (nativeAssembly.EndsWith(".dll"))
+                    if (!nativeAssembly.EndsWith(".pdb"))
                         platformNative.Add($"runtimes/{rid}/native/{nativeAssembly}", new JObject());
                 }
                 catch (System.ArgumentException)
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
```diff
@@ -110,52 +110,56 @@ public override bool Execute()
...
+                JProperty packageProperty = package as JProperty;
+                if (packageProperty != null && packageProperty.Name.Contains($"runtime.{rid}.Microsoft.NETCore.App"))
+                {
+                    platformPackageName = packageProperty.Name;
+                    break;
+                }
+            }
+
+            if (string.IsNullOrEmpty(platformPackageName))
+                Log.LogError($"Unable to find platform package that matches runtime.{rid}.Microsoft.NETCore.App when generating deps.json file.");
```
weshaggard: I take it there was too much work to make both scenarios work?

src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
```diff
@@ -110,52 +110,56 @@ public override bool Execute()
...
-                    targetValue.Add(key, runtimes);
+                    platformRuntime.Add($"runtimes/{rid}/lib/netcoreapp2.0/{assembly.FileName}", new JObject());
                 }
                 catch (System.ArgumentException)
                 {
 
                 }
+            }
+            // Add mscorlib and System facades to platformRuntime
+            platformRuntime.Add($"runtimes/{rid}/lib/netcoreapp2.0/mscorlib.dll", new JObject());
```
weshaggard: Why are these special? They should be in the shared framework.

src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
```diff
@@ -110,52 +110,56 @@ public override bool Execute()
...
+            platformRuntime.Add($"runtimes/{rid}/lib/netcoreapp2.0/System.dll", new JObject());
 
+            foreach (var nativeAssembly in nonManagedAssemblyFilePaths)
+            {
                 try
                 {
-                    libraryValue.Add(key, ConstructLibraryNode());
+                    if (!nativeAssembly.EndsWith(".pdb"))
```
weshaggard: Why do we need to check for pdb's here but not in the managed assembly case?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BclRewriter.runtimeconfig.json
```diff
@@ -2,7 +2,7 @@
   "runtimeOptions": {
     "framework": {
       "name": "Microsoft.NETCore.App",
-      "version": "1.1.0"
+      "version": "2.0.0"
```
weshaggard: I"m fine moving these up to 2.0.0 but we should probably make changes in Buildtools to overwrite this version with whatever we restore otherwise we will have to keep BuildTools in lock step with what CLI we are using. That can be done in another change if needed.

weshaggard: Actually it cannot be done in another change because we don't want folks to not be able to consume Buildtools after your commit and before they move to the 2.0.0 SDK.

src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
joperezr: I didn't try given that I was moving all of the runtimeconfig.jsons to point to 2.0 so that meant that the old scenario wouldn't work anymore. Now that I'll try to make that happen on a script, I'll make both scenarios work here. If it's not easy to autodetect which deps.json should be generated, I could always fallback to an extra parameter to indicate which generation of deps.json should be built.

src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
joperezr: Because these two are the manual shims that get built *after* src.builds is done, which means that they won't be present in the folder by the time that we run this task. That is why manually I added the two shims into the json. You do bring up a good point which is the case of a full rebuild scenario, I believe it will fail given that the platformRuntime will already have a mscorlib and System.dll assemblies. I'll fix this.

src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
joperezr: nativeAssembly is basically a bag where we throw anything that is not a managed assembly, which means that all the pdbs are going here along with the native binaries. That is why we only have the check in the nativeAssemblies.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/BclRewriter.runtimeconfig.json
joperezr: I'll make this change as we discussed yesterday.

joperezr:
With these changes, I have now a successful build on Windows and non-Windows in corefx when calling build.cmd/sh

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
```diff
@@ -12,6 +12,7 @@ foreach ($file in Get-ChildItem $ToolRuntimePath *.runtimeconfig.json)
 {
     Write-Host "Correcting runtime version of" $file.FullName
     $text = (Get-Content $file.FullName) -replace "1.1.0","$highestVersion"
+    $text = (Get-Content $file.FullName) -replace "2.0.0-preview1-001913-00","$highestVersion"
```
weshaggard: Instead of this why not make all the versions consistent?

src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
```diff
@@ -29,6 +29,8 @@ public FileNameAssemblyPair(AssemblyName assemblyName, string fileName)
             }
         }
 
+        public bool GenerateLegacyDepsJson { get; set; }
```
weshaggard: If we want to be backwards compatible then we need to reverse the default.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
joperezr: I was going to do that but then I saw that you explicitly set this the msbuild.runtimeconfig.json only to this version, which led me to think that you did that for a very specific reason which is why I chose to instead fix the scripts. If there is no particular reason why you updated it on PR: #1460 then I'll just go ahead and move it back to 1.1.0 as well.

src/Microsoft.DotNet.Build.Tasks/GenerateDepsJson.cs
joperezr: you are right, will do.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -109,9 +109,14 @@ cp $__TOOLRUNTIME_DIR/runtimes/any/native/* $__TOOLRUNTIME_DIR/
 cp ${__TOOLRUNTIME_DIR}/csc.dll ${__TOOLRUNTIME_DIR}/csc.exe
 cp ${__TOOLRUNTIME_DIR}/vbc.dll ${__TOOLRUNTIME_DIR}/vbc.exe
 
+#Copy RID specific assets to the tools dir since we don't have a deps.json for .NETCore msbuild
+cp ${__TOOLRUNTIME_DIR}/runtimes/unix/lib/netstandard1.3/*.dll $__TOOLRUNTIME_DIR/
```
ericstj: If you published for the current RID you wouldn't need to do this.  Do we have the RID at this point?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
weshaggard: I actually don't recall that was probably me porting back what was changed in the individual repos. I suspect it doesn't matter as we are updating anyway.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: I don't think that we know the publish RID at this point, which is why we publish only using the framework but not the RID. That said, the only library that really is RID specific out of all of the things we publish is System.Text.Encoding.CodePages.dll which only has a Win and Unix implementation, so I guess that we could just use unix-x64 everywhere and that should probably work.

ericstj: What about using the portable rid (win|linux|osx-<arch>) as well as letting it be specified?  Make sure selfcontained isn't set to true (so that you don't get another copy of the shared framework).

weshaggard: @joperezr can you file a tracking issue for this as I agree we should consider other options then making this assumption but for now I want to make forward progress on the SDK update.

joperezr: #1697

PR #1688, opened 2017-09-21
Avoid linefeed in /proc/self/coredump_filter

am11:
On Alpine Linux, the default shell is ash. To build .NET Core, we
explicitly install bash.

If we run the following command:

```sh
echo 0x3F > /proc/self/coredump_filter
```

in bash, it raises the exception:

> echo: write error: Invalid argument

while with ash, it works fine.

Problem is that, when `echo` is run from `bash`, it adds a trailing
linefeed which kernel seems to reject.

Adding `-n` avoids adding the linefeed.

Original exception while building CoreRT:

```sh
chmod +x /corert/Tools/dotnetcli//dotnet
/corert/bin/Linux.x64.Debug/ILCompiler.DependencyAnalysisFramework.Tests/RunTests.sh: line 67: echo: write error: Invalid argument
/corert/Tools/dotnetcli//dotnet xunit.console.netcore.exe ILCompiler.DependencyAnalysisFramework.Tests.dll  -xml testResults.xml -notrait Benchmark=true -notrait category=nontests -notrait category=nonlinuxtests  -notrait category=OuterLoop -notrait category=failing
```


am11:
cc @jkotas

PR #1690, opened 2017-09-21
Add Mono to TargetFrameworkMonikers

EgorBo:
For several namespaces we try to reuse .NET Core sources including xunit tests and in most cases mono can pretend to be NetFramework
but in some places it completely uses .NET Core code and behaves more like .NET Core rather than .NET.

Example: https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Numerics/tests/BigInteger/Comparison.cs#L369-L373 - we can't pretend to be NetFramework here as we also have that netcore fix for this case.

/cc: @marek-safar @akoeplinger

### DIFF
--- a/src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
+++ b/src/xunit.netcore.extensions/Discoverers/ActiveIssueDiscoverer.cs
@@ -51,6 +51,8 @@ namespace Xunit.NetCore.Extensions
             {
                 if (frameworks.HasFlag(TargetFrameworkMonikers.NetFramework))
                     yield return new KeyValuePair<string, string>(XunitConstants.Category, XunitConstants.NonNetfxTest);
+                if (frameworks.HasFlag(TargetFrameworkMonikers.Mono))
+                    yield return new KeyValuePair<string, string>(XunitConstants.Category, XunitConstants.NonMonoTest);
                 if (frameworks.HasFlag(TargetFrameworkMonikers.Netcoreapp))
                     yield return new KeyValuePair<string, string>(XunitConstants.Category, XunitConstants.NonNetcoreappTest);
                 if (frameworks.HasFlag(TargetFrameworkMonikers.UapNotUapAot))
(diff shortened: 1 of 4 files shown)


marek-safar:
@karelz could you help us to move this forward

karelz:
@ericstj @weshaggard can you please take a look at this one?

ericstj:
Seems OK to me, we've done similar things for psuedo-TFMs in the past (aot / corert).  I do think this will end up in a lot of `SkipOnTargetFramework(Mono)` sprinkled throughout the repo next to the same for desktop and netcore.  Presumably we are OK with that? /cc @danmosemsft

EgorBo:
@ericstj currently there are just a few tests (3 afair) need this attribute (we use xunit tests for System.Linq, System.Numerics, System.Reflection.Emit and a few other namespaces).

danmosemsft:
We'll now need a build tools update. @joperezr this is OK to do now?

joperezr:
Not yet. I haven't made the change yet to have buildtools go to use the Open key and disable signing so buildtools build is still down.

EgorBo:
@joperezr @danmosemsft I see packages are released very often but I don't see this change in them :(
https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.xunit.netcore.extensions

ericstj:
Note the different version strings.  This will be in 1.0.1-*prerelease*-02026-04 and later.

@weshaggard / @karajas is it intentional that you guys didn't bump the versions of these packages in master when enabling them in servicing?

EgorBo:
@ericstj @weshaggard @karajas can I update it in corefx myself? here https://github.com/dotnet/corefx/blob/master/dependencies.props#L44
we really need this fix, the PR was merged 22 days ago

ericstj:
> can I update it in corefx myself?

Yeah, that should be fine.  @eerhardt it looks like this package isn't getting updated as a result of https://github.com/dotnet/corefx/pull/23894, is that expected?

eerhardt:
> @eerhardt it looks like this package isn't getting updated as a result of dotnet/corefx#23894, is that expected?

I wasn't aware this package was being shipped from buildtools.  Someone could hook up the auto-dependency updates for this package as well.  They would just need to follow the existing pattern used for the other packages.

PR #1692, opened 2017-09-22
Add some options to GenerateRuntimeGraph for testing

ericstj:
This adds support for emitting a DGML and a compatibility mapping from a runtime graph.

The DGML is handy for visualizing the runtime graph and the compatibility mapping
helps identify what the effects of a RID graph change will be on package asset resolution.

/cc @weshaggard @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
@@ -10,6 +10,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Xml.Linq;
 
 namespace Microsoft.DotNet.Build.Tasks.Packaging
@@ -34,9 +35,15 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
         ///   AdditionalQualifiers: A list of strings delimited by semi-colons that represent the additional qualifiers for this RID.
         ///     Additional qualifers do not stack, each only applies to the qualifier-less RIDs (so as not to cause combinatorial 
         ///     exponential growth of RIDs).
+        ///
+        ///   *Use at your own risk*
         ///   OmitRIDs: A list of strings delimited by semi-colons that represent RIDs calculated from this RuntimeGroup that should
-        ///     be omitted from the RuntimeGraph.  This is useful in cases where overlapping RuntimeGroups are used and one is 
-        ///     designated to be "best" for some set of calculated RIDs, the others may omit the overlapping RIDs.
+        ///     be omitted from the RuntimeGraph.  These RIDs will not be referenced nor defined.
+        ///   OmitRIDDefinitions: A list of strings delimited by semi-colons that represent RIDs calculated from this RuntimeGroup 
+        ///     that should be omitted from the RuntimeGraph.  These RIDs will not be defined by this runtimegroup, but will be 
+        ///     refereenced: usefull in case some other RuntimeGroup (or runtime.json template) defines them.
+        ///   OmitRIDReferences: A list of strings delimited by semi-colons that represent RIDs calculated from this RuntimeGroup 
+        ///     that should be omitted from the RuntimeGraph.  These RIDs will be defined but not referenced by this runtimeGroup.
         /// </summary>
         public ITaskItem[] RuntimeGroups
         {
@@ -45,28 +52,9 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
         }
 
         /// <summary>
-        /// A set of QualifierPolicies that can be used to define the ordering of qualifer-specific RIDs in imports.
-        ///   Identity: the qualifier name.  This should match what is used in RuntimeGroups.AdditionalQualifiers
-        ///   VersionPrecedence: default is false.  When set to true, a qualifier-specific RID should import the 
-        ///     qualifier-less RID before any other RIDs that are not of the same version.  When false, the qualifier-
-        ///     less RID should be imported after all qualifier specific RIDs.
-        ///     
-        ///     Given the RID of  foo.1.0-x64 : [ foo.1.0, foo.0.8-x64 ]
-        ///     Applying the qualifier "bar" the following is the result for each value of VersionPrecedence:
-        ///       True: foo.1.0-x64-bar : [ foo.1.0-bar, foo.1.0-x64, foo.0.8-x64-bar ]
-        ///       False: foo.1.0-x64-bar : [ foo.1.0-bar, foo.0.8-x64-bar, foo.1.0-x64 ]
-        /// </summary>
-        public ITaskItem[] QualifierPolicies
-        {
-            get;
-            set;
-        }
-
-
-        /// <summary>
         /// Optional source Runtime.json to use as a starting point when merging additional RuntimeGroups
         /// </summary>
-        public ITaskItem SourceRuntimeJson
+        public string SourceRuntimeJson
         {
             get;
             set;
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
```diff
@@ -75,21 +75,38 @@ public ITaskItem SourceRuntimeJson
...
         public ITaskItem RuntimeJson
         {
             get;
             set;
         }
 
+        /// <summary>
+        /// When defined, specifies the file to write compatibility precedence for each RID in the graph.
+        /// </summary>
+        public ITaskItem CompatibilityMap
```
eerhardt: Any reason why these are TaskItems instead of just `string` properties?

ericstj: Nope, I can make them strings.

ericstj:
Using this I've actually found some compat issues with the way I'm generating the RID heirarchy for windows.  I'm going to roll in a fix for that to this PR.  Stay tuned.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
```diff
@@ -33,88 +35,81 @@ public class GenerateRuntimeGraph : PackagingTask
...
         ///     exponential growth of RIDs).
+        ///
+        ///   *Use at your own risk*
         ///   OmitRIDs: A list of strings delimited by semi-colons that represent RIDs calculated from this RuntimeGroup that should
-        ///     be omitted from the RuntimeGraph.  This is useful in cases where overlapping RuntimeGroups are used and one is 
-        ///     designated to be "best" for some set of calculated RIDs, the others may omit the overlapping RIDs.
+        ///     be omitted from the RuntimeGraph.  These RIDs will not be referenced nor defined.
+        ///   OmitRIDDefinitions: A list of strings delimited by semi-colons that represent RIDs calculated from this RuntimeGroup 
+        ///     that should be omitted from the RuntimeGraph.  These RIDs will not be defined by this runtimegroup, but will be 
+        ///     refereenced: usefull in case some other RuntimeGroup (or runtime.json template) defines them.
```
eerhardt: `refereenced`

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
```diff
@@ -33,88 +35,81 @@ public class GenerateRuntimeGraph : PackagingTask
         ///   AdditionalQualifiers: A list of strings delimited by semi-colons that represent the additional qualifiers for this RID.
         ///     Additional qualifers do not stack, each only applies to the qualifier-less RIDs (so as not to cause combinatorial 
         ///     exponential growth of RIDs).
+        ///
+        ///   *Use at your own risk*
```
eerhardt: I'm not sure I understand this warning.  Can it be explained further?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
```diff
@@ -194,6 +190,8 @@ public RuntimeGroup(ITaskItem item)
                 Architectures = item.GetStrings(nameof(Architectures));
                 AdditionalQualifiers = item.GetStrings(nameof(AdditionalQualifiers));
                 OmitRIDs = new HashSet<string>(item.GetStrings(nameof(OmitRIDs)));
+                OmitRIDs = new HashSet<string>(item.GetStrings(nameof(OmitRIDDefinitions)));
```
eerhardt: I think this is supposed to write to `OmitRIDDefinitions` and `OmitRIDReferences`, not overwrite OmitRIDs each time.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
```diff
@@ -205,137 +203,336 @@ public RuntimeGroup(ITaskItem item)
...
-                            imports.Add(runtimeDescription.RuntimeIdentifier);
+                            lastVersion = version;
                         }
+                    }
+                }
+            }
+
+            private bool IsNullOrRoot(string rid)
+            {
+                return rid == null || rid == rootRID;
```
eerhardt: I'm a bit confused at the "rootRID".  In the code above, it is "any".  But in the [runtime.json file](https://github.com/dotnet/corefx/blob/master/pkg/Microsoft.NETCore.Platforms/runtime.json), it seems that "any" imports "base".

I would think "base" is the rootRID.

ericstj: `any` is the true root.  `base` was introduced as a private RID that only we would use back when we tried to use runtime.json to lift up our package versions to a minimum version https://github.com/dotnet/buildtools/blob/96ee49b1f47f1a61dc040b2be2d976bbaa7903c4/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeDependencies.cs#L195-L207.  This technique never really worked perfectly so we stopped using it, then NuGet broke it (intentionally) by allowing downgrade of package versions referenced by runtime.json if the version referenced was higher in the graph than the package which brought in the runtime package.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
ericstj: Omitting RIDs from the graph can breaks the compatibility precedence we ensure when generating RIDs/imports from RuntimeGroups.  This makes it easy to create a RID fallback chain that doesn't match the rest of the RIDs and therefore is hard for developers/package authors to reason about.  Only use these options for cases where you know what you are doing and have carefully reviewed the resulting RID fallbacks.

eerhardt: Sounds like a perfect comment to put into this code 😉

ericstj: I'll add a comment to the source.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
ericstj: Doh, good catch

ericstj:
> If in the future we need more precedences we'll have to refactor this, one thing to watch
out for in particular is that if version doesn't have the highest precedence we'll have to
flatten the version list to imports for a single RID for compatible versioned groups.  That
will not scale well which is another reason I've avoided exposing such functionality.

I was worried that we'd get this wrong if we tried to do it later.  I went ahead and brought back QualifierPolicies and implemented qualifier-first-precedence, since it seems closer to what corert was trying to do before.  I inverted the default since using qualifier-first-precedence is not necessarily something we want to do everywhere due to how big it grows the runtime graph.

ericstj:
Ok, that's a failed experiment.  It's not as simple as just saying qualifiers have precedence.  Since the RIDs with qualifiers must also import the RIDs without qualifiers (in order to associate at the right os/version/arch) and runtime-compatibility is a breadth first traversal, it means that qualified RIDs must contain their *entire* flattened qualifier specific compatibility mappings (all the way up the graph) as ordered imports.  This is not scalable in a static representation so I'm cutting it.

ericstj:
You'll probably want to review the diffs with this change: https://github.com/dotnet/corefx/pull/24260

PR #1693, opened 2017-09-25
Allow projects to disable the Roslyn in BuildTools.

eerhardt:
Port #1651 to release/2.0.0

When using BuildTools in a repo like core-setup, where we want to use the dotnet-cli to build managed projects, we need to be able to not use the Roslyn properties that come in BuildTools.

For example, setting `<RoslynTargetsPath>$(ToolRuntimePath)</RoslynTargetsPath>` breaks any SDK project using the dotnet-cli.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -52,7 +52,7 @@
 
   <Import Project="$(MSBuildThisFileDirectory)BuildVersion.targets" />
 
-  <Import Project="$(MSBuildThisFileDirectory)Roslyn.Common.props" />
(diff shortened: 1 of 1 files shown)


danmosemsft:
Doesn't matter for this one, but generally speaking I've found it less confusing to avoid negatives in property names so eg this would be
`Condition="'$(EnableBuildToolsRoslynVersion)' == 'true'"`
which of course would need somewhere
```xml
<PropertyGroup>
    <EnableBuildToolsRoslynVersion Condition="'$(EnableBuildToolsRoslynVersion)' == ''">true</>
```

Mentally negating a negative tends to lead to bugs.

eerhardt:
I've usually found it easier for consumers if the properties are named in such a way that `false` is the default value.

I agree there is a tradeoff here though.  Sometimes making the default value this way does lead to less readable implementation.

PR #1699, opened 2017-09-27
Fix release/1.0.0 official build

joperezr:
cc: @weshaggard @Chrisboh 

Fixing the build.cmd script so that it works with the existing build definition.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,5 +1,5 @@
 @echo off
-setlocal
+setlocal EnableDelayedExpansion 
 
 :: Note: We've disabled node reuse because it causes file locking issues.
 ::       The issue is that we extend the build with our own targets which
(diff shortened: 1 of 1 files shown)


build.cmd
```diff
@@ -22,6 +22,29 @@ if not defined VisualStudioVersion (
...
+set "__args= %*"
+set __ConfigurationGroup=Debug
+set processedArgs=
+set __Platform=AnyCPU
+:Arg_Loop
+if "%1" == "" goto ArgsDone
+
+if /i "%1" == "-release"                 (set __ConfigurationGroup=Release&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
+if /i "%1" == "-debug"                 (set __ConfigurationGroup=Debug&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
+if /i "%1" == "-platform"                 (set __Platform=%2&set processedArgs=!processedArgs! %1 %2&shift&shift&goto Arg_Loop)
```
weshaggard: What do we pass in platform for?

joperezr: I'm not sure why we pass it, but it's in our build definition. From what I can tell, we always seem to pass `AnyCPU`

PR #1700, opened 2017-09-27
Add task to download files from url

safern:
This task will be used in order to download CoreCLR symbol package in CoreFX. 

cc: @weshaggard @danmosemsft

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
@@ -27,6 +27,7 @@ namespace Microsoft.DotNet.Build.Tasks
 
         public override bool Execute()
         {
+            bool success = false;
             try
             {
                 Log.LogMessage(MessageImportance.High, $"Downloading {DownloadSource} -> {DestinationFile}");
@@ -44,15 +45,10 @@ namespace Microsoft.DotNet.Build.Tasks
                                 {
                                     result.Content.CopyToAsync(stream).GetAwaiter().GetResult();
                                     Log.LogMessage(MessageImportance.High, $"Finished downloading: {DownloadSource}");
+                                    success = true;
                                 }
                                 else
                                 {
-                                    if (File.Exists(DestinationFile))
-                                    {
-                                        // If we fail to download, we want to do cleanup to not leave empty files.
-                                        File.Delete(DestinationFile);
-                                    }
-
                                     if (TreatErrorsAsWarnings)
                                     {
                                         Log.LogWarning($"Downloading {DownloadSource} failed with status code: {result.StatusCode}");
@@ -77,6 +73,17 @@ namespace Microsoft.DotNet.Build.Tasks
                 {
                     Log.LogError($"Downloading {DownloadSource} failed with exception: ");
                     Log.LogErrorFromException(e, showStackTrace: true);
+                }
+            }
+            finally
+            {
+                if (!success)
+                {
+                    if (File.Exists(DestinationFile))
+                    {
+                        // if we fail, do cleanup and delete file.
+                        File.Delete(DestinationFile);
+                    }
                 }
             }
 
--- a/src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
@@ -2,71 +2,65 @@ using Microsoft.Build.Framework;
 using System;
 using System.IO;
 using System.Net.Http;
+using System.Net;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
     public sealed class DownloadFileFromUrl : BuildTask
     {
-        
         /// <summary>
         /// The url to download the file from.
         /// </summary>
         [Required]
-        public string DownloadSource { get; set;}
+        public string DownloadSource { get; set; }
         
         /// <summary>
         /// The directory where the file will be downloaded to.
         /// </summary>
         [Required]
-        public string DestinationDirectory { get; set; }
+        public string DestinationFile { get; set; }
 
         /// <summary>
-        /// The file where the source will be downloaded to.
+        /// The default is to fail on error, set this to true if you want to just warn.
         /// </summary>
-        [Required]
-        public string DestinationFileName { get; set; }
-
-        /// <summary>
-        /// The default is to emit warnings and not fail, the caller can fail if output parameter DownloadedFile is empty.
-        /// Set this to true to fail instead of just warn.
-        /// </summary>
-        public bool FailOnError { get; set; }
-
-        /// <summary>
-        /// The file that was downloaded, if there is an error this will be empty.
-        /// </summary>
-        [Output]
-        public string DownloadedFile { get; set; }
+        public bool TreatErrorsAsWarnings { get; set; }
 
         public override bool Execute()
         {
             try
             {
-                Log.LogMessage(MessageImportance.High, $"Downloading {DownloadSource} -> {DestinationFileName}");
-                Directory.CreateDirectory(DestinationDirectory);
-                string destinationPath = Path.Combine(DestinationDirectory, DestinationFileName);
-                using (FileStream stream = File.Create(destinationPath))
+                Log.LogMessage(MessageImportance.High, $"Downloading {DownloadSource} -> {DestinationFile}");
+                var directory = Path.GetDirectoryName(DestinationFile);
+                Directory.CreateDirectory(directory);
+                using (FileStream stream = File.Create(DestinationFile))
                 {
-                    using (var client = new HttpClient())
+                    using (var handler = GetHttpHandler())
                     {
-                        using (var result = client.GetAsync(DownloadSource).GetAwaiter().GetResult())
+                        using (var client = new HttpClient(handler))
                         {
-                            if (result.IsSuccessStatusCode)
+                            using (var result = client.GetAsync(DownloadSource).GetAwaiter().GetResult())
                             {
-                                var bytes = result.Content.ReadAsByteArrayAsync().GetAwaiter().GetResult();
-                                stream.Write(bytes, 0, bytes.Length);
-                                DownloadedFile = destinationPath;
-                                Log.LogMessage(MessageImportance.High, $"Finished downloading: {DownloadSource}");
-                            }
-                            else
-                            {
-                                if (FailOnError)
+                                if (result.IsSuccessStatusCode)
                                 {
-                                    Log.LogError($"Downloading {DownloadSource} failed with status code: {result.StatusCode}");                                    
+                                    result.Content.CopyToAsync(stream).GetAwaiter().GetResult();
+                                    Log.LogMessage(MessageImportance.High, $"Finished downloading: {DownloadSource}");
                                 }
                                 else
                                 {
-                                    Log.LogWarning($"Downloading {DownloadSource} failed with status code: {result.StatusCode}");
+                                    if (File.Exists(DestinationFile))
+                                    {
+                                        // If we fail to download, we want to do cleanup to not leave empty files.
+                                        File.Delete(DestinationFile);
+                                    }
+
+                                    if (TreatErrorsAsWarnings)
+                                    {
+                                        Log.LogWarning($"Downloading {DownloadSource} failed with status code: {result.StatusCode}");
+                                    }
+                                    else
+                                    {
+                                        Log.LogError($"Downloading {DownloadSource} failed with status code: {result.StatusCode}");
+                                    }
                                 }
                             }
                         }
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// The url to download the file from.
+        /// </summary>
+        [Required]
+        public string DownloadSource { get; set;}
+        
+        /// <summary>
+        /// The directory where the file will be downloaded to.
+        /// </summary>
+        [Required]
+        public string DestinationDirectory { get; set; }
```
danmosemsft: If I have to specify both anyway, why not just have Destination that takes a path? Taht way if I have a path, I don't have to split it to pass it in.

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// The file where the source will be downloaded to.
+        /// </summary>
+        [Required]
+        public string DestinationFileName { get; set; }
+
+        /// <summary>
+        /// The default is to emit warnings and not fail, the caller can fail if output parameter DownloadedFile is empty.
+        /// Set this to true to fail instead of just warn.
+        /// </summary>
+        public bool FailOnError { get; set; }
```
danmosemsft: maybe TreatErrorsAsWarnings, it is used elsewhere

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
safern: Makes sense. I did it to make sure that the directory existed before creating the file, but I can achieve that by getting the parent directory from the full path, and create it before creating the file. (This is because File.Create throws if part of the path doesn't exist)

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
```diff
@@ -0,0 +1,92 @@
+using Microsoft.Build.Framework;
+using System;
+using System.IO;
+using System.Net.Http;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public sealed class DownloadFileFromUrl : BuildTask
+    {
+
```
danmosemsft: nit.. line

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
```diff
@@ -0,0 +1,92 @@
...
+                string destinationPath = Path.Combine(DestinationDirectory, DestinationFileName);
+                using (FileStream stream = File.Create(destinationPath))
+                {
+                    using (var client = new HttpClient())
+                    {
+                        using (var result = client.GetAsync(DownloadSource).GetAwaiter().GetResult())
+                        {
+                            if (result.IsSuccessStatusCode)
+                            {
+                                var bytes = result.Content.ReadAsByteArrayAsync().GetAwaiter().GetResult();
```
danmosemsft: Can you rewrite this to write to the file async? And avoid the intermediate byte array with ReadAsStreamAsync.

I realize that perf may not be critical but it's nice to do it the "right" way as it's no more lines.

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
```diff
@@ -0,0 +1,92 @@
...
+            try
+            {
+                Log.LogMessage(MessageImportance.High, $"Downloading {DownloadSource} -> {DestinationFileName}");
+                Directory.CreateDirectory(DestinationDirectory);
+                string destinationPath = Path.Combine(DestinationDirectory, DestinationFileName);
+                using (FileStream stream = File.Create(destinationPath))
+                {
+                    using (var client = new HttpClient())
+                    {
+                        using (var result = client.GetAsync(DownloadSource).GetAwaiter().GetResult())
```
weshaggard: please include some logic for handling the proxy. See https://github.com/dotnet/corefx/pull/23467 as example.

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// The default is to emit warnings and not fail, the caller can fail if output parameter DownloadedFile is empty.
+        /// Set this to true to fail instead of just warn.
+        /// </summary>
+        public bool FailOnError { get; set; }
+
+        /// <summary>
+        /// The file that was downloaded, if there is an error this will be empty.
+        /// </summary>
+        [Output]
+        public string DownloadedFile { get; set; }
```
mellinoe: If we combine the two "destination" input parameters into "DestinationPath" (for example), then this will just be equivalent to that.

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
safern: I've fixed this. Talked to @davidsh and there is a difference on how this is done in .NET Core vs Desktop. The way to do it in .NET Core is not available until .NET4.7.1 so I had to use if def.

safern:
Addressed PR Feedback. I've added a cleanup at the end to delete the file created if the download failed in order to track it in the targets to see if we need to unzip the file or not. Instead I could add an Output param but I'll wait for feedback.

danmosemsft:
You were going to make it async?

danmosemsft:
Oh, nm, I see your comment

safern:
> You were going to make it async?

We need to wait for the file to be created because we are going to unzip it and then extract the contents from another task. I’m using CopyToFileAsync but I have to wait for it anyway so that we can get the contents in the target.

danmosemsft:
You can wait for the file to finish writing, but it should be possible for it to write while the download completes.

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// The default is to fail on error, set this to true if you want to just warn.
+        /// </summary>
+        public bool TreatErrorsAsWarnings { get; set; }
+
+        public override bool Execute()
+        {
+            bool success = false;
+            try
+            {
+                Log.LogMessage(MessageImportance.High, $"Downloading {DownloadSource} -> {DestinationFile}");
```
joperezr: NIT: do we really want this to be `MessageImportance.High`? I would expect this type of messages to not be present on minimal logging.

joperezr:
I'm wondering if it would be useful to add also a .targets file called `DownloadSymbols.targets` or something and have that do all the logic of given an itemGroup of sources (defined by each repo) it will do the download packages, mark them as zip files, extract them, and get the .pdbs out. I say this becuase it seems that this logic could potentially be shared across different repos, and as you can see there are several steps that we might not want to have duplicated. Just of the top of my head, WCF jumps to mind, since they will probably want to do this too.

joperezr:
If we do agree to add the targets file though, that doesn't need to block this PR, it could be done on a subsequent PR.

src/Microsoft.DotNet.Build.Tasks/DownloadFileFromUrl.cs
safern: I made it like this because downloading a file can take ~5 or ~10 seconds, so that when we are building we know we are downloading something, but I can remove it and leave it to the targets to log that.

joperezr: that is true for this specific use because it's downloading a nuget package, but this task can/will be reused to download other things in the future, and I'm not sure if we will always want to have the verbosity in all cases. That's why I would suggest to keep logging a message, just not make it Importance.High

safern: Sounds reasonable. We could also pass an optional parameter to enable High verbosity and log with Importance.High if that is enabled.

weshaggard: I wouldn't do that. Just make the logging normal verbosity and if folks want that information they can always enable that level of logging.

safern: Ok, sounds good. Will do that then.

safern:
I've changed it to write to the file while downloading instead of first download and then write to the file, and also to be able to receive multiple items to download and custom metadata per item like `DestinationDir`. 

Could you guys please take a look?

src/Microsoft.DotNet.Build.Tasks/DownloadFilesFromUrl.cs
```diff
@@ -0,0 +1,144 @@
...
+                        continue;
+                    }
+                    else
+                    {
+                        Log.LogError($"Item {item.ItemSpec} is missing Url to download from in it's metadata.");
+                        return false;
+                    }
+                }
+
+                string fileName = item.GetMetadata("DestinationFile");
```
weshaggard: don't we need to set this if it isn't found? Wouldn't we default to the filename of the requested file.

safern: But we actually don't have the filename of the requested file, as it is a url and the url doesn't always in the filename that we want to download i.e to download a symbol package it is `https://feedurl/symbols/packageId/packageVersion` so we don't really have a way to know the default filename.

safern: There will be some cases were the Url will have the filename at the end, but that is not always the case, that is why I wanted to force people to give the destination file name (which they should always know).

weshaggard: What do you download in that case?

safern: Whatever the HTTP GET request returns, which in the case of the symbol server it returns a symbols.nupkg file. If the request doesn't return anything or throws an error then we would just output a a warning or an error, depending on the `TreatErrorsAsWarnings` parameter.

weshaggard: OK perhaps default the filename based on the url if it has a filename.

safern: OK if we can't find a filename in the URL and the DestinationFile metadata is empty then we throw the error.

weshaggard: Yeah otherwise we have an empty filename.

safern: But empty file name will cause an exception when trying to create the file, do we want to handle that error ourselves or should we leave it to `File.OpenWrite` ?

weshaggard: I think we should handle it ourselves so we can give a better error message.

safern: OK.

src/Microsoft.DotNet.Build.Tasks/DownloadFilesFromUrl.cs
```diff
@@ -0,0 +1,174 @@
...
+                }
+
+                Uri downloadUri = new Uri(downloadSource);
+                string fileName = item.GetMetadata("DestinationFile");
+                if (string.IsNullOrWhiteSpace(fileName))
+                {
+                    // starts with file://
+                    if (downloadUri.IsFile)
+                    {
+                        fileName = Path.GetFileName(downloadUri.LocalPath);
```
weshaggard: I think you can actually use Path.GetFileName on both cases instead of trying to parse the Uri yourself below.

src/Microsoft.DotNet.Build.Tasks/DownloadFilesFromUrl.cs
```diff
@@ -0,0 +1,174 @@
...
+                            downloadSource = downloadSource.Substring(0, downloadSource.Length - 1);
+                        }
+
+                        string[] urlContents = downloadSource.Split('/');
+
+                        // we set the filename to whatever is at the end of the url.
+                        // with this we ensure that if the metadata doesn't contain a file name, we always get one.
+                        fileName = urlContents[urlContents.Length - 1];
+
+                        if (string.IsNullOrEmpty(fileName))
```
weshaggard: I would move this check one level up just to be sure that the file case also gets a filename that isn't empty.

src/Microsoft.DotNet.Build.Tasks/DownloadFilesFromUrl.cs
safern: Ok, will try that. In that case we wouldn't need the downloadUri.IsFile check. Will try it.

safern: Fixed. Thanks for the feedback.

safern:
Thanks for the feedback!

PR #1703, opened 2017-09-28
Make GenerateRuntimeGraph diff files by default

ericstj:
This makes the task do a diff with existing files by default and only updates the files
when UpdateRuntimeFiles is set to true.

When updating the files we'll clear the readonly bit to handle the case where someone
has made the files readonly (since they are explicitly setting UpdateRuntimeFiles).

This is a buildtools implementation of https://github.com/dotnet/corefx/pull/24306.

/cc @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
@@ -82,6 +82,18 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
             set;
         }
 
+
+        /// <summary>
+        /// True to write the generated runtime.json to RuntimeJson and compatibility map to CompatibilityMap, otherwise files are read and diffed 
+        /// with generated versions and an error is emitted if they differ.
+        /// Setting UpdateRuntimeFiles will overwrite files even when the file is marked ReadOnly.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateRuntimeGraph.cs
```diff
@@ -140,6 +195,18 @@ public override bool Execute()
...
 
+        private void EnsureWritable(string file)
+        {
+            if (File.Exists(file))
+            {
+                var existingAttributes = File.GetAttributes(file);
+
+                if ((existingAttributes & FileAttributes.ReadOnly) != 0)
+                {
+                    File.SetAttributes(file, existingAttributes &= ~FileAttributes.ReadOnly);
```
weshaggard: Cannot you still force overwrite it without changing the file attributes if it is readonly?

ericstj: How would you do that?  CreateFile will fail when opening read only file for write.  It will also fail when CREATE_ALWAYS is specified.  I could move/copy but that's more work than just clearing the bit. I've tested this and it does the right thing.

weshaggard: OK. I'm fine with this just thought there was a force overwrite option.

PR #1706, opened 2017-09-29
Pare down include lists for dump collection

DrewScoggins:
We started by just collecting all of the SOs in the lib and usr/lib
directories, but that proved to be too big.  This brings the list to
just the files that we need.  This also turns off dump triaging as it is
currently broken and blocking regular collection.


DrewScoggins:
@mellinoe @danmosemsft PTAL

mellinoe:
Looks okay to me -- are these paths the same on all distros?

DrewScoggins:
They are the same on RHEL as they are on Ubuntu.  I have also added the paths for MacOSX as well.

PR #1707, opened 2017-09-29
Add Repo API common implementation

dagood:
Takes APIs from https://github.com/dotnet/source-build/blob/dev/release/2.0/Documentation/auto-dependency-flow/api.md and sets BuildTools properties/items to implement them.

I used these branches in source-build submodules to try it out:

 * https://github.com/dotnet/standard/compare/dev/release/2.0.0...dagood:flow-api
 * https://github.com/dotnet/corefx/compare/dev/release/2.0.0...dagood:flow-api

I copied the props file directly below the `Build.Common.props` import in those branches to test this out quickly without updating the entire BuildTools package (which could have its own problems).

Confirmed that it works by checking the blob feed dir for the `nupkg`/`symbols.nupkg`s and using `strings` to peek at the DLLs that CoreFX binplaced to `bin/ref/netstandard`.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RepoAPI.Mapping.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RepoAPI.Mapping.props
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+    Provides a default API implementation where possible by setting BuildTools properties/items.
+
+    API documentation: https://github.com/dotnet/source-build/tree/dev/release/2.0/Documentation/auto-dependency-flow
+  -->
+
+  <!-- Produce packages into the specified blob feed. -->
+  <PropertyGroup Condition="'$(DotNetOutputBlobFeedDir)' != ''">
+    <PackageOutputPath>$(DotNetOutputBlobFeedDir)packages/</PackageOutputPath>
+    <SymbolPackageOutputPath>$(DotNetOutputBlobFeedDir)assets/</SymbolPackageOutputPath>
+  </PropertyGroup>
+
+  <!--
+    Import the restore source props file if passed to get "DotNetRestoreSources".
+
+    Note the "!= ''" check rather than Exists. If this API argument is passed but the file doesn't
+    exist, something is wrong and an error should be bubbled up as soon as we know about it.
+  -->
+  <Import Project="$(DotNetRestoreSourcePropsPath)"
+          Condition="'$(DotNetRestoreSourcePropsPath)' != ''"/>
+
+  <!-- Use passed NuGet restore sources, if any. -->
+  <PropertyGroup>
+    <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
+  </PropertyGroup>
+
+  <!--
+    Populate DotnetSourceList items based on NuGet restore sources from the Repo API plus any
+    additions the repo makes to the RestoreSources property.
+
+    This is used by the "dotnet restore" Exec that uses "source" args. When a repo uses msbuild
+    restore tooling, this ItemGroup is unnecessary.
+  -->
+  <ItemGroup>
+    <DotnetSourceList Include="$(RestoreSources)" />
+  </ItemGroup>
+</Project>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/RepoAPI.Mapping.props
```diff
@@ -0,0 +1,39 @@
...
+
+    Note the "!= ''" check rather than Exists. If this API argument is passed but the file doesn't
+    exist, something is wrong and an error should be bubbled up as soon as we know about it.
+  -->
+  <Import Project="$(DotNetRestoreSourcePropsPath)"
+          Condition="'$(DotNetRestoreSourcePropsPath)' != ''"/>
+
+  <!-- Use passed NuGet restore sources, if any. -->
+  <PropertyGroup>
+    <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
```
eerhardt: Are projects who import this .props file expected to use `RestoreSources` at all?  If so, I think there is an issue here.

If the project sets `RestoreSources` before importing this file, this will overwrite that value.  If the project sets `RestoreSources` after importing this file, their values won't get added to `DotnetSourceList` below.

I would think this would append to the `RestoreSources` property instead of overwriting it.

dagood: I'll change it to append. You're right, no reason to restrict people from defining sources before importing this file.

> If the project sets RestoreSources after importing this file, their values won't get added to DotnetSourceList below.

Experimentally, this isn't the case: items are all evaluated after properties. For clarity, it might be nice to move it into a target to make the ordering explicit, but some Core repos and parts of BuildTools depend directly on the static values.

eerhardt: > items are all evaluated after properties

I keep forgetting about this...

dagood:
@dotnet-bot test this please

dagood:
Oh, no CI builds here. Merging.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/RepoAPI.Mapping.props
```diff
@@ -0,0 +1,39 @@
...
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+    Provides a default API implementation where possible by setting BuildTools properties/items.
+
+    API documentation: https://github.com/dotnet/source-build/tree/dev/release/2.0/Documentation/auto-dependency-flow
+  -->
+
+  <!-- Produce packages into the specified blob feed. -->
+  <PropertyGroup Condition="'$(DotNetOutputBlobFeedDir)' != ''">
+    <PackageOutputPath>$(DotNetOutputBlobFeedDir)packages/</PackageOutputPath>
```
weshaggard: Do we think this is going to be setup in the correct order?

Sorry for not catching this earlier but don't we think that the override property should be named a little more general then "BlobFeed"? I expect us to use this for local builds as well so I wonder if this should just be "DotNetPackageOutputRootPath" or something similar.

dagood: Well, I know it is in the correct order for Standard and CoreFX, at least. As long as Build.Common.props is imported before packaging msbuild data, then it will work... seems like a safe assumption, but I don't know if we have a generalized import order structure/convention to make that obvious.

About the name, currently "blob feed" [is defined](https://github.com/dotnet/source-build/tree/dev/release/2.0/Documentation/auto-dependency-flow#terms) as being local (disk) or remote (blob storage), so it is pretty general. I think it needs to be more specific than `DotNetPackageOutputRootPath` because it is *not* the directory where you should put your packages. [The property defines the root of the blob feed, and NuGet packages go in `<blob feed>/packages/`](https://github.com/dotnet/source-build/blob/dev/release/2.0/Documentation/auto-dependency-flow/api.md#-pdotnetoutputblobfeeddirtarget-directory).

weshaggard: Fair enough. This is mostly about the artifacts that will be taken from the output. We will also need to define some properties for bin and obj directories to share a common output but that is orthogonal to these.

PR #1702, opened 2017-09-27
Create Microsoft.DotNet.Build.Tasks.VersionTools package

chcosta:
Create Microsoft.DotNet.Build.Tasks.VersionTools package which is repackaging the VersionTools tasks from Microsoft.DotNet.Build.Tasks.

This allows me to use the Microsoft.DotNet.Build.Tasks.VersionTools package for updating the dotnet/versions repo rather than the full Microsoft.DotNet.BuildTools package.

https://github.com/dotnet/core-eng/issues/1733

Issue #1489, opened 2017-05-11
VersionTools and BuildTools have inconsistent version numbers


eerhardt:
Inspect the package versions at:

https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.DotNet.VersionTools
https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.DotNet.BuildTools

VersionTools still has `1.0.27`.
BuildTools has `2.0.0`.

This is confusing to understand which versions are in sync.

@dagood also points out that branching can be affected here, because branch build version numbers could overlap without a fix, and we lose version coherency.

weshaggard:
Do we really want all the version numbers to match across all packages we ship in BuildTools? If so then we should probably switch to using the pkgproj's so we an share a common property for it.

eerhardt:
> Do we really want all the version numbers to match across all packages we ship in BuildTools?

I guess I'm not 100% sure on our plans for branching/servicing BuildTools.  It does seem to make things easier to understand when all `X.Y.Z` versions of packages come from the same branch.  Just like in corefx when we moved to all packages to `4.3.0` versions.

> If so then we should probably switch to using the pkgproj's so we an share a common property for it.

Can't we just use a common `.props` file that has the version prefix in it? ... Ah I see that we have checked-in .nuspec files.  An alternative to pkgproj's is to use `dotnet pack` or `msbuild /t:Pack` using the new Microsoft.NET.Sdk.  Or to use a replacement value `$version$`.

dagood:
The versions don't need to be exactly the same, but we do need to bump each version across the entire repo. For example, if we are building BuildTools 1.1.0 and 2.0.0 Git branches with VersionTools `1.0.0-beta-*` in both branches, the best that can happen is thrashing: adjacent VersionTools `1.0.0-beta-*` builds can come from different branches. If we don't have unique build ids (and I don't think we will) we would get last-wins MyGet overwriting behavior.

Keeping everything in exact sync seems the easiest way to maintain this.

dagood:
This is fixed by https://github.com/dotnet/buildtools/pull/1755 in master: all packages are now `2.1.0-prerelease-*`. It should be easy to port to other branches if we encounter overlap (or if the clarification itself is useful enough).

PR #1710, opened 2017-09-29
Release/2.0.0 Create Microsoft.DotNet.Build.Tasks.VersionTools package

chcosta:
Moving PR from https://github.com/dotnet/buildtools/pull/1702 to target release/2.0.0

This addresses feedback in that PR, except for changing BuildTask and the open question of how to handle versioning.

### DIFF
--- a/src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
+++ b/src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
@@ -15,21 +15,11 @@
     </description>
     <copyright>Copyright © Microsoft Corporation</copyright>
     <tags></tags>
-    <dependencies>
-      <group targetFramework="netstandard1.5">
-        <dependency id="NETStandard.Library" version="1.6.0" />
-        <dependency id="Newtonsoft.Json" version="9.0.1" />
-        <dependency id="NuGet.Packaging" version="4.3.0-preview2-4095" />
-        <dependency id="NuGet.Versioning" version="4.3.0-preview2-4095" />
-        <dependency id="System.Diagnostics.Process" version="4.1.0" />
-        <dependency id="System.Diagnostics.TraceSource" version="4.0.0" />
-      </group>
-    </dependencies>  </metadata>
+  </metadata>
   <files>
     <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DotNet.Build.Tasks.dll" target="lib\netstandard1.5" />
-    <file src="Microsoft.DotNet.VersionTools\Microsoft.DotNet.VersionTools.dll" target="lib\netstandard1.5" />
+    <file src="Microsoft.DotNet.VersionTools\*.dll" target="lib\netstandard1.5" />
     <file src="Microsoft.DotNet.Build.Tasks.net45\Microsoft.DotNet.Build.Tasks.dll" target="lib\net46" />
-    <file src="Microsoft.DotNet.Build.Tasks.Packaging.Desktop\NuGet.*.dll" target="lib\net46" />
     <file src="Microsoft.DotNet.VersionTools.net45\*.dll" target="lib\net46" />
     <file src="Microsoft.DotNet.Build.Tasks\PackageFiles\VersionTools.targets" target="build" />
     <file src="..\obj\version.txt" target="" />
--- a/src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
+++ b/src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
@@ -1,0 +1,37 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.8.1">
+    <id>Microsoft.DotNet.Build.Tasks.VersionTools</id>
+    <version>2.0.0-servicing</version>
+    <title>DotNet Build VersionTools</title>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
+    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=518631</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <summary>DotNet Build VersionTools</summary>
+    <description>This package provides support for updating the versions repo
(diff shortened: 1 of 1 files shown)


src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
```diff
@@ -0,0 +1,31 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.8.1">
+    <id>Microsoft.DotNet.Build.Tasks.VersionTools</id>
+    <version>1.0.0-prerelease</version>
```
dagood: Yeah, about versioning: I do think all BuildTools packages should be on the same version path for reasonable branching. So here, `2.0.0-servicing`. Some discussion here: https://github.com/dotnet/buildtools/issues/1489. (So don't worry about updating everything here, just this new package.)

chcosta: I updated the prerelease label.  I'll update the version to be 2.0.0 and ignore the other packages.

src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
```diff
@@ -0,0 +1,31 @@
...
+     <!-- Do not add any dependencies to this package as we want to install it versionless (i.e. -ExcludeVersion) and we don't want the dependencies to be installed that way -->
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Microsoft.DotNet.Build.Tasks.VersionTools\Microsoft.DotNet.Build.Tasks.VersionTools.dll" target="lib\netstandard1.5" />
+    <file src="Microsoft.DotNet.Build.Tasks.VersionTools.net45\*.dll" target="lib\net46" />
+    <file src="Microsoft.DotNet.VersionTools\Microsoft.DotNet.VersionTools.dll" target="lib\netstandard1.5" />
+    <file src="Microsoft.DotNet.VersionTools.net45\Microsoft.DotNet.VersionTools.dll" target="lib\net46" />
+    <file src="Microsoft.DotNet.Build.Tasks\PackageFiles\VersionTools.targets" target="lib\netstandard1.5" />
+    <file src="Microsoft.DotNet.Build.Tasks\PackageFiles\VersionTools.targets" target="lib\net46" />
```
dagood: Hmm, is there a hard reason to duplicate this file? In the BuildTools package these files live in `lib\` and are shared. Then again, in the BuildTools package the netstandard assets live in `lib\` so maybe that's what's changing here.

I'm also wondering if we should start using the `build` dir for this, but probably not yet: https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package

src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
dagood: Thanks!

src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
chcosta: If this was in `build`, I wouldn't be as keen to duplicate this.  As-is, it simplifies some things to reference the target in the same location as the assembly.  I like the suggestion of `build`, I'll make that change.

weshaggard:
I will port my comment from the other PR because I do think it is worth talking about:

I wonder if instead of this we just package the task library in its own package for folks that want to consume it directly. That will mean things stay in the same place and I don't think it is a big deal to have more stuff in the task library as it won't add much in size and it will move us closer to componentizing buildtools. I think that is better then one-off creating a new task library for one small task.

dagood:
I think a reason it would make sense to split the task lib itself is to avoid forcing people to take dependencies they don't want to take, e.g. NuGet. (I don't know if the split in this PR helps someone there.) I agree though that task library size doesn't seem like the problem.

weshaggard:
The dependencies don't really matter either unless you use some API from that dependency. If you scope to just one task you only need the dependencies that task uses, assuming we are JIT'ing that isn't true if we ngen.

dagood:
Interesting, good point. That would basically mean we have a full BuildTools package that delivers dependencies (and fetches some when you call init-tools inside), but also the option to use the lightweight tools package and provide your own deps for what you plan to call. Right?

chcosta:
> I wonder if instead of this we just package the task library in its own package for folks that want to consume it directly.

Are you referring to packaging buildtask.cs in a separate package, or doing something else?

weshaggard:
>Are you referring to packaging buildtask.cs in a separate package, or doing something else?

I'm suggesting that we don't create a new library but instead just package the existing task library instead of creating a new VersionTools task library.

chcosta:
i.e. package `Microsoft.DotNet.Build.Tasks.dll', 'Microsoft.DotNet.Versions.dll', and 'VersionTools.targets', and ignoring the fact that `Microsoft.DotNet.Build.Tasks.dll` contains a lot of surface area that we don't necessarily care about?

weshaggard:
Exactly.

chcosta:
That changes everything!  I like it.  Thanks for the clarification

src/nuget/Microsoft.DotNet.Build.Tasks.VersionTools.nuspec
```diff
@@ -0,0 +1,37 @@
...
+        <dependency id="NuGet.Versioning" version="4.3.0-preview2-4095" />
+        <dependency id="System.Diagnostics.Process" version="4.1.0" />
+        <dependency id="System.Diagnostics.TraceSource" version="4.0.0" />
+      </group>
+    </dependencies>  </metadata>
+  <files>
+    <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DotNet.Build.Tasks.dll" target="lib\netstandard1.5" />
+    <file src="Microsoft.DotNet.VersionTools\Microsoft.DotNet.VersionTools.dll" target="lib\netstandard1.5" />
+    <file src="Microsoft.DotNet.Build.Tasks.net45\Microsoft.DotNet.Build.Tasks.dll" target="lib\net46" />
+    <file src="Microsoft.DotNet.Build.Tasks.Packaging.Desktop\NuGet.*.dll" target="lib\net46" />
```
chcosta: I'm going to remove this line and re-validate.  When I was testing on Saturday, I was seeing weird behavior which ended up being because the package cache was tripping me up.  Looks like I unintentionally left this line in.

chcosta:
@weshaggard , PTAL

PR #1717, opened 2017-10-03
Pass relativePath instead of feedUrl in BlobFeed constructor

karajas:
/cc @weshaggard @wtgodbe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
@@ -35,16 +35,19 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
 
         bool IsPublic { get; set; }
 
+        string FeedUrl { get; set; }
+
         private static readonly CancellationTokenSource TokenSource = new CancellationTokenSource();
         private static readonly CancellationToken CancellationToken = TokenSource.Token;
 
-        public BlobFeed(string accountName, string accountKey, string containerName, string relativePath, string indexDirectory, MSBuild.TaskLoggingHelper loggingHelper, bool isPublic)
+        public BlobFeed(string accountName, string accountKey, string containerName, string relativePath, string feedUrl, string indexDirectory, MSBuild.TaskLoggingHelper loggingHelper, bool isPublic)
         {
             AccountName = accountName;
             AccountKey = accountKey;
             ContainerName = containerName;
             IndexDirectory = indexDirectory;
             RelativePath = relativePath;
+            FeedUrl = feedUrl;
             IsPublic = isPublic;
             Log = loggingHelper;
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -36,10 +36,9 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
                 string accountName = m.Groups["accountname"].Value;
                 string containerName = m.Groups["containername"].Value;
                 string relativePath = m.Groups["relativepath"].Value;
-                string feedUrl = m.Groups["feedurl"].Value;
 
                 bool isPublic = string.IsNullOrWhiteSpace(m.Groups["token"].Value);
-                this.feed = new BlobFeed(accountName, accountKey, containerName, feedUrl, string.IsNullOrWhiteSpace(indexDirectory) ? Path.GetTempPath() : indexDirectory, Log, isPublic);
+                this.feed = new BlobFeed(accountName, accountKey, containerName, relativePath, string.IsNullOrWhiteSpace(indexDirectory) ? Path.GetTempPath() : indexDirectory, Log, isPublic);
```
weshaggard: I don't think this is going to work. We need should use the exact full feedUrl not a relativePath.

karajas: BlobFeed.RelativePath is set by this, which calculates the relativeUrl (which is the part of the url after the feed.blob.core.windows.net/containername). It seems to work from my testing, what makes you suspect otherwise?

karajas: Addressed the concern of a different url, by writing feedUrl into the index.json.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -204,7 +207,7 @@ private string GenerateRootServiceIndex(string redirectUrl)
                 writer.WriteStartArray();
                 writer.WriteStartObject();
                 writer.WritePropertyName("@id");
-                writer.WriteRawValue($"\"{FeedContainerUrl}/{redirectUrl}/\"");
+                writer.WriteRawValue($"\"{FeedUrl}\"");
```
weshaggard: Why wouldn't FeedUrl completely replace FeedContainerUrl? 

This looks like it is writing to the root index.json file but what about the individual package index.json files?

karajas: We still need the FeedContainerUrl to make calls to the storage account(createcontainer, checkifblobexists, etc.) I've left it in there for now.

the individual package index.json don't have urls in them, just the versions available for that package.

PR #1720, opened 2017-10-03
Fix Flat upload of assets

karajas:
/cc @wtgodbe @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
@@ -153,13 +153,13 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             if (Path.GetExtension(item) == ".nupkg")
             {
                 Tuple<string, NuGetVersion> nugetPackage = GetNugetPackageVersion(item);
-                blobPath = Path.Combine(Path.Combine(relativePath, Path.Combine(nugetPackage.Item1, nugetPackage.Item2.ToFullString())), Path.GetFileName(item));
+                blobPath = Path.Combine(Path.Combine(relativePath, Path.Combine(nugetPackage.Item1, nugetPackage.Item2.ToFullString())), Path.GetFileName(item)).ToLowerInvariant();
             }
             else
             {
-                blobPath = item.Substring(item.IndexOf(relativePath));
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -90,13 +90,16 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
 
                 await UploadAsync(CancellationToken, item, uploadPath, clientThrottle, allowOverwrite);
 
-                List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, packageDirectory);
-                if (!listAzureBlobs.Any(x => x.Contains(uploadPath)))
+                if (!isFlat)
```
karajas: Skipping generation of index.json uploads

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -153,13 +153,13 @@ public string CalculateBlobPath(string item, string relativePath)
...
                 Tuple<string, NuGetVersion> nugetPackage = GetNugetPackageVersion(item);
-                blobPath = Path.Combine(Path.Combine(relativePath, Path.Combine(nugetPackage.Item1, nugetPackage.Item2.ToFullString())), Path.GetFileName(item));
+                blobPath = Path.Combine(Path.Combine(relativePath, Path.Combine(nugetPackage.Item1, nugetPackage.Item2.ToFullString())), Path.GetFileName(item)).ToLowerInvariant();
             }
             else
             {
-                blobPath = item.Substring(item.IndexOf(relativePath));
+                blobPath = Path.Combine(relativePath, Path.GetFileName(item));
             }
-            return blobPath.ToLowerInvariant().Replace("\\", "/");
```
karajas: Moving toLower to just the nupkg section to preserve cases for files

karajas:
This also assumes that we can only upload flat files, never folders.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
weshaggard: why does the path need to be tolower for the nuget files?

PR #1722, opened 2017-10-04
Update timeout and delay time for lease during upload

karajas:
/cc @weshaggard @wtgodbe

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -133,7 +133,8 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
 
             await clientThrottle.WaitAsync();
             string leaseId = string.Empty;
-            AzureBlobLease blobLease = new AzureBlobLease(feed.AccountName, feed.AccountKey, string.Empty, feed.ContainerName, uploadPath, Log, "60", "10");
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -133,7 +133,7 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
 
             await clientThrottle.WaitAsync();
             string leaseId = string.Empty;
-            AzureBlobLease blobLease = new AzureBlobLease(feed.AccountName, feed.AccountKey, string.Empty, feed.ContainerName, uploadPath, Log, "60", "10");
+            AzureBlobLease blobLease = new AzureBlobLease(feed.AccountName, feed.AccountKey, string.Empty, feed.ContainerName, uploadPath, Log, "15", "3000");
```
weshaggard: Can you add a comment in the commit to state what these new times are? Are they all in milliseconds? which one is delay and which one is timeout.

karajas: Sure, the first one is in seconds and the second in milliseconds.

karajas:
The only way to test that this works is to ensure that it works in the context of official build, testing there and will update with results.

karajas:
This is now working in the intended manner in the CoreCLR build.

Issue #1724, opened 2017-10-05
DownloadFilesFromUrl task cannot be used twice in the same build process


danmosemsft:
The bug is that HttpClient is stored in a static field
`private static readonly HttpClient s_client = new HttpClient(GetHttpHandler());`
but it is disposed after being used once (eg., at the end of `Execute()`).

Either it needs to be not disposed (leaving it to be torn down with the process) or it needs to be created by each instance of the task. The latter is probably the sensible approach - just make it not static.

@safern 

```
D:\j\workspace\windows-TGrou---c60886e1\Tools\Symbols.targets(533,7): warning : Downloading https://dotnet.myget.org/F/dotnet-core/symbols/runtime.win10-x86-aot.Microsoft.NETCore.Runtime.CoreCLR/2.1.0-preview1-25805-02/ failed with exception:  [D:\j\workspace\windows-TGrou---c60886e1\external\runtime\runtime.depproj]
D:\j\workspace\windows-TGrou---c60886e1\Tools\Symbols.targets(533,7): warning : Response status code does not indicate success: 404 (Not Found). [D:\j\workspace\windows-TGrou---c60886e1\external\runtime\runtime.depproj]
D:\j\workspace\windows-TGrou---c60886e1\Tools\Symbols.targets(533,7): warning :    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task) [D:\j\workspace\windows-TGrou---c60886e1\external\runtime\runtime.depproj]
D:\j\workspace\windows-TGrou---c60886e1\Tools\Symbols.targets(533,7): warning :    at Microsoft.DotNet.Build.Tasks.DownloadFilesFromUrl.Execute() [D:\j\workspace\windows-TGrou---c60886e1\external\runtime\runtime.depproj]
D:\j\workspace\windows-TGrou---c60886e1\external\runtime\runtime.depproj(94,7): warning : Failed to download CoreCLR symbols
D:\j\workspace\windows-TGrou---c60886e1\Tools\Symbols.targets(533,7): warning : Downloading https://dotnet.myget.org/F/dotnet-core/symbols/runtime.win10-x86.Microsoft.NETCore.Runtime.CoreCLR/2.1.0-preview1-25805-02/ failed with exception:  [D:\j\workspace\windows-TGrou---c60886e1\external\runtime\runtime.depproj]
[...]

PR #1725, opened 2017-10-06
Fix disposed exception when downloading a file and calling the task multiple times on the same msbuild process

safern:
Fixes: https://github.com/dotnet/buildtools/issues/1724

This should be enough to fix the disposed exception you saw.

cc: @danmosemsft


safern:
> needed to add ?w=1 to review this!

Ha, that is a nice trick to delete the indenting and spaces additions.

PR #1728, opened 2017-10-09
Update ReportGenerator to 3.0.1

danmosemsft:
See https://github.com/dotnet/corefx/issues/23588 - this works around a stack overflow.

### DIFF
--- a/src/common/test-runtime/project.json
+++ b/src/common/test-runtime/project.json
@@ -9,7 +9,7 @@
     "System.Runtime.InteropServices.RuntimeInformation": "4.4.0-beta-24813-03",
     "System.Linq.Parallel": "4.0.1",
     "OpenCover": "4.6.519",
-    "ReportGenerator": "2.5.0",
+    "ReportGenerator": "3.0.1",
     "Microsoft.DotNet.xunit.performance.analysis": "1.0.0-alpha-build0035",
     "Microsoft.DotNet.xunit.performance.analysis.cli": "1.0.0-alpha-build0035",
     "Microsoft.DotNet.xunit.performance.runner.cli": "1.0.0-alpha-build0035",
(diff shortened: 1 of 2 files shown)


src/common/test-runtime/project.json
```diff
@@ -9,7 +9,7 @@
     "System.Runtime.InteropServices.RuntimeInformation": "4.4.0-beta-24813-03",
     "System.Linq.Parallel": "4.0.1",
     "OpenCover": "4.6.519",
-    "ReportGenerator": "2.5.0",
```
danmosemsft: @safern I'm not sure what this file does: is it superseded by xunit.runtime.depproj in corefx? Nevertheless, changing it here also.

stephentoub:
The last time I went to change this, @weshaggard had this feedback: https://github.com/dotnet/buildtools/pull/1612#discussion_r127760325

stephentoub:
Do you need to replace it here as well?
https://github.com/dotnet/wcf/blob/58a8c43aa8a26897d199cddadd3158b9b87c1353/external/test-runtime/XUnit.Runtime.depproj#L55

src/common/test-runtime/project.json
weshaggard: this file is old and I don't believe we use it any longer since all our consuming repo's are now using msbuild based CLI.

safern: I think to update it in our repos you would have to change the Xunit.Runtime.depproj, not here.

https://github.com/dotnet/corefx/blob/master/external/test-runtime/XUnit.Runtime.depproj#L65

safern: I see you already have a PR to do that. So looks good.

danmosemsft:
I'll merge as-is and update dependencies.props at some future time.
I'l put up another PR to remove test-runtime.

PR #1729, opened 2017-10-09
Remove dead test-runtime

danmosemsft:


danmosemsft:
@weshaggard I admit I am not familiar with this at all: if it's dead, we can submit this, if not, I won't.

src/Microsoft.DotNet.Build.Tasks.Packaging/test/project.json
```diff
@@ -8,10 +8,6 @@
     "NETStandard.Library": "1.6.0",
```
weshaggard: I suspect this file is also dead.

weshaggard:
>@weshaggard I admit I am not familiar with this at all: if it's dead, we can submit this, if not, I won't

I do think this is dead and we should to the work to clean it up but I cannot be for sure without testing. There is more then just deleting these files that is needed to clean this up. For example the duel code paths in the init-tools scripts https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd#L70.

Issue #1735, opened 2017-10-12
Fix name of "ConvertPdbsToPortablePdbs" task


dagood:
[`ConvertPdbsToPortablePdbs`](https://github.com/dotnet/buildtools/blob/583046217f68333c04521e119e9cf97eca6f2494/src/Microsoft.DotNet.Build.Tasks/ConvertPdbsToPortablePdbs.cs) is backwards: the task actually converts Portable PDBs to Windows PDBs. Getting it renamed before anyone outside BuildTools directly uses it would be ideal.

PR #1736, opened 2017-10-12
Rename ConvertPdbsToPortablePdbs to ConvertPortablePdbsToWindowsPdbs

dagood:
Fixes https://github.com/dotnet/buildtools/issues/1735. I want to fix this before someone gets a chance to start relying on this build task's name outside of the BuildTools targets.

Thanks for spotting this, @vancem!


vancem:
LGTM.  Thanks for the quick turn-around Davis.

Issue #1737, opened 2017-10-13
Investigate Portable PDB to Windows PDB conversion AVs


dagood:
The Portable PDB to Windows PDB conversion task sometimes hit AVs, so I implemented a retry workaround in https://github.com/dotnet/buildtools/pull/1623. Errors look like:

```
Converting portable PDB '[...]System.Private.Xml.dll'...
##[error]Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Microsoft.DiaSymReader.ISymUnmanagedWriter5.Close()
   at Microsoft.DiaSymReader.SymUnmanagedWriter.WriteTo(Stream stream)
   at Microsoft.DiaSymReader.Tools.PdbConverter.ConvertPortableToWindows(PEReader peReader, MetadataReader pdbReader, Stream targetPdbStream, PdbConversionOptions options)
   at Microsoft.DiaSymReader.Tools.PdbConverter.ConvertPortableToWindows(PEReader peReader, Stream sourcePdbStream, Stream targetPdbStream, PdbConversionOptions options)
   at Microsoft.DotNet.Build.Tasks.ConvertPdbsToPortablePdbs.Execute()
```

I set the workaround to only retry once, which seems to work most of the time. It failed again during the retry at least once, though: https://github.com/dotnet/core-eng/issues/1782. However, it hasn't failed on retry again since then, so this isn't currently considered a hot issue as far as engineering services goes.

I was not able to repro except in the exact circumstances of the build machines hitting the issue, and even then, it only seemed to repro on the first attempt on the machine in a while. @vancem offered some tips to help try to repro consistently:

> Turning on the environment variable COMPLUS_GCStress=3 and running a non-trivial symbol conversion scenario for a long time (e.g. over a weekend), is likely to cause the bug to reproduce.     You would not want to just do Pdb2Pdb since that closes down the process (and hides corruption),   But you could have just directory full of portable PDBS that you write a little harness that simply converts them over and over in one process.   Set that up with Windbg to stop on an AV, and you probably will get a repo.   
> 
> A variation on GCStress=3 (which is considerably cheaper) is to use these managed debugging assistants that cause GCs every time you transition from manage to unmanaged or back.   
> 
> https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/gcmanagedtounmanaged-mda
[...]

PR #1740, opened 2017-10-13
Allow Code Coverage to work on Portable PDBs

vancem:
Currently OpenCover requires MSPDBs.  This change makes the coverage task generate MSPDBs from
portable ones so code coverage continues to work (soon OpenCover will support Portable PDBS directly).

We need this so we can switch CoreFX over to using Portable PDBs.

Basically I add a new Target that creates the MSPDB in the WindowsPDB directory, and then add that directory to the search path the OpenCover uses.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -6,7 +6,7 @@
   -->
   <PropertyGroup>
     <OpenCoverVersion>4.6.519</OpenCoverVersion>
-    <ReportGeneratorVersion>2.5.0</ReportGeneratorVersion>
+    <ReportGeneratorVersion>3.0.1</ReportGeneratorVersion>
```
vancem: The ReportGenerator version is 3.0.1 in the CoreFX version of this.  I am not sure why that is (does something update this file as it gets installed into CoreFx\Tools?) but it seems better to move forward since CoreFX is the major (only?) client of code coverage.

vancem:
@dagood @stephentoub   I have tested this change in CoreFX (which as far as I know is the only repo that does code coverage).    I have also confirmed that it works properly if you turn off

vancem:
Including @weshaggard  @danmosemsft  @karajas  for visibility.

At a high level, this is here so that we can move CoreFX to use Portable PDBs instead of Windows PDBs.

Currently OpenCover requires Windows PDBS, so this generates them on the fly during coverage runs.  

This task can be REMOVED in a few months when OpenCover supports Portable PDBs.  

The change is conceptually very simple.  Before running OpenCover convert the portable PDB to a Windows PDB (that is what the new task does).  

It also works if the build generates PortablePDBs or the conversion already happened (it does nothing).  

I want to check this in later today, so speak now if you have issues.

danmosemsft:
We also need System.Private.Corelib.pdb to work with code coverage. If you change that to portable, we'll need the same trick.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -50,6 +50,39 @@
...
+      </PortableDllsToConvert>
+
+      <ExistingPortableDllsToConvert Include="@(PortableDllsToConvert)" Condition="Exists('%(PortableDllsToConvert.Identity)') AND Exists('%(PortableDllsToConvert.PdbPath)')"/>
+    </ItemGroup>
+
+    <!-- There are two file in CoreFX, (System.Security.Permissions and Microsoft.Win32.Registry.AccessControl.pdb)  
+         are generated by the GenFacades task and lack a Debug diretory (see https://github.com/dotnet/buildtools/issues/1739)
+         This causes ConvertPortablePdbsToWindowsPdbs to fail.    Work around it right now by using WarnAndContinue 
+         Soon we won't need this code, (since OpenCover will support portable PDBs) so it is OK to just give up for now.  -->
+    <ConvertPortablePdbsToWindowsPdbs Files="@(ExistingPortableDllsToConvert)" ContinueOnError="WarnAndContinue"/>
```
dagood: Is the known AV issue (https://github.com/dotnet/buildtools/issues/1737) likely to cause much pain in this context? I don't know when code coverage runs and who failures here would impact.

Here's a the retry workaround if you don't have a link handy: https://github.com/dotnet/buildtools/blob/e639cfa061b4f490f60baabc499123eb826055b6/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets#L248-L273.

vancem:
> We also need System.Private.Corelib.pdb to work with code coverage. If you change that to portable, we'll need the same trick.

I did some snooping around on this, and there does not seem to be much automation for coverage support for System.Private.Corelib directly.    Here is some docs in coreFX that tell you about it. 

https://github.com/dotnet/corefx/blob/master/Documentation/building/code-coverage.md#code-coverage-with-systemprivatecorelib-code

And I could not find anything in CoreCLR (there is coverage, but it is for native (and seems to be Linux only).   

If anyone knows of anything that I should be testing let me know, but to the best of my knowledge the only thing are these instructions for doing coverage of things with System.private.Corelib by hand.  

I can update those instructions, but I will do that as part of the CoreFX checkin.

danmosemsft:
@vancem yes correct, coverage with s.p.corelib requires manual steps. It's improving (eg., we now get the PDB for you) but there are still issues (eg., will not load crossgen PDB last I tried). My goal is to make it "just work". If temporarily the corelib PDB needs to be manually rebuilt with a flag we could live with that if we have to.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
vancem: I would liked to avoid preemptively working around #1737.    If an AV happens my expectation is that conversion of the PDB will fail (note the WarnOnError).  This means code coverage is likely to produce no results, but only for that one DLL, and it is not likely to be the same DLL each time.  

Finally this fix itself should only be useful for a few months.  I think it is OK to simply live with this in the mean time (or fix #1737)

dagood: I'm pretty sure the AV crashes the entire process in an uncatchable way. At least, on Desktop people say it's a bad idea, and on Core the API doesn't seem to exist.

I don't know who it will impact, but living with the flakiness is ok to me.

vancem: You are right that WarnOnError will probably not work.  It will probably crash the whole coverage run.   

I am willing to monitor it to see how it goes, if it happens once a week (or even once a month), I will react to it.

PR #1741, opened 2017-10-15
Add "UpdateToRemote" concept for better submodule update

dagood:
There are two significant VersionTools API changes here to support submodule update (https://github.com/dotnet/source-build/issues/151) better:
 * Extract `UpdateToRemoteDependencies` from `UpdateDependenciesAndSubmitPullRequest`. This makes it much easier to run an auto-update dry run locally, simulating what changes would show up in an auto-update PR.
   * Updating submodules in a checked-out repo is a significant use case. (Although this is actually useful for any updater... surprised nothing drove this change sooner.)
 * Add `IDependencyInfo` so the core code can support updates that are unrelated to `DependencyBuildInfo` (a dotnet/versions `build-info`). I cleaned up the code to consistently use this interface.
   * This allows for `SubmoduleDependencyInfo`, which points to a repo/branch/commit combination on GitHub.

The submodule updater behavior now matches the package updater's: `UpdateDependencies` checks to see if your working directory state matches the commit you have checked out. `UpdateToRemoteDependencies[AndSubmitPullRequest]` actually unpins from the state stored in the commit (in the index).

The earlier submodule updater would always update to latest on `UpdateDependencies`, which would work terribly combined with another type of updater.

I've been developing this against source-build dev/release/2.0 with submodule updaters. I also tested a few scenarios (including full PR flow) in corefx release/2.0.0 to make sure VersionTools still generally works after these changes.

FYI @MichaelSimons this will require some changes in the dotnet-docker-nightly updaters when it goes to buildtools/master. If you think it's worthwhile, I believe this will let you clean up [the synthetic build infos](https://github.com/dotnet/dotnet-docker-nightly/blob/e0c1809a605c3c840e1e6c0064ceea2f7e65d23f/update-dependencies/Program.cs#L88-L96) and instead implement your own `IDependencyInfo`.


src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
```diff
@@ -30,18 +36,21 @@ public class GitHubClient : IDisposable
...
         {
-            _auth = auth;
+            Auth = auth;
 
             _httpClient = new HttpClient();
             _httpClient.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
-            _httpClient.DefaultRequestHeaders.Add("Authorization", $"token {auth.AuthToken}");
-            _httpClient.DefaultRequestHeaders.Add("User-Agent", auth.User);
+            _httpClient.DefaultRequestHeaders.Add("User-Agent", auth?.User ?? DefaultUserAgent);
+            if (auth?.AuthToken != null)
```
eerhardt: Should all the usages of `Auth` be updated to allow for `null`?

src/Microsoft.DotNet.VersionTools/Dependencies/BuildOutput/DependencyBuildInfo.cs
```diff
@@ -7,9 +7,9 @@
 using System.Collections.Generic;
 using System.Linq;
 
-namespace Microsoft.DotNet.VersionTools.Dependencies
+namespace Microsoft.DotNet.VersionTools.Dependencies.BuildOutput
 {
-    public class DependencyBuildInfo
+    public class DependencyBuildInfo : IDependencyInfo
```
eerhardt: I think this name should be flipped around.  This isn't a "BuildInfo".  This is a "DependencyInfo"

src/Microsoft.DotNet.VersionTools/Dependencies/IDependencyUpdater.cs
```diff
@@ -14,6 +14,6 @@ public interface IDependencyUpdater
         /// <summary>
         /// Updates based on the given build infos and returns build infos used during update.
```
eerhardt: Need to update this comment `build infos and returns build infos`

src/Microsoft.DotNet.VersionTools/Dependencies/IDependencyUpdater.cs
```diff
@@ -14,6 +14,6 @@ public interface IDependencyUpdater
...
         /// </summary>
-        IEnumerable<DependencyUpdateTask> GetUpdateTasks(IEnumerable<DependencyBuildInfo> dependencyBuildInfos);
+        IEnumerable<DependencyUpdateTask> GetUpdateTasks(IEnumerable<IDependencyInfo> dependencyBuildInfos);
```
eerhardt: should rename the parameter - `dependencyBuildInfos` to `dependencyInfos`

src/Microsoft.DotNet.VersionTools/Dependencies/IDependencyInfoProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.VersionTools.Dependencies
+{
+    public interface IDependencyInfoProvider : IDependencyUpdater
+    {
+        /// <summary>
+        /// Create a dependency info that this updater depends on. An updater implements this when
```
eerhardt: I wonder if this is mixing responsibilities too much.  Should an "updater" be responsible for creating these dependency infos?

It seems to me that the Updater's responsibility should be:  "given a set of dependency infos, update the repo".  But here we are adding the responsibility of also creating those dependency infos.  It seems like that should be separate.

src/Microsoft.DotNet.VersionTools/Dependencies/DependencyUpdateTask.cs
```diff
@@ -18,21 +18,21 @@ public class DependencyUpdateTask : Task<DependencyUpdateResults>
         /// <summary>
         /// The build infos that would be used during this update.
```
eerhardt: update comment ` build infos`

src/Microsoft.DotNet.VersionTools/Automation/DependencyUpdateResults.cs
```diff
@@ -12,22 +14,30 @@ namespace Microsoft.DotNet.VersionTools.Automation
 {
     public class DependencyUpdateResults
     {
+        public IEnumerable<IDependencyInfo> UsedInfos { get; }
+
         public IEnumerable<BuildInfo> UsedBuildInfos { get; }
```
eerhardt: I think we can/should get rid of `UsedBuildInfos` and just have `IEnumerable<IDependencyInfo> UsedInfos`.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
```diff
@@ -66,22 +63,47 @@
...
-                                            GitHubAuthor="$(GitHubAuthor)"
-                                            NotifyGitHubUsers="@(NotifyGitHubUsers)"
-                                            CurrentRefXmlPath="$(CurrentRefXmlPath)"
-                                            AlwaysCreateNewPullRequest="$(AlwaysCreateNewPullRequest)" />
+  <!--
+    Deprecated auto-upgrade PR target. Redirects to the target's new name.
+  -->
+  <Target Name="UpdateDependenciesAndSubmitPullRequest"
+          DependsOnTargets="UpdateToRemoteDependenciesAndSubmitPullRequest">
+    <Message Text="'UpdateDependenciesAndSubmitPullRequest' has been renamed to 'UpdateToRemoteDependenciesAndSubmitPullRequest'. The old name is deprecated. Ran the new target instead."
```
weshaggard: I don't think it is worth breaking usages of this. Instead I would just keep it working or just stick with this name.

weshaggard: Nevermind it does continue to work you just have a message. That's reasonable.

src/Microsoft.DotNet.VersionTools/Automation/DependencyUpdateUtils.cs
```diff
@@ -11,34 +12,53 @@ namespace Microsoft.DotNet.VersionTools.Automation
...
         {
-            return updaters.SelectMany(updater => updater.GetUpdateTasks(buildInfoDependencies));
+            IDependencyInfo[] updaterDependencyInfos = updaters
+                .OfType<IDependencyInfoProvider>()
+                .SelectMany(updater => updater.CreateDependencyInfos())
+                .ToArray();
+
+            foreach (var info in updaterDependencyInfos)
+            {
+                modifyUpdaterDependencyInfo?.Invoke(info);
```
eerhardt: This is a case where I think having the caller create the DependencyInfos would help.  The caller can manipulate the DependencyInfo anyway it wants, and we don't need this `modifyUpdaterDependencyInfo` action at all here.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
eerhardt: (nit) `Ran the new target instead` => `Run`

eerhardt: Why the change at all?  Just to add the word `remote`?  Is it necessary?

src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
dagood: There are 3 other usages I can find, and I don't think `null` makes sense for them (but I do think they deserve handling):
 * 2 in `PutGitHubFileAsync`, which I think I should add an early descriptive exception for to be clear it's not possible anonymously. (And I think I should add that to other methods that depend on the headers being set too.)
 * 1 passing `client.Auth` when creating a `PullRequestCreator` in `UpdateToRemoteDependenciesAndSubmitPullRequest`. `PullRequestCreator` goes on to use `.User` and `.Email`. The build task would have already failed before this point, but might as well guard against it in the `PullRequestCreator` constructor to show intent and for external use.

src/Microsoft.DotNet.VersionTools/Dependencies/BuildOutput/DependencyBuildInfo.cs
dagood: Agreed, `BuildDependencyInfo` is fine by me.

If we wanted to a bit more descriptive, `BuildInfoDependencyInfo` would be the "full name", imo... but repeating `Info` like that seems bad. (Maybe `IDependencyInfo` (or `BuildInfo`) could use a better name.) I'm fine with `BuildDependencyInfo` so I'll stick with that unless someone has a problem with it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
dagood: This is intended to describe the automatic redirect MSBuild did with `DependsOnTargets` rather than tell the user what to do. Maybe I should use `CallTarget` and say `Running` before to make it clearer.

dagood: Here are the steps that landed me here:

 1. There needs to be a way for a dev to run an update that actually pulls in the remote status--this is specifically a goal for submodule updating. `UpdateDependenciesAndSubmitPullRequest` had that logic embedded inside, so I refactored it into a `UpdateToRemoteDependencies` target. Then I had:
    * `UpdateDependencies`
    * `UpdateToRemoteDependencies`
    * `UpdateDependenciesAndSubmitPullRequest`
 2. Now `UpdateDependenciesAndSubmitPullRequest` is misleading. It's actually adding onto `UpdateToRemoteDependencies`, so I renamed it with a back-compat target:
    * `UpdateDependencies`
    * `UpdateToRemoteDependencies`
    * `UpdateToRemoteDependenciesAndSubmitPullRequest`
    * (`deprecated`) `UpdateDependenciesAndSubmitPullRequest`

The `...And` is the source of the clunkiness here. I've kept it so far because it seems more reasonable to keep it than try to export all the info we need from `UpdateToRemoteDependencies` and stuff it back into a "`SubmitPullRequest`" task. I haven't seriously considered making that split, though... now that I'm thinking about it, it actually seems like it could be reasonable. I'll take a stab at it when I'm done with the other comments.

src/Microsoft.DotNet.VersionTools/Dependencies/IDependencyInfoProvider.cs
dagood: Agreed, the provider idea was a lopsided attempt at cohesion, and it would be better to keep infos and updaters distinct. Changing.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -22,14 +22,12 @@ public abstract class BaseDependenciesTask : BuildTask
...
-        internal const string CurrentBranchMetadataName = "CurrentBranch";
         internal const string PackageIdMetadataName = "PackageId";
         internal const string VersionMetadataName = "Version";
-
-        internal const string AutoUpgradeBranchMetadataName = "Version";
+        internal const string DependencyTypeMetadataName = "DependencyType";
 
         [Required]
-        public ITaskItem[] DependencyBuildInfo { get; set; }
+        public ITaskItem[] DependencyInfo { get; set; }
```
weshaggard: Are there any targets calling this task that don't exist in BuildTools? If so then changing this will break them.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateToRemoteDependencies.cs
```diff
@@ -0,0 +1,128 @@
...
+            // Use the commit hash of the remote dotnet/versions repo master branch.
+            string versionsCommitHash = client
+                .GetReferenceAsync(new GitHubProject("versions", "dotnet"), "heads/master")
+                .Result.Object.Sha;
+
+            DependencyUpdateResults updateResults = DependencyUpdateUtils.Update(
+                CreateUpdaters().ToArray(),
+                CreateDependencyInfos(true, versionsCommitHash).ToArray());
+
+            // Update CurrentRef and CurrentBranch for each applicable build info used.
```
weshaggard: Seems like you aren't using CurrentBranch any longer, so perhaps the comment is stale.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateToRemoteDependencies.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+
+            return updateResults;
+        }
+
+        private ITaskItem FindBuildDependencyInfoItem(string name)
+        {
+            return DependencyInfo.SingleOrDefault(item =>
+                item.ItemSpec == name &&
```
weshaggard: Should these compares be case-insensitive?

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
dagood: Not that I'm aware of, and searching the `dotnet` org for `UpdateDependencies` and `VerifyDependencies` only shows the usages in the BuildTools targets.

src/Microsoft.DotNet.VersionTools/Dependencies/BuildOutput/BuildDependencyInfo.cs
```diff
@@ -7,9 +7,9 @@
 using System.Collections.Generic;
 using System.Linq;
 
-namespace Microsoft.DotNet.VersionTools.Dependencies
+namespace Microsoft.DotNet.VersionTools.Dependencies.BuildOutput
 {
-    public class DependencyBuildInfo
+    public class BuildDependencyInfo : IDependencyInfo
```
weshaggard: Note: These are breaking public API changes that will need to be consumed by anyone that is calling this library directly.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateToRemoteDependencies.cs
dagood: The `BuildInfo`'s name was originally set from the `DependencyInfo`'s `ItemSpec` in `CreateBuildInfo`, so there's no reason for the case to differ.

That said, there's no good reason to do a search here to "rediscover" the connection. Really, the item and the `DependencyInfo` should be stored together.

What comes to mind to fix that is subclassing `BuildDependencyInfo` in `Microsoft.DotNet.Build.Tasks` to tack on the `ITaskItem` that created the info. (Subclassing to avoid referencing MSBuild libs in VersionTools.) I'll work on changing to this.

src/Microsoft.DotNet.VersionTools/Dependencies/BuildOutput/BuildDependencyInfo.cs
dagood: Noted, the two known places are dotnet-docker-nightly and CLI.

weshaggard: Didn't Roslyn also start using this library?

dagood: It looks like [RoslynPublish](https://github.com/dotnet/roslyn/tree/3d57b165d1eb872389b60459493ebd583f5dc5f9/src/Tools/RoslynPublish) uses it to publish to dotnet/versions, but dependency changes are still done with [RepoUtil](https://github.com/dotnet/roslyn/tree/3d57b165d1eb872389b60459493ebd583f5dc5f9/src/Tools/RepoUtil).

dagood:
I have commit https://github.com/dagood/buildtools/commit/5df0ad3933aefb4ee8ba2bdaf664eb1c86b6b2e4 with the change to split off `CreatePullRequest` I talked about here: https://github.com/dotnet/buildtools/pull/1741#discussion_r145286213. Instead of pushing that here, though, I think it would be better to just keep the name `UpdateDependenciesAndSubmitPullRequest` for now, and I can submit the split in another PR.

PR #1742, opened 2017-10-16
Use Sleet for package management

jcagme:
Since we decided we wanted to support nuget install(s) some changes in how we publish and delete packages were needed. We decided to use Sleet(https://github.com/emgarten/Sleet) for the heavy lifting operations of updating index.json files, create/removing registration versions, etc.

With Sleet some pre-setup of the blob is needed but it offers init commands to do so.


src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -22,84 +22,64 @@ sealed class BlobFeedAction
...
-                string feedUrl = m.Groups["feedurl"].Value;
-
-                bool isPublic = string.IsNullOrWhiteSpace(m.Groups["token"].Value);
-                this.feed = new BlobFeed(accountName, accountKey, containerName, relativePath, feedUrl, string.IsNullOrWhiteSpace(indexDirectory) ? Path.GetTempPath() : indexDirectory, Log, isPublic);
-            }
-            else
-            { 
-                throw new Exception("Unable to parse expected feed. Please check ExpectedFeedUrl.");
-            }
+            this.feed = new BlobFeed(accountName, accountKey, containerName, packagesPath, Log);
```
weshaggard: This now means we are going to need to duplication information in our build definitions. One URL for restore and then a bunch of container names and such that needs to match for the publishing. I would much prefer if we can have just one set of information that doesn't need to be duplicated and remain in sync.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -22,84 +22,64 @@ sealed class BlobFeedAction
...
-                string packageDirectory = feed.CalculateRelativeUploadPath(item, relativePath);
-
-                await UploadAsync(CancellationToken, item, uploadPath, clientThrottle, allowOverwrite);
-
-                if (!isFlat)
+                Source source = new Source
+                {
+                    Name = feed.ContainerName,
+                    Type = "azure",
+                    Path = $"https://{feed.AccountName}.blob.core.windows.net/{feed.ContainerName}/",
```
weshaggard: Why can't this just be the feed URL that we pass in instead of constructing it manually here?

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Publish.targets
```diff
@@ -14,19 +14,18 @@
...
-                      AccountKey="$(AccountKey)"
-                      ItemsToPush="@(ItemsToPush)"
-                      IndexDirectory="$(IndexDirectory)"
-                      PublishFlatContainer="$(PublishFlatContainer)"
-                      Overwrite="$(Overwrite)" />
-    </Target>
+    <Message Importance="High" Text="Publish to $(ContainerName) started" />
+    <PushToBlobFeed ExpectedFeedUrl="$(ExpectedFeedUrl)"
+                    AccountKey="$(AccountKey)"
+                    ItemsToPush="$(ItemsToPush)"
```
dagood: Should `ItemsToPush="$(ItemsToPush)"` use `@`?

src/Microsoft.DotNet.Build.Tasks.Feed/SleetSettings.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.DotNet.Build.Tasks.Feed
+{
+    public class SleetSettings
+    {
+        public List<Source> Sources { get; set; }
+    }
+
+    public class Source
```
dagood: This class should have its own file.

src/Microsoft.DotNet.Build.Tasks.Feed/SleetLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        public void Log(ILogMessage message)
+        {
+            _log.LogMessage(message.Message, message.Level);
+        }
+
+        public Task LogAsync(LogLevel level, string data)
+        {
+            return Task.Run(() => _log.LogMessage(data, level));
```
dagood: [In NugetMsBuildLogger](https://github.com/dotnet/buildtools/blob/9bc085cef38e12857f93a94ba2c9fd1faef4052a/src/Microsoft.DotNet.Build.Tasks/NugetMsBuildLogger.cs#L96) we do the logging then return `Task.CompletedTask`. Is there a reason to use `Task.Run` here?

src/Microsoft.DotNet.Build.Tasks.Feed/Microsoft.DotNet.Build.Tasks.Feed.csproj
```diff
@@ -20,18 +20,19 @@
...
     <Compile Include="BlobFeed.cs" />
     <Compile Include="BlobFeedAction.cs" />
     <Compile Include="ConfigureInputFeed.cs" />
     <Compile Include="PushToBlobFeed.cs" />
+    <Compile Include="SleetLogger.cs" />
+    <Compile Include="SleetSettings.cs" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="PackageFiles\**\*.*" />
+    <None Include="config\sleet.json" />
```
dagood: Should this file be in the PR?

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Publish.targets
```diff
@@ -14,19 +14,18 @@
 
         PublishDependsOn can be customized at build time on the commandline to determine the pre steps required to publish 
     -->
+  <Target Name="PublishOutputLeg">
+    <Error Text="The ExpectedFeedUrl property must be set on the command line."
+          Condition="'$(ExpectedFeedUrl)' == ''" />
+    <Error Text="The AccountKey property must be set on the command line."
+          Condition="'$(AccountKey)' == ''" />
+    <Error Text="The PackagesPath property must be set on the command line."
+          Condition="'$(PackagesPath)' == ''" />
```
dagood: `Condition` is misaligned a bit.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -2,16 +2,17 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Build.Framework;
+using Microsoft.WindowsAzure.Storage;
```
karajas: This was an initial concern I had especially in the context of build from source, but we can just opt out of building this from the build.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -22,28 +23,25 @@ sealed class BlobFeedAction
...
         {
             this.Log = Log;
             Match m = Regex.Match(expectedFeedUrl, feedRegex);
-
             if (m.Success)
             {
                 string accountName = m.Groups["accountname"].Value;
                 string containerName = m.Groups["containername"].Value;
-                string relativePath = m.Groups["relativepath"].Value;
-                string feedUrl = m.Groups["feedurl"].Value;
```
weshaggard: I expect we are still going to need the relativepath support for multiple feeds in one container and the feedurl support for our secure feeds.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -22,28 +23,25 @@ sealed class BlobFeedAction
...
-
                 bool isPublic = string.IsNullOrWhiteSpace(m.Groups["token"].Value);
-                this.feed = new BlobFeed(accountName, accountKey, containerName, relativePath, feedUrl, string.IsNullOrWhiteSpace(indexDirectory) ? Path.GetTempPath() : indexDirectory, Log, isPublic);
+                feed = new BlobFeed(accountName, accountKey, containerName, Log);
+                feedUrl = expectedFeedUrl;
```
weshaggard: expectedFeedUrl generally also contains the index.json in it which is why we were parsing that out.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: Sleet depends on it. What was the issue with it?

src/Microsoft.DotNet.Build.Tasks.Feed/SleetLogger.cs
jcagme: No reason in particular. When I tried to change it to Task.CompletedTask it failed at run time claiming the property did not exist hence left it as Task.Run

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
```diff
@@ -5,9 +5,10 @@
     "System.Security.Cryptography.Algorithms": "4.2.0",
     "System.Reflection.Metadata": "1.3.0",
     "NETStandard.Library": "1.6.0",
-    "Newtonsoft.Json": "9.0.1",
-    "NuGet.Versioning": "4.3.0-preview2-4095",
-    "NuGet.Packaging": "4.3.0-preview2-4095"
+    "Newtonsoft.Json": "10.0.3",
```
dagood: Should we upgrade everything to 10.0.3 rather than just the Feed project? I'm not sure if there are any concrete problems, but it seems odd to depend on two versions.

jcagme: Yeah I did think about doing that but did not know if there were hard dependencies on a particular version. In my case, Sleet depends on 10.0.3. I'll upgrade all.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: Nothing, just wanted to confirm that this is a new dependency being included.

dependencies.props
```diff
@@ -41,7 +41,7 @@
       <CurrentRef>$(CoreClrCurrentRef)</CurrentRef>
     </RemoteDependencyBuildInfo>
     <RemoteDependencyBuildInfo Include="ProjectKTfs">
-      <BuildInfoPath>$(BaseDotNetBuildInfo)projectk-tfs/$(DependencyBranch)</BuildInfoPath>
+      <BuildInfoPath>$(BaseDotNetBuildInfo)projectk-tfs/release/1.0.0</BuildInfoPath>
```
weshaggard: This change looks like a merge issue as I merged it into master yesterday.

weshaggard: I suggest rebasing your changes on master.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -22,28 +23,26 @@ sealed class BlobFeedAction
...
             {
                 string accountName = m.Groups["accountname"].Value;
                 string containerName = m.Groups["containername"].Value;
                 string relativePath = m.Groups["relativepath"].Value;
-                string feedUrl = m.Groups["feedurl"].Value;
 
-                bool isPublic = string.IsNullOrWhiteSpace(m.Groups["token"].Value);
-                this.feed = new BlobFeed(accountName, accountKey, containerName, relativePath, feedUrl, string.IsNullOrWhiteSpace(indexDirectory) ? Path.GetTempPath() : indexDirectory, Log, isPublic);
+                feed = new BlobFeed(accountName, accountKey, containerName, relativePath, Log);
+                feedUrl = expectedFeedUrl;
```
weshaggard: Isn't this still going to have the index.json in the url? If so is that what Sleet is expecting?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -22,28 +23,26 @@ sealed class BlobFeedAction
...
         public BlobFeed feed;
-        public int MaxClients { get; set; } = 8;
-        
-        const string feedRegex = @"(?<feedurl>https:\/\/(?<accountname>[^\.-]+)(?<domain>[^\/]*)\/((?<token>[a-zA-Z0-9+\/]*?\/\d{4}-\d{2}-\d{2})\/)?(?<containername>[^\/]+)\/(?<relativepath>.*)\/)index\.json";
+        public string feedUrl;
 
-        public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDirectory, MSBuild.TaskLoggingHelper Log)
+        const string feedRegex = @"(?<feedurl>https:\/\/(?<accountname>[^\.-]+)(?<domain>[^\/]*)\/((?<token>[a-zA-Z0-9+\/]*?\/\d{4}-\d{2}-\d{2})\/)?(?<containername>[^\/]+)\/(?<relativepath>.*\/))index\.json";
+
+        public BlobFeedAction(string expectedFeedUrl, string accountKey, ITaskItem[] itemstoPush, MSBuild.TaskLoggingHelper Log)
```
weshaggard: Where is itemsToPush used?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -19,13 +19,8 @@ public class PushToBlobFeed : MSBuild.Task
         [Required]
         public string AccountKey { get; set; }
 
-        [Required]
```
weshaggard: Why is this not required any longer?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -19,13 +19,8 @@ public class PushToBlobFeed : MSBuild.Task
...
         public ITaskItem[] ItemsToPush { get; set; }
 
-        public string IndexDirectory { get; set; }
```
weshaggard: Does Sleet use any temp directories that we need to override the location for? If so we should still have a version of this property to repoint those.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -19,13 +19,8 @@ public class PushToBlobFeed : MSBuild.Task
...
-
-        public bool PublishFlatContainer { get; set; }
```
weshaggard: There are still potential scenarios where we want to publish to a flat directory. What is the plan for those cases?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -42,26 +37,10 @@ public override bool Execute()
                 {
```
weshaggard: If we don't make ItemsToPush required again then we should handle the null/empty case instead of throwing. FWIW I think we should make it required.

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
weshaggard: This is going to cause issues because it is higher then the version that lives in the CLI we are using. Can we avoid bumping the version?

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
```diff
@@ -5,9 +5,10 @@
...
+    "NuGet.Versioning": "4.3.0",
+    "NuGet.Packaging": "4.3.0",
+    "SleetLib": "2.2.7"
```
weshaggard: Please make sure this dependency ends up in the package dependency for the feed package.

Also do remember that given we need to build all this from source we will need to make sure Sleet conforms to being built from source in our overall efforts.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: It's fine. I remove index.json from the URL on line 75

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: Actually it is not needed here since I am using the passed in actual value in PushToBlobFeed.cs (44)

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
jcagme: Sleet uses its own internal temp directories but there is no way to override the directory

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
jcagme: @weshaggard  Sleet has a hard dependency on 10.0.3. I did try updating in the code base but it fails to build. How hard would it be to update CLI's?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
weshaggard: Why do it there as opposed to here?

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
weshaggard: That will be a large undertaking. Plus we want to stick with the 2.0.0 RTM CLI for most if not all of this release.

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
jcagme: As per our offline conversation we'll leave the package dependency instead of the DLL. I've updated the nuspec

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
jcagme: In Sleet all packages live in flatcontainers so no need to make this configurable

dependencies.props
jcagme: I did rebase and this remained, master has release/1.0.0. Should I change it to the vairable?

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
jcagme: Starting on Sleet 2.0.0 where a lot of perf improvements were done a dependency on 10.x was introduced to use Json.NET async pattern. Moving back to 9.0.1 will require several code changes and most likely some perf trade offs. @weshaggard what do you suggest would be a good middle ground?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
weshaggard: This is exactly the confusion I was talking about. This flat container is a different concept then then flatcontainer folder Sleet produces. This flatcontainer option was just about putting all files in the root directory with no indexing at all.

src/Microsoft.DotNet.Build.Tasks.Feed/project.json
weshaggard: I thought we decided to just include the dll in the Feed package instead of adding the package dependency.

dependencies.props
weshaggard: This matches master but it shouldn't be showing up in your diff if you did a rebase.

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
```diff
@@ -16,7 +16,9 @@
     <copyright>Copyright © Microsoft Corporation</copyright>
     <tags></tags>
     <dependencies>
-     <!-- Do not add any dependencies to this package as we want to install it versionless (i.e. -ExcludeVersion) and we don't want the dependencies to be installed that way -->
+     <group targetFramework=".NETStandard1.5">
```
weshaggard: I'm good either way. Either a package dependency or including the dll, I will leave it to you.

src/Microsoft.DotNet.Build.Tasks/EncryptedConfigNuGetRestore.cs
```diff
@@ -35,7 +35,6 @@ public override bool Execute()
                 GlobalPackagesFolder = PackagesDir,
 
                 Sources = new List<string>(),
-                FallbackSources = new List<string>(),
```
karajas: Is this a relevant change?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -57,133 +56,79 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
...
-            if (isLeaseRequired)
-            {
-                try
-                {
-                    leaseId = blobLease.Acquire();
-                }
-                catch (Exception)
+                // In case the first Push attempt fails with an InvalidOperationException we Init the feed and retry the Push command once.
+                // Sleet internally retries 5 times on each package when the push operation fails so we don't need to retry ourselves.
+                for (int i = 0; i <= 1; i++)
```
karajas: Sleet does the retries. Is this loop needed?

jcagme: Yes in case an Init is needed, this loop does not retry on op failure

src/Microsoft.DotNet.Build.Tasks/EncryptedConfigNuGetRestore.cs
jcagme: This property is not there in "4.3.0" a change made in project.json

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
jcagme: I'll go with the package ref

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
karajas: But it only loops once?

jcagme: Think of it as a do/while without the need of having extra variables in place. If the call works then we just return if it fails with an expected exception it tried to init just once, else it fails

dagood: It seems like it would be clearer/simpler to make the calls more self-contained and write out the logic.

```C#
try
{
    return await push();
}
catch (InvalidOperationException ex) when (ex.Message.Contains("init"))
{
    if (await init())
    {
        return await push();
    }
    return false;
}
```

(I'm imagining using async funcs as a minimal-change way of getting it done.)

src/Microsoft.DotNet.Build.Tasks.Feed/Microsoft.DotNet.Build.Tasks.Feed.csproj
```diff
@@ -20,18 +20,19 @@
...
+    <Compile Include="SleetLogger.cs" />
+    <Compile Include="SleetSettings.cs" />
+    <Compile Include="SleetSource.cs" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="PackageFiles\**\*.*" />
+    <None Include="PackageFiles\Publish.targets">
+      <SubType>Designer</SubType>
+    </None>
+    <None Include="PackageFiles\Restore.targets" />
```
dagood: I think you should keep the `PackageFiles\**\*.*` rather than listing each one. (I know VS sometimes likes to expand it.)

src/Microsoft.DotNet.Build.Tasks.Feed/SleetSettings.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
```
dagood: Missing copyright header (+other new files).

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -57,133 +54,76 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
...
-                            item,
-                            uploadPath,
-                            leaseId);
+                    Log.LogWarning($"Sub-feed {source.FeedSubPath} has not been initialized. Initializing now...");
+                    bool result = await InitCommand.RunAsync(settings, fileSystem, true, true, new SleetLogger(Log), CancellationToken);
 
+                    if (result)
+                    {
+                        Log.LogMessage($"Initializing sub-feed {source.FeedSubPath} succeeded!");
+                        result = await PushCommand.RunAsync(settings, fileSystem, items.ToList(), allowOverwrite, !allowOverwrite, new SleetLogger(Log));
```
dagood: The reason I suggested async funcs was to avoid repeating all the RunAsync parameters and encapsulate the logging lines for maintainability. (My understanding is the `for` was there for this reason.)

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
weshaggard: What library/package are these API's in? Do we already pull them into the dependency list?

jcagme: @weshaggard  Yes, Microsoft.Build.Framework is part of the Microsoft.Build.Framework package and Microsoft.WindowsAzure.Storage is contained in Sleet which is part of the dependency list

weshaggard: >Microsoft.WindowsAzure.Storage is contained in Sleet which is part of the dependency list

By contained I assume you mean a dependency? Hopefully that dependency works on dotnet core.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -20,29 +21,35 @@ sealed class BlobFeedAction
...
             {
                 string accountName = m.Groups["accountname"].Value;
                 string containerName = m.Groups["containername"].Value;
                 string relativePath = m.Groups["relativepath"].Value;
-                string feedUrl = m.Groups["feedurl"].Value;
-
-                bool isPublic = string.IsNullOrWhiteSpace(m.Groups["token"].Value);
-                this.feed = new BlobFeed(accountName, accountKey, containerName, relativePath, feedUrl, string.IsNullOrWhiteSpace(indexDirectory) ? Path.GetTempPath() : indexDirectory, Log, isPublic);
+                feed = new BlobFeed(accountName, accountKey, containerName, relativePath, Log);
+                feedUrl = expectedFeedUrl.Replace("index.json", string.Empty);
```
weshaggard: I still think the feedUrl should come out from the regex. The goal was we might end up needing the pass the regex as an extra parameter in the build. It is also better then a blind Replace which doesn't take into consideration if "index.json" exists anywhere else in the URL. It is unlikely but still I think good practice to not assume.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -109,83 +115,47 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
...
-            string leaseId = string.Empty;
-            //this defines the lease for 15 seconds (max is 60) and 3000 milliseconds between requests
-            AzureBlobLease blobLease = new AzureBlobLease(feed.AccountName, feed.AccountKey, string.Empty, feed.ContainerName, uploadPath, Log, "15", "5000");
+            return settings;
+        }
 
-            bool isLeaseRequired = allowOverwrite && await feed.CheckIfBlobExists(uploadPath);
+        private AzureFileSystem GetAzureFileSystem()
+        {
+            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(source.ConnectionString);
```
weshaggard: Can and should this be cached for all the pushes?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
weshaggard: At least CoreCLR will be broken by not having this "flat" option, so they will need to be fixed once your change flows. I do still think we may need to have this option to publish to a flat container, essentially by going around sleet and publish things to the Assets folder for example. What is your expectation for how we publish just to an azure blob container? Would we use a different task for that?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -20,29 +21,35 @@ sealed class BlobFeedAction
...
+        private string feedUrl;
+        private SleetSource source;
 
         public BlobFeed feed;
-        public int MaxClients { get; set; } = 8;
-        
-        const string feedRegex = @"(?<feedurl>https:\/\/(?<accountname>[^\.-]+)(?<domain>[^\/]*)\/((?<token>[a-zA-Z0-9+\/]*?\/\d{4}-\d{2}-\d{2})\/)?(?<containername>[^\/]+)\/(?<relativepath>.*)\/)index\.json";
 
-        public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDirectory, MSBuild.TaskLoggingHelper Log)
+        public BlobFeedAction(string expectedFeedUrl, string accountKey, ITaskItem[] itemstoPush, MSBuild.TaskLoggingHelper Log)
```
weshaggard: You are still taking itemsToPush here and not using it. You should remove this parameter.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: Oh! Yeah your are right, I did not understand what you meant at the beginning but now I see what you mean

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -31,231 +23,17 @@ public sealed class BlobFeed
...
-                    Thread.Sleep(3000);
-                    if (retry == 1)
-                        throw;
-                }
-            }
-            return rootIndexJsonLocation;
+            RelativePath = relativePath;
         }
 
         public bool IsSanityChecked(IEnumerable<string> items)
```
weshaggard: Should this move over to where is is actually used now?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: Yeah, rebase/merge snafu

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -57,49 +64,48 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
...
-                    List<string> listAzureBlobs = await ListAzureBlobs.ListBlobs(Log, feed.AccountName, feed.AccountKey, feed.ContainerName, packageDirectory);
-                    if (!listAzureBlobs.Any(x => x.Contains(uploadPath)))
+                    Log.LogWarning($"Sub-feed {source.FeedSubPath} has not been initialized. Initializing now...");
+                    bool result = await InitAsync();
+
+                    if (result)
                     {
-                        throw new Exception($"Uploaded package {uploadPath} is not present on feed. Cannot update index.json.");
+                        Log.LogMessage($"Initializing sub-feed {source.FeedSubPath} succeeded!");
+                        result = await PushAsync(items.ToList(), allowOverwrite);
```
weshaggard: Don't do this in the catch block. Use a retry pattern.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: Talked offline.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -20,86 +22,87 @@ sealed class BlobFeedAction
...
+                    try
                     {
-                        throw new Exception($"Uploaded package {uploadPath} is not present on feed. Cannot update index.json.");
+                        bool result = await PushAsync(items.ToList(), allowOverwrite);
+                        return result;
+                    }
+                    catch (InvalidOperationException ex) when (ex.Message.Contains("init"))
+                    {
+                        Log.LogWarning($"Sub-feed {source.FeedSubPath} has not been initialized. Initializing now...");
+                        bool result = await InitAsync();
```
weshaggard: Doing this InitAsync in the catch block is probably also not a good idea. Does sleet not have a way to check to see if the feed is initialized?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
dagood: Is this a general style, or are there issues that come up? (Is it because awaiting in catch is a recent C# feature?) I suggested in my review to move away from the `for (int i = 0; i <= 1; i++)` because it was raising some questions.

weshaggard: In general you shouldn't be calling code that might throw inside a catch block as if it throws there it will terminate the process.

weshaggard: As for the for loop retry pattern I agree it is a little odd but not a big deal. I especially find the `<=1` part odd instead of just a `<2`.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
weshaggard: @jcagme please do this in a clean-up change.

PR #1743, opened 2017-10-18
Remove project.json from init-tools

danmosemsft:
The "dotnet --version" check has been broken by reversioning dotnet.exe to 15. 
After discussion with @weshaggard, I'm just removing it and all project.json support.

Need to figure out how to test this.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -49,9 +27,8 @@ else
     <PackageReference Include=\"Microsoft.Portable.Targets\" Version=\"$__PORTABLETARGETS_VERSION\" />
   </ItemGroup>
 </Project>"
-    __PROJECT_EXTENSION=csproj
-    __PUBLISH_TFM=netcoreapp2.0
-fi
+__PROJECT_EXTENSION=csproj
```
safern: I don't think `__PROJECT_EXTENSION` is used anymore after this change, so can be removed.

danmosemsft:
Tested Unix and Windows.

PR #1744, opened 2017-10-18
Remove project.json from init-tools (#1743)

weshaggard:
Porting https://github.com/dotnet/buildtools/pull/1743 to release/2.0.0

cc @danmosemsft @eerhardt


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -1,4 +1,4 @@
-@echo off
+@if not defined _echo @echo off
```
danmosemsft: I wonder what the Unix equivalent is

weshaggard: not sure.

PR #1746, opened 2017-10-19
Fix GenApi so that when creating a PNSE Assembly Deconstructors throw as well

safern:
Currently for deconstructors we're just producing an empty body and this causes the build to fail if we are generating a PNSE Assembly because in implementation deconstroctors must have a body. This doesn't hurt the regular GenApi scenario of creating an AssemblyReference since `_platformNotSupportedExceptionMessage` will be null and it will produce an empty body. I've tested both scenarios locally and they work fine. 

cc: @weshaggard 

FYI: @pjanotti


safern:
> LGTM - Please verify that the destructor case correctly goes through WriteMethodBody.

Yes I already did for both cases, when creating a reference assembly and when creating a platform not soportes assembly and works fine!

PR #1748, opened 2017-10-19
Add feed package to buildtools init

chcosta:
1. Adds the feed package to .toolsversions so that we can publish in the official build def and not have a chicken-egg problem which may occur if we used the version of feeds from the current build

2. ie, it'd be nice if there was a way to use this package without having to always define a wrapper project in the repo which is used simply to define itemstopush and to pass through the rest of the variables..

### DIFF
--- a/.toolversions
+++ b/.toolversions
@@ -1,3 +1,4 @@
 Microsoft.DotNet.BuildTools=1.0.27-prerelease-01205-03
 Microsoft.DotNet.BuildTools.Run=1.0.1-prerelease-01205-03
+Microsoft.DotNet.Build.Tasks.Feed=1.0.0-prerelease-02119-02
 NuGet.CommandLine=3.4.3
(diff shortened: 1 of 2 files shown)


chcosta:
@weshaggard @karajas , I'll likely merge this before this afternoon unless there are any concerns to address.

weshaggard:
(2) - That sounds like something worth putting into the default targets file for this package.

.toolversions
```diff
@@ -1,3 +1,4 @@
 Microsoft.DotNet.BuildTools=1.0.27-prerelease-01205-03
 Microsoft.DotNet.BuildTools.Run=1.0.1-prerelease-01205-03
+Microsoft.DotNet.Build.Tasks.Feed=1.0.0-prerelease-02119-02
```
weshaggard: Is this actually being used in BuildTools?

chcosta: Not at the moment, this would be used for publishing to the transport feed during official builds.  The alternative I considered was to just use the built packages from the build, but I figured that would lead to a chicken and egg problem so I'd rather use a stable method.

weshaggard: I'm fine with it being used by I don't see you plugging it into the publish step. Am I missing that somewhere?

chcosta: The publishing happens in the VSTS definition (which is not checked in for BuildTools), so that's a subsequent (hidden) update, unfortunately.

PR #1750, opened 2017-10-19
Include props that imports custom feed task

karajas:
/cc @eerhardt @chcosta @weshaggard


src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
```diff
@@ -20,11 +20,11 @@
...
   <files>
-    <file src="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll" target="lib\netstandard1.5" />
-    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib\netstandard1.5" Exclude="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll"/>
-    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib\net46" />
-    <file src="Microsoft.DotNet.Build.Tasks.Feed\PackageFiles\**\*.*" target="lib\netstandard1.5" />
-    <file src="Microsoft.DotNet.Build.Tasks.Feed\PackageFiles\**\*.*" target="lib\net46" />
+    <file src="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll" target="build\netstandard1.5" />
+    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="build\netstandard1.5" Exclude="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll"/>
+    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="build\net46" />
+    <file src="Microsoft.DotNet.Build.Tasks.Feed\PackageFiles\*.props" target="build\netstandard1.5" />
```
weshaggard: While I think this is fine I'm curious to understand what your plan is for importing the correct props file in BuildTools based repos?

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
```diff
@@ -20,11 +20,11 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll" target="lib\netstandard1.5" />
-    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib\netstandard1.5" Exclude="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll"/>
-    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib\net46" />
-    <file src="Microsoft.DotNet.Build.Tasks.Feed\PackageFiles\**\*.*" target="lib\netstandard1.5" />
-    <file src="Microsoft.DotNet.Build.Tasks.Feed\PackageFiles\**\*.*" target="lib\net46" />
+    <file src="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll" target="build\netstandard1.5" />
+    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="build\netstandard1.5" Exclude="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll"/>
```
weshaggard: Why are you putting the net45 files into the netstandard1.5 package folder?

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
karajas: Isn't the plan to make a project that has a reference to this package?

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
karajas: copying the dependencies of the feed.dll

chcosta: Wes' point is that those would be the wrong dependencies though.

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Publish.targets
```diff
@@ -1,32 +0,0 @@
...
-         
-         Optional parameters:
-         IndexDirectory - the directory in which the temporary index.json files live
-         PublishFlatContainer - push without generating the feed useful for flat assets like installers
-         Overwrite - Allow overwrite, will skip if file already exists when false
-
-        PublishDependsOn can be customized at build time on the commandline to determine the pre steps required to publish 
-    -->
-
-    <Target Name="PublishOutputLeg" DependsOnTargets="@(PublishDependsOn)" >
```
chcosta: Why are we losing the ability to use these targets?

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
weshaggard: that project would be init-tools.msbuild and it would be used to publish but there isn't necessarily a clear way to do the import in the other projects like publish.proj. I think may need to figure out something based on BuildToolsTaskDir perhaps.

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
karajas: Ah will fix.

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
karajas: I was thinking we move towards the model of treating it like an actual package reference. Have the repo's publish.proj reference it, restore for the correct tfm and auto import the appropriate task based on the framework it restore for.

karajas: The current model seems to be treating it like a zip file, and unpacking it into a directory, which makes things more confusing.

weshaggard: I don't really want a bunch of random projects in corefx (outside of external and init-tools) to have PackageReferences and require being restored. Instead these tools packages I want them to be restored and published via init-tools.msbuild and it will need to produce the correct layout under the Tools directory so that the assumptions of BuildTools still holds.

chcosta: My interpretation of the model that Wes is proposing...

Have a Common.props file like...

```
<Project ToolsVersion="12.0" DefaultTargets="PublishOutputLeg" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <RunningOnCore Condition="'$(MSBuildRuntimeType)' == 'Core'">true</RunningOnCore>
    <_TaskDir>$(MSBuildThisFileDirectory)../lib/net46/</_TaskDir>
    <_TaskDir Condition="'$(RunningOnCore)' == 'true'">$(MSBuildThisFileDirectory)../lib/netstandard1.5/</_TaskDir>
  </PropertyGroup> 
 </Project>
```

The nuspec looks like...
```
  <files>
    <file src="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll" target="lib\netstandard1.5" />
    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib\netstandard1.5" Exclude="Microsoft.DotNet.Build.Tasks.Feed\Microsoft.DotNet.Build.Tasks.Feed.dll"/>
    <file src="Microsoft.DotNet.Build.Tasks.Feed.net45\*.dll" target="lib\net46" />
    <file src="Microsoft.DotNet.Build.Tasks.Feed\PackageFiles\**\*.*" target="build" />
    <file src="SomeCommonLocation\Common.props" target="build" />
    <file src="..\obj\version.txt" target="" />
  </files>
```

The Publish.targets and Restore.targets include...
```
  <Import Project="$(MSBuildThisFileDirectory)Common.props" />
  <UsingTask TaskName="PushToBlobFeed" AssemblyFile="$(_TaskDir)Microsoft.DotNet.Build.Tasks.Feed.dll"/>
```
Then the user just does either `msbuild Publish.targets blah` or `dotnet msbuild Publish.targets blah` and things just work.

Regarding Common.props: I guess there's no consistency (currently), for what desktop or core framework we are targeting (ie, lib/net45 or lib/net46 for desktop; lib/netstandard13 or lib/netstandard1.5 or something else for core), but initially lib/net46 for desktop and lib/netstandard1.5 for core would work.  I could imagine a slightly adjusted model where each package can explicitly provide its desktop or core target frameworks.

weshaggard: Yes that is close to what I'm expecting. I would put the UsingTask in the targets file in the package as well. 

Given our Tools layout though I'm not certain the relative paths you specified will work after we do the publish into the Tools directory. However I think we can figure out how to make those work based on the Tools layout as well as the package layout.  If not then perhaps a targets file in each TFM directly like @karajas has and then we could in a common place have select the correct import.

src/nuget/Microsoft.DotNet.Build.Tasks.Feed.nuspec
karajas: The bin folder for the netstandard version doesn't have any deps in it, I think this is how packaging has worked. Will test before confirming.

karajas: was missing CopyNugetImplementaions

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Common.props
```diff
@@ -0,0 +1,7 @@
+<Project ToolsVersion="12.0" DefaultTargets="PublishOutputLeg" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <RunningOnCore Condition="'$(MSBuildRuntimeType)' == 'Core'">true</RunningOnCore>
```
weshaggard: I wouldn't even define this RunningOnCore property just for this one usage.

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Common.props
```diff
@@ -0,0 +1,7 @@
...
+    <_TaskDir>$(MSBuildThisFileDirectory)net46/</_TaskDir>
```
weshaggard: I don't think these paths are going to work for both of our scenarios.

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Publish.targets
```diff
@@ -1,5 +1,6 @@
 <Project ToolsVersion="12.0" DefaultTargets="PublishOutputLeg" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <UsingTask TaskName="PushToBlobFeed" AssemblyFile="$(MSBuildThisFileDirectory)Microsoft.DotNet.Build.Tasks.Feed.dll"/>
+  <Import Project="$(MSBuildThisFileDirectory)Common.props" />
+  <UsingTask TaskName="PushToBlobFeed" AssemblyFile="$(_TaskDir)Microsoft.DotNet.Build.Tasks.Feed.dll"/>
```
weshaggard: Why not just do the UsingTask instead Common?

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Common.props
```diff
@@ -0,0 +1,7 @@
+<Project ToolsVersion="12.0" DefaultTargets="PublishOutputLeg" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
weshaggard: I wouldn't call this Common, but instead make it match the name of the package.

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Common.props
karajas: Removed.

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Common.props
karajas: This seems to work based on testing

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Publish.targets
karajas: Got rid of Common.props

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Common.props
karajas: Updated.

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Microsoft.DotNet.Build.Tasks.Feed.targets
```diff
@@ -33,4 +38,18 @@
...
     </Target>
-</Project>
+
+  <!-- This is the target used to configure feeds for restore. It can be used to specify the sources from which you would restore your packages from.
+       Parameters to specify:
+       EnableFeeds - ItemGroup which contains sources which are allowed
+      ConfigureInputDependsOn can be customized at build time on the commandline to determine the pre steps required
+  -->
+
+  <Target Name="ConfigureInputFeeds" DependsOnTargets="@(ConfigureInputDependsOn)" >
```
weshaggard: Do we still have a needed for this target?

karajas:
PTAL @weshaggard @chcosta

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Microsoft.DotNet.Build.Tasks.Feed.targets
karajas: Unclear if we will at this point, all repos that we did update have a checked in Nuget.config.

weshaggard: OK we can keep it for now and wait to see if folks end up needing it.

src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Microsoft.DotNet.Build.Tasks.Feed.targets
```diff
@@ -33,4 +38,18 @@
                       PublishFlatContainer="$(PublishFlatContainer)"
                       Overwrite="$(Overwrite)" />
     </Target>
-</Project>
```
weshaggard: I expect that our core repos will likely just import this targets file and call the PushToBlobFeed directly.

PR #1686, opened 2017-09-20
Allow maintainers to modify auto PRs by default.

eerhardt:
This change checks this box by default when auto-PRs are created.

image

That way people don't need access to `dotnet-maestro-bot`'s fork to update the auto-PR.

PR #1752, opened 2017-10-20
[release/2.0.0] Allow maintainers to modify auto PRs by default.

dagood:
Port https://github.com/dotnet/buildtools/pull/1686 to `release/2.0.0`. Porting this will help with merge conflicts vs. `master` when forward-porting VersionTools features. (It's also a very useful feature.)

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
@@ -103,6 +103,10 @@
   -->
   <Target Name="UpdateDependenciesAndSubmitPullRequest"
           DependsOnTargets="CreateDefaultDependencyInfos">
+    <PropertyGroup>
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
```diff
@@ -13,6 +13,10 @@
     <BuildInfoCacheDir>$(ToolsDir)BuildInfoCache/</BuildInfoCacheDir>
   </PropertyGroup>
 
+  <PropertyGroup>
```
eerhardt: Why the difference from #1686 here?  The reasoning for putting the property inside the target is that it gives everyone a chance to set the property before the target runs.  If you set the property in a property group, someone that is imported after this targets file would have to overwrite the default and couldn't do the `== ''` check.

dagood: Ah, good catch, this was pretty messy in the conflict resolution and I somehow didn't think to check the original state when deciding where to put it. No reason, putting it back.

Issue #1727, opened 2017-10-06
Allow dependency update pull requests to create a new commit each time


mconnew:
Sometimes updating dependencies will cause the WCF tests to fail due to an issue in a dependency. The automated tool to create a PR when dependencies build a new version replaces the commit if there's an existing pending update PR. This can make it hard to diagnose as the PR gets updated frequently and we lose the commit as it gets replaced. As the GitHub UI now allows to squash and rebase a PR, having a new commit for each update wouldn't cause lots of unnecessary commits in the log.  
It would also be useful to add a comment listing the CI build checks which had failed along with the link provided against the check to the failure details so that we don't lose that information when a new PR is submitted.

dagood:
This is the same as https://github.com/dotnet/buildtools/issues/1658 (see there for the reasons it does fresh commits + force pushes now). /cc @danmosemsft @StephenBonikowsky 

A way to keep history I thought up more recently than that other issue is: after making the update changes on the target branch, do a `git reset --soft` onto the common PR branch and *then* commit. This makes sure the update is fresh (previous update commit attempts on the branch can't ruin subsequent update attempts) but also keeps commit history. I don't think this would be difficult to add.

dagood:
I realized a problem with keeping the old commits using `git reset --soft`: when the target branch (say, `master`) gets new changes, the next update to run includes those changes in its commit. Just like trying to merge changes from `master` into a PR branch, the changes show up in GitHub and make the PR basically unreviewable.

To make this slightly better, I could separate each PR update into two commits: first, bring the PR branch up to date with `master`. Then, apply the new update changes. The PR's "files changes" still shows all the updates in `master` since the PR was created, but you can filter down to only see changes from the latest commit for reviewing.

Either way, you'd have to trust that the updater did the right thing when bringing the changes from `master`... if some unwanted change got in, the only way to reasonably detect it would be to examine the commit *after* using GitHub's "squash merge" feature, and at that point, if it's really bad, it's too late.

## Alternative 1:

> [dagood] This makes sure the update is fresh (previous update commit attempts on the branch can't ruin subsequent update attempts)

If we drop that goal, and instead accumulate updates over time on top of each other, changes from `master` would never appear in the PR. If something goes bad (merge conflict, update changes something unexpected) the PR would need to be closed--then the next update would be fresh from `master`.

This is the kind of implementation I was thinking about on the other issue. (I didn't fully understand the freshness issue at that point, though.)

## Alternative 2:

[...]

PR #1754, opened 2017-10-23
Split PR creation into new target, track discarded auto-PR commits

dagood:
* Have `CreatePullRequest` as its own task to clean up the target names. See https://github.com/dotnet/buildtools/pull/1741#issuecomment-337946286.
 * First iteration on https://github.com/dotnet/buildtools/issues/1727 "Allow dependency update pull requests to create a new commit each time".
   * Example run on source-build where I ran it multiple times to try saving off old commits: https://github.com/dagood/source-build/pull/4. I made up some CI statuses to try saving them, too.

The second feature depends on the first, so to save time I'm submitting as one PR. These changes cleanly apply on release/2.0.0 as well--will submit a PR there too after this review.

/cc @mconnew @danmosemsft @StephenBonikowsky @MichaelSimons


src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateToRemoteDependencies.cs
```diff
@@ -59,7 +86,7 @@ protected GitHubClient CreateClient(bool allowAnonymous)
             }
             else
             {
-                gitHubAuth = new GitHubAuth(GitHubAuthToken, GitHubUser, GitHubEmail);
+                gitHubAuth = new GitHubAuth(GitHubAuthToken, GitHubUser, null);
```
eerhardt: Why don't you want/need to pass email here?

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateToRemoteDependencies.cs
```diff
@@ -25,17 +23,46 @@ public class UpdateToRemoteDependencies : BaseDependenciesTask
...
-                    $"Suggested commit message: '{updateResults.GetSuggestedCommitMessage()}'");
+                    $"No value for '{nameof(GitHubAuthToken)}'. " +
+                    "Accessing GitHub API anonymously.");
+            }
+            else
+            {
+                auth = new GitHubAuth(GitHubAuthToken, GitHubUser, null);
+            }
+
+            using (var client = new GitHubClient(auth))
```
eerhardt: Should this use "CreateClient" below?

src/Microsoft.DotNet.Build.Tasks/VersionTools/SubmitPullRequest.cs
```diff
@@ -0,0 +1,94 @@
...
+                string body = Body ?? string.Empty;
+
+                if (NotifyGitHubUsers != null)
+                {
+                    body += PullRequestCreator.NotificationString(NotifyGitHubUsers.Select(item => item.ItemSpec));
+                }
+
+                var prCreator = new PullRequestCreator(client.Auth, origin, upstreamBranch, GitHubAuthor)
+                {
+                    ForceCreate = AlwaysCreateNewPullRequest,
```
eerhardt: Why are these moving to properties on the class?  It seems like all 3 of these should be parameters to `CreateOrUpdateAsync`.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -125,38 +146,68 @@ public class PullRequestCreator
...
+                insertionMarker,
+                StringComparison.Ordinal);
+
+            if (elementBegin != -1)
+            {
+                return pullRequestToUpdate.Body.Insert(
+                    elementBegin + insertionMarker.Length,
+                    oldCommitLine);
+            }
+            return pullRequestToUpdate.Body +
```
eerhardt: I'm a bit confused here.  I would have thought the new commit would have just applied on top of the old commit.  But it appears here that the old commits just get appended to the new commit's message.

What's the user scenario for "TrackDiscardedCommits"?  Do people want to be able to see the CI run details for each separate commit?  Or do they just want a list of them?

dagood:
Moving some submodule tasks/targets here, too, as followup from https://github.com/dotnet/source-build/pull/254.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateToRemoteDependencies.cs
dagood: Email isn't used for anything `UpdateToRemoteDependencies` does, so I'm making that clear by removing that task property.

I just noticed that this argument defaults to `null`, so I'm removing the explicit `null` here now.

src/Microsoft.DotNet.Build.Tasks/VersionTools/UpdateToRemoteDependencies.cs
dagood: Good catch, `CreateClient` should actually be removed. I didn't see enough overlap between `UpdateToRemoteDependencies` and `CreatePullRequest` to justify keeping it--the split simplified it quite a bit.

src/Microsoft.DotNet.Build.Tasks/VersionTools/SubmitPullRequest.cs
dagood: I'm not familiar with any guideline people generally use for this--I figured that these bools describe the behavior of the `PullRequestCreator`, and the parameters describe pull request data. Adding many arguments to a function makes me uneasy, but I can't explain that. I don't see a reason for `UpstreamBranch` and `Origin` to be properties if nothing else is. Do you have a methodology for figuring out where to put things?

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
dagood: I described the problem with putting the new commit on top here: https://github.com/dotnet/buildtools/issues/1727#issuecomment-337669300.

This is the PR description, not the commit message, if that clarifies anything.

The end goal is to see CI run details and source code changes, and the links go to the page where you can get to both. This isn't the best workflow (a few extra clicks), so a future improvement would be to include the CI statuses in the list entry for each discarded commit.

src/Microsoft.DotNet.Build.Tasks/VersionTools/SubmitPullRequest.cs
eerhardt: On a class that only has a single method, it is pretty hard to differentiate between where to use properties and where to use parameters to the one method it has.

My suggestion would be to think about if you wanted to create multiple PRs using a single instance of the class.  If it makes sense that every PR created by this class shares these values, then I would say move it to a property.  But if you can imagine a scenario where on PR would want something one way, and another PR would want the option a different way, I'd put the option as a parameter to the method.

Another way of thinking about it - does this parameter only apply to the functionality of the method?

If you are concerned about the number of parameters growing, we can create a "PullRequestOptions" class that encapsulates all the possible options.

dagood: Thanks, that's very helpful.

For creating multiple PRs, it's really hard to imagine a reason to do that... going with something crazy like submitting for multiple repos, branches, etc. at once, I would expect `_auth` and `GitAuthorName` to be the same: generally I'd say the same person/bot is responsible for all those PRs. Probably nothing else. Good call on the options class.

dagood:
PTAL, I've added review feedback and added CI status to the discarded commit tracking.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
```diff
@@ -44,6 +46,51 @@
...
+  <Target Name="CreateDefaultSubmoduleUpdateSteps"
+          Condition="'$(GitModulesPath)' != ''">
+    <ReadGitConfigFile File="$(GitModulesPath)">
+      <Output TaskParameter="SubmoduleConfiguration" ItemName="SubmoduleConfiguration" />
+    </ReadGitConfigFile>
+
+    <ItemGroup>
+      <_UpgradableSubmodule Include="@(SubmoduleConfiguration)"
+                            Condition="'%(SubmoduleConfiguration.Branch)' != '' AND
+                                       '%(SubmoduleConfiguration.SourceBuildAutoUpdate)' == 'true'" />
```
eerhardt: I think I'm of the opinion that if you have a "Branch" configured in your .gitmodules, and you are using `UpdateDependencies`, then the submodule should be updated by default.  I would think that generally all the submodules should be updatable normally.  And it would only be a special case where we needed to turn off auto-updates on a submodule.

This way we don't have to put `SourceBuildAutoUpdate=true` on every single submodule.

eerhardt: Is `SourceBuildAutoUpdate` the best name for this property, considering that we are no longer in the `dotnet/source-build` repo, but in BuildTools?

dagood: From https://github.com/dotnet/source-build/pull/254#discussion_r146281494, it didn't seem like it was clear enough that adding `branch` was the thing that enables this feature. I'm fine putting it back the way I had it, though. Also, Git config files support comments, which would help in source-build's case.

dagood: Good catch, renaming to `VersionToolsAutoUpdate`.

eerhardt: My reasoning is - "what do we expect the majority of usages to be?"  If we expect the majority of usages to not auto-update, then opt-in sounds good. But if we expect the majority of usages to want auto-update, then to me it feels like it should be on by default.

When we have all the submodules patches removed and auto-dependency flow, I would expect every submodule to have auto-update turned on.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
```diff
@@ -64,7 +64,7 @@
     <ItemGroup>
       <_UpgradableSubmodule Include="@(SubmoduleConfiguration)"
                             Condition="'%(SubmoduleConfiguration.Branch)' != '' AND
-                                       '%(SubmoduleConfiguration.SourceBuildAutoUpdate)' == 'true'" />
+                                       '%(SubmoduleConfiguration.VersionToolsAutoUpdate)' != 'false'" />
```
eerhardt: I think you need to change the places in ReadGitConfigFile as well.

dagood: Indeed, fixed up and tested in source-build.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
dagood: Agreed, that's why it was that way in the first place. I think I overcorrected towards self-documentation.

dagood:
Rebased to clean up history, no diff.

PR #1755, opened 2017-10-23
Update all package outputs to 2.1.0-prerelease

dagood:
This allows us to avoid conflicts between BuildTools branches in the most straightforward way. See https://github.com/dotnet/buildtools/issues/1489.

I noticed this was still out of date when reviewing https://github.com/dotnet/corefx/pull/24784.

### DIFF
--- a/src/nuget/Microsoft.DotNet.Build.CloudTest.nuspec
+++ b/src/nuget/Microsoft.DotNet.Build.CloudTest.nuspec
@@ -2,7 +2,7 @@
 <package>
   <metadata minClientVersion="2.8.1">
     <id>Microsoft.DotNet.Build.CloudTest</id>
-    <version>1.0.0-prerelease</version>
+    <version>$PackageVersion$</version>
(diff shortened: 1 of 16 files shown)


weshaggard:
Can we figure out a way to pass the common version in a common location instead of updating all the nuspec's manually each time we branch?

karajas:
Should this be updated? @dagood @weshaggard

dagood:
Updated to/for what? I'm not familiar with anything that's made this PR out of date.

This is a low priority change that *may* affect any downstream users making bad assumptions about the versions, so I don't think it's something to worry about until we have buildtools up and running normally again.

dagood:
@weshaggard @karajas, does this look ok to merge now?

FYI @jcagme this puts the Feed package version in line with the rest.

dagood:
> We may want to consider porting to other branches if problems arise.

Agreed. It's possible we have some version overlaps between the 1.x branches and 2.0 that we just don't know about yet. BuildTools packaging hasn't changed much over time so porting should be straightforward.

PR #1758, opened 2017-10-25
Update PDB Conversion Task to fix random crash

vancem:
There was a COM bug in the PDB writing code used in the task that
converts portable PDBs to windows PDBs (see https://github.com/dotnet/symreader/pull/70)
This picks up a version with this fix in the build tools.

@tmat  @dagood 

I have confirmed that with this update the fix (which is in the SymUnmanagedWriterImpl.CloseSymWriter method), shows up in the Microsoft.DiaSymReader.dll placed in the bin directory.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -86,6 +86,9 @@
     <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
     <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
+    <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/symreader/api/v3/index.json" />
+    <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/symreader-native/api/v3/index.json" />
+    <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/symreader-converter/api/v3/index.json" />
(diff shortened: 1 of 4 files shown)


vancem:
Note that after we are comfortable with this, we can pull out the msbuild hacks we have that run the PDB conversion in a sub-process.   It should not be needed anymore.

dir.props
```diff
@@ -86,6 +86,9 @@
     <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
     <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
+    <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/symreader/api/v3/index.json" />
+    <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/symreader-native/api/v3/index.json" />
+    <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/symreader-converter/api/v3/index.json" />
```
vancem: We were using really old version of symreader that you could get from nuget.org.   To get the newer myget versions you need to add these feeds.

vancem:
Checking in so I can deploy for CoreFX.

dir.props
dagood: I actually mirrored the packages on https://dotnet.myget.org/gallery/dotnet-buildtools to make sure old versions don't get dropped off the symreader feeds and break the BuildTools build. I don't think any versions of these packages are available on NuGet.org.

I've gone ahead and mirrored the new versions, too.

PR #1759, opened 2017-10-26
Allow UAP remote runner service to run methods that return void

danmosemsft:
Matching change for https://github.com/dotnet/corefx/pull/24890


safern:
In order for this to work we would need to check if `CheckExitCode` in RemoteExecutor for Uap is true to check for the exit code. https://github.com/dotnet/corefx/blob/master/src/CoreFx.Private.TestUtilities/src/System/Diagnostics/RemoteExecutorTestBase.uap.cs#L69

Because right now we are always checking it and if we return 0, wouldn't it fail?

danmosemsft:
@safern thanks, fixed that in dotnet/corefx#24890

danmosemsft:
@safern Verified private buildtools for UAP

PR #1761, opened 2017-10-26
Fix the tool-runtime restore warnings

weshaggard:
Moved BuildTools back to a version of CCI which depended on all stable
package versions. This version happens to align with .NET Core 1.0 but
we don't currently have a newer version of that package with stable
dependencies.

Disable NU1701 that we are getting for Microsoft.Net.Compilers.Targets.NetCore
as we know it is safe to resolve its assets as net461. We also plan to remove
this package when we start compiling against the CLI itself instead of the
targets/msbuild we carry with BuildTools.

Delete tool-runtime\project.json because it is no longer used after
https://github.com/dotnet/buildtools/pull/1743

cc @eerhardt @dagood @joperezr


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
```diff
@@ -19,13 +24,13 @@
...
     <PackageReference Include="Microsoft.Net.Compilers.Targets.NetCore" Version="0.1.5-dev" />
-    <PackageReference Include="Microsoft.Cci" Version="4.0.0-rc4-24217-00" />
+    <PackageReference Include="Microsoft.Cci" Version="4.0.0-rc3-24214-00" />
     <PackageReference Include="System.Composition" Version="1.1.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
-    <PackageReference Include="Microsoft.Cci" Version="4.0.0-rc4-24217-00" />
+    <PackageReference Include="Microsoft.Cci" Version="4.0.0-rc3-24214-00" />
```
joperezr: just curious, why the downgrade?

dagood: Here's where we decided to make this change to fix some warnings: https://github.com/dotnet/core-setup/pull/3350#discussion_r147025196. The downgrade's explained in this PR's description:

> Moved BuildTools back to a version of CCI which depended on all stable package versions. This version happens to align with .NET Core 1.0 but we don't currently have a newer version of that package with stable dependencies.

weshaggard: I also looked at where we upgraded the version and it was just a side-effect of switching to master branch and ProjectK push newer version from master so we go this version, but we never ended up with another one that I could find with stable dependencies. Given this I just moved it back to the last  one with stable dependencies.

joperezr: lol sorry, it would have helped if I actually looked at the description of the change, but I went into the changes directly without reading the description.

PR #1764, opened 2017-10-26
Fix failure on Unix associate with converting PDBs

vancem:
Code to convert portable pdbs to windows pdbs was being run on UNIX machines.
This is clearly not needed and was causing failures.   Fix it to only run on non Windows
and only for coverage (which is the only thing we know that needs the Windows  PDBs).

Basically just make the conversion condition stricter.  

See also  https://github.com/dotnet/corefx/issues/24807.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -50,10 +50,13 @@
     </PropertyGroup>
   </Target>
 
-  <Target Name="GenerateWindowsPdbsForAssemblyBeingTested" BeforeTargets="GenerateTestExecutionScripts">
+  <!-- *********************************************************************************************** -->
     <!-- As of 10/2017 OpenCover does not support portable PDBs, but we want the builds to generate
          portable PDBs.  Thus we generate windows PDBs from portable PDBs here.   Can be removed
          when OpenCover directly supports Portable PDBs (probably by early 2018) -->
+
+  <Target Name="GenerateWindowsPdbsForAssemblyBeingTested" BeforeTargets="GenerateTestExecutionScripts"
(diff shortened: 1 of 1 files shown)


vancem:
This is to fix bug https://github.com/dotnet/corefx/issues/24807

@dagood

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -50,11 +50,14 @@
...
   </Target>
 
-  <Target Name="GenerateWindowsPdbsForAssemblyBeingTested" BeforeTargets="GenerateTestExecutionScripts">
+  <!-- *********************************************************************************************** -->
     <!-- As of 10/2017 OpenCover does not support portable PDBs, but we want the builds to generate
          portable PDBs.  Thus we generate windows PDBs from portable PDBs here.   Can be removed
          when OpenCover directly supports Portable PDBs (probably by early 2018) -->
 
+  <Target Name="GenerateWindowsPdbsForAssemblyBeingTested" BeforeTargets="GenerateTestExecutionScripts"
+    Condition="$(CodeCoverageEnabled) == 'true' and '$(TargetOS)'=='Windows_NT'">
```
dagood: Not that we have a style guide, but I've usually seen chopping done like this in MSBuild:

```xml
<Target Name="GenerateWindowsPdbsForAssemblyBeingTested"
        BeforeTargets="GenerateTestExecutionScripts"
        Condition="'$(CodeCoverageEnabled)' == 'true' and '$(TargetOS)'=='Windows_NT'">
```

dagood: `$(CodeCoverageEnabled)` is missing quotes in the condition.

vancem:
@dotnet-bot test this please

PR #1765, opened 2017-10-27
Noop to force recreation of Nuget package

vancem:


dagood:
This is fine, but not needed: you can just requeue the https://devdiv.visualstudio.com/DevDiv/_build/index?definitionId=587 definition with default parameters. There is no dependency on the commit count/hash.

vancem:
Thanks.  Good to know.

PR #1766, opened 2017-10-27
Move OpenCommPort to api set module

pjanotti:
OpenCommPort should use api set instead of kernel base.


src/Microsoft.DotNet.CodeAnalysis/PackageFiles/PinvokeAnalyzer_Win32UWPApis.txt
```diff
@@ -149,6 +149,7 @@ api-ms-win-appmodel-runtime-l1-1-2.dll!VerifyPackageFamilyName
 api-ms-win-appmodel-runtime-l1-1-2.dll!VerifyPackageFullName
 api-ms-win-appmodel-runtime-l1-1-2.dll!VerifyPackageId
 api-ms-win-appmodel-runtime-l1-1-2.dll!VerifyPackageRelativeApplicationId
+api-ms-win-core-comm-l1-1-1.dll!OpenCommPort
```
weshaggard: Why use the api-set in this case?

pjanotti: Windows team request: they want to avoid us using kernelbase directly.

pjanotti: Let me check the other kernelbase entries in this file...

pjanotti: It seems that we can remove all others from kernelbase, I will do some tests before adding it to the change.

pjanotti: Yes, we can.

weshaggard: Should we be moving them to kernal32 instead of api-sets then? In general we are not using api-sets any more.

pjanotti: `OpenCommPort` does not exist on kernel32. We should avoid API sets in general but in this case there is no alternative since it is a new API that doesn't exist on downlevel.

pjanotti: The other entries for kernelbase are not being used at all.

weshaggard: Ok. We should decide whether or not we want to depend on API that isn't downlevel. That means that library will not work on Win7 which we still want to support.

pjanotti: This API is for UWP only, on Win7 we use kernel32!CreateFile.

Issue #1677, opened 2017-09-14
Throw actionable error when VersionTools can't find a build-info


dagood:
There's some helpful info VersionTools has at that point that should be shown: dotnet/versions url requested and which BuildInfo is being checked. The error can also describe (or point to some help describing) some diagnostic steps.

Right now the user gets a stack trace and a 404. It's quick to reproduce given a commit but it's hard to tell who to ask about it without already knowing about the dependency tooling.

PR #1768, opened 2017-10-27
Improve build-info Get failure error message

dagood:
Fixes https://github.com/dotnet/buildtools/issues/1677


src/Microsoft.DotNet.VersionTools/BuildInfo.cs
```diff
@@ -180,6 +189,23 @@ public static string RawBuildInfoBaseUrl(string rawRepoUrl, string gitRef, strin
...
+            HttpClient client,
+            string buildInfoName,
+            string url)
+        {
+            HttpResponseMessage response = await client.GetAsync(url);
+            if (!response.IsSuccessStatusCode)
+            {
+                Trace.TraceError(
+                    $"Failure response code while fetching BuildInfo: name '{buildInfoName}', " +
+                    $"url: '{url}'. Ensure the repository is correct and the file exists at " +
```
MichaelSimons: nit - why a colon after url but not name?

dagood: Whoops, just bad proofreading. Fixing.

PR #1769, opened 2017-10-27
Revert "Add PackageRestoreLogger to log restore urls"

karajas:
This reverts commit 98e4d4ebf8e6e6fa68389c71f529d29307372896.

Will move the project into a separate tool.

@chcosta @weshaggard


chcosta:
> do you know if this is still needed?

Do you mean this PR or the functionality?

At the moment, nobody is using this functionality.  Long term, there may be value because there is a desire to remove nuget from the equation.  It's not clear what the timeline is for that (likely not for quite a while).

weshaggard:
Ok I'd be inclined to get this PR fixed and merged then to remove dead code. We can always include it or something similar again if and when we need it.

Issue #1739, opened 2017-10-13
GenFacades generates partial facades that don't have Debugging Entry in them


vancem:
Currently in the CoreFX build if you generate either the dlls
* Microsoft.Win32.Registry.AccessControl 
* System.Security.Permissions

If you inspect the resulting DLL (using link /dump /headers DLL)  you will see that it does not have a Debug directory associated with it.  This entry is used to remember the PDB signature and is needed to look up the DLL on a symbol server.   As a result finding the PDBs for these DLLs will not work unless the PDB is right next to the DLL.  

In addition it turns out that our portable PDB to Windows PDB tool fails when DLLs dont have Debug entry, so this makes it problematic to generate Windows PDBs (which in turn is needed for code coverage).   

The reason these DLLs don't have this entry is because they were morphed by the GetFacades task (see OutputFacadeToFile in  src\GenFacades\GenFacades.Core\GenFacades.cs).   This code needs to write a Debug entry when it makes the new DLL/PDB.

vancem:
I am not completely sure, but I think what is happening is that GenFacades uses CCI which does not support portable PDBs.   There is logic in partialfacades.task.targets that simply tells GenFacades to not produce a PDB (which is why the DLL does not have a Debug entry).  Presumably the idea is that the PDB from the 'before' GenFacades will work (at least for thing that were not added by GenFacades).  

The problem is that this does not work if the goal is to have the PDB served by a symbol server.  

The right approach is to make GenFacades generate portable PDBs.   Lets see how hard that is...

vancem:
CCI does not have a Portable PDB writer.   Probably not worth making one.  

@tmat  suggests that we change GenFacades to emit C# that we feed through the C# compiler.   

Question: what do we do on Linux?  

Another option is to use the Symbol conversion code so that GenFacades makes the Windows PDBs and then use the symreader\converter logic (which is already in build-tools) to generate a portable PDB from it.  This should be pretty straightforward (but is a bit ugly).

vancem:
This has been fixed by https://github.com/dotnet/buildtools/pull/1792 (but also see https://github.com/dotnet/buildtools/pull/1799 for followon fix.

PR #1771, opened 2017-10-30
Warn and skip PPDB conversion if a DLL has no Debug directory

dagood:
This unblocks the symbol publish leg from a known GenFacades issue: https://github.com/dotnet/buildtools/issues/1739.


vancem:
Looks OK.  Do you know if this works in the larger context (is the fact that there is no output file going to mess something up later on?   It is OK for Coverage runs, but I don't know about your other use of ConvertPortablePDBToWIndowsPdb.

dagood:
Good question: the later steps for the symbol indexing leg are just symbol indexing/archiving, and they pick up the set of files to work on based on glob patterns, which will work fine.

PR #1772, opened 2017-10-30
Fix renew lease function and enable more logging

karajas:
@chcosta @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
@@ -55,7 +55,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                     string leaseId = AcquireLeaseOnBlobAsync().GetAwaiter().GetResult();
                     _cancellationTokenSource = new CancellationTokenSource();
                     _leaseRenewalTask = Task.Run(() =>
-                    { AutoRenewLeaseOnBlob(this, _accountName, _accountKey, _containerName, _blobName, leaseId, _leaseUrl, _log); },
+                    { AutoRenewLeaseOnBlob(this, _log); },
                       _cancellationTokenSource.Token);
                     _leaseId = leaseId;
                     return _leaseId;
@@ -117,7 +117,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
 
             return leaseId;
         }
-        private static void AutoRenewLeaseOnBlob(AzureBlobLease instance, string accountName, string accountKey, string containerName, string blob, string leaseId, string leaseUrl, Microsoft.Build.Utilities.TaskLoggingHelper log)
+        private static void AutoRenewLeaseOnBlob(AzureBlobLease instance, Microsoft.Build.Utilities.TaskLoggingHelper log)
         {
             TimeSpan maxWait = TimeSpan.FromSeconds(s_MaxWaitDefault);
             TimeSpan delay = TimeSpan.FromMilliseconds(s_DelayDefault);
@@ -130,13 +130,13 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
 
                 try
                 {
-                    log.LogMessage(MessageImportance.Low, $"Requesting lease for container/blob '{containerName}/{blob}'.");
+                    log.LogMessage(MessageImportance.Low, $"Requesting lease for container/blob '{instance._containerName}/{instance._blobName}'.");
                     using (HttpClient client = new HttpClient())
                     {
                         Tuple<string, string> leaseAction = new Tuple<string, string>("x-ms-lease-action", "renew");
-                        Tuple<string, string> headerLeaseId = new Tuple<string, string>("x-ms-lease-id", leaseId);
+                        Tuple<string, string> headerLeaseId = new Tuple<string, string>("x-ms-lease-id", instance._leaseId);
                         List<Tuple<string, string>> additionalHeaders = new List<Tuple<string, string>>() { leaseAction, headerLeaseId };
-                        var request = AzureHelper.RequestMessage("PUT", leaseUrl, accountName, accountKey, additionalHeaders);
+                        var request = AzureHelper.RequestMessage("PUT", instance._leaseUrl, instance._accountName, instance._accountKey, additionalHeaders);
                         using (HttpResponseMessage response = AzureHelper.RequestWithRetry(log, client, request).GetAwaiter().GetResult())
                         {
                             if (!response.IsSuccessStatusCode)
@@ -149,7 +149,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                 }
                 catch (Exception e)
                 {
-                    Console.WriteLine($"Rerying lease renewal on {containerName}, {e.Message}");
+                    Console.WriteLine($"Rerying lease renewal on {instance._containerName}, {e.Message}");
                     waitFor = delay;
                 }
                 token.ThrowIfCancellationRequested();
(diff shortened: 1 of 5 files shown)


karajas:
PTAL

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -190,38 +190,52 @@ public string GeneratePackageServiceIndex(string redirectUrl, IEnumerable<string
...
-                writer.WriteStartObject();
-                writer.WritePropertyName("@vocab");
-                writer.WriteRawValue("\"http://schema.nuget.org/schema#\"");
-                writer.WritePropertyName("comment");
-                writer.WriteRawValue("\"http://www.w3.org/2000/01/rdf-schema#comment\"");
-                writer.WriteEndObject();
-                writer.WriteEndObject();
+            int retry = 5;
+            bool isFileCreated = false;
+            while (!isFileCreated && --retry > 0) {
```
chcosta: delays are generally helpful in retry loops to help avoid transient issues.

src/Microsoft.DotNet.Build.CloudTestTasks/AzureHelper.cs
```diff
@@ -290,7 +290,7 @@ private static bool IsWithinRetryRange(HttpStatusCode statusCode)
                 response.Dispose();
             }
 
-            throw new HttpRequestException(string.Format("Request failed with status {0} response {1}", statusCode, contentStr));
+            throw new HttpRequestException(string.Format("Request failed with status {0} response {1} request url {2}", statusCode, contentStr, createRequest().RequestUri));
```
chcosta: You could probably just move this to an interpolated string for readability - `$"Request failed with status {statusCode} response {contentStr} ..."`

src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
```diff
@@ -134,7 +134,7 @@ private static void AutoRenewLeaseOnBlob(AzureBlobLease instance, string account
                     using (HttpClient client = new HttpClient())
                     {
                         Tuple<string, string> leaseAction = new Tuple<string, string>("x-ms-lease-action", "renew");
-                        Tuple<string, string> headerLeaseId = new Tuple<string, string>("x-ms-lease-id", leaseId);
+                        Tuple<string, string> headerLeaseId = new Tuple<string, string>("x-ms-lease-id", instance._leaseId);
```
chcosta: What's the purpose of all the variables passed to the method, if they are just being acquired from the insance of AzureBlobLease?  ie, are accountName, accountKey, containerName, etc... needed or could they be replaced with instance._accountName, instance._blah?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -190,38 +190,52 @@ public string GeneratePackageServiceIndex(string redirectUrl, IEnumerable<string
...
+                        writer.WritePropertyName("comment");
+                        writer.WriteRawValue("\"http://www.w3.org/2000/01/rdf-schema#comment\"");
+                        writer.WriteEndObject();
+                        writer.WriteEndObject();
+                    }
+                    isFileCreated = true;
+                }
+                catch (Exception)
+                {
+                    Log.LogMessage($"Failed to get file {rootIndexJsonLocation}. Retrying");
```
chcosta: nit, Failed to "write" file?

src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
karajas: I think we wanted to be able to call this function outside of autorenew as well.

chcosta: Makes sense

karajas: Actually there isn't anyone consuming this except us, I went ahead and removed the extra parameters. The renewal should be automatic anyways.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
karajas: Added in the catch

src/Microsoft.DotNet.Build.CloudTestTasks/AzureBlobLease.cs
chcosta: I see you changed it.  Given that it's a private method, I think you're right, it's ok to change it.  Just make sure you do plenty of validation.  It'd be nice to get this in tonight, but I'd rather not rush it if we have any concerns that it could actually make publishing worse.

karajas:
This was the fix @jcagme

karajas:
Merging this in, to grab a fix for CoreCLR official build.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -165,63 +165,95 @@ public string CalculateBlobPath(string item, string relativePath)
...
             {
-                writer.WriteStartObject();
-                writer.WritePropertyName("versions");
-                writer.WriteStartArray();
-                foreach (string version in values)
+                try
+                {
+                    if (Directory.Exists(pathToTempFolder))
+                    {
+                        Directory.Delete(pathToTempFolder, true);
```
weshaggard: Why are you calling delete each time in the loop? Do keep in mind the more often we call Delete the more likely we are going to fail, as anytime a file still has an open handle it will fail this Delete call.

chcosta:
Looks like this may have caused coreclr publishing failures.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
karajas: but we do need to ensure the old index.json is deleted if there is one present, is there a better way to address a file handle in use?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -134,7 +134,7 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, string indexDir
             await clientThrottle.WaitAsync();
             string leaseId = string.Empty;
             //this defines the lease for 15 seconds (max is 60) and 3000 milliseconds between requests
-            AzureBlobLease blobLease = new AzureBlobLease(feed.AccountName, feed.AccountKey, string.Empty, feed.ContainerName, uploadPath, Log, "15", "3000");
+            AzureBlobLease blobLease = new AzureBlobLease(feed.AccountName, feed.AccountKey, string.Empty, feed.ContainerName, uploadPath, Log, "15", "5000");
```
karajas: This change may be hiding more errors, if we continue to see more.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
weshaggard: You are deleting the entire directory. It would be better to try and just delete or rename the file if it exists.

weshaggard: Assuming your directory is shared. 

Another thing we should probably do (assuming we stick with this) is to set the IndexDirectory to make sure it is shared on the machine and is scoped to the repo.

PR #1775, opened 2017-10-31
Auto-update discarded commit list: use combined status; only include Details link if a target exists

### LINKED ISSUE #1773: Auto-update discarded commit tracker should handle CI statuses with no link
Right now, it adds a `[Details]()` link, which reloads the current page. Instead, the details link shouldn't show up at all.

The details link text just needs to be made conditional on whether there is a link:

https://github.com/dotnet/buildtools/blob/868a2dd719771247293471b7dad0a911528e178d/src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs#L248-L250

### LINKED ISSUE #1774: Auto-update discarded commit tracker should only list combined status entries
When the status on a PR is updated, the existing entry isn't modified, rather, an additional entry is added later. The CI saver currently lists all these entries, which makes the final status hard or impossible to determine.

The CI saver currently lists all statuses using https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref, but it should use https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref.

dagood:
Resolves #1773
Resolves #1774

Example of these issues in a live auto-update PR: https://github.com/dotnet/core-setup/pull/3372. Tested fix with https://github.com/dagood/source-build/pull/6, where I applied more than one status with the same context.

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubApiModel.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubApiModel.cs
@@ -50,6 +50,12 @@ namespace Microsoft.DotNet.VersionTools.Automation.GitHubApi
         public string Content { get; set; }
     }
 
+    public class GitHubCombinedStatus
+    {
+        public string State { get; set; }
+        public GitHubStatus[] Statuses { get; set; }
+    }
+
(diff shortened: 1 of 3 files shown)


dagood:
Thanks, merging.

PR #1721, opened 2017-10-04
Changes required in order to onboard 2.0.0 CLI to corefx

joperezr:
cc: @weshaggard @ericstj 

Porting required changes to release/2.0 branch so that we can update the CLI used in 2.0 corefx

Issue #1697, opened 2017-09-26
Remove workaround that copies RID specific assets when initializing buildtools in non-Windows


joperezr:
On PR #1687 we had to make a workaround that manually copied a few assets that were getting published to RID specific locations into the tools directory when running init-tools on non-Windows. This issue is to track the work of removing that workaround, and instead trying publishing the tool runtime to the right RID so this is won't be required.

Workaround is here: https://github.com/dotnet/buildtools/pull/1687/files#diff-519820c77125188495c6a22d71f859ceR112

ViktorHofer:
// Auto-generated message

Going forward, the .NET team is using https://github.com/dotnet/arcade to develop the code and issues formerly in this repository. Feel free to reopen/move this issue if it still applies to servicing branches in this repository, or source code which now resides in the Arcade repository.

PR #1777, opened 2017-10-31
[release/2.0.0] tool-runtime dependency upgrade followup

dagood:
Seems like the CodePages copy was missed in https://github.com/dotnet/buildtools/pull/1721, or considered not needed because MSBuild was older. This started causing errors in this source-build buildtools upgrade: https://github.com/dotnet/source-build/pull/271.

https://github.com/dotnet/buildtools/issues/1697 tracks removing the `Copy System.Text.Encoding.CodePages runtimes` workaround. See https://github.com/dotnet/buildtools/pull/1687/files#diff-519820c77125188495c6a22d71f859ceR112 for the exact line I'm porting there.

The other commits are catching release/2.0.0 up with the compiler upgrades that were done in master. (To make it clean, I resolved conflicts towards master in the first cherry pick which resulted in an apparent downgrade of the tool-runtime dependency, then the next two cherry-picks were conflictless.)


dagood:
I've confirmed these changes get source-build CoreCLR to build on my Fedora 26 machine, PTAL.

dagood:
I'm going to go ahead with this merge because BuildTools release/2.0.0 is broken in source-build already, these are straightforward cherry-picks from master, and I want to unblock potentially moving ahead.

joperezr:
I didn't add these changes into the release/2.0.0 because I wans't sure we wanted the new compiler on that branch, but changes look ok.

PR #1779, opened 2017-11-01
Make public the default SignType

weshaggard:
As part of our repo API we are using 'public' as the signtype to
indicate what we formally referred to as oss signing. We still work
if oss is specified for back-compat but we now default to public
if it isn't set.

cc @dagood @mmitche

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
@@ -8,7 +8,7 @@
 
   <PropertyGroup>
     <BuildToolsTaskDir Condition="'$(BuildToolsTaskDir)' == ''">$(ToolsDir)</BuildToolsTaskDir>
-    
+
     <!-- For UAP, we need to bin place the resources as resw files to create the runner's pri file. -->
     <ResourcesFolderPath Condition="'$(ResourcesFolderPath)' == ''">$(RuntimePath)resw</ResourcesFolderPath>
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
```diff
@@ -295,7 +295,7 @@
     import the MicroBuild boot-strapper project (only relevant for shipping binaries)
     NOTE: we import this at the end as it will override some dummy targets (e.g. SignFiles)
   -->
-  <Import Project="$(MSBuildThisFileDirectory)MicroBuild.Core.targets" Condition="'$(IsTestProject)'!='true' and '$(SignType)' != 'oss'" />
+  <Import Project="$(MSBuildThisFileDirectory)MicroBuild.Core.targets" Condition="Exists('$(MSBuildThisFileDirectory)MicroBuild.Core.targets') and $(IsTestProject)' != 'true'" />
```
dagood: Missing opening `'` in `$(IsTestProject)' != 'true'`

src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets
```diff
@@ -45,13 +45,16 @@
...
+    <!-- applicable values for SignType are public, test or real -->
+    <SignType Condition="'$(SignType)' == ''">public</SignType>
 
     <ShouldWriteSigningRequired Condition="'$(IsTestProject)' == 'true'">false</ShouldWriteSigningRequired>
     <ShouldWriteSigningRequired Condition="'$(SkipSigning)' == 'true'">false</ShouldWriteSigningRequired>
-    <ShouldWriteSigningRequired Condition="'$(SignType)' == 'oss'">false</ShouldWriteSigningRequired>
+    <ShouldWriteSigningRequired Condition="$(SignType)' == 'public' or '$(SignType)' == 'oss'">false</ShouldWriteSigningRequired>
     <ShouldWriteSigningRequired Condition="'$(ShouldWriteSigningRequired)'==''">true</ShouldWriteSigningRequired>
+
+    <ShouldOpenSourceSign Condition="$(SignType)' == 'public' or '$(SignType)' == 'oss'">$(DelaySign)</ShouldOpenSourceSign>
```
dagood: Missing opening `'`s on the `SignType`s here.

PR #1780, opened 2017-11-02
Fix missing quote in sign.targets

karajas:
@stephentoub @weshaggard


weshaggard:
Fixed in https://github.com/dotnet/buildtools/pull/1781.

PR #1781, opened 2017-11-02
Fix typo

danmosemsft:
Direct commit to fix other build break in corefx.


danmosemsft:
Clearly the buildtools build does not validate this file content anyway...

PR #1782, opened 2017-11-02
Skipping apicompat error on System.ComponentModel.Composition

maryamariyan:
cc: @danmosemsft @weshaggard

### DIFF
--- a/src/Microsoft.Cci.Extensions/HostEnvironment.cs
+++ b/src/Microsoft.Cci.Extensions/HostEnvironment.cs
@@ -367,8 +367,11 @@ namespace Microsoft.Cci.Extensions
             }
 
             var unresolved = new UnresolvedReference<IUnit, AssemblyIdentity>(referringUnit, referencedAssembly);
-
-            OnUnableToResolve(unresolved);
+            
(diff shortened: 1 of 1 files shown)


danmosemsft:
It may be better to do this patch in apicompat itself. But I defer to @weshaggard

src/Microsoft.Cci.Extensions/HostEnvironment.cs
```diff
@@ -367,8 +367,11 @@ public override AssemblyIdentity ProbeAssemblyReference(IUnit referringUnit, Ass
             }
 
             var unresolved = new UnresolvedReference<IUnit, AssemblyIdentity>(referringUnit, referencedAssembly);
-
-            OnUnableToResolve(unresolved);
+            
+            if (!referencedAssembly.Name.Value.Equals("System.Private.CoreLib", StringComparison.OrdinalIgnoreCase))
```
weshaggard: This is not a good fix for this issue it will hide other issues.

weshaggard:
For now I suggest you just disable apicompat for your project and file and issue.

PR #1785, opened 2017-11-02
Fix race associted with SourceLink

vancem:
The sourcelink file is generated one for the whole repo so it is possible that
differnet parts of the build will race to write it.   We have seen this show up as a file
access error in builds.

It is already the case that there is logic that avoids running the code most of the time
(see VersionPropsImported).  but races are still possible.  They are avoided in that
scheme by testing for the file existing and simply warning if there is a error trying to
write the file.   We follow suit here.

As a unrelated change there was already a variable SourceLinkFilePath, for the source
link file so I use it instead of building it out of more primitive variable.

@dagood

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -294,8 +294,10 @@
 
     <!-- Generate the SourceLink file that will be passed to the C# compiler (using ProjectDir and GitHubRepositoryUrl as data) -->
     <MakeDir Directories="$(BaseIntermediateOutputPath)" />
-    <WriteLinesToFile Condition="'$(UseSourceLink)' == 'true' AND '$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A'"
-      File="$(BaseIntermediateOutputPath)source_link.json" 
+    <WriteLinesToFile 
+      Condition="'$(UseSourceLink)' == 'true' AND '$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A' AND !Exists('$(SourceLinkFilePath)')"
+      ContinueOnError="WarnAndContinue"
+      File="$(SourceLinkFilePath)" 
       Overwrite="true" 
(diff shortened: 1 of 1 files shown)


vancem:
@weshaggard  FYI

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -294,8 +294,10 @@
 
     <!-- Generate the SourceLink file that will be passed to the C# compiler (using ProjectDir and GitHubRepositoryUrl as data) -->
     <MakeDir Directories="$(BaseIntermediateOutputPath)" />
-    <WriteLinesToFile Condition="'$(UseSourceLink)' == 'true' AND '$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A'"
-      File="$(BaseIntermediateOutputPath)source_link.json" 
+    <WriteLinesToFile 
+      Condition="'$(UseSourceLink)' == 'true' AND '$(GitHubRepositoryUrl)' != '' AND '$(LatestCommit)' != 'N/A' AND !Exists('$(SourceLinkFilePath)')"
```
dagood: About `!Exists('$(SourceLinkFilePath)'`: is it ok that this won't update an existing file anymore?

The target name is `CreateOrUpdateCurrentVersionFile`, so intuitively it should update. Not sure if it's a problem in practice, though: it seems like it would only cause a problem if an unclean repo checks out a new commit and builds, but our official builds are cleaned pretty rigorously.

vancem: The CreateOrUpdateCurrentVersionFile has a condition '$(VersionPropsImported)' != 'true'", which is set by importing a $(BuildVersionFile) which is written by this task.  This sets up a 'write once' dynamic (if you exclude races).  The $(BuildVersionFile) also has a !Exists condition on it so it will only be written once) and includes the commit ID that is used (so that too is frozen after first write).    I was just trying to be consistent with that.  

However, I could just as happy removing it (it should not matter).   Let me know if you want it changed.

dagood: Ah, that makes sense, thanks for explaining.

PR #1786, opened 2017-11-03
Enabling the Feed task to copy nuget dependencies for net15

jcagme:
We found that for net15 based builds there was an error about Sleet not being found since the dll was not being copied along with the Feed dll. This should fix this.


karajas:
Tested on my rhel box, looks like it has the right closure.

PR #1787, opened 2017-11-03
Add init-tools PackageVersionProps download

dagood:
This implementation is in `init-tools.ps1` and `init-tools.sh`, which runs after the BuildTools package is restored. `RepoAPI.Mapping.props` then picks up on the downloaded file inside the MSBuild call.

This downloads the file pointed to by the environment variable `PACKAGEVERSIONPROPSURL` and uses the props file during the build.

CoreFX only needs to add a VSTS variable `PACKAGEVERSIONPROPSURL` with the value of `PB_PackageVersionPropsUrl` to put the URL value in the environment var. (This only works if it's not a secret. For now, this flow does not support a package version props URL that requires auth.)

I rely on the environment to pass the arg deep into the init-tools flow to avoid adding more args that need to pass all the way through. Using the env is bad for discoverability, but makes my changes contained and less likely to break other scenarios.

The init-tools flow seems like a reasonable place for this for the Core repos: it is already behind a semaphore, so there's no need to worry about downloading the file more than once. It's also a common entry point that happens before the MSBuild call, which is necessary so that the `Import` can find the `DotNetPackageVersionPropsPath` file.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
@@ -21,22 +21,11 @@ foreach ($file in Get-ChildItem $ToolRuntimePath *.runtimeconfig.json)
 New-Item -Force -Type Directory (Join-Path $ToolRuntimePath (Split-Path -Leaf (Split-Path $BuildToolsPackageDir)))
 
 # Download the package version props file, if  was passed in the environment.
-$packageVersionPropsUrl = $env:PB_PACKAGEVERSIONPROPSURL
-$packageVersionPropsPath = $env:PackageVersionPropsDownloadPath
+$packageVersionPropsUrl = $env:PACKAGEVERSIONPROPSURL
+$packageVersionPropsPath = Join-Path $ToolRuntimePath "DownloadedPackageVersions.props"
 
 if ($packageVersionPropsUrl)
 {
-    if (-not $packageVersionPropsPath)
-    {
-        throw "Url '$packageVersionPropsUrl' (PB_PACKAGEVERSIONPROPSURL) specified, but no download path (PackageVersionPropsDownloadPath)."
-    }
-
-    $dir = Split-Path -Parent $packageVersionPropsPath
-    if (-not (Test-Path $dir))
-    {
-        mkdir $dir
-    }
-
     Write-Host "Downloading package version props from '$packageVersionPropsUrl' to '$packageVersionPropsPath'..."
 
     # Copied from init-tools.cmd in CoreFX
@@ -71,5 +60,6 @@ if ($packageVersionPropsUrl)
         }
     } while ($success -eq $false);
 
-    Write-Host "Downloaded package version props."
+    Write-Host "Downloaded package version props:"
+    Get-Content $packageVersionPropsPath
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
```diff
@@ -19,3 +19,57 @@ foreach ($file in Get-ChildItem $ToolRuntimePath *.runtimeconfig.json)
 # the init-tools.cmd (that is checked into each repository that uses buildtools) can write the semaphore
 # marker into this file once tool initialization is complete.
 New-Item -Force -Type Directory (Join-Path $ToolRuntimePath (Split-Path -Leaf (Split-Path $BuildToolsPackageDir)))
+
+# Download the package version props file, if  was passed in the environment.
+$packageVersionPropsUrl = $env:PB_PACKAGEVERSIONPROPSURL
```
weshaggard: Lets not use PB_ variables directly in our build scripts. Just remove the PB_ and it is up to the build definitions to set the correct value in the environment. 

PB_ means it comes from the build definition and should be passed or set to something else we don't want to start depending on them directly outside of our build definitions.

dagood: Agreed, didn't consider how simple the mapping can be. Just this in `variables`, for a simple impl:

```
"PACKAGEVERSIONPROPSURL": {
  "value": "$(PB_PackageVersionPropsUrl)"
},
```

(I figure all caps is easier to understand later because you don't need to know/realize that VSTS uppercases everything.)

weshaggard:
>I set PackageVersionPropsDownloadPath in run.cmd/sh rather than using a default here in init-tools. I figure this is more resilient, because a default in init-tools would need to be matched in run.cmd/sh to detect if something was downloaded, and it would break if we change the default. (This is assuming there's no simple way to pass a value out of init-tools.cmd/sh, which I haven't seen a way to do.)

Instead of needing to set this value in run scripts lets simply put put a well known file into the Tools directory and conditionally import it in dependencies.props. Something like:

```
<DotNetPackageVersionPropsPath Condition="'$(DotNetPackageVersionPropsPath)' == '' and Exists('$(ToolsDir)PackageVersions.props')">$(ToolsDir)PackageVersions.props</DotNetPackageVersionPropsPath >
```

Then we always just download it an put it in that location. That reduces the number of env variables and scopes changes to just here and dependencies.props.

dagood:
Good idea, I think we can go even further and add that detection to [`RepoAPI.Mapping.props`](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RepoAPI.Mapping.props) so no dependencies.props changes are needed either.

weshaggard:
Yes if RepoAPI.Mapping.props is imported in the right spot then yes lets do that.

dagood:
@weshaggard the latest commit made the changes to make this entirely implemented in BuildTools (using existing source-build Repo API entrypoint in dependencies.props). PTAL

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
```diff
@@ -19,3 +19,47 @@ foreach ($file in Get-ChildItem $ToolRuntimePath *.runtimeconfig.json)
...
+
+    # Copied from init-tools.cmd in CoreFX
+    $retryCount = 0
+    $success = $false
+    $proxyCredentialsRequired = $false
+    do
+    {
+        try
+        {
+            $wc = New-Object Net.WebClient
```
weshaggard: While I don't like the duplicated logic for downloading things both here and in the init-tools scripts in the repo's I guess we can live with it until we get our auto-updating scripts workflow going and this can be shared.

dagood: Yeah, pained me to put this in here but I think it's the reasonable way in the short term while keeping the retries for reliability.

PR #1789, opened 2017-11-06
Add asset upload support

jcagme:
Currently we just support uploading nukpg files to the feeds. With this change if we find non-nupkg files in ItemsToPush, we upload them using Azure APIs and place them in {container}/{relativePath}/assets/.

ItemsToPush can be composed of only nupkgs, only non-nupkgs or a mixture of both

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
@@ -39,13 +39,7 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             RelativePath = relativePath;
         }
 
-        public string FeedContainerUrl
-        {
-            get
-            {
-                return AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
-            }
-        }
+        public string FeedContainerUrl => AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
 
         public async Task<bool> CheckIfBlobExists(string blobPath)
         {
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
@@ -3,10 +3,13 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Build.Framework;
+using Microsoft.DotNet.Build.CloudTestTasks;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net.Http;
 using System.Threading;
+using System.Threading.Tasks;
 using MSBuild = Microsoft.Build.Utilities;
 
 namespace Microsoft.DotNet.Build.Tasks.Feed
@@ -34,6 +37,40 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             ContainerName = containerName;
             Log = loggingHelper;
             RelativePath = relativePath;
+        }
+
+        public string FeedContainerUrl
+        {
+            get
+            {
+                return AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
+            }
+        }
+
+        public async Task<bool> CheckIfBlobExists(string blobPath)
+        {
+            string url = $"{FeedContainerUrl}/{blobPath}?comp=metadata";
+            using (HttpClient client = new HttpClient())
+            {
+                client.DefaultRequestHeaders.Clear();
+                var request = AzureHelper.RequestMessage("GET", url, AccountName, AccountKey).Invoke();
+                using (HttpResponseMessage response = await client.SendAsync(request))
+                {
+                    if (response.IsSuccessStatusCode)
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Blob {blobPath} exists for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    else
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Blob {blobPath} does not exist for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    return response.IsSuccessStatusCode;
+                }
+            }
         }
     }
 }
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
@@ -41,6 +41,32 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
 
         public string FeedContainerUrl => AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
 
+        public async Task<bool> CheckIfContainerExists()
+        {
+            string url = $"{FeedContainerUrl}?restype=container";
+            using (HttpClient client = new HttpClient())
+            {
+                client.DefaultRequestHeaders.Clear();
+                var request = AzureHelper.RequestMessage("GET", url, AccountName, AccountKey).Invoke();
+                using (HttpResponseMessage response = await client.SendAsync(request))
+                {
+                    if (response.IsSuccessStatusCode)
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Container {ContainerName} exists for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    else
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Low,
+                            $"Container {ContainerName} does not exist for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
+                    }
+                    return response.IsSuccessStatusCode;
+                }
+            }
+        }
+
         public async Task<bool> CheckIfBlobExists(string blobPath)
         {
             string url = $"{FeedContainerUrl}/{blobPath}?comp=metadata";
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
@@ -38,59 +38,5 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             Log = loggingHelper;
             RelativePath = relativePath;
         }
-
-        public string FeedContainerUrl => AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
-
-        public async Task<bool> CheckIfContainerExists()
-        {
-            string url = $"{FeedContainerUrl}?restype=container";
-            using (HttpClient client = new HttpClient())
-            {
-                client.DefaultRequestHeaders.Clear();
-                var request = AzureHelper.RequestMessage("GET", url, AccountName, AccountKey).Invoke();
-                using (HttpResponseMessage response = await client.SendAsync(request))
-                {
-                    if (response.IsSuccessStatusCode)
-                    {
-                        Log.LogMessage(
-                            MessageImportance.Low,
-                            $"Container {ContainerName} exists for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
-                    }
-                    else
-                    {
-                        Log.LogMessage(
-                            MessageImportance.Low,
-                            $"Container {ContainerName} does not exist for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
-                    }
-                    return response.IsSuccessStatusCode;
-                }
-            }
-        }
-
-        public async Task<bool> CheckIfBlobExists(string blobPath)
-        {
-            string url = $"{FeedContainerUrl}/{blobPath}?comp=metadata";
-            using (HttpClient client = new HttpClient())
-            {
-                client.DefaultRequestHeaders.Clear();
-                var request = AzureHelper.RequestMessage("GET", url, AccountName, AccountKey).Invoke();
-                using (HttpResponseMessage response = await client.SendAsync(request))
-                {
-                    if (response.IsSuccessStatusCode)
-                    {
-                        Log.LogMessage(
-                            MessageImportance.Low,
-                            $"Blob {blobPath} exists for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
-                    }
-                    else
-                    {
-                        Log.LogMessage(
-                            MessageImportance.Low,
-                            $"Blob {blobPath} does not exist for {AccountName}: Status Code:{response.StatusCode} Status Desc: {await response.Content.ReadAsStringAsync()}");
-                    }
-                    return response.IsSuccessStatusCode;
-                }
-            }
-        }
     }
 }
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -4,6 +4,7 @@
 
 using Microsoft.Build.Framework;
 using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
 using Newtonsoft.Json.Linq;
 using Sleet;
 using System;
@@ -27,6 +28,7 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
         private SleetSource source;
 
         public BlobFeed feed;
+        public int MaxClients { get; set; } = 8;
 
         public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLoggingHelper Log)
         {
@@ -112,6 +114,14 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             return !Log.HasLoggedErrors;
         }
 
+        public async Task UploadAssets(IEnumerable<string> items, bool allowOverwrite)
+        {
+            using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
+            {
+                await Task.WhenAll(items.Select(item => UploadAsync(item, $"{feed.RelativePath}assets/", clientThrottle, allowOverwrite, CancellationToken)));
+            }
+        }
+
         private bool IsSanityChecked(IEnumerable<string> items)
         {
             Log.LogMessage(MessageImportance.Low, $"START checking sanitized items for feed");
@@ -137,8 +147,6 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
 
         private LocalSettings GetSettings()
         {
-            
-
             SleetSettings sleetSettings = new SleetSettings()
             {
                 Sources = new List<SleetSource>
@@ -176,6 +184,76 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             AzureFileSystem fileSystem = GetAzureFileSystem();
             bool result = await InitCommand.RunAsync(settings, fileSystem, true, true, new SleetLogger(Log), CancellationToken);
             return result;
+        }
+
+        private async Task UploadAsync(string item, string uploadPath, SemaphoreSlim clientThrottle, bool allowOverwrite, CancellationToken cancellationToken)
+        {
+            if (!File.Exists(item))
+                throw new Exception(string.Format("The file '{0}' does not exist.", item));
+
+            string fileName = Path.GetFileName(item);
+            uploadPath += fileName;
+            await clientThrottle.WaitAsync();
+            string leaseId = string.Empty;
+
+            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(source.ConnectionString);
+            CloudBlobClient client = storageAccount.CreateCloudBlobClient();
+            CloudBlobContainer container = client.GetContainerReference(feed.ContainerName);
+            CloudBlockBlob blob = container.GetBlockBlobReference(uploadPath);
+            AzureBlobLease blobLease = new AzureBlobLease(blob);
+            bool blobExists = await blob.ExistsAsync();
+            bool isLeaseRequired = allowOverwrite && blobExists;
+
+            if (isLeaseRequired)
+            {
+                try
+                {
+                    leaseId = blobLease.LeaseId;
+                    Log.LogMessage($"Obtained lease ID {leaseId} for {uploadPath}.");
+                }
+                catch (Exception)
+                {
+                    Log.LogError($"Unable to obtain lease on {uploadPath}");
+                }
+            }
+
+            try
+            {
+                if (!blobExists || allowOverwrite)
+                {
+                    Log.LogMessage($"Uploading {item} to {uploadPath}.");
+                    AccessCondition accessCondition = new AccessCondition
+                    {
+                        LeaseId = leaseId
+                    };
+
+                    OperationContext opContext = new OperationContext
+                    {
+                        LogLevel = LogLevel.Informational
+                    };
+
+
+                    await blob.UploadFromFileAsync(item, accessCondition, new BlobRequestOptions(), opContext, cancellationToken);
+                }
+                else
+                {
+                    Log.LogMessage($"Skipping uploading of {item} to {uploadPath}. Already exists.");
+                }
+            }
+            catch (Exception)
+            {
+                Log.LogError($"Unable to upload to {uploadPath}");
+                throw;
+            }
+            finally
+            {
+                if (isLeaseRequired)
+                {
+                    blobLease.Release();
+                }
+
+                clientThrottle.Release();
+            }
         }
     }
 }
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -81,7 +81,6 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
         public async Task<bool> PushItemsToFeedAsync(IEnumerable<string> items, bool allowOverwrite)
         {
             Log.LogMessage(MessageImportance.Low, $"START pushing items to feed");
-            bool requiresInit = false;
 
             try
             {
@@ -89,6 +88,8 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
                 // We also retry in case Sleet is not able to get a lock on the feed since it does not retry in this case.
                 for (int i = 0; i < retries; i++)
                 {
+                    bool requiresInit = false;
+
                     try
                     {
                         bool result = await PushAsync(items.ToList(), allowOverwrite);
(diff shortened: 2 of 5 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -51,13 +63,22 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
-        private List<string> ConvertToStringLists(ITaskItem[] taskItems)
+        private List<string> ConvertToStringLists(ITaskItem[] taskItems, bool isNugetPackage)
```
weshaggard: I would change isNugetPackage to fileextension to make this method a little more general.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -51,13 +63,22 @@ public override bool Execute()
...
             {
-                stringList.Add(item.ItemSpec);
+                string fileSpec = item.ItemSpec;
+                string extension = Path.GetExtension(fileSpec);
+
+                // If packages is set to true it will only add files with ".nupkg" extension to the list (true == true)
+                // else this will add non-nupkg files to the list (false == false)
+                if ((extension == ".nupkg") == isNugetPackage)
+                {
+                    stringList.Add(item.ItemSpec);
```
weshaggard: You should also support a relative path passed in as item metadata for the non-nuget packages.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -177,5 +193,75 @@ private AzureFileSystem GetAzureFileSystem()
             bool result = await InitCommand.RunAsync(settings, fileSystem, true, true, new SleetLogger(Log), CancellationToken);
             return result;
         }
+
+        private async Task UploadAsync(string item, string uploadPath, SemaphoreSlim clientThrottle, bool allowOverwrite, CancellationToken cancellationToken)
```
weshaggard: Can we please share code with the UploadToAzure task instead of duplicating it?

dagood:
At the moment, Core-Setup puts nupkgs in blob storage as well as in feeds. I don’t know if any downstream repos consume it that way, but there’s a possibility for a small feature gap here. I suggest allowing the repo to pass metadata to bypass detection.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
weshaggard: I think we are going to want to support something like https://github.com/dotnet/buildtools/blob/78c57e11af6a055c734e180ae3c40a5967429be6/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets#L24 to ensure we preserve the same directory structure when we upload.

jcagme:
@dagood after the changes now one can upload all sort of files (including nupks) to a flat container

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +97,17 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+                    }
+                    catch (InvalidOperationException ex) when (ex.Message.Contains("Unable to obtain a lock on the feed."))
+                    {
+                        Log.LogWarning($"Sleet was not able to get a lock on the feed. Sleeping {delay} seconds and retrying.");
+                        await Task.Delay(delay);
+                    }
+
+                    if (requiresInit)
+                    {
+                        requiresInit = false;
```
dagood: Can avoid keeping track of this state by putting the bool in the loop.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +97,17 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
                         bool result = await InitAsync();
```
dagood: Does Init ever fail with an exception that should cause a retry?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +133,70 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
             return !Log.HasLoggedErrors;
         }
 
+        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false)
+        {
+            string relativeBlobPath = $"{feed.RelativePath}{item.GetMetadata("RelativeBlobPath").Replace("\\", "/")}"; 
+            if (string.IsNullOrEmpty(relativeBlobPath))
+                throw new Exception(string.Format("Metadata 'RelativeBlobPath' is missing for item '{0}'.", item.ItemSpec));
```
dagood: That doesn't seem quite accurate because `feed.RelativePath` might be set but not `RelativeBlobPath`. Also, any reason you used `string.Format` here explicitly?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +133,70 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
             return !Log.HasLoggedErrors;
         }
 
+        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false)
+        {
+            string relativeBlobPath = $"{feed.RelativePath}{item.GetMetadata("RelativeBlobPath").Replace("\\", "/")}";
```
dagood: Null ref exception if metadata is null due to Replace?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +133,70 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+        {
+            string relativeBlobPath = $"{feed.RelativePath}{item.GetMetadata("RelativeBlobPath").Replace("\\", "/")}"; 
+            if (string.IsNullOrEmpty(relativeBlobPath))
+                throw new Exception(string.Format("Metadata 'RelativeBlobPath' is missing for item '{0}'.", item.ItemSpec));
+
+            Log.LogMessage($"Uploading {relativeBlobPath}");
+
+            await clientThrottle.WaitAsync();
+
+            CloudTestTasks.AzureBlobLease blobLease = new CloudTestTasks.AzureBlobLease(feed.AccountName, feed.AccountKey, string.Empty, feed.ContainerName, relativeBlobPath, Log, "15", "5000");
```
dagood: Why 15 and 5000 rather than defaults 60 and 500?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +133,70 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+            if (isLeaseRequired)
+            {
+                try
+                {
+                    leaseId = blobLease.Acquire();
+                    Log.LogMessage($"Obtained lease ID {leaseId} for {relativeBlobPath}.");
+                }
+                catch (Exception)
+                {
+                    Log.LogError($"Unable to obtain lease on {relativeBlobPath}");
```
dagood: Original exception text should be (at least) logged as well.

dagood: If this method doesn't exit here, doesn't that means it continues to try to upload something while it knows that someone else has the lease?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
dagood: Could you please chop the args to new lines? Hard to read. (FWIW, I personally go for 100 columns as a soft limit.)

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -56,8 +78,9 @@ private List<string> ConvertToStringLists(ITaskItem[] taskItems)
             List<string> stringList = new List<string>();
             foreach (var item in taskItems)
             {
-                stringList.Add(item.ItemSpec);
+                     stringList.Add(item.ItemSpec);
```
dagood: Extra indent.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -35,5 +38,39 @@ public BlobFeed(string accountName, string accountKey, string containerName, str
             Log = loggingHelper;
             RelativePath = relativePath;
         }
+
+        public string FeedContainerUrl
+        {
+            get
+            {
+                return AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
```
dagood: `public string FeedContainerUrl => AzureHelper.GetContainerRestUrl(AccountName, ContainerName);`?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -74,12 +81,13 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
             Log.LogMessage(MessageImportance.Low, $"START pushing items to feed");
+            bool requiresInit = false;
 
             try
             {
                 // In case the first Push attempt fails with an InvalidOperationException we Init the feed and retry the Push command once.
-                // Sleet internally retries 5 times on each package when the push operation fails so we don't need to retry ourselves.
-                for (int i = 0; i <= 1; i++)
+                // We also retry in case Sleet is not able to get a lock on the feed since it does not retry in this case.
+                for (int i = 0; i < retries; i++)
```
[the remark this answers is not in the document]
jcagme: retries default value is 5. I've added code to make sure that if init was called we don't count that as an iteration.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: Karthik made this change and I brought it from file's history. I'll add the comment he had. In summary seems that we are looking to expire the lease in 15 seconds instead of the default 60

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +98,22 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
                     catch (InvalidOperationException ex) when (ex.Message.Contains("init"))
                     {
                         Log.LogWarning($"Sub-feed {source.FeedSubPath} has not been initialized. Initializing now...");
+                        requiresInit = true;
+                    }
+                    catch (InvalidOperationException ex) when (ex.Message.Contains("Unable to obtain a lock on the feed."))
```
weshaggard: Why are you checking for string messages? That is troublesome given they might change at anytime and they also might change if they are localized.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -39,6 +43,9 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
                 string relativePath = m.Groups["relativepath"].Value;
                 feed = new BlobFeed(accountName, accountKey, containerName, relativePath, Log);
                 feedUrl = m.Groups["feedurl"].Value;
+                retries = retryAttempts;
+                delay = TimeSpan.FromSeconds(retryDelay);
```
weshaggard: Do we think we need some random variation in the retry delay?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -24,6 +25,15 @@ public class PushToBlobFeed : MSBuild.Task
 
         public bool Overwrite { get; set; }
```
weshaggard: From my recent testing in coreclr I think we also need an option to also create the container if it doesn't already exist. That will help in creating feeds on the fly.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: I've added logic below to increase the delay on every iteration since, depending on what's being pushed it might take up to minutes. I don't think randomizing will make a difference but making the delay longer

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: While I agree there is no other way to catch this particular scenario. This is what being thrown "throw new InvalidOperationException("Unable to obtain a lock on the feed.");" and InvalidOperationException is thrown almost everywhere else

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
jcagme: Ok, I'll add that. I thought we did not want to support automatically creating containers

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +139,83 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+            if (string.IsNullOrEmpty(relativeBlobPath))
+                throw new Exception($"Metadata 'RelativeBlobPath' is missing for item '{item.ItemSpec}'.") ;
+            relativeBlobPath = $"{feed.RelativePath}{relativeBlobPath.Replace("\\", "/")}"; 
+            
+            Log.LogMessage($"Uploading {relativeBlobPath}");
+
+            await clientThrottle.WaitAsync();
+
+            // this defines the lease for 15 seconds (max is 60) and 3000 milliseconds between requests
+            CloudTestTasks.AzureBlobLease blobLease = new CloudTestTasks.AzureBlobLease(
```
weshaggard: Is this pulling in a new binary dependency or is this source sharing?

weshaggard: Nevermind I see it is source you pulled in.

jcagme: Shared code. Namespace is different so I had to add it but is not a binary dependency

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
dagood: In other BT code we've done exponential stepback, but not randomness (yet):

https://github.com/dotnet/buildtools/blob/fffbc00cd3e7cc8bd5724a662e4c1a4c2cf8755f/src/Microsoft.DotNet.Build.Tasks/ExecWithRetries.cs#L26

https://github.com/dotnet/buildtools/blob/fffbc00cd3e7cc8bd5724a662e4c1a4c2cf8755f/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh#L22

I don't think anyone's actually experimented to get there, but it kinda makes sense that the first tries should be relatively quick, then later make them longer to account for potential network congestion.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +98,22 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+                        // Pushing packages might take more than just 60 seconds, so on each iteration we add the defined value to itself so wait for
+                        // a bit more the next iterations
+                        delay += delay;
+                    }
+
+                    // If the feed has not been Init'ed this will be caught in the first iteration
+                    if (requiresInit && i == 0)
+                    {
+                        i--;
                         bool result = await InitAsync();
```
dagood: (Duplicating outdated comment to make it visible again:)

> Does Init ever fail with an exception that should cause a retry?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +139,83 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+            string relativeBlobPath = item.GetMetadata("RelativeBlobPath");
+            if (string.IsNullOrEmpty(relativeBlobPath))
+                throw new Exception($"Metadata 'RelativeBlobPath' is missing for item '{item.ItemSpec}'.") ;
+            relativeBlobPath = $"{feed.RelativePath}{relativeBlobPath.Replace("\\", "/")}"; 
+            
+            Log.LogMessage($"Uploading {relativeBlobPath}");
+
+            await clientThrottle.WaitAsync();
+
+            // this defines the lease for 15 seconds (max is 60) and 3000 milliseconds between requests
```
dagood: Can you explain why?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +139,83 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+            {
+                try
+                {
+                    leaseId = blobLease.Acquire();
+                    Log.LogMessage($"Obtained lease ID {leaseId} for {relativeBlobPath}.");
+                }
+                catch (Exception exc)
+                {
+                    Log.LogError($"Unable to obtain lease on {relativeBlobPath} due to exception: {exc}");
+                    throw;
```
dagood: Is it a problem that `clientThrottle` isn't released?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: We want to fail fast if this fails. I've added return code to it

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
jcagme: @karajas can you help with this comment?

karajas: This is just the shortest lease possible, which will auto renew if needed.

jcagme: Regardless, we are not leasing anymore on asset files

src/Microsoft.DotNet.Build.Tasks.Feed/SleetLogger.cs
```diff
@@ -43,7 +43,7 @@ public void LogDebug(string data)
 
         public void LogError(string data)
         {
-            _log.LogError(data);
+            _log.LogWarning(data);
```
dagood: I think you should add some extra message here saying it was originally sent as an error, for diagnosability.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +99,21 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
                     {
                         Log.LogWarning($"Sub-feed {source.FeedSubPath} has not been initialized. Initializing now...");
+                        requiresInit = true;
+                    }
+                    catch (InvalidOperationException ex) when (ex.Message.Contains("Unable to obtain a lock on the feed."))
+                    {
+                        Log.LogWarning($"Sleet was not able to get a lock on the feed. Sleeping {delay} seconds and retrying.");
+
+                        // Pushing packages might take more than just 60 seconds, so on each iteration we add the defined value to itself so wait for
+                        // a bit more the next iterations
```
dagood: Log and comment are no longer accurate.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
```diff
@@ -35,5 +38,59 @@ public BlobFeed(string accountName, string accountKey, string containerName, str
             Log = loggingHelper;
             RelativePath = relativePath;
         }
+
+        public string FeedContainerUrl => AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
+
+        public async Task<bool> CheckIfContainerExists()
```
weshaggard: Do we have this code anywhere else? If so we might want to share source with it as well.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +99,21 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+                        await Task.Delay(TimeSpan.FromSeconds(rnd.Next(1, 5) * delay));
+                    }
+
+                    // If the feed has not been Init'ed this will be caught in the first iteration
+                    if (requiresInit && i == 0)
+                    {
+                        i--;
```
weshaggard: this `i--` through me for a loop it might be worth a comment.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +140,71 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+                relativeBlobPath = $"{feed.RelativePath}{recursiveDir}{fileName}";
+            }
+
+            relativeBlobPath = relativeBlobPath.Replace("\\", "/");
+
+            Log.LogMessage($"Uploading {relativeBlobPath}");
+
+            await clientThrottle.WaitAsync();
+
+            bool blobExists = await feed.CheckIfBlobExists(relativeBlobPath);
```
weshaggard: We don't need to check if the blob exists if we are allowing overwrite. It just doing more work then we need to.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +140,71 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+                        relativeBlobPath);
+                }
+                else
+                {
+                    Log.LogError($"Item '{item}' already exists in {relativeBlobPath}.");
+                }
+            }
+            catch (Exception)
+            {
+                Log.LogError($"Unable to upload to {relativeBlobPath}");
```
weshaggard: We should include the exception message in this error.

weshaggard: Actually the exception toString.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeed.cs
jcagme: Actually the "CreateAzureContainer" has a bool to avoid failing in case the container exist already so there is no need to check for that.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -112,6 +142,66 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+                    UploadClient uploadClient = new UploadClient(Log);
+                    await uploadClient.UploadBlockBlobAsync(
+                        CancellationToken,
+                        feed.AccountName,
+                        feed.AccountKey,
+                        feed.ContainerName,
+                        item.ItemSpec,
+                        relativeBlobPath);
+                }
+                else
```
weshaggard: You still need to check for existence if allowOverwrite is not true.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +99,23 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
                     catch (InvalidOperationException ex) when (ex.Message.Contains("init"))
                     {
                         Log.LogWarning($"Sub-feed {source.FeedSubPath} has not been initialized. Initializing now...");
+                        requiresInit = true;
+                    }
+                    catch (InvalidOperationException ex) when (ex.Message.Contains("Unable to obtain a lock on the feed."))
+                    {
+                        Log.LogWarning($"Sleet was not able to get a lock on the feed. Sleeping {delay} seconds and retrying.");
```
dagood: This prints out the value pre-multiplication.

src/Microsoft.DotNet.Build.Tasks.Feed/SleetLogger.cs
```diff
@@ -43,7 +43,9 @@ public void LogDebug(string data)
 
         public void LogError(string data)
         {
-            _log.LogError(data);
+            // There are cases where Sleet fails and we retry on our side causing things to actually work but if Sleet logs an error the whole build leg
+            // is marked as failed even though it actually succeeded, hence we log a warning here but we will log an error if the retry did not help
```
dagood: I meant something should be added in the warning message to alert anyone trying to diagnose sleet errors that show up as warnings.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -89,6 +99,23 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
+
+                        // Pushing packages might take more than just 60 seconds, so on each iteration we multiply the defined delay to a random factor
+                        // Using the defaults this could range from 30 seconds to 12.5 minutes.
```
dagood: 12.5 minutes seems excessive for potentially the first retry. The more I think about this, the more I like the exponential approach intuitively (perhaps with a random sleep time reduction).

jcagme: Sorry, looks like I need to work on my comment, top delay amount on each iteration is 2.5 minutes. If we use the default 5 retried this adds up to 12.5

src/Microsoft.DotNet.Build.Tasks.Feed/SleetLogger.cs
jcagme: These comments have been addressed in https://github.com/dotnet/buildtools/pull/1791

PR #1791, opened 2017-11-07
Checking for blob existence

jcagme:
In case overwrite is not set to true then we have to check whether the blob exists.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -103,11 +103,13 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
                     }
                     catch (InvalidOperationException ex) when (ex.Message.Contains("Unable to obtain a lock on the feed."))
                     {
-                        Log.LogWarning($"Sleet was not able to get a lock on the feed. Sleeping {delay} seconds and retrying.");
+                        int delayInSeconds = rnd.Next(1, 5) * delay;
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -174,7 +174,12 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
                 }
                 else
                 {
-                    Log.LogError($"Item '{item}' already exists in {relativeBlobPath}.");
```
weshaggard: This isn't the correct logic. if allowOverride == true and !blobExists then upload. Just don't do the exist check if allowOverride is set to true.

weshaggard:
```
bool blobExists = false;

if (!allowOverride) 
{
   blobExists = await feed.CheckIfBlobExists(relativeBlobPath);
}

if (allowOverwrite || !blobExists)
{
     Log.LogMessage($"Uploading {item} to {relativeBlobPath}.");
     UploadClient uploadClient = new UploadClient(Log);
     await uploadClient.UploadBlockBlobAsync(
         CancellationToken,
         feed.AccountName,
         feed.AccountKey,
         feed.ContainerName,
         item.ItemSpec,
         relativeBlobPath);
}
else
{
    Log.LogError($"Item '{item}' already exists in {relativeBlobPath}.");
}
```

PR #1792, opened 2017-11-07
Make the GenFacades task work for Portable Pdbs

vancem:
Basically we add code so that if portable PDBs are detected,
we run a portable -> windows conversion, do the old
GenFacades, and then convert windows -> Portable.

Previously we just gave up and did not produce  PDB.
Fixes issue  https://github.com/dotnet/buildtools/issues/1739

@dagood @stephentoub


src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
```diff
@@ -41,12 +41,12 @@
...
     <Move SourceFiles="$(PartialFacadeSymbols)"
           DestinationFolder="$(GenFacadesInputPath)"
-          Condition="'$(DebugSymbols)' != 'false' AND '$(DebugType)' != 'Portable'"
+          Condition="'$(DebugSymbols)' != 'false'"
     />
 
     <PropertyGroup>
-      <ProducePdb>true</ProducePdb>
-      <ProducePdb Condition="'$(DebugSymbols)' == 'false' OR '$(DebugType)'=='Portable'">false</ProducePdb>
+	  <ProducePdb>true</ProducePdb>
```
weshaggard: nit: indention

src/GenFacades/GenFacades.Core/project.json
```diff
@@ -1,6 +1,7 @@
 {
   "dependencies": {
     "Microsoft.Cci": "4.0.0-rc3-24214-00",
+    "Microsoft.DiaSymReader.Converter": "1.1.0-beta1-62225-01",
```
weshaggard: Where does this dependency come from? We will ultimately need to be able to build it from source for our build-from-source efforts.

dagood: It's in symreader-portable (https://github.com/dotnet/symreader-portable/tree/0b38450a75c9ea28b40020ac9b1a14a092ed1621/src/Microsoft.DiaSymReader.Converter) which we already have a submodule of in source-build.

vancem: @weshaggard  also note that Microsoft.DotNet.Build.Tasks.dll (which is also build in BuildTools), already has a dependency on Microsoft.DiaSymReader.Converter.  Thus there is no really new dependency here.

vancem:
If there are no objections I am going to merge this (so I can get the fix to CoreFX).

src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -198,6 +216,60 @@ private static void OutputFacadeToFile(string facadePath, HostEnvironment seedHo
...
+                    converter.ConvertPortableToWindows(peStream, pdbReadStream, pdbWriteStream, PdbConversionOptions.SuppressSourceLinkConversion);
+                }
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Convert the windows PDB winPdbInputPath to portablePdbOutputPath. 
+        /// 'dllInputPath' is the DLL that goes along with 'winPdbInputPath'.  
+        /// </summary
```
dagood: Missing closing `>`

src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -198,6 +216,60 @@ private static void OutputFacadeToFile(string facadePath, HostEnvironment seedHo
...
+            }
+        }
+
+        /// <summary>
+        /// Given dllInputPath determine if it is portable.  If so convert it *\WindowsPdb\*.pdb and update 'pdbInputPath' to 
+        /// point to this converted file.
+        /// Returns true if the conversion was done (that is the original file was protable).  
+        /// 'dllInputPath' is the DLL that goes along with 'pdbInputPath'.  
+        /// </summary>
+        private static bool ConvertFromPortableIfNecessary(string dllInputPath, ref string pdbInputPath)
```
dagood: This API seems odd to me--my first thought is to return `string` or `null` depending on it if converted or not. Since it's private and only used in one place, I don't object to merging it this way, though.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
```diff
@@ -67,7 +67,7 @@
       Seeds="@(GenFacadesSeeds, ',')"
       FacadePath="$(GenFacadesOutputPath.TrimEnd('/'))"
       SeedTypePreferencesUnsplit="@(SeedTypePreference)"
-      ProducePdb="$(ProducePdb)"
+	  ProducePdb="$(ProducePdb)"
```
dagood: Nit: tab character causing misalignment

src/GenFacades/GenFacades.Core/GenFacades.cs
vancem: I was willing to change this but when I started it was actually less convenient to use (because you want to update in place AND remember if you did the up). so I have left it.

PR #1793, opened 2017-11-08
Enable download flat files and download in parallel

karajas:
@weshaggard @jcagme

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
@@ -5,12 +5,11 @@
 using Microsoft.Build.Framework;
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.IO;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs
```diff
@@ -113,44 +151,45 @@ public override bool Execute()
...
+                        }
                     }
                 }
-                Log.LogMessage($"{failureCount} errors seen downloading blobs.");
             }
-            catch (Exception e)
+            catch (Exception ex)
             {
-                Log.LogErrorFromException(e, true);
+                Log.LogError(ex.Message);
```
jcagme: Log.LogErrorFromException

karajas: Fixed.

karajas:
Merging to unblock coreclr. Will fix if something comes up.

PR #1794, opened 2017-11-08
Track discarded commits in a comment, not PR description

### LINKED ISSUE #1783: Use a comment to save discarded auto-update commits, not the description
The description's put into an environment variable by Jenkins, and the description gets very long over time. There's a limit to how long the env vars can be and this ends up breaking CI until someone intervenes.

Instead, VersionTools should post a comment and update that.

dagood:
This prevents the auto-PR description from growing in size over time. The description is fed into an environment variable in CI, which has a length limit.

When the auto-PR code needs to save a discarded commit's info, it looks for the first comment posted by the bot in the PR. If it hasn't posted a comment yet, it posts one.

Resolves #1783 
Resolves https://github.com/dotnet/core-eng/issues/2014 (FR) by working around the problem. The root bug could still be hit by a dev if they manually make an extremely long PR description.

I exercised this code on this testing PR: https://github.com/dagood/corefx/pull/32

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
@@ -87,7 +87,7 @@ namespace Microsoft.DotNet.VersionTools.Automation
                         {
                             if (options.TrackDiscardedCommits)
                             {
-                                await UpdateDiscardedCommitListAsync(
+                                await PostDiscardedCommitCommentAsync(
                                     baseBranch.Project,
                                     pullRequestToUpdate,
                                     headCommit,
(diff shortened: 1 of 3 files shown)


dagood:
@jcagme, this is kind of related to reliability because it fixes a problem that can prevent uptake. Maybe you could take a look?

@weshaggard if you have a moment, a review would be much appreciated.

src/Microsoft.DotNet.VersionTools/Automation/PullRequestCreator.cs
```diff
@@ -167,24 +167,54 @@ public class PullRequestCreator
...
 
+            // Find comment to update, if one exists.
+            GitHubComment[] prIssueComments = await client.GetAllIssueCommentsAsync(
+                baseProject,
+                pullRequestToUpdate.Number);
+
+            GitHubComment commentToUpdate = prIssueComments
+                .FirstOrDefault(c => c.User.Login == _auth.User);
+
+            string commentBodyToUpdate = commentToUpdate?.Body ?? string.Empty;
```
weshaggard: Why not just create a new comment each time? Is there any value in updating an existing comment?

dagood:
Hmm, the reason it tries to get an existing comment is to avoid increasing the number of notifications the auto-PRs are sending out. I don't know if this is actually a requirement. Just posting a comment would definitely be simpler, and it also works better with comment order if people are actively discussing the PR.

@stephentoub @MichaelSimons @eerhardt @dotnet/corefxlab-contrib @dotnet/coreclr-auto-update-notify
I think you're all getting auto-PR notifications: what do you think? Is it ok to get a new auto-upgrade PR comment each time a new upgrade is pushed, or would you prefer a single comment that gets updated?

If nobody objects before, say, Tuesday 11/14, I'm changing this so that a new comment shows up with the old CI results whenever a new auto-update commit gets pushed. (This PR is not urgent: a known workaround is to remove the PR description when CI fails when it gets too long.)

weshaggard:
People already get notifications when a new commit is pushed so the noise is there for the most part so I'm not sure this will make it any worse.

eerhardt:
Agreed.  I'd have no objection to a new comment on every update.

ahsonkhan:
> Is it ok to get a new auto-upgrade PR comment each time a new upgrade is pushed, or would you prefer a single comment that gets updated?

A new comment on every update is fine by me.

dagood:
Thanks, all. The comments look like this:

---

<details><summary>

Discarded [`16aa292`](https://github.com/dagood-bot/corefx/commit/16aa292b2952843c9e6a04543a3cf49a6047af26) (1:x: 1:hourglass: 1:heavy_check_mark:) Update CoreClr, CoreFx, CoreSetup, ProjectNTfs, ProjectNTfsTestILC to preview1-25914-01, preview1-25913-03, preview1-25912-02, beta-25914-00, beta-25914-00, respectively
</summary>

 * :x: **testing-fail** Description
 * :hourglass: **testing-pending** Description
 * :heavy_check_mark: **testing-success** Description

</details>

PR #1795, opened 2017-11-08
Create Public containers through PushToFeed

jcagme:
If a token is not provided in the ExpectedFeedUrl when we create a container we make them public.

/cc: @weshaggard @karajas

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -274,7 +274,7 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
         {
             LocalSettings settings = GetSettings();
             AzureFileSystem fileSystem = GetAzureFileSystem();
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -271,7 +274,7 @@ private AzureFileSystem GetAzureFileSystem()
         {
             LocalSettings settings = GetSettings();
             AzureFileSystem fileSystem = GetAzureFileSystem();
-            bool result = await InitCommand.RunAsync(settings, fileSystem, true, true, new SleetLogger(Log), CancellationToken);
+            bool result = await InitCommand.RunAsync(settings, fileSystem, true, false, new SleetLogger(Log), CancellationToken);
```
weshaggard: What is this bool change?

jcagme: It was my test yesterday of push/not push symbols. I added that in my change but somehow the commit is not here

PR #1797, opened 2017-11-09
Fix for CoreFX build break on Linux

vancem:
PR #1792 had a break on Linux because of course windows PDBs don't exist there (and neither does the converter).     Add condition to simply not call the converter on non-windows platforms.  

@dagood

### DIFF
--- a/src/GenFacades/GenFacades.Core/GenFacades.cs
+++ b/src/GenFacades/GenFacades.Core/GenFacades.cs
@@ -13,6 +13,7 @@ using Microsoft.Cci.Extensions;
 using Microsoft.Cci.MutableCodeModel;
 using System.Globalization;
 using Microsoft.DiaSymReader.Tools;
+using System.Runtime.InteropServices;
 
 namespace GenFacades
 {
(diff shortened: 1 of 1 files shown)


src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -189,7 +190,8 @@ private static void OutputFacadeToFile(string facadePath, HostEnvironment seedHo
                         // Convert from portable to windows PDBs if necessary.  If we convert
                         // set the pdbOutput path to a *.windows.pdb file so we can convert it back 
                         // to 'finalPdbOutputPath'.
-                        needsConversion = ConvertFromPortableIfNecessary(facade.Location, ref pdbLocation);
+                        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                            needsConversion = ConvertFromPortableIfNecessary(facade.Location, ref pdbLocation);
```
dagood: Style guides usually say to include braces, but I see this is done elsewhere in the file too so no big deal.

stephentoub:
Thanks, @vancem.

PR #1798, opened 2017-11-10
Publish only symbols.nupkg as symbols

jcagme:
Updating Sleet version so we don't publish packages as symbols when the extension is not symbols.nupkg

Removed retry logic on push since now the retries when the feed is locked are native in Sleet


dependencies.props
```diff
@@ -89,7 +89,7 @@
     <NuGetDependency Include="NuGet.RuntimeModel"/>
     <NuGetDependency Include="NuGet.Versioning"/>
     <StaticDependency Include="@(NuGetDependency)">
-      <Version>4.3.0</Version>
+      <Version>4.4.0</Version>
```
weshaggard: We need to avoid bumping nuget to 4.4.0 without extensive testing

karajas: What's the current version in  2.0 cli? 4.3.0 right?

weshaggard: Yes, which is why I believe this will cause failures.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -103,16 +101,6 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
                         Log.LogWarning($"Sub-feed {source.FeedSubPath} has not been initialized. Initializing now...");
                         requiresInit = true;
                     }
```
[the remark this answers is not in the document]
jcagme: Yes, we don't need to do anything on our side now. We also added better locking handling so the whole locking should be better now

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -28,10 +28,6 @@ public class PushToBlobFeed : MSBuild.Task
 
         public bool PublishFlatContainer { get; set; }
 
-        public int RetryAttempts { get; set; } = 5;
```
weshaggard: remove these from the targets file.

jcagme:
As a reference, Sleet commits on 11/09 have the retry logic, dependency changes and sub-feed feedlock improvements: https://github.com/emgarten/Sleet/commits/master

PR #1799, opened 2017-11-10
Fix to unbreak Linux build.

vancem:
This is a second attempt.   The first attempt was a misunderstanding of how Linux builds worked.

Before trying to support protable PDBs, GenFacades had a condition that it was never called if
Portable PDBs are used.   However we WANTED to used portable PDBs so we added support for
that and dropped the condition to skip GenFacades if portable PDBs were used.

However this had the effect of running GenFacades on Linux, where previously it never ran there
(because Linux always used portable PDBs).

This fix basically puts the behavior back EXACTLY as it was before on non-windows platforms.
This will fix errors (they happen to be index out of range errors) on Linux builds (because we
are running GenFacades where we should not be).

@dagood  @stephentoub

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
@@ -38,17 +38,17 @@
           DestinationFolder="$(GenFacadesInputPath)"
     />
 
-    <!-- Move the PDB into a subdirectory for GenFacades if we are producing PDBs -->
-    <Move SourceFiles="$(PartialFacadeSymbols)"
-          DestinationFolder="$(GenFacadesInputPath)"
-          Condition="'$(DebugSymbols)' != 'false' AND '$(OSGroup)' == 'Windows_NT'" />
-
     <PropertyGroup>
       <ProducePdb>true</ProducePdb>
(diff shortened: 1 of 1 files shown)


vancem:
@weshaggard

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
```diff
@@ -41,12 +41,11 @@
     <!-- Move the PDB into a subdirectory for GenFacades if we are producing PDBs -->
     <Move SourceFiles="$(PartialFacadeSymbols)"
           DestinationFolder="$(GenFacadesInputPath)"
-          Condition="'$(DebugSymbols)' != 'false'"
-    />
+          Condition="'$(DebugSymbols)' != 'false' AND '$(OSGroup)' == 'Windows_NT'" />
```
weshaggard: I would suggest moving the ProducePdb property above this and using that in the condition.

vancem:
For what it is worth, I also feel bad for breaking things multiple times to try to achieve my original goal.   Part of the problem is that I am not set up for Linux builds.  I test this on Windows by dropping in and building CoreFX, but I should do the same for Linux.   I should have been thinking more about Linux than I was.   On my next non-trivial update, I will be more thorough...

vancem:
followon fix to https://github.com/dotnet/buildtools/pull/1792 
(first attempt #1797 did not work)

Issue #1363, opened 2017-03-02
Automatically restore and publish ILAsm in buildtools initialization


mellinoe:
Right now, repos wanting to use ilasm outside of Windows need to have a custom initialization step that restores and publishes ilasm for the correct operating system. We should include that in the initialization of buildtools itself.

@weshaggard

weshaggard:
hopefully this can be part of the tools bootstraping work to make this easier to pull down.

MichalStrehovsky:
I think the only thing that's left here is changes to `init-tools.sh` to restore ILAsm on Unix too.

weshaggard:
Reopening this to track switching to this model in corefx.

weshaggard:
Updated corefx to use ilasm restored from buildtools package.

PR #1806, opened 2017-11-16
Add ILAsm to buildtools

MichalStrehovsky:
Windows-only for now since that's what I need. I think we would want the bootstrapper (the thing that invokes init-tools) to pass us the RID it used to restore .NET CLI so that we can restore the appropriate version of the native tools. RID in the Windows batch file was easy to hardcode since we seem to hardcode x64 on Windows everywhere.

Contributes to #1363.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
@@ -1,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- suppress the attempt to copy build output. -->
+    <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
+  </PropertyGroup>
+
+  <Target Name="CoreCompile">
+    <Copy SourceFiles="@(NativeCopyLocalItems)" DestinationFolder="$(MSBuildThisFileDirectory)" />
+  </Target>
+
+  <!-- Required by Common.Targets -->
+  <Target Name="CreateManifestResourceNames" />
+
+  <PropertyGroup>
+    <NuGetTargetMoniker>.NETCoreApp,Version=v2.0</NuGetTargetMoniker>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RidSpecificAssets>true</RidSpecificAssets>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.Runtime.CoreCLR" Version="2.1.0-preview1-25916-01" />
+    <PackageReference Include="Microsoft.NETCore.ILAsm" Version="2.1.0-preview1-25916-01" />
+  </ItemGroup>
+</Project> 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <!-- suppress the attempt to copy build output. -->
@@ -19,7 +19,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETCore.Runtime.CoreCLR" Version="2.1.0-preview1-25916-01" />
-    <PackageReference Include="Microsoft.NETCore.ILAsm" Version="2.1.0-preview1-25916-01" />
+    <PackageReference Include="Microsoft.NETCore.ILAsm" Version="$(ILAsmPackageVersion)" />
+
+    <!-- ILAsm has a very unfortunate runtime dependency on CoreCLR, so we need to grab that too -->
+    <!-- https://github.com/dotnet/coreclr/issues/15059 -->
+    <PackageReference Include="Microsoft.NETCore.Runtime.CoreCLR" Version="$(ILAsmPackageVersion)" />
   </ItemGroup>
 </Project> 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
     <!-- suppress the attempt to copy build output. -->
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
```diff
@@ -0,0 +1,25 @@
...
+  <Target Name="CreateManifestResourceNames" />
+
+  <PropertyGroup>
+    <NuGetTargetMoniker>.NETCoreApp,Version=v2.0</NuGetTargetMoniker>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RidSpecificAssets>true</RidSpecificAssets>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.Runtime.CoreCLR" Version="2.1.0-preview1-25916-01" />
```
weshaggard: We need to be careful about hardcoding these versions in BuildTools. If the output for this is isolated it might be OK but if it isn't isolated it might conflict with the other tools. We also need a strategy about updating this version.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -95,6 +95,19 @@ Robocopy "%PACKAGES_DIR%\MicroBuild.Core\%MICROBUILD_VERSION%\build\." "%TOOLRUN
 Robocopy "%PACKAGES_DIR%\Microsoft.Net.Compilers\%ROSLYNCOMPILERS_VERSION%\." "%TOOLRUNTIME_DIR%\net46\roslyn\." /E
 
 @echo on
+call "%DOTNET_CMD%" build "%TOOLRUNTIME_DIR%\ilasm\ilasm.depproj" -r win-x64 --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --packages "%PACKAGES_DIR%\."
```
weshaggard: While I agree x64 is the common default there might be issues with anyone that may or may not be using this in Razzle because that requires x86 tools.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -95,6 +95,19 @@ Robocopy "%PACKAGES_DIR%\MicroBuild.Core\%MICROBUILD_VERSION%\build\." "%TOOLRUN
 Robocopy "%PACKAGES_DIR%\Microsoft.Net.Compilers\%ROSLYNCOMPILERS_VERSION%\." "%TOOLRUNTIME_DIR%\net46\roslyn\." /E
 
 @echo on
```
weshaggard: I feel like we need to make this optional in some way because not all consumers of BuildTools require ilasm. I wonder if we should hook this initialization up as part of https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets, so it is pay-for-play.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
jkotas: > using this in Razzle

Is there anybody still using recent enough buildtools in Razzle?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
MichalStrehovsky: Is this any different from the version of Roslyn hardcoded in `init-tools.cmd`? I can hardcode it in the same place as `ROSLYNCOMPILERS_VERSION` if that will make you feel better about this but it will be equally unconfigurable.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
MichalStrehovsky: You mean lazily downloading it when something ends up needing it? How does it gel with "build from source" effort and building without internet connection?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: I hope not but I'm not entirely sure. My main point is that there are some envirments that still force x86 toolsets. I think it is OK to assume x64 but we may need to figure out a way to make it an option somehow. I think my other suggestion about moving this into the il.targets instead might help mitigate that issue as then we can set an msbuild property to control it if needed.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
weshaggard: If we move this logic to the il.targets we could control this via a property and we could have a default value for it but we having a way to override it would be good.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
weshaggard: They will either have the ilasm package in their cache or they won't. If they don't then it will fail. While I don't mind this being in init-tools I want to make sure it is optional in some way and not required for every repo using BuildTools which is what you currently have.

jaredpar:
It looks like this PR is struggling with the same problems that Roslyn had using ILASM in our Unix builds. The tool is not published as a stand alone NuGet package. Instead it's only usable if you use NuGet to publish the artifacts to an output directory. 

This has already caused a lot of unneeded complexity in the Roslyn build. Now it looks like it's adding friction here too.

Have we considered just fixking the NuGet package here? Virtually every other NuGet package that is a tool publishes as a stand alone artifact that can be consumed by other repos. Why does ILASM want to be different here and how does it justify complicating all the downstream consumers?

weshaggard:
That is a great point @jaredpar. I didn't realize there were others using that package. @RussKeldorph is this something we could follow-up on?

One issue that we will hit with this particular tool is that it is a native exe and I don't know if there is a great way for individual projects to consume it without defining a RID in their project, which kind of sucks for IL projects who's purpose is to be runtime agnostic.

RussKeldorph:
@weshaggard Wouldn't it be great if we had a 100% compatible managed ILAsm built on Roslyn?  After all, it's just another front end on an IL code generator. :)

@jaredpar I'm not familiar with the history, but I suspect ILAsm doesn't "want to be different."  I'd guess that we simply didn't know the proper way to package things...or that way didn't exist at the time and ILAsm is behind.

If there's a better way to do things, let us know.  Happy to discuss offline.

jkotas:
> a 100% compatible managed ILAsm

+1

> built on Roslyn

It would not make sense to build it on Roslyn. ilasm does not need any intermediate form, it just reads the text and emits the IL instructions. Roslyn infrastructure would be overkill for it.

MichalStrehovsky:
I'm all up for doing something to make the ILAsm package actually useful (that's why I filed dotnet/coreclr#15059 yesterday), but is this something considered blocking for this PR, or can we make progress on getting this PR merged without it?

I think the public interface with buildtools clients that I'm proposing here will stay the same no matter what happens to the package (or acquisition mechanism) underneath.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -94,6 +94,27 @@ Robocopy "%PACKAGES_DIR%\MicroBuild.Core\%MICROBUILD_VERSION%\build\." "%TOOLRUN
 :: Copy Roslyn Compilers Over to ToolRuntime
 Robocopy "%PACKAGES_DIR%\Microsoft.Net.Compilers\%ROSLYNCOMPILERS_VERSION%\." "%TOOLRUNTIME_DIR%\net46\roslyn\." /E
 
+:: Restore ILAsm if the caller asked for it by setting the environment variable
+if [%ILASMCOMPILER_VERSION%]==[] goto :afterILAsmRestore
+if [%NATIVE_TOOLS_RID%]==[] (
```
weshaggard: I think it is reasonable to default this to win-x64 if it isn't set.

weshaggard:
>but is this something considered blocking for this PR, or can we make progress on getting this PR merged without it?

We should not block this PR.

PR #1808, opened 2017-11-16
Update Roslyn to 2.6.0-beta3-62316-02

eerhardt:
Updating to the latest Roslyn compiler.  This is necessary in order to use the `in` parameter keyword as the current version still uses `ref readonly`.


weshaggard:
@VSadov @jaredpar is this a OK version of the compiler package to use?

joperezr:
> is this a OK version of the compiler package to use?

I believe usually what @VSadov recommends is to use it if we were able to use it in corefxlab and build successfully.

VSadov:
Yes. Subject to passing tests of course. 
Compiler is pretty stable these days since there is no much churn in the shipping branch.

eerhardt:
> LGTM assuming we've tried using it in corefx and coreclr and built succesfully

I tried it locally just in corefx and on my Windows machine.  I didn't run it through everything.  There were 2 places that need to change in corefx.  One is [already changed](https://github.com/dotnet/corefx/pull/25300).  [The other](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/ref/System.Runtime.CompilerServices.Unsafe.cs#L19) will need to wait until we get the new compiler.

Honestly - if the new compiler causes problems, that should be a huge red flag that needs to get addressed very quickly.

VSadov:
I will make a PR in corfxlab. That should be quick. It is a good idea to update there as well anyways.

joperezr:
@eerhardt already did: https://github.com/dotnet/corefxlab/pull/1940#pullrequestreview-77285938

VSadov:
https://github.com/dotnet/corefxlab/pull/1941    lets see if that works.

VSadov:
@joperezr - I will close my PR then. And it looks like yours is passing.

PR #1809, opened 2017-11-16
Publish symbols to flat containers

jcagme:
We've decided that we'll publish symbols to flat containers by default instead of using Sleet. This change disables Sleet's symbols publishing and adds logic to identify and categorize packages and symbols

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -97,15 +97,14 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
         {
             string relativeBlobPath = item.GetMetadata("RelativeBlobPath");
 
-            if (string.IsNullOrEmpty(relativeBlobPath) || symbols)
+            if (string.IsNullOrEmpty(relativeBlobPath))
             {
-                string symbolsFolder = symbols ? "symbols/" : string.Empty;
                 string fileName = Path.GetFileName(item.ItemSpec);
                 string recursiveDir = item.GetMetadata("RecursiveDir");
-                relativeBlobPath = $"{feed.RelativePath}{symbolsFolder}{recursiveDir}{fileName}";
+                relativeBlobPath = $"{recursiveDir}{fileName}";
             }
 
-            relativeBlobPath = relativeBlobPath.Replace("\\", "/");
+            relativeBlobPath = $"{feed.RelativePath}{relativeBlobPath}".Replace("\\", "/");
 
             Log.LogMessage($"Uploading {relativeBlobPath}");
 
@@ -170,7 +169,6 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             {
                 try
                 {
-
                     bool result = await InitAsync();
 
                     if (result)
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -79,11 +79,9 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
         public async Task<bool> PushItemsToFeedAsync(IEnumerable<string> items, bool allowOverwrite)
         {
             Log.LogMessage(MessageImportance.Low, $"START pushing items to feed");
-            Random rnd = new Random();
 
             try
             {
-
                 bool result = await PushAsync(items.ToList(), allowOverwrite);
                 return result;
             }
@@ -95,14 +93,16 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             return !Log.HasLoggedErrors;
         }
 
-        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false)
+        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false, bool symbols = false)
         {
             string relativeBlobPath = item.GetMetadata("RelativeBlobPath");
-            if (string.IsNullOrEmpty(relativeBlobPath))
+
+            if (string.IsNullOrEmpty(relativeBlobPath) || symbols)
             {
+                string symbolsFolder = symbols ? "symbols/" : string.Empty;
                 string fileName = Path.GetFileName(item.ItemSpec);
                 string recursiveDir = item.GetMetadata("RecursiveDir");
-                relativeBlobPath = $"{feed.RelativePath}{recursiveDir}{fileName}";
+                relativeBlobPath = $"{feed.RelativePath}{symbolsFolder}{recursiveDir}{fileName}";
             }
 
             relativeBlobPath = relativeBlobPath.Replace("\\", "/");
@@ -148,7 +148,7 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             }
         }
 
-        public async Task CreateContainerAsync(IBuildEngine buildEngine)
+        public async Task CreateContainerAsync(IBuildEngine buildEngine, bool publishFlatContainer)
         {
             Log.LogMessage($"Creating container {feed.ContainerName}...");
 
@@ -166,23 +166,26 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
 
             Log.LogMessage($"Creating container {feed.ContainerName} succeeded!");
 
-            try
+            if (!publishFlatContainer)
             {
-
-                bool result = await InitAsync();
-
-                if (result)
+                try
                 {
-                    Log.LogMessage($"Initializing sub-feed {source.FeedSubPath} succeeded!");
+
+                    bool result = await InitAsync();
+
+                    if (result)
+                    {
+                        Log.LogMessage($"Initializing sub-feed {source.FeedSubPath} succeeded!");
+                    }
+                    else
+                    {
+                        throw new Exception($"Initializing sub-feed {source.FeedSubPath} failed!");
+                    }
                 }
-                else
+                catch (Exception e)
                 {
-                    throw new Exception($"Initializing sub-feed {source.FeedSubPath} failed!");
+                    Log.LogErrorFromException(e);
                 }
-            }
-            catch (Exception e)
-            {
-                Log.LogErrorFromException(e);
             }
         }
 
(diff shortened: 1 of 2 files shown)


jaredpar:
Why did we decide against sleet here?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -80,15 +77,27 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
-        private List<string> ConvertToStringLists(ITaskItem[] taskItems)
+        private List<string> GetPackageStringLists(ITaskItem[] taskItems)
         {
             List<string> stringList = new List<string>();
-            foreach (var item in taskItems)
+            foreach (var item in taskItems.Where(i => !i.ItemSpec.Contains("symbols.nupkg")))
```
weshaggard: We shouldn't have this extension check in multiple places. We should have it both together outside this function.

jcagme:
@jaredpar I've just added you to the email thread. In summary, symbol publishing through Sleet was adding time to the publish leg due to the indexing and creation of temp blobs so we decided to use flat containers instead.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -53,22 +52,21 @@ public override bool Execute()
...
+                        await PublishToFlatContainerAsync(ItemsToPush, blobFeedAction);
                     }
                     else
                     {
-                        using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
-                        {
-                            Log.LogMessage($"Uploading {ItemsToPush.Length} items...");
-                            await Task.WhenAll(ItemsToPush.Select(item => blobFeedAction.UploadAssets(item, clientThrottle, Overwrite)));
-                        }
+                        ITaskItem[] symbolItems = ItemsToPush.Where(i => i.ItemSpec.Contains("symbols.nupkg")).ToArray();
```
weshaggard: What about the relative path setting it to "symbols"?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -95,14 +93,16 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
 
-        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false)
+        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false, bool symbols = false)
         {
             string relativeBlobPath = item.GetMetadata("RelativeBlobPath");
-            if (string.IsNullOrEmpty(relativeBlobPath))
+
+            if (string.IsNullOrEmpty(relativeBlobPath) || symbols)
             {
+                string symbolsFolder = symbols ? "symbols/" : string.Empty;
```
weshaggard: No reason to special case this here. Just set the RelativeBlobPath metadata to "symbols/$(filename)" in the root directory below.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -95,17 +93,18 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
             return !Log.HasLoggedErrors;
         }
 
-        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false)
+        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, bool allowOverwrite = false, bool symbols = false)
```
weshaggard: Don't need the symbols parameter.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -246,7 +247,7 @@ private AzureFileSystem GetAzureFileSystem()
         {
             LocalSettings settings = GetSettings();
             AzureFileSystem fileSystem = GetAzureFileSystem();
-            bool result = await InitCommand.RunAsync(settings, fileSystem, true, true, new SleetLogger(Log), CancellationToken);
+            bool result = await InitCommand.RunAsync(settings, fileSystem, false, false, new SleetLogger(Log), CancellationToken);
```
weshaggard: What are these bool parameters? Can you use named parameters there to help document them?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -79,11 +79,9 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
         public async Task<bool> PushItemsToFeedAsync(IEnumerable<string> items, bool allowOverwrite)
```
weshaggard: We should probably exit early if there are no items in the list.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -79,11 +79,15 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
         {
             Log.LogMessage(MessageImportance.Low, $"START pushing items to feed");
-            Random rnd = new Random();
 
-            try
+            if (items.Count() == 0)
```
jaredpar: Prefer the following

``` c#
if (!items.Any()) 
```

It's more efficient than `.Count()` as it will iterate at most 1 element.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -79,11 +79,15 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
         public async Task<bool> PushItemsToFeedAsync(IEnumerable<string> items, bool allowOverwrite)
         {
             Log.LogMessage(MessageImportance.Low, $"START pushing items to feed");
-            Random rnd = new Random();
 
-            try
+            if (!items.Any())
             {
+                Log.LogError("No items to push found in the items list.");
```
weshaggard: This should not be an error. Just make it a message.

In a case where someone only pushes symbol.nupkg items this would be empty and we don't want that to be an error.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -53,22 +53,26 @@ public override bool Execute()
...
-                        using (var clientThrottle = new SemaphoreSlim(this.MaxClients, this.MaxClients))
+                        ITaskItem[] symbolItems = ItemsToPush.Where(i => i.ItemSpec.Contains("symbols.nupkg")).Select(i => 
                         {
-                            Log.LogMessage($"Uploading {ItemsToPush.Length} items...");
-                            await Task.WhenAll(ItemsToPush.Select(item => blobFeedAction.UploadAssets(item, clientThrottle, Overwrite)));
-                        }
+                            string fileName = Path.GetFileName(i.ItemSpec);
+                            i.SetMetadata("RelativeBlobPath", $"symbols/{fileName}");
+                            return i;
+                        }).ToArray();
```
weshaggard: From an efficiency point of view you are calling ToArray and then passing an IEnumerable and later calling ToList you are making a number of copies of these lists. That may start to add up for larger file lists.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -53,22 +53,26 @@ public override bool Execute()
...
                         {
-                            Log.LogMessage($"Uploading {ItemsToPush.Length} items...");
-                            await Task.WhenAll(ItemsToPush.Select(item => blobFeedAction.UploadAssets(item, clientThrottle, Overwrite)));
-                        }
+                            string fileName = Path.GetFileName(i.ItemSpec);
+                            i.SetMetadata("RelativeBlobPath", $"symbols/{fileName}");
+                            return i;
+                        });
+                        var packages = ItemsToPush.Where(i => !symbolItems.Contains(i));
+                        List<string> packageItems = GetPackageStringLists(packages);
```
weshaggard: Does this need to be a list as opposed to an IEnumerable?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -79,11 +79,15 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
...
-            try
+            if (!items.Any())
             {
+                Log.LogMessage("No items to push found in the items list.");
+                return true;
+            }
 
+            try
+            {
                 bool result = await PushAsync(items.ToList(), allowOverwrite);
```
weshaggard: Do we need to call ToList here?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -238,15 +245,16 @@ private AzureFileSystem GetAzureFileSystem()
         {
             LocalSettings settings = GetSettings();
             AzureFileSystem fileSystem = GetAzureFileSystem();
-            bool result = await PushCommand.RunAsync(settings, fileSystem, items.ToList(), allowOverwrite, false, new SleetLogger(Log));
+            bool result = await PushCommand.RunAsync(settings, fileSystem, items.ToList(), allowOverwrite, skipExisting: false, log: new SleetLogger(Log));
```
weshaggard: Do we need to call ToList here?

jcagme: In this case Sleet actually needs a List to be passed in so we need to convert the IEnumerable to List

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -68,8 +68,11 @@ public override bool Execute()
                             i.SetMetadata("RelativeBlobPath", $"symbols/{fileName}");
                             return i;
                         });
-                        var packages = ItemsToPush.Where(i => !symbolItems.Contains(i));
-                        List<string> packageItems = GetPackageStringLists(packages);
+
+                        var packageItems = ItemsToPush.Where(i => !symbolItems.Contains(i)).Select(i => 
+                        {
```
weshaggard: You don't really need the brackets here. I would just do `Select(i => i.ItemSpec);`

PR #1812, opened 2017-11-20
Restore ILAsm on Unix

### LINKED ISSUE #1363: Automatically restore and publish ILAsm in buildtools initialization

MichalStrehovsky:
Also restore ILAsm on Unix. Turns out this will make my life easier, so just doing that too. Fixes #1363.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -93,6 +93,21 @@ __MNCA_FOLDER=$(dirname $__DOTNET_CMD)/shared/Microsoft.NETCore.App
 __HIGHEST_RUNTIME_VERSION=`ls $__MNCA_FOLDER | sed 'r/\([0-9]\+\).*/\1/g' | sort -n | tail -1`
 sed -i -e "s/1.1.0/$__HIGHEST_RUNTIME_VERSION/g" $__TOOLRUNTIME_DIR/*.runtimeconfig.json
 
+# Restore ILAsm, if requested in the environment.
+__ILASM_PACKAGE_VERSION="${ILASMCOMPILER_VERSION:-}"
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
@crummel @weshaggard please take a look

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -93,6 +93,21 @@ __MNCA_FOLDER=$(dirname $__DOTNET_CMD)/shared/Microsoft.NETCore.App
 __HIGHEST_RUNTIME_VERSION=`ls $__MNCA_FOLDER | sed 'r/\([0-9]\+\).*/\1/g' | sort -n | tail -1`
 sed -i -e "s/1.1.0/$__HIGHEST_RUNTIME_VERSION/g" $__TOOLRUNTIME_DIR/*.runtimeconfig.json
 
+# Restore ILAsm, if requested in the environment.
+__ILASM_PACKAGE_VERSION="${ILASMCOMPILER_VERSION:-}"
+if [ "$__ILASM_PACKAGE_VERSION" ]; then
+    echo "Restoring ILAsm version '$__ILASM_PACKAGE_VERSION'..."
+	
+    __ILASM_PACKAGE_RID="${NATIVE_TOOLS_RID:-}"
```
weshaggard: I think it is probably OK to default to `linux-x64` as the RID.

MichalStrehovsky: Erroring out right here might be a bit more easy to troubleshoot. I'm not sure running a Linux executable on e.g. macOS produces a useful error message that would be easy to backtrack to this place, should someone hit this.

PR #1814, opened 2017-11-21
Adding 5 minute timeout to HttpClient for UploadToAzure scenarios

jcagme:
Some core-setup and corefx build are recently failing when trying to upload big symbol packages. At this point HttpClient uses the default timeout of 100 seconds which in some cases is not enough

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
@@ -46,6 +46,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             string ContainerName,
             string filePath,
             string destinationBlob,
+            int uploadTimeout,
             string leaseId = "")
         {
             string resourceUrl = AzureHelper.GetContainerRestUrl(AccountName, ContainerName);
@@ -86,7 +87,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                         client.DefaultRequestHeaders.Clear();
 
                         // In random occassions the request fails if the network is slow and it takes more than 100 seconds to upload 4MB. 
-                        client.Timeout = TimeSpan.FromMinutes(5);
+                        client.Timeout = TimeSpan.FromMinutes(uploadTimeout);
                         Func<HttpRequestMessage> createRequest = () =>
                         {
                             DateTime dt = DateTime.UtcNow;
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
@@ -84,6 +84,9 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                     using (HttpClient client = new HttpClient())
                     {
                         client.DefaultRequestHeaders.Clear();
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -84,6 +84,9 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
                     using (HttpClient client = new HttpClient())
                     {
                         client.DefaultRequestHeaders.Clear();
+
+                        // In random occassions the request fails if the network is slow and it takes more than 100 seconds to upload 4MB. 
+                        client.Timeout = TimeSpan.FromMinutes(5);
```
jaredpar: Did we consider making the timeout an argument to the build task? Would amke it easer to configure in the future: change a build property vs. rebuilding the DLL.

jaredpar:
Why are we using HTTP to upload to Azure here instead of the Azure APIs?

jcagme:
@jaredpar the original idea was to have one less dependency on an external library. Today, that remains true since we would have to have the azure-storage-net repo be part of our orchestrated build. I think we would highly benefit of using the APIs but it also comes with a cost. @weshaggard, do you have other reasons?

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
jcagme: I did but thought this might not be a thing we'd like to change that often but I'll just add 5 as default and then the value can be overridden if needed

jaredpar:
> Today, that remains true since we would have to have the azure-storage-net repo be part of our orchestrated build.

Why would that be the case? This is the orchestrated build, not source build, hence we can take dependencies as we see fit.

jcagme:
> Why would that be the case? This is the orchestrated build, not source build, hence we can take dependencies as we see fit.
Are we going to have a different fork for source build where we reference external repos and not import the packages?

jaredpar:
> Are we going to have a different fork for source build where we reference external repos and not import the packages?

The transport feeds are not necessary for a successful completion of source build. We should not make our orchestrated build less reliable due to the existence of source build. We should either:

1. Change the build tools build such that we condition out the transport feeds during source build
1. Remove all code related to orchestrated build out of build tools into a separate repo

jcagme:
@jaredpar Makes sense. Is it ok to merge this change to be covered while we implement the change with the APIs?

jaredpar:
Sure. Just want to understand why we're doing the extra work here. Seems like we could just switch to the Azure SDK in a future change.

markwilkie:
@weshaggard - comments on using the API instead of http?  I too would rather use the hardened surface area of the API - especially now that it runs on core.

weshaggard:
>The transport feeds are not necessary for a successful completion of source build. 

That isn't exactly true, at least not at this point. We would need to do more refactoring work to make that true and I'm not sure it is worth it because it would be another fork in our engineering system.

Just to give some background context we have had this dependency since before Azure SDK supported .NET Core but now that they do we should definitely look at switching to it. We just need to keep in mind that potentially causes future debt of either making it build from source or refactoring our tools to break the dependency. 

The work to switch is tracked with https://github.com/dotnet/buildtools/issues/1816.

PR #1818, opened 2017-11-23
Unify IL.targets with CoreCLR's version

MichalStrehovsky:
See https://github.com/dotnet/coreclr/blob/06828ca1eea8af9b2a0ac147ba8827f128a306e2/tests/src/IL.targets.


MichalStrehovsky:
@weshaggard please take a look

PR #1819, opened 2017-11-24
Support optional ContentType item metadata in UploadToAzure

trylek:
Prefer explicit content type specification in the item metadata
and only use autodetection when not available. We're using this
for uploading test logs and results to Azure - without proper
content type some browsers have trouble opening them.

This is my first contribution to this repo so I'll be grateful for
any advice regarding testing and the review process in general.
The change passes the default testing as part of building the repo
and I have verified the correct content types in the Azure
Storage Explorer.


trylek:
Thanks Jan!

PR #1821, opened 2017-11-28
Add symbol uploader build task package to optional tooling list.

mikem8361:
This is needed by PR https://github.com/dotnet/core-setup/pull/3465

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/optional-tool-runtime/optional.json
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/optional-tool-runtime/optional.json
@@ -2,7 +2,7 @@
   "frameworks": {
     "net45": {
       "dependencies": {
-	"Microsoft.SymbolUploader.Build.Task": "1.0.0-alpha-00001",
+        "Microsoft.SymbolUploader.Build.Task": "1.0.0-alpha-00001",
         "Microsoft.DotNet.BinInspector": "1.0.0-alpha-00001",
         "Microsoft.DotNet.IBCMerge": "4.6.0-alpha-00001",
         "Microsoft.DotNet.UAP.TestTools" : "1.0.2"
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/optional-tool-runtime/optional.json
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/optional-tool-runtime/optional.json
@@ -2,6 +2,7 @@
   "frameworks": {
     "net45": {
       "dependencies": {
+	"Microsoft.SymbolUploader.Build.Task": "1.0.0-alpha-00001",
         "Microsoft.DotNet.BinInspector": "1.0.0-alpha-00001",
         "Microsoft.DotNet.IBCMerge": "4.6.0-alpha-00001",
         "Microsoft.DotNet.UAP.TestTools" : "1.0.2"


mikem8361:
@dagood not sure how to bump the package version number (and how to upload the new build tools package).

mikem8361:
@weshaggard, @dagood This PR should only affect the publish tasks. I also hard coded my private VSTS symbol server path in core-setup-publish.json because I want to test using that first before switching to the Microsoft one (https://microsoft.artifacts.visualstudio.com/DefaultCollection) making them available on symweb.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/optional-tool-runtime/optional.json
```diff
@@ -2,6 +2,7 @@
   "frameworks": {
     "net45": {
       "dependencies": {
+	"Microsoft.SymbolUploader.Build.Task": "1.0.0-alpha-00001",
```
jaredpar: Nit: indentation is one space off here.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/optional-tool-runtime/optional.json
```diff
@@ -2,6 +2,7 @@
   "frameworks": {
     "net45": {
       "dependencies": {
+        "Microsoft.SymbolUploader.Build.Task": "1.0.0-alpha-00001",
```
dagood: Where can I get the packages this package depends on? I can't find them on the VSTS feed or on any public feeds.

mikem8361:
It shouldn’t depend on any other packages but it looks like from the nuspec generated that are. I should have checked that. Two of them are from the symstore repo and I included them in the package. The other is Microsoft.Build.Utilities.Core which should be fine because it is part of msbuild.

The repo it is built from is: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/DotNet-SymStore

dagood:
> @dagood not sure how to bump the package version number (and how to upload the new build tools package).

That's automatic--once this is merged, the build starts with a new number and it'll show up here: https://github.com/dotnet/versions/blob/master/build-info/dotnet/buildtools/master/Last_Build_Packages.txt

There will also be auto-PRs submitted to a few repos including Core-Setup (after a 10 minute delay) to uptake the new package.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/optional-tool-runtime/optional.json
mikem8361: Are those dependencies going to cause a problem?  I'm using the msbuild pack logic which doesn't seem to have a simple way to suppress them unless I want to manually create the nuspec file.

dagood: I'd expect the tool restore to fail when it can't find them. I'm not aware of a small way to get rid of them either (but it's not an area I know much about).

mikem8361:
It didn’t fail when I tried this locally in the core-setup PR. It was a little hacky because I pointed  OptionalToolSource property to a local directory with my package.

dagood:
Hmm, are you sure your package cache(s) didn't have the dependencies already somehow?

mikem8361:
I’ve never created packages for the two DotNet-SymStore dependences: Microsoft.SymbolStore and Microsoft.SymbolUploader.

    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.SymbolUploader" version="1.0.0-dev" exclude="Build,Analyzers" />
        <dependency id="Microsoft.SymbolStore" version="1.0.0-dev" exclude="Build,Analyzers" />
        <dependency id="Microsoft.Build.Utilities.Core" version="15.3.409" exclude="Build,Analyzers" />
      </group>
    </dependencies>

mikem8361:
@dagood Is there any way to test this change (optional tooling)?

weshaggard:
@dagood we really need to remove this project.json file :)

@dagood can correct me if I'm wrong but I think you will need to add this to an override file in corefx as well. 

As for testing there isn't any great way beyond trying to do a private run which isn't trivial.

mikem8361:
I don’t think we need to put in the corefx override because the symbol upload build task is only needed by core-setup.

>>>@dagood<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fdagood&data=02%7C01%7Cmikem%40microsoft.com%7C2bb3e192fedd432a78c708d5374dee72%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636475730208427889&sdata=WTHsHSMhuKaS%2BwcpLQX%2Be5rXWiekoDhkR%2BT4V9xqUvs%3D&reserved=0> can correct me if I'm wrong but I think you will need to add this to an override file in corefx as well.

weshaggard:
>I don’t think we need to put in the corefx override because the symbol upload build task is only needed by core-setup.

Why is it only needed for core-setup? Aren't we planning to use this for indexing all our symbols for anything we ship now? Not everything flows through core-setup.

mikem8361:
>>>>Why is it only needed for core-setup? Aren't we planning to use this for indexing all our symbols for anything we ship now? Not everything flows through core-setup.

I wanted to start with one repo first like core-setup and I did assume that all the files we want to publish on the symbol servers for coreclr and corefx would flow through core-setup.  Was I wrong about that?

PR #1823, opened 2017-12-01
Handle additional false-positive NuGet push scenarios

chcosta:
Addresses https://github.com/dotnet/core-eng/issues/1422

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PublishProduct.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/PublishProduct.targets
@@ -31,13 +31,19 @@
     <ItemGroup>
       <NuGetPushCommand Include="@(PackagesToPush -> '$(NuGetExePath) $(NuGetPushArgsBase) %(Identity)')" />
 
-      <!-- There's a very special failure scenario that we want to ignore.  That scenario is
-           when NuGet hits a timeout on one "push" attempt, and then gets a "Forbidden" response
-           because the package "already exists" on the next response.  This indicates that the
-           timeout occurred, but the push was actually successful.
+      <!-- There are special failure scenarios that we want to ignore.  Those scenarios are
+           when NuGet reports a failure during a push attempt for something server or timeout related, 
+           and then gets a "Forbidden" response on the next attempt because the package "already exists".  
+           This indicates that the failure was reported, but the push was actually successful.
       -->
       <IgnorableErrorMessages Include="Overwriting existing packages is forbidden according to the package retention settings for this feed.">
         <ConditionalErrorMessage>Pushing took too long</ConditionalErrorMessage>
+      </IgnorableErrorMessages>
+      <IgnorableErrorMessages Include="Overwriting existing packages is forbidden according to the package retention settings for this feed.">
+        <ConditionalErrorMessage>The connection with the server was terminated abnormally</ConditionalErrorMessage>
+      </IgnorableErrorMessages>
+      <IgnorableErrorMessages Include="Overwriting existing packages is forbidden according to the package retention settings for this feed.">
+        <ConditionalErrorMessage>An existing connection was forcibly closed by the remote host</ConditionalErrorMessage>
       </IgnorableErrorMessages>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/PublishProduct.targets
```diff
@@ -31,14 +31,20 @@
...
-           timeout occurred, but the push was actually successful.
+      <!-- There are special failure scenarios that we want to ignore.  Those scenarios are
+           when NuGet reports a failure during a push attempt for something server or timeout related, 
+           and then gets a "Forbidden" response on the next attempt because the package "already exists".  
+           This indicates that the failure was reported, but the push was actually successful.
       -->
       <IgnorableErrorMessages Include="Overwriting existing packages is forbidden according to the package retention settings for this feed.">
         <ConditionalErrorMessage>Pushing took too long</ConditionalErrorMessage>
       </IgnorableErrorMessages>
+      <IgnorableErrorMessages Include="Overwriting existing packages is forbidden according to the package retention settings for this feed.">
```
dagood: I'm leaning towards putting this in the propertygroup. I doubt it'll change, but IMO it would make it slightly easier to know they're exactly the same when reading.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/PublishProduct.targets
```diff
@@ -31,14 +31,20 @@
     <ItemGroup>
       <NuGetPushCommand Include="@(PackagesToPush -> '$(NuGetExePath) $(NuGetPushArgsBase) %(Identity)')" />
 
-      <!-- There's a very special failure scenario that we want to ignore.  That scenario is
-           when NuGet hits a timeout on one "push" attempt, and then gets a "Forbidden" response
-           because the package "already exists" on the next response.  This indicates that the
-           timeout occurred, but the push was actually successful.
+      <!-- There are special failure scenarios that we want to ignore.  Those scenarios are
+           when NuGet reports a failure during a push attempt for something server or timeout related, 
+           and then gets a "Forbidden" response on the next attempt because the package "already exists".
```
weshaggard: Do we need to worry about the package being in a bad state?

Will these ignored messages cause us to stop retrying for the given package and move on to the next one?

chcosta: I'm not certain if that's possible or not.  It's not something that we have seen to date.  Retrying wouldn't fix anything though because we would just continue to hit a "package already exists" type error.

weshaggard: I agree retry isn't useful in these errors.

dagood: The only bad state I think we can get in is having a green build that actually should have been a 409... pushing a package, getting one of these networking errors instead of the real 409, then getting the 409 on the retry.

A few things mitigate this risk: we haven't seen that sequence before, and it would have to happen for every pushed package with a real 409. In some cases there's only one package with a given version, but I think it's more typical for there to be a bunch.

weshaggard: Other risk to think about is if we respin stable builds the second build might end up passing even though it didn't end up successfully pushing the packages it built. So we would end up with the first stable build set of packages instead of the second.

dagood: I think that's the same risk I'm talking about. We only ignore the 409's if we get one of these specific transient network errors in a previous retry, and it seems like a very slim chance to hit one of these for every single package, then also then hit the 409 on a retry. If there's a persistent network error, I wouldn't expect it to alternate between network-based failure and valid 409 response like that.

weshaggard: OK fair enough.

PR #1828, opened 2017-12-06
Provide a way to ignore BuildConfigurations through _ prefix in the build configuration

safern:
This will allow to add a build configuration with an _ at the beginning of it and it will be a negation of that build configuration. 

The build system will find those build configurations, will turn them to a valid configuration by removing the prefix and if that configuration is present in the best build configurations then we will return an empty set of best build configurations and will turn in a noop for that project.

Changes will need to be made in corefx to add this negative build configurations to the projects that are inbox on netfx.

Contributes to: https://github.com/dotnet/corefx/issues/24903

cc: @danmosemsft

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/Configuration/ConfigurationFactory.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/Configuration/ConfigurationFactory.cs
@@ -13,6 +13,7 @@ namespace Microsoft.DotNet.Build.Tasks
     public class ConfigurationFactory
     {
         internal const char PropertySeparator = '-';
+        internal const string NopConfigurationPrefix = "_";
 
         private Dictionary<string, PropertyInfo> Properties { get; }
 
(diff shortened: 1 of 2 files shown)


danmosemsft:
Would `!` be more intuitive than `_` ?

safern:
> Would ! be more intuitive than _ ?

I think yes, @weshaggard and I try to come up with the best way but we didn't think of `!`

wfurt:
! may be special character - it is at least for bash.

build@toweinfu-ubu14:~/github/corefx/src/System.Net.Sockets/tests/FunctionalTests$ echo _foo
_foo
build@toweinfu-ubu14:~/github/corefx/src/System.Net.Sockets/tests/FunctionalTests$ echo !foo
-bash: !foo: event not found

That may be unpleasant to deal with even if our system is not directly impacted.

weshaggard:
I thought about `!` as well but I didn't like that one. The reason I think `_` is better is because it mapps more closely to the nuget convention of `_._` which means it is a place holder and doesn't contain an asset.

danmosemsft:
Another option is `-` which is sometimes used in command lines to "undefine" a symbol.

But this is your change, I have no opinion...

weshaggard:
>Another option is - which is sometimes used in command lines to "undefine" a symbol.

We wanted to avoid `-` as well because it is already used as a separator in these configs.

safern:
@weshaggard any thoughts on the fix/change?

PR #1829, opened 2017-12-07
Omit restore during publish of tool-runtime

chcosta:
The issue reported in https://github.com/dotnet/core-eng/issues/2213, initially included a callstack...

```
2017-12-05T13:06:52.4893036Z System.ComponentModel.Win32Exception (0x80004005): Access is denied
2017-12-05T13:06:52.4893036Z    at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
2017-12-05T13:06:52.4893036Z    at System.Diagnostics.Process.Start()
2017-12-05T13:06:52.4893036Z    at Microsoft.DotNet.Cli.Utils.ProcessStartInfoExtensions.Execute(ProcessStartInfo startInfo)
2017-12-05T13:06:52.4893036Z    at Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingApp.Execute()
2017-12-05T13:06:52.4893036Z    at Microsoft.DotNet.Tools.Restore.RestoreCommand.Run(String[] args)
2017-12-05T13:06:52.4893036Z    at Microsoft.DotNet.Tools.RestoringCommand.Execute()
2017-12-05T13:06:52.4893036Z    at Microsoft.DotNet.Tools.Publish.PublishCommand.Run(String[] args)
2017-12-05T13:06:52.4893036Z    at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, ITelemetry telemetryClient)
2017-12-05T13:06:52.4893036Z    at Microsoft.DotNet.Cli.Program.Main(String[] args)
```

That callstack implies that the `dotnet publish` of the tool-runtime is doing a `dotnet restore` and the dotnet restore that is failing with "Access is denied".  This change disables the restore from the publish step because we've already done a `dotnet restore` on the tool runtime in a separate step of init-tools previously.  

While I think this may reduce the occurences of "Access is denied", slightly, I don't think it's a complete solution as the majority of cases where we hit this type of failure, we don't get a callstack included and creating a repro is difficult.  I do think that we're likely hitting some sort of process handle issue related to the global package cache, but I have yet to be able to repro the problem so I can't address that further at the moment.


weshaggard:
This is likely not the reason for the "Access Denied" but reasonable clean-up.

PR #1836, opened 2017-12-20
 Create build manifest in PushToBlobFeed for orchestrated build final publish

dagood:
This adds functionality to `PushToBlobFeed` to create build output manifests. See [core-eng/publish.md](https://github.com/dotnet/core-eng/blob/master/Documentation/Orchestrated-Build/Api/publish.md#cookbook-using-pushtoblobfeed-to-generate-a-manifest) for details on usage.

https://github.com/dotnet/core-eng/issues/2287 "Add output manifest generation to PushToBlobFeed task"

---

This is intended for single/low-use feeds: the manifest(s) would build up over time on a central blob feed. Central feeds can ensure manifests are *never* pushed to them by putting a blob called `disable-manifest-push` at the root of the feed. (The blob's contents aren't examined and don't matter.)

> I initially wanted to use Sleet feed settings for this. However, `PushToBlobFeed` by default runs `Init` on every call, which clears the settings. I think it's infeasible and error-prone to distribute an update that disables the `Init` calls, so I decided to go with disabling via a blob.

I'll add this blob to the `dotnet-core` central feed before merging this change.
(Done: https://dotnetfeed.blob.core.windows.net/dotnet-core/disable-manifest-push)

FYI @jcagme @weshaggard 

---

I also added a small VersionTools test project. It doesn't match the existing BuildTools infra because it uses the new Core SDK project format (it won't run in CI). I wasn't able to quickly figure out how to get the existing BuildTools test project patterns to work any better than sometimes working with F5 in Visual Studio.

I also created a `LaunchDebugger` task that IMO is overdue. It makes it really easy to debug BuildTools tasks without temporarily hard-coding a `Debugger.Launch()`.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
@@ -3,18 +3,25 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Build.Framework;
+using Microsoft.DotNet.VersionTools.Automation;
+using Microsoft.DotNet.VersionTools.BuildManifest.Model;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 using MSBuild = Microsoft.Build.Utilities;
 
 namespace Microsoft.DotNet.Build.Tasks.Feed
 {
     public class PushToBlobFeed : MSBuild.Task
     {
+        private static readonly char[] ManifestDataPairSeparators = { ';' };
+        private const string DisableManifestPushConfigurationBlob = "disable-manifest-push";
+        private const string AssetsVirtualDir = "assets/";
+
         [Required]
         public string ExpectedFeedUrl { get; set; }
 
@@ -33,6 +40,22 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
         public bool SkipCreateContainer { get; set; } = false;
 
         public int UploadTimeoutInMinutes { get; set; } = 5;
+
+        public bool SkipCreateManifest { get; set; }
+
+        public string ManifestName { get; set; } = "anonymous";
+        public string ManifestBuildId { get; set; } = "no build id provided";
+        public string ManifestBranch { get; set; }
+        public string ManifestCommit { get; set; }
+
+        /// <summary>
+        /// When publishing build outputs to an orchestrated blob feed, do not change this property.
+        /// 
+        /// The virtual dir to place the manifest XML file in, under the assets/ virtual dir. The
+        /// default value is the well-known location that orchestration searches to find all
+        /// manifest XML files and combine them into the orchestrated build output manifest.
+        /// </summary>
+        public string ManifestAssetOutputDir { get; set; } = "orchestration-metadata/manifests/";
 
         public override bool Execute()
         {
@@ -53,6 +76,9 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
                 {
                     BlobFeedAction blobFeedAction = new BlobFeedAction(ExpectedFeedUrl, AccountKey, Log);
 
+                    IEnumerable<BlobArtifactModel> blobArtifacts = Enumerable.Empty<BlobArtifactModel>();
+                    IEnumerable<PackageArtifactModel> packageArtifacts = Enumerable.Empty<PackageArtifactModel>();
+
                     if (!SkipCreateContainer)
                     {
                         await blobFeedAction.CreateContainerAsync(BuildEngine, PublishFlatContainer);
@@ -61,6 +87,7 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
                     if (PublishFlatContainer)
                     {
                         await PublishToFlatContainerAsync(ItemsToPush, blobFeedAction);
+                        blobArtifacts = ConcatBlobArtifacts(blobArtifacts, ItemsToPush);
                     }
                     else
                     {
@@ -69,12 +96,24 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
                             string fileName = Path.GetFileName(i.ItemSpec);
                             i.SetMetadata("RelativeBlobPath", $"symbols/{fileName}");
                             return i;
-                        });
+                        }).ToArray();
 
-                        var packageItems = ItemsToPush.Where(i => !symbolItems.Contains(i)).Select(i => i.ItemSpec);
+                        ITaskItem[] packageItems = ItemsToPush
+                            .Where(i => !symbolItems.Contains(i))
+                            .ToArray();
 
-                        await blobFeedAction.PushToFeed(packageItems, Overwrite);
+                        var packagePaths = packageItems.Select(i => i.ItemSpec);
+
+                        await blobFeedAction.PushToFeed(packagePaths, Overwrite);
                         await PublishToFlatContainerAsync(symbolItems, blobFeedAction);
+
+                        packageArtifacts = ConcatPackageArtifacts(packageArtifacts, packageItems);
+                        blobArtifacts = ConcatBlobArtifacts(blobArtifacts, symbolItems);
+                    }
+
+                    if (!SkipCreateManifest)
+                    {
+                        await PushBuildManifestAsync(blobFeedAction, blobArtifacts, packageArtifacts);
                     }
                 }
             }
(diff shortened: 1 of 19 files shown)


mmitche:
@dagood Do you have a generated sample manifest around?

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -97,5 +197,84 @@ public override bool Execute()
...
+            {
+                virtualPath = virtualPath.Substring(AssetsVirtualDir.Length);
+            }
+            else
+            {
+                // Don't include assets in the manifest if they aren't in "assets/".
+                return null;
+            }
+
+            string[] virtualPathSegments = virtualPath.Split('/');
```
mmitche: Do you need RemoveEmptyEntries here?

dagood:
Here's one that I created during testing:

```xml
<Build Name="anonymous">
  <Package Id="Feed.Initialization.Placeholder" Version="1.0.0-do-not-use" />
  <Dir Name="proj">
    <Dir Name="relative">
      <Blob foo="bar" more="testing" Id="foo.nupkg" />
    </Dir>
  </Dir>
</Build>
```

More "real" ones are in the tests: https://github.com/dotnet/buildtools/pull/1836/files#diff-4590351532b923e82d4a615c7d317a82R115 (besides formatting)

src/Microsoft.DotNet.VersionTools/BuildManifest/Model/BlobArtifactModel.cs
```diff
@@ -0,0 +1,53 @@
...
+        public string Id
+        {
+            get { return Attributes[nameof(Id)]; }
+            set { Attributes[nameof(Id)] = value; }
+        }
+
+        public string VirtualDirectory
+        {
+            get { return string.Join("/", VirtualDirectorySegments); }
+            set { VirtualDirectorySegments = value.Split('/'); }
```
mmitche: RemoveEmptyEntries?

src/Microsoft.DotNet.VersionTools/BuildManifest/Model/BlobArtifactModel.cs
```diff
@@ -0,0 +1,53 @@
...
+        private static string[] ElementVirtualDirSegments(XElement xml)
+        {
+            // Get all ancestor dirs from top to bottom.
+            IEnumerable<XElement> dirs = xml.Ancestors("Dir").Reverse();
+
+            // Get dir name strings, in order. Allow null and empty to avoid crashing if we have
+            // an element like <Dir> or <Dir Name="">. This can happen if someone passes an extra
+            // slash like "foo/bar//baz". The "//" is a valid virtual dir, even though it breaks the
+            // virtual dir metaphor.
+            return dirs.Select(dir => dir.Attribute("Name")?.Value).ToArray();
```
mmitche: What purpose do the empty dirs serve and how could we end up with them, other than bad path processing?  We could just make a hard requirement not to have empty dirs.  There doesn't seem to be a reason that they should be generated by the repos.

src/Microsoft.DotNet.VersionTools/BuildManifest/Model/BuildIdentity.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public class BuildIdentity
+    {
+        public BuildIdentity(
+            string name,
+            string buildId = null,
+            string branch = null,
+            string commit = null)
+        {
+            Name = name;
```
mmitche: Throw ArgumentNullException on a null name here?

src/Microsoft.DotNet.VersionTools/BuildManifest/Model/OrchestratedBuildModel.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        public OrchestratedBuildModel(BuildIdentity identity)
+        {
+            if (identity == null)
+            {
+                throw new ArgumentNullException(nameof(identity));
+            }
+            if (string.IsNullOrEmpty(identity.BuildId))
+            {
+                throw new ArgumentException(
```
mmitche: With ArgumentNullException/empty on the BuildIdentity constructor this check would be duplicate.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -3,18 +3,25 @@
...
+using System.Xml.Linq;
 using MSBuild = Microsoft.Build.Utilities;
 
 namespace Microsoft.DotNet.Build.Tasks.Feed
 {
     public class PushToBlobFeed : MSBuild.Task
     {
+        private static readonly char[] ManifestDataPairSeparators = { ';' };
+        private const string DisableManifestPushConfigurationBlob = "disable-manifest-push";
+        private const string AssetsVirtualDir = "assets/";
```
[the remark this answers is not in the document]
mmitche: If constant, constant should be used in here as much as possible.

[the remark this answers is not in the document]
dagood: There was a bit of confusion earlier about whether things should go in `assets/` or not, and what we want to pass to repos as `AssetsRootUrl`. (Currently we pass `.../assets/` to aspnet for runtime installers and `.../` to corefx/core-setup so they can download symbol packages.) IMO we should standardize and say all addressable flat assets should go in `assets/`, and since that's all we'll have, I planned to chop off the `assets/` part in the manifest because this name is well-known. I have this issue out to standardize the symbol package placement: https://github.com/dotnet/core-eng/issues/2274.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -86,6 +116,76 @@ public override bool Execute()
...
+
+            if (disabledByBlob)
+            {
+                Log.LogMessage(
+                    MessageImportance.Normal,
+                    $"Skipping manifest push: feed has '{DisableManifestPushConfigurationBlob}'.");
+                return;
+            }
+
+            string blobPath = $"assets/orchestration-metadata/manifests/{ManifestName}.xml";
```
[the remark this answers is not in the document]
dagood: I suppose there's no harm in that, although I don't expect anyone to change it--this is a well-known location that the orchestrator harvests to assemble the combined manifest to put on dotnet/versions. I'll change to a parameter with a default and this info in a comment.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
dagood: Obsolete because I'm removing the tree structure, but no, reason was to be compatible with unnamed virtual dirs (`foo//bar` being `'foo'` `''` `'bar'`) if they happen to be pushed.

src/Microsoft.DotNet.VersionTools/BuildManifest/Model/BlobArtifactModel.cs
dagood: Compatibility here could mean the difference between a build being weird but usable vs. tanked. With the change to the flattened approach I'm keeping the compatibility because it would take work to not be compatible, but probably worth adding a warning in `PushToBlobFeed` because it's certainly not good or needed to have empty-name virtual dirs.

mmitche: Sounds good.  I always get worried with string parsing of paths because it's so easy to make mistakes with trailing paths, reconstructing paths, etc.

src/Microsoft.DotNet.VersionTools/BuildManifest/Model/OrchestratedBuildModel.cs
dagood: Yeah, originally I didn't think Builds should require it, just OrchestratedBuilds, but I'll just pass a placeholder in `PushToBlobFeed` if nothing's provided. This won't be a mainline scenario in any case.

src/Microsoft.DotNet.VersionTools/BuildManifest/Model/BuildIdentity.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public class BuildIdentity
+    {
+        public BuildIdentity(
+            string name,
+            string buildId,
+            string branch = null,
+            string commit = null)
+        {
+            if (name == null)
```
mmitche: Should this be IsNullOrEmpty?

dagood: Yeah, no reason to support an empty build name. Changing.

src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs
```diff
@@ -34,6 +41,22 @@ public class PushToBlobFeed : MSBuild.Task
 
         public int UploadTimeoutInMinutes { get; set; } = 5;
 
+        public bool SkipCreateManifest { get; set; }
+
+        public string ManifestName { get; set; } = "anonymous";
```
weshaggard: Why not make this an error if not set instead of defaulting to "anonyomous"? That seems like it will be a costlier fix to try and understand that you needed to pass a name.

weshaggard: OK that seems reasonable to me given the extra context.

PR #1837, opened 2017-12-20
Refuse to create an unnamed virtual directory

dagood:
Surfacing this as an error will make it easier to spot and correct, rather than browsing blob storage.

Use LogError rather than throwing an exception so that every bad item gets an error message.

@natemcmaster


dagood:
New package is available with this change:
`Microsoft.DotNet.Build.Tasks.Feed` `2.1.0-prerelease-02320-01`

PR #1840, opened 2017-12-28
Update OneCore allowed list

danmosemsft:
Phillipe checked this against the OneCore subset and noted 3 dlls ("hosts") that are not actually in it. 

The only one we use is httpapi, which is used by HttpListener. I'm going to leave it here as apparently we're OK with requiring the "EnterpriseNetworking" package is installed on your SKU to use HttpListener.


danmosemsft:
cc @pjmsft

PR #1842, opened 2017-12-29
Add osGroup scenario to placeholder build configurations to FindBestConfiguration task

safern:
Currently the only scenario that was supported was to add a default placeholder configuration (no osgroup) `_netfx` and in debug mode. If we build for Release, the configuration infrastructure would add a -Release or Windows_NT-Release suffix when parsing the build configuration and it will not be ignored. 

With this change it will allow to add placeholder configurations including the osgroup (`_netfx-Windows_NT`) and to fallback to the default targetgroup (`_netfx`). So this way if I add a `_netfx` configuration all the `_netfx` configurations will be ignored when building for that project. If I add `_netcoreapp-Windows_NT`, it will be ignored only in Windows for netcoreapp. 

Contributes to: https://github.com/dotnet/corefx/issues/24903

cc: @weshaggard @danmosemsft

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
@@ -32,7 +32,7 @@ namespace Microsoft.DotNet.Build.Tasks
                 SupportedConfigurations.Where(c => !string.IsNullOrWhiteSpace(c)).Select(c => ConfigurationFactory.ParseConfiguration(c)),
                 Configuration.CompatibleComparer);
 
-            var ignoredBuildConfigurations = new HashSet<string>(SupportedConfigurations.Where(c => c.StartsWith(ConfigurationFactory.NopConfigurationPrefix)));
+            var placeHolderBuildConfigurations = new HashSet<string>(SupportedConfigurations.Where(c => c.StartsWith(ConfigurationFactory.NopConfigurationPrefix)));
 
             var bestConfigurations = new List<ITaskItem>();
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
```diff
@@ -51,7 +51,12 @@ public override bool Execute()
                 }
                 else
                 {
-                    if (ignoredBuildConfigurations.Contains($"{ConfigurationFactory.NopConfigurationPrefix}{bestConfiguration}"))
+                    string targetGroup = bestConfiguration.Values[0].Value; // TargetGroup is in property value index 0
```
weshaggard: While I don't know that this is wrong exactly it feels a little to specific to targetgroup/osgroup which feels a little hacky to me. Is there a more general places to do it in the Configuration class itself?

safern:
That is one way I found I could do it without changing `supportedProjectConfigurations` to be a Dictionary and keeping logic we're currently using, another way I found was: 

```cs
var bestConfiguration = compatibleConfigurations.Select(c => 
{
       var supportedConfiguration = supportedProjectConfigurations.FirstOrDefault(sc => Configuration.CompatibleComparer.Equals(sc, c));
       if (supportedConfiguration != null)
       {
           c.IsPlaceHolderConfiguration = supportedConfiguration.IsPlaceHolderConfiguration;
       }

       return c;
}).FirstOrDefault(c => supportedProjectConfigurations.Contains(c));
```

And with the above described bestConfiguration will have the flag set itself with no extra flag in the picture, but I think the one I pushed into the branch is more efficient.

src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
```diff
@@ -42,7 +40,18 @@ public override bool Execute()
 
                 var compatibleConfigurations = ConfigurationFactory.GetCompatibleConfigurations(buildConfiguration, DoNotAllowCompatibleValues);
 
-                var bestConfiguration = compatibleConfigurations.FirstOrDefault(c => supportedProjectConfigurations.Contains(c));
+                bool isPlaceHolderConfiguration = false;
+                var bestConfiguration = compatibleConfigurations.FirstOrDefault(c =>
+                {
+                    var supportedConfiguration = supportedProjectConfigurations.FirstOrDefault(sc => Configuration.CompatibleComparer.Equals(sc, c));
```
weshaggard: Searching linearly might end up being a perf issue but for now I don't think it it really matters.

safern: I changed to use Dictionary instead to avoid linear search and avoid future issues.

safern:
Thanks @weshaggard

PR #1843, opened 2018-01-02
Allow disabling fxcop analyzers

MichalStrehovsky:
Restores #1350 that got trampled over in e23965130c93c3cc66846efc60f6034e5710ef51.


MichalStrehovsky:
@jkotas @weshaggard PTAL

PR #1845, opened 2018-01-04
When creating the supportedProjectConfigurations dictionary we need to make sure they are distinc elements

safern:
If not we would fail with an ArgumentException that the key is already present if a project duplicates a build configuration for some reason. We're currently hitting that on: 

https://ci3.dot.net/job/dotnet_corefx/job/master/job/windows-TGroup_netcoreapp+CGroup_Release+AGroup_x86+TestOuter_false_prtest/6768/console

PR: https://github.com/dotnet/corefx/pull/26138

cc: @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
@@ -28,9 +28,10 @@ namespace Microsoft.DotNet.Build.Tasks
         {
             LoadConfiguration();
 
-            var supportedProjectConfigurations = new Dictionary<Configuration, Configuration>(
-                SupportedConfigurations.Where(c => !string.IsNullOrWhiteSpace(c)).Distinct(StringComparer.OrdinalIgnoreCase).Select(c => ConfigurationFactory.ParseConfiguration(c)).ToDictionary(c => c, Configuration.CompatibleComparer),
-                Configuration.CompatibleComparer);
+            Dictionary<Configuration, Configuration> supportedProjectConfigurations = SupportedConfigurations.Where(c => !string.IsNullOrWhiteSpace(c))
+                .Distinct(StringComparer.OrdinalIgnoreCase)
+                .Select(c => ConfigurationFactory.ParseConfiguration(c))
+                .ToDictionary(c => c, Configuration.CompatibleComparer);
 
             var bestConfigurations = new List<ITaskItem>();
 
--- a/src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
@@ -29,7 +29,7 @@ namespace Microsoft.DotNet.Build.Tasks
             LoadConfiguration();
 
             var supportedProjectConfigurations = new Dictionary<Configuration, Configuration>(
-                SupportedConfigurations.Where(c => !string.IsNullOrWhiteSpace(c)).Select(c => ConfigurationFactory.ParseConfiguration(c)).ToDictionary(c => c, Configuration.CompatibleComparer),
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/FindBestConfigurations.cs
```diff
@@ -29,7 +29,7 @@ public override bool Execute()
             LoadConfiguration();
 
             var supportedProjectConfigurations = new Dictionary<Configuration, Configuration>(
-                SupportedConfigurations.Where(c => !string.IsNullOrWhiteSpace(c)).Select(c => ConfigurationFactory.ParseConfiguration(c)).ToDictionary(c => c, Configuration.CompatibleComparer),
+                SupportedConfigurations.Where(c => !string.IsNullOrWhiteSpace(c)).Distinct(StringComparer.OrdinalIgnoreCase).Select(c => ConfigurationFactory.ParseConfiguration(c)).ToDictionary(c => c, Configuration.CompatibleComparer),
```
weshaggard: I think it is actually worth adding a check that ensures the configurations are unique as they shouldn't be duplicated anyway, but an error better then what the duplicate key message is giving.

safern: Ok. The .csproj is not having duplicates itself. I need to figure out why it is passing duplicate values to the task. https://github.com/dotnet/corefx/blob/master/src/System.Linq.Expressions/src/Configurations.props

Any thoughts?

weshaggard: We might be concatenating more to them somewhere. I don't see anything obvious. If you cannot figure out quickly I'm fine with your approach as it doesn't really harm anything. 

I wonder if uap and uapaot configurations are being treated as the same because of the comparer and that is causing a duplicate.

safern: After investigating I found the root cause of why duplicate build configurations where found. The reason was because [System.Linq.Expressions](https://github.com/dotnet/corefx/blob/master/src/System.Linq.Expressions/src/System.Linq.Expressions.csproj#L30) has ProjectReferences (I don't know why) instead of regular references, maybe it needs some APIs that are not exposed in the reference assembly. What was causing this is that we had a duplicate ProjectReference and when calling [GetBuildConfigurations](https://github.com/dotnet/corefx/blob/master/buildvertical.targets#L146) for that ProjectReference it was passed as two projects to get the build configurations from.
 It looked something like this:
 
`..\..\System.Diagnostics.Tools\src\System.Diagnostics.Tools.csproj;..\..\System.Diagnostics.Tools\src\System.Diagnostics.Tools.csproj` it was iterating over the ItemGroup and appending the BuildConfigurations twice.

Should we detect duplicate ProjectReferences? 

Also, I debugged with passing duplicate BuildConfigurations to `FindBestConfigurations` task creating a `HashSet` and a `Dictionary` using the `Distinct` method when creating the `Dictionary` and both had the same entries. So using `Distinct` will emulate what `HashSet` was doing. Should we keep it to avoid future issues that are hard to detect where the duplicate is coming from?

Also I tried the uap and uapaot scenario, and both build configurations are added into the dictionary.

weshaggard: Yes go ahead and go with the Distinct approach and go ahead and fix the duplicate project reference as well independent of this issue. I don't think we really need to care about adding extra validation to ensure we don't have duplicate references as it isn't really that big an issue.

safern: I fixed the duplicate project reference already in corefx and was able to consume the new buildtools.

To add Distinct approach would be this PR. Looks good?

safern:
Thanks for the review :)

PR #1849, opened 2018-01-05
Update CCI extensions to understand new C# lang features

weshaggard:
With this change our tools like genapi and apicompat will understand

in params
ref readonly returns
readonly structs
ref structs

Also added apicompat rules to ensure the new annotations are compatible

Addresses https://github.com/dotnet/corefx/issues/25039.

cc @jkotas @jaredpar @VSadov 

These updates were used to find the incompatible annotations and regenerate ref files in PR https://github.com/dotnet/corefx/pull/26175.

### DIFF
--- a/src/ApiCompat/Rules/Compat/TypeCannotChangeClassification.cs
+++ b/src/ApiCompat/Rules/Compat/TypeCannotChangeClassification.cs
@@ -27,10 +27,10 @@ namespace Microsoft.Cci.Differs.Rules
                 return DifferenceType.Changed;
             }
 
-            if (impl.Attributes.HasIsReadOnlyAttribute() && !contract.Attributes.HasIsReadOnlyAttribute())
+            if (contract.Attributes.HasIsReadOnlyAttribute() && !impl.Attributes.HasIsReadOnlyAttribute())
             {
                 differences.AddIncompatibleDifference(this,
-                    "Type '{0}' is marked as readonly in the implementation so it must also be marked readonly in the contract.",
+                    "Type '{0}' is marked as readonly in the contract so it must also be marked readonly in the implementation.",
                     impl.FullName());
 
                 return DifferenceType.Changed;
--- a/src/ApiCompat/Rules/Compat/TypeCannotChangeClassification.cs
+++ b/src/ApiCompat/Rules/Compat/TypeCannotChangeClassification.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Cci.Extensions;
+using Microsoft.Cci.Extensions.CSharp;
 
 namespace Microsoft.Cci.Differs.Rules
 {
@@ -20,8 +21,19 @@ namespace Microsoft.Cci.Differs.Rules
             if (implObjType != contractObjType)
             {
                 differences.AddIncompatibleDifference(this,
-                    "Type '{0}' is a {1} in the implementation but is a {2} in the contract.",
+                    "Type '{0}' is a '{1}' in the implementation but is a '{2}' in the contract.",
                     impl.FullName(), implObjType, contractObjType);
+
+                return DifferenceType.Changed;
+            }
+
+            if (impl.Attributes.HasIsReadOnlyAttribute() && !contract.Attributes.HasIsReadOnlyAttribute())
+            {
+                differences.AddIncompatibleDifference(this,
+                    "Type '{0}' is marked as readonly in the implementation so it must also be marked readonly in the contract.",
+                    impl.FullName());
+
+                return DifferenceType.Changed;
             }
 
             return DifferenceType.Unknown;
@@ -33,7 +45,12 @@ namespace Microsoft.Cci.Differs.Rules
                 return "class";
 
             if (type.IsValueType)
+            {
+                if (type.Attributes.HasIsByRefLikeAttribute())
+                    return "ref struct";
+
                 return "struct";
+            }
 
             if (type.IsInterface)
                 return "interface";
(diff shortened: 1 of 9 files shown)


src/ApiCompat/Rules/Compat/TypeCannotChangeClassification.cs
```diff
@@ -20,8 +21,19 @@ public override DifferenceType Diff(IDifferences differences, ITypeDefinition im
...
+                    "Type '{0}' is a '{1}' in the implementation but is a '{2}' in the contract.",
                     impl.FullName(), implObjType, contractObjType);
+
+                return DifferenceType.Changed;
+            }
+
+            if (impl.Attributes.HasIsReadOnlyAttribute() && !contract.Attributes.HasIsReadOnlyAttribute())
+            {
+                differences.AddIncompatibleDifference(this,
+                    "Type '{0}' is marked as readonly in the implementation so it must also be marked readonly in the contract.",
```
jkotas: I think this should be other way around: 
- readonly in implementation only is ok
- readonly in contract only is not ok

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
```diff
@@ -387,7 +387,24 @@ private static bool ExcludeSpecialAttribute(ICustomAttribute c)
...
                 case "System.Runtime.CompilerServices.ExtensionAttribute": return true;
                 case "System.Runtime.CompilerServices.DynamicAttribute": return true;
-            }
+                case "System.Runtime.CompilerServices.IsByRefLikeAttribute": return true;
+                case "System.Runtime.CompilerServices.IsReadOnlyAttribute": return true;
+                case "System.ObsoleteAttribute":
+                    {
+                        var arg = c.Arguments.OfType<IMetadataConstant>().FirstOrDefault();
+
+                        if (arg != null && arg.Value is string)
```
justinvp: Nit: Could be shortened to:

```c#
case "System.ObsoleteAttribute":
    return c.Arguments.OfType<IMetadataConstant>().FirstOrDefault()?.Value is string argValue &&
        argValue == "Types with embedded references are not supported in this version of your compiler.";
```

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
```diff
@@ -387,7 +387,24 @@ private static bool ExcludeSpecialAttribute(ICustomAttribute c)
...
+                            string argValue = (string)arg.Value;
+
+                            if (argValue == "Types with embedded references are not supported in this version of your compiler.")
+                            {
+                                return true;
+                            }
+                        }
+                        break;
+                    }
+    }
```
justinvp: Nit: Indentation of this `}` looks off.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
weshaggard: I'm not going turn it all the way into a statement like you have but I will use a few of the shortened code fragments.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Attributes.cs
weshaggard: fixed.

src/ApiCompat/Rules/Compat/TypeCannotChangeClassification.cs
weshaggard: Fixed.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs
```diff
@@ -213,19 +225,31 @@ private void WriteMethodBody(IMethodDefinition method)
...
-            else if (!TypeHelper.TypesAreEquivalent(method.Type, method.ContainingTypeDefinition.PlatformType.SystemVoid))
-            {
-                WriteKeyword("throw null;");
+                Write("throw null; ");
             }
 
             WriteSymbol("}");
         }
 
+        private bool NeedsMethodBodyForCompilation(IMethodDefinition method)
```
jkotas: Nit: You can keep it simple and just emit `throw null` all the time. `throw null` vs. empty body has no impact on the binary size - the binary size will be same in both cases.

weshaggard: I guess that is true. I wasn't as concerned wit the binary size as much as I was the source code size. Ideally I would just be empty bodies all the time. Perhaps the new compiler /refonly flag might allow that I should check.

VSadov: " => throw null; "    is not a lot of code :-) 

Allowing empty bodies (or no bodies at all - like in interfaces) was considered, but I am not sure it was done (yet?). 
CC: @jcouv

VSadov:
CC:@jbevain @tmat  @jaredpar  - do we need to do similar changes in the context of cecil and/or metadata reader?

tmat:
I don't think we would need to update Cecil, since it operates on IL and shouldn't have any code that would need to understand C# language features. 
What will need to be updated is decompilers like ILSpy. FYI @sharwell

weshaggard:
test Innerloop Windows_NT Debug

PR #1852, opened 2018-01-05
Include manifest info properties in feed targets

dagood:
Same addition as https://github.com/dotnet/coreclr/pull/15705 and https://github.com/dotnet/corefx/pull/26192. Adds manifest info properties to the `PushToBlobFeed` task call. Core-Setup needs this changed in BuildTools because it gets the targets from the Feeds package. (CoreCLR and CoreFX implement this targets file themselves.)

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Microsoft.DotNet.Build.Tasks.Feed.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Microsoft.DotNet.Build.Tasks.Feed.targets
@@ -63,7 +63,11 @@
                     AccountKey="$(AccountKey)"
                     ItemsToPush="@(_ItemsToPush)"
                     Overwrite="$(PushToBlobFeed_Overwrite)"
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Microsoft.DotNet.Build.Tasks.Feed.targets
```diff
@@ -63,7 +63,11 @@
                     AccountKey="$(AccountKey)"
                     ItemsToPush="@(_ItemsToPush)"
                     Overwrite="$(PushToBlobFeed_Overwrite)"
-                    MaxClients="$(PushToBlobFeed_MaxClients)" />
```
weshaggard: please document them above in the comment.

dagood: Documenting it made me notice that `ManifestName` mismatched `GitHubRepositoryName`, so I changed it to make the behavior effectively the same but allow specifying `ManifestName` directly.

PR #1854, opened 2018-01-09
Create final publish tooling tasks/targets

dagood:
This provides utilities to build final publish steps on. Produced this commit on a partial/testing manifest: https://github.com/dagood/versions/commit/a95f4bb5600b682845aeb856f3455cb2f5c6423e

See https://github.com/dotnet/core-eng/blob/master/Documentation/Orchestrated-Build/Api/publish.md

https://github.com/dotnet/core-eng/issues/2331


src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Final.PackagePublish.targets
```diff
@@ -0,0 +1,63 @@
+<Project ToolsVersion="12.0" DefaultTargets="PublishFinalPackages" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
weshaggard: Is the expectation that this targets file is every only used by the final orchestrated build publish? If so I assume we need a proj file somewhere to pull this in as well. Is there any reason you just didn't include this with that as well instead of in the Feed package?

dagood: Yeah, only for final package publish. At first I actually did have these targets in the publish proj in the orchestration utils VSTS repo--I moved them here to centralize. I was thinking of the maintenance cost of synchronizing buildtools package updates with project file changes. Thinking about it again, changes to the util targets should be changed in backward compatible ways anyway, because they'll be used for various other final publish processes. I'll move these targets back to the VSTS repo.

PR #1859, opened 2018-01-11
Update CSharpWriter to handle private fields in structs

weshaggard:
For compile-time compat, the following rules should work for producing a reference assembly. We drop all private fields, but add back certain synthesized private fields for a value type (struct) as follows:
- If there are any private fields that are or contain any value type members, add a single private field of type int
- If there are any private fields that are or contain any reference type members, add a single private field of type object.
- If the type is generic, then for every type parameter of the type, if there are any private fields that are or contain any members whose type is that type parameter, we add a direct private field of that type.

For more details see issue https://github.com/dotnet/corefx/issues/6185
this blog is helpful as well http://blog.paranoidcoding.com/2016/02/15/are-private-members-api-surface.html

PTAL @jkotas @VSadov @jaredpar @gafter 

To get an idea of what is produced see the updated System.Runtime with https://github.com/dotnet/corefx/commit/a2d4a328c5e6f297ddf93643b8d565cd2c56574c.

### DIFF
--- a/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
+++ b/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
@@ -123,16 +123,16 @@ namespace Microsoft.Cci.Extensions.CSharp
                 var typeToCheck = typesToCheck.Dequeue();
                 visited.Add(typeToCheck);
 
-                if (!typeToCheck.IsValueType)
-                    return true;
-
                 var resolvedType = typeToCheck.ResolvedType;
 
                 // If it is dummy we cannot really check so assume it does because that is will be the most conservative 
                 if (resolvedType is Dummy)
                     return true;
 
-                foreach (var field in resolvedType.Fields)
+                if (resolvedType.IsReferenceType)
+                    return true;
+
+                foreach (var field in resolvedType.Fields.Where(f => !f.IsStatic))
                 {
                     if (!visited.Contains(field.Type))
                     {
--- a/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
+++ b/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
@@ -111,6 +111,39 @@ namespace Microsoft.Cci.Extensions.CSharp
             throw new InvalidOperationException("All enums should have a value__ field!");
         }
 
+        public static bool IsOrContainsReferenceType(this ITypeReference type)
+        {
+            Queue<ITypeReference> typesToCheck = new Queue<ITypeReference>();
+            HashSet<ITypeReference> visited = new HashSet<ITypeReference>();
+
+            typesToCheck.Enqueue(type);
+
+            while (typesToCheck.Count != 0)
+            {
+                var typeToCheck = typesToCheck.Dequeue();
+                visited.Add(typeToCheck);
+
+                if (!typeToCheck.IsValueType)
+                    return true;
+
+                var resolvedType = typeToCheck.ResolvedType;
+
+                // If it is dummy we cannot really check so assume it does because that is will be the most conservative 
+                if (resolvedType is Dummy)
+                    return true;
+
+                foreach (var field in resolvedType.Fields)
+                {
+                    if (!visited.Contains(field.Type))
+                    {
+                        typesToCheck.Enqueue(field.Type);
+                    }
+                }
+            }
+
+            return false;
+        }
+
         public static bool IsConversionOperator(this IMethodDefinition method)
         {
             return (method.IsSpecialName &&
@@ -420,7 +453,7 @@ namespace Microsoft.Cci.Extensions.CSharp
                     if (IsAssembly(baseMethod) && !InSameUnit(baseMethod, method))
                         continue;
 
-                    if (filter != null && !filter.Include(baseMethod))
+                    if (filter != null && !filter.Include(baseMethod.UnWrapMember()))
                         continue;
 
                     // NOTE: Do not check method.IsHiddenBySignature here. C# is *always* hide-by-signature regardless of the metadata flag.
--- a/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
+++ b/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
@@ -132,6 +132,10 @@ namespace Microsoft.Cci.Extensions.CSharp
                 if (resolvedType.IsReferenceType)
                     return true;
 
+                // ByReference<T> is a special type understood by runtime to hold a ref T.
+                if (resolvedType.AreEquivalent("System.ByReference<T>"))
+                    return true;
+
                 foreach (var field in resolvedType.Fields.Where(f => !f.IsStatic))
                 {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
```diff
@@ -111,6 +111,39 @@ public static ITypeReference GetEnumType(this ITypeDefinition type)
...
+                if (!typeToCheck.IsValueType)
+                    return true;
+
+                var resolvedType = typeToCheck.ResolvedType;
+
+                // If it is dummy we cannot really check so assume it does because that is will be the most conservative 
+                if (resolvedType is Dummy)
+                    return true;
+
+                foreach (var field in resolvedType.Fields)
```
VSadov: Just to be sure - are these all fiends or only instance fields? (we should only care about instance fields)

weshaggard: That is all fields. I can filter out the statics.

jkotas:
Does this also check that the implementation is consistent with the contract; or does this just generate?

weshaggard:
>Does this also check that the implementation is consistent with the contract; or does this just generate?

Right now it is just generating. I will see if I can write up an APICompat rule for it but I think that is going to be even more complex to reason about. 

I did compare results between generating with coreclr vs uapaot. It generated the same results with the only difference being https://github.com/dotnet/corefx/pull/26264/files#diff-b3d39e42f75b9efe89d21eb385ed61f6R2001 RuntimeArgumentHandle where CoreCLR had a field and UAPAOT didn't. You can see the UAPAOT generated ref here https://github.com/weshaggard/corefx/commit/ca4a3d3db6ebe789cf38378a2ca230f842bdab7e if you are interested.

weshaggard:
@dotnet-bot test this please

src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
```diff
@@ -128,6 +134,63 @@ public override void Visit(IEnumerable<ITypeDefinitionMember> members)
...
 
+        public override void Visit(ITypeDefinition parentType, IEnumerable<IFieldDefinition> fields)
+        {
+            if (parentType.IsStruct && !_apiOnly)
+            {
+                // For compile-time compat, the following rules should work for producing a reference assembly. We drop all private fields, but add back certain synthesized private fields for a value type (struct) as follows:
+                // - If there are any private fields that are or contain any value type members, add a single private field of type int
+                // - If there are any private fields that are or contain any reference type members, add a single private field of type object.
+                // - If the type is generic, then for every type parameter of the type, if there are any private fields that are or contain any members whose type is that type parameter, we add a direct private field of that type.
+
```
VSadov: "contain any members whose type is that type parameter"   -  I do not understand how this part is handled. 

Are the input "fields" actually a flattened recursive list of all instance fields and fields of struct fields?
Or is there is some other way this recurses into fields of fields?

Do we actually have a case where fields of generic types are contained indirectly.
Like:

```C#
struct ContainsFieldOfT<T>
{
      // no fields of type T contained directly, but there is one nested 1 level deeper
      private Nullable<T> _field;
}

**Comment by jkotas** [comment]
> Right now it is just generating. 

Ok, sounds good.

> RuntimeArgumentHandle where CoreCLR had a field and UAPAOT didn't

I would expect to have more differences. E.g. RuntimeTypeHandle should have object on CoreCLR and IntPtr on CoreRT.

File: src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs (hunk 1 above)
Comment (VSadov) [reply src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs]: I think we should filter out statics. Presence of object static fields does not make a struct managed.

---
In reply to: [160838104](https://github.com/dotnet/buildtools/pull/1859#discussion_r160838104) [](ancestors = 160838104)

File: src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs (hunk 2 above)
Comment (VSadov) [reply src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs]: I see. Makes sense. Thanks!

---
In reply to: [160841720](https://github.com/dotnet/buildtools/pull/1859#discussion_r160841720) [](ancestors = 160841720)

**Comment by weshaggard** [comment]
>I would expect to have more differences. E.g. RuntimeTypeHandle should have object on CoreCLR and int on CoreRT.

RuntimeTypeHandle on CoreRT has a EETypePtr which as a pointer field https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/src/System/EETypePtr.cs#L27 so it gets flagged as being a reference and ends up with an object dummy field as well.

**Comment by jkotas** [comment]
> RuntimeTypeHandle on CoreRT has a EETypePtr which as a pointer 

Unmanaged pointers should be treated as `int`. They should not be treated as `object`.

**Comment by weshaggard** [comment]
@jkotas your right there are a few more differences in UAPAOT implementation. See https://github.com/weshaggard/corefx/commit/71347e71f41a7218a2358386736cd8ddb032739e to see the diff between System.Runtime generated by CoreCLR and then generated by UAPAOT. I think the changes for CoreCLR are the most general and should handle the cases for UAPAOT as well, but let me know if you think there are any differences that are worrisome.

File: src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs [hunk 3]
```diff
@@ -128,6 +134,74 @@ public override void Visit(IEnumerable<ITypeDefinitionMember> members)
...
+                        bool hasRefPrivateField = excludedFields.Any(f => f.Type.IsOrContainsReferenceType());
+                        ITypeReference fieldType = parentType.PlatformType.SystemInt32;
+
+                        // If at least one the private fields contains a reference type then we need to
+                        // set this field type to object or reference field to inform the compiler to block
+                        // taking pointers to this struct because the GC will not track updating those references
+                        if (hasRefPrivateField)
+                            fieldType = parentType.PlatformType.SystemObject;
+
+                        // For primiative types that have a field of their type set the dummy field to that type
```
jkotas: Typo: primiative

src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
```diff
@@ -128,6 +134,74 @@ public override void Visit(IEnumerable<ITypeDefinitionMember> members)
...
+
+                    if (!genericTypedFields.Any())
+                    {
+                        // For definiteassignment checks the compiler needs to know there is a private field
+                        // that has not be initialized so if there are any we need to add a dummy private
+                        // field to help the compiler do its job and error about uninitialized structs
+                        bool hasRefPrivateField = excludedFields.Any(f => f.Type.IsOrContainsReferenceType());
+                        ITypeReference fieldType = parentType.PlatformType.SystemInt32;
+
+                        // If at least one the private fields contains a reference type then we need to
```
jkotas: Missing "of: at least one *of* the private fields ?

jkotas:
> to see the diff between System.Runtime generated by CoreCLR and then generated by UAPAOT

I have fixed differences that were just plain bugs in https://github.com/dotnet/corert/pull/5238

PR #1860, opened 2018-01-11
Update run.cmd and update CI to not call VS2015 any longer

weshaggard:
Also add the dotnet cli env variables to both run.cmd/sh.

cc @dagood


weshaggard:
@dotnet-bot test ci please

PR #1862, opened 2018-01-11
Limit inner iteration count to a max of 2

ahsonkhan:
Helps resolve the timeout issue here: https://github.com/dotnet/corefx/issues/26265

The reason we are seeing this issue with System.Memory Performance tests, in particular, is that they heavily use the InnerIterationCount, and hence running those perf tests, especially in debug, will take longer than 20 minutes.

Increasing the Max Iteration from 1 to 2 as per @jorive's suggestion since the data from the first iteration is ignored.

cc @jorive, @danmosemsft, @KrzysztofCwalina, @joperezr, @Priya91, @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -437,12 +437,14 @@ robocopy /S /NP %EXECUTION_DIR%int\%(Identity)\ %EXECUTION_DIR%native\
       <!-- On Windows, call prevents the test command from making execution end prematurely -->
       <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(TestGCStressLevel)' != ''" Include="set COMPlus_GCStress=$(TestGCStressLevel)"/>
       <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(OuterLoop)' == 'true'" Include="set XUNIT_PERFORMANCE_MIN_ITERATION=1"/>
-      <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(OuterLoop)' == 'true'" Include="set XUNIT_PERFORMANCE_MAX_ITERATION=1"/>
+      <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(OuterLoop)' == 'true'" Include="set XUNIT_PERFORMANCE_MAX_ITERATION=2"/>
+      <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(OuterLoop)' == 'true'" Include="set XUNIT_PERFORMANCE_MAX_ITERATION_INNER_SPECIFIED=2"/>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -437,12 +437,14 @@ robocopy /S /NP %EXECUTION_DIR%int\%(Identity)\ %EXECUTION_DIR%native\
       <!-- On Windows, call prevents the test command from making execution end prematurely -->
       <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(TestGCStressLevel)' != ''" Include="set COMPlus_GCStress=$(TestGCStressLevel)"/>
       <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(OuterLoop)' == 'true'" Include="set XUNIT_PERFORMANCE_MIN_ITERATION=1"/>
-      <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(OuterLoop)' == 'true'" Include="set XUNIT_PERFORMANCE_MAX_ITERATION=1"/>
+      <TestCommandLines  Condition="'$(TargetOS)'=='Windows_NT' and '$(OuterLoop)' == 'true'" Include="set XUNIT_PERFORMANCE_MAX_ITERATION=2"/>
```
joperezr: Can we instead have this to be set to a property that is defaulted to 1 but can be overridden? That way, we can only have it be two on the tests where we need it, but keep it 1 for the rest of the cases.

ahsonkhan: > That way, we can only have it be two on the tests where we need it, but keep it 1 for the rest of the cases.

In what case would we want to keep it at 1?

joperezr: Well, all cases only required 1 iteration before, so I believe we shouldn't change all of them to be 2 now only because of some tests that need to run twice. That is why I would set the default as 1 so most tests will continue to only execute once, and special case the ones we know that need to run 2 times.

ahsonkhan: The change to 2 is not test specific. None of the test require 2 iterations. Changing to 2 allows us to verify the correctness of the test run (i.e. data is being produced), which I believe is part of the intention of running the performance tests briefly as part of outerloop. If that is not important, I could revert all back to 1 (since there isn't any requirement, from the tests perspective, to run more than that).

ahsonkhan: To unblock CI for now, I could keep the change minimal (and revert back to 1 iteration). Would that be preferable?

joperezr: I mean the other reason why I wanted to have it parameterized is that that way we don't require a change in buildtools if we want to change it in the future. I know that it is not test-specific but instead this is a setting that gets set per test project.

PR #1867, opened 2018-01-17
Synchronize manifest updates using feed lock

dagood:
This prevents race conditions for builds that have multiple legs potentially pushing manifest changes.

Diff is much better ignoring whitespace changes: https://github.com/dotnet/buildtools/pull/1867/files?w=1

FYI @johnbeisner this fixes an edge case CLI could hit.

---

Tested:
 * Existing Sleet feed with the lock blob initially unleased or leased. Behavior is as expected: this task will wait for me to free my lease before taking its own lease and continuing.
 * Virtual dir with no blobs at all. I was concerned that Sleet would require a feed to exist to lock it, but it just creates the lock blob if needed. (And *only* the lock blob, no other Sleet feed blobs.)


dagood:
Abandoning: this problem has never actually happened as far as I know, and I don't think the feed package is being consumed from BuildTools after it got copied to Arcade.

PR #1871, opened 2018-01-22
Add Finalize/Copy to Latest task for final publish step

dagood:
Adds `CopyBlobsToLatest` that will be used to finalize both Core-Setup and CLI blobs in product builds. (https://github.com/dotnet/core-eng/issues/2407.) I started with Core-Setup's `FinalizeBuild` with CLI's `CopyBlobsToLatest` as a reference and made changes to generalize. We should be able to use this central implementation in Core-Setup and CLI. We'll need a way to give CLI CloudTestTasks. (Right now, the focus is final publish.)

~`GetBlobProducts` determines what products a set of blobs represents. It's useful to parse this for final publish because we don't have the product version of Core-Setup at this point in the build.~ (Removed: ProductVersion will be passed in from the manifest for final publish.)

Tested on a cut-down manifest in my own `dagood` container on Core-Setup and CLI in `dotnetcli` and `dotnetclichecksums`, and it works as expected.

/cc @johnbeisner

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
@@ -19,6 +19,9 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             "latest.version"
         };
 
+        private static readonly Regex VersionRegex = new Regex(
+            @"(?<version>\d+\.\d+\.\d+)(-(?<prerelease>[^-]+-)?(?<major>\d+)-(?<minor>\d+))?");
+
         [Required]
         public string ContainerName { get; set; }
 
@@ -53,8 +56,6 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
         /// has one finalization attempt, e.g. PipeBuild.
         /// </summary>
         public bool EnableVersionHint { get; set; }
-
-        private Regex _versionRegex = new Regex(@"(?<version>\d+\.\d+\.\d+)(-(?<prerelease>[^-]+-)?(?<major>\d+)-(?<minor>\d+))?");
 
         public override bool Execute()
         {
@@ -102,7 +103,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                     // Delete old version files.
                     GetBlobList(channelDir)
                         .Select(s => s.Replace($"/{ContainerName}/", ""))
-                        .Where(w => _versionRegex.Replace(Path.GetFileName(w), "") == "")
+                        .Where(w => VersionRegex.Replace(Path.GetFileName(w), "") == "")
                         .ToList()
                         .ForEach(f => TryDeleteBlob(f));
 
@@ -140,7 +141,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             return !Log.HasLoggedErrors;
         }
 
-        public bool CopyBlobs(string sourceFolder, string destinationFolder)
+        private bool CopyBlobs(string sourceFolder, string destinationFolder)
         {
             // List of versions that need to be replaced with "latest" when copying blobs.
             var versions = new List<string> { ProductVersion };
@@ -168,12 +169,12 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             return returnStatus;
         }
 
-        public bool TryDeleteBlob(string path)
+        private bool TryDeleteBlob(string path)
         {
             return DeleteBlob(ContainerName, path);
         }
 
-        public void CreateBlobIfNotExists(string path)
+        private void CreateBlobIfNotExists(string path)
         {
             var blobList = GetBlobList(path);
             if (blobList.Count() == 0)
@@ -182,13 +183,13 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             }
         }
 
-        public bool IsLatestSpecifiedVersion(string versionFile)
+        private bool IsLatestSpecifiedVersion(string versionFile)
         {
             var blobList = GetBlobList(versionFile);
             return blobList.Count() != 0;
         }
 
-        public bool DeleteBlob(string container, string blob)
+        private bool DeleteBlob(string container, string blob)
         {
             return DeleteAzureBlob.Execute(
                 AccountName,
@@ -200,7 +201,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                 HostObject);
         }
 
-        public Task<bool> CopyBlobAsync(string sourceBlobName, string destinationBlobName)
+        private Task<bool> CopyBlobAsync(string sourceBlobName, string destinationBlobName)
         {
             return CopyAzureBlobToBlob.ExecuteAsync(
                 AccountName,
@@ -213,7 +214,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                 HostObject);
         }
 
-        public string[] GetBlobList(string path)
+        private string[] GetBlobList(string path)
         {
             return ListAzureBlobs.Execute(
                 AccountName,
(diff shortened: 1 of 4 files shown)


dagood:
Reviewers PTAL, it would be good to start running this regularly on the product builds.

dagood:
New commit is renaming `GetBlobProducts` to `ParseBlobProducts`.

> I assume most of the code is just copied

Mostly--`ParseBlobProducts` is new, and I made some surface changes to `CopyBlobsToLatest` but the deeper logic is the same.

eerhardt:
> It's useful to parse this for final publish because we don't have the product version of Core-Setup at this point in the build.

I'm not sure I understand the `we don't have the product version of Core-Setup at this point in the build.` part.  What point of the build is this in?  Is it after all the builds have completed?

src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
```diff
@@ -29,11 +29,13 @@
     <Compile Include="AzureConnectionStringBuildTask.cs" />
     <Compile Include="AzureBlobLease.cs" />
     <Compile Include="AzureHelper.cs" />
+    <Compile Include="CopyBlobsToLatest.cs" />
```
eerhardt: I'm sure I've asked this question before, but why is this assembly named `CloudTestTasks`?  This doesn't have anything to do with `Test`.

dagood:
> What point of the build is this in? Is it after all the builds have completed?

Yeah, this is for a post-product-build step, where all we have is the manifest. [Example manifest](https://github.com/dotnet/versions/blob/65c23578e86c84950377cda078b791a02bc0c579/build-info/dotnet/product/cli/master/build.xml). For CLI, BuildId = product version, but not for Core-Setup.

src/Microsoft.DotNet.Build.CloudTestTasks/Microsoft.DotNet.Build.CloudTestTasks.csproj
dagood: I have no good answer. Ideally the generic cloud/azure tasks would be split out, IMO.

dagood: Filed https://github.com/dotnet/buildtools/issues/1878 to track this.

src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
```diff
@@ -0,0 +1,231 @@
...
+        [Required]
+        public string ProductVersion { get; set; }
+
+        [Required]
+        public string Channel { get; set; }
+
+        [Required]
+        public string Commit { get; set; }
+
+        public bool Coherent { get; set; }
```
eerhardt: @livarcocc @nguerrera - Do we still need/want this `Coherent` terminology?  I assume with the latest product construction, every build should be "coherent".

nguerrera: I think you're right. In this context, is this always a prodcon build?

src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
```diff
@@ -0,0 +1,231 @@
...
+                    .ToList()
+                    .ForEach(f => TryDeleteBlob(f));
+
+                // Drop the version file signaling such for any race-condition builds (see above comment).
+                CreateBlobIfNotExists(targetVersionFile);
+
+                CopyBlobs(sourceDir, channelDir);
+
+                // Generate the latest version text file
+                string sfxVersion =
```
eerhardt: (nit) `sfx` is probably a copy-paste leftover from core-setup.  It stands for "shared framework".

eerhardt:
>Yeah, this is for a post-product-build step, where all we have is the manifest. Example manifest. For CLI, BuildId = product version, but not for Core-Setup.

I think we should add the product version to the manifest file.  It is a useful piece of information.  Then we don't need `ParseBlobProducts` at all.  We can just get the ProductVersion from the manifest file.

Just a note, from looking at the code in core-setup, it produces multiple different "versions" - shared fx version, host version, and UWP version.  Hopefully the UWP stuff goes to a different repo at some point. But you should possibly think about how to handle multiple product versions.

src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
dagood: At the moment, yes, but the goal is to use this for Core-Setup and *eventually* CLI builds, too--including isolated builds.

So far I was planning on setting Coherent to false for Core-Setup and true for CLI for all prodcon builds.

If the term is going away for all 2.1+ builds, though, makes sense to remove it (and never publish the "latest coherent version" file).

src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
dagood: Ah, yep, changed to `versionText`.

dagood:
> But you should possibly think about how to handle multiple product versions.

How I read it, those products all live in the `ProductVersion` virtual dir, they only have different versions in their filenames. The FinalizeBuild task just replaces their filename versions with `latest`. I handled that here with `FullVersions`--Core-Setup can pass `$(SharedFrameworkNugetVersion);$(UWPCoreRuntimeSdkFullVersion);$(HostVersion)` to get them all replaced.

For now though, all the artifacts that show up in the manifest are `ProductVersion`ed, so prodcon usage doesn't need to pass any `FullVersions`.

dagood:
I reduced this PR to bringing in `CopyBlobsToLatest` plus the `AzureBlobLease` fix. Should be good to go, PTAL.

johnbeisner:
@dagood 
How have you solved the likely collision of SDK assets given that the CLI build number is commit based? 
Scenario: 
1) daily build issues a new build number & publishes assets
2) orchestrated build uses that same build number and tries to publish to the same path

I don't see this is the code; have I missed it?

dagood:
@johnbeisner We just aren't going to run that scenario: isolated builds (the current daily builds) and orchestrated builds are never turned on at the same time for the same branch/channel.

src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
```diff
@@ -0,0 +1,231 @@
...
+                        $"Version '{Commit}' is already published, skipping finalization. " +
+                            $"Hint file: '{targetVersionFile}'");
+
+                    return true;
+                }
+
+                // Delete old version files
+                GetBlobList(channelDir)
+                    .Select(s => s.Replace($"/{ContainerName}/", ""))
+                    .Where(w => _versionRegex.Replace(Path.GetFileName(w), "") == "")
```
dagood: Aha, I didn't understand why this was different between Core-Setup and CLI for a while, but I see now. This conflicts with my change to `string targetVersionFile = $"{channelDir}{Commit}";`, which should be `string targetVersionFile = $"{channelDir}{ProductVersion}"`. Fixing.

Since Core-Setup doesn't use commit-based versions and may build the same commit multiple times in a row, this can't be based on commit for Core-Setup.

Even though it's a bit different, this should work fine for CLI, too, since ProductVersion is based on commit depth.

src/Microsoft.DotNet.Build.CloudTestTasks/CopyBlobsToLatest.cs
```diff
@@ -0,0 +1,242 @@
...
+            finally
+            {
+                Log.LogMessage($"Releasing lease on semaphore blob '{semaphoreBlob}'");
+                blobLease.Release();
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+
+        public bool CopyBlobs(string sourceFolder, string destinationFolder)
```
eerhardt: (minor) Do these methods need to be public?

dagood: Forgot to look at these after the copy--they don't. Fixing.

PR #1873, opened 2018-01-23
Adding UAP RS3 package validation

joperezr:
@ericstj This will just add extra package validation for UAP TFM RS3.

FYI: @weshaggard


joperezr:
@ericstj 15138 is the TFM contained in our packages, but the actual one that is used for restore is uap10.0.16299, do you think that validation should check against that instead?

ericstj:
Validation should match what is used by the shipping VS/project-templates

joperezr:
makes sense, I'll change this then.

PR #1875, opened 2018-01-23
Make CreateTrimDependencyGroups use NuGet.Frameworks

ericstj:
Previously this task used a bunch of custom logic to decide when to copy dependencies.

This has been simplified to the following:
1. For all frameworks supported by the package that have inbox assemblies.
2. Determine which dependency group applies to them.
3. Copy that dependency group to a framework-specific group and remove all inbox assemblies.
4. Remove any redundant dependency groups.

Here are the changes to corefx nuspecs after making this change: https://github.com/ericstj/scratch/commit/1e1daf516d57feeadce446c3c6273ad0238eacc6

Issue: https://github.com/dotnet/corefx/issues/26162

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/CreateTrimDependencyGroups.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/CreateTrimDependencyGroups.cs
@@ -5,6 +5,9 @@
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using NuGet.Frameworks;
+using NuGet.Packaging;
+using NuGet.Packaging.Core;
+using NuGet.Versioning;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -14,6 +17,8 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
 {
     public class CreateTrimDependencyGroups : PackagingTask
     {
+        private const string PlaceHolderDependency = "_._";
+
         [Required]
         public ITaskItem[] Dependencies
         {
@@ -46,8 +51,6 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
             set;
         }
 
-        public bool UseNetPlatform { get; set; }
-
         /* Given a set of available frameworks ("InboxOnTargetFrameworks"), and a list of desired frameworks,
         reduce the set of frameworks to the minimum set of frameworks which is compatible (preferring inbox frameworks. */
         public override bool Execute()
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/test/CreateTrimDependencyGroupsTests.cs
```diff
@@ -311,7 +312,39 @@ public void AddInboxFrameworkGroupsAndDependencies()
...
+                Files = files,
+                Dependencies = dependencies,
+                PackageIndexes = packageIndexes
+            };
+
+            _log.Reset();
+            task.Execute();
+            Assert.Equal(0, _log.ErrorsLogged);
+            Assert.Equal(0, _log.WarningsLogged);
+            Assert.Equal(1, task.TrimmedDependencies.Length);
```
joperezr: NIT: do you have to check the length? Wouldn't Single() throw an exception if that is not the case?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/FrameworkUtilities.cs
```diff
@@ -36,7 +36,8 @@ public static bool IsGenerationMoniker(NuGetFramework dependencyGroup)
         }
         public static bool IsPortableMoniker(NuGetFramework nuGetFramework)
         {
-            return nuGetFramework == null ? false : nuGetFramework.GetShortFolderName().StartsWith("portable-");
+            return nuGetFramework == null ? false : nuGetFramework.Framework == FrameworkConstants.FrameworkIdentifiers.Portable ||
```
joperezr: Wouldn't the starts with already cover this other case? If not, what is the FrameworkConstants.FrameworkIdentifiers.Portable value?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/CreateTrimDependencyGroups.cs
```diff
@@ -215,54 +175,70 @@ private bool IsSupported(NuGetFramework inboxFx, NuGetAssetResolver resolver)
...
+                string versionString = dependency.GetMetadata("AssemblyVersion");
+                if (string.IsNullOrEmpty(versionString))
                 {
-                    version = version.Substring(0, prereleaseIndex);
+                    versionString = dependency.GetMetadata("Version");
+
+                    int prereleaseIndex = versionString.IndexOf('-');
+                    if (prereleaseIndex != -1)
+                    {
+                        versionString = versionString.Substring(0, prereleaseIndex);
```
joperezr: What if '-' is not found?

joperezr: oh nvm I just saw bellow that we default to 0.0.0.0

src/Microsoft.DotNet.Build.Tasks.Packaging/src/CreateTrimDependencyGroups.cs
```diff
@@ -65,127 +68,84 @@ public override bool Execute()
...
-                    Dependencies = dg.ToArray()
-                });
-
-            List<ITaskItem> addedDependencies = new List<ITaskItem>();
-
-            // Exclude any non-portable frameworks that already have specific dependency groups.
-            var frameworksToExclude = dependencyGroups.Select(dg => dg.Framework).Where(fx => !FrameworkUtilities.IsGenerationMoniker(fx));
+            // Retrieve the list of dependency group TFM's
+            var dependencyGroups = Dependencies
+                .Select(i => new TaskItemPackageDependency(i))
```
joperezr: NIT: perhaps choosing better linq variables would make this line a bit more readable? for example using `dependencies` instead of `i`

weshaggard:
I made a few comments/questions on the nuspec diff commit, otherwise looks good.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/FrameworkUtilities.cs
ericstj: no, NuGet started using portable45- and portable46- depending on how we construct the NuGetFramework type (if it's parsed from a string it doesn't get a version, when constructed from a FrameworkName it has the version).  Its potentially a bug that they're including version in portable tfms, but they've been doing it for a couple releases.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/CreateTrimDependencyGroups.cs
ericstj: sure I can fix that.

src/Microsoft.DotNet.Build.Tasks.Packaging/test/CreateTrimDependencyGroupsTests.cs
ericstj: Sure, I just prefer to fail with an assert.

ericstj:
@joperezr I updated those lambda params to be more readable.  @weshaggard I replied to your comments.  Will merge once green.

ericstj:
See updated diff: https://github.com/ericstj/scratch/commit/ce5764c71d1e2f2abb60c0df9e96bdeb4d4e332f

I needed to do this to address the precedence for NETStandard being different for PCLs vs actual frameworks.

PR #1876, opened 2018-01-25
Push manifest summary/readme along with manifest

dagood:
When these changes are used by the utility repo, we'll get readmes in dotnet/versions like [README.md](https://github.com/dagood/versions/blob/905a3d202ecdcd9293a5f8756ffa522a14b79566/build-info/dotnet/product/test-push-manifest/master/README.md) uploaded alongside the orchestrated build manifest.

@shawnro, does that example readme file have all the info you're looking for? Should I add in the table of Core-Setup (Runtime) installer links, too?

The table is copied from https://github.com/dotnet/cli/blob/master/README.md, but the links are build-specific, acquired from the manifest.

I hard-coded the blob to link associations into the task to keep it simple for now. In the future, I think the best approach is to have the CLI repo pass along this info in the manifest so we don't have to associate at all.

https://github.com/dotnet/core-eng/issues/2473

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BuildManifest/WriteOrchestratedBuildManifestSummaryToFile.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BuildManifest/WriteOrchestratedBuildManifestSummaryToFile.cs
@@ -1,0 +1,134 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.DotNet.VersionTools.BuildManifest.Model;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Xml.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks.Feed.BuildManifest
+{
+    public class WriteOrchestratedBuildManifestSummaryToFile : Task
+    {
+        private static readonly Dictionary<string, string> SdkFileEndingLinks = new Dictionary<string, string>
+        {
+            ["alpine.3.6-x64.tar.gz"] = "alpine-3.6-targz",
+            ["linux-x64.tar.gz"] = "linux-targz",
+            ["osx-x64.pkg"] = "osx-installer",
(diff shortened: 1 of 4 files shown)


shawnro:
How hard would it be to add the runtime links?  If it's not easy, it doesn't have to hit the EOW deadline, but I think we should have it eventually.

dagood:
I'll add that now, will take less than an hour for sure. No new concepts, just more of the same code for the Runtime blobs.

dagood:
There are vastly more links in core-setup than cli, and I couldn't figure out good patterns to set up core-setup, so I simplified and went for template files. (The template files are fairly large and repetitive, but you can generate them by copy-pasting from the Core-Setup and CLI readmes and doing a few find-replaces.)

[New example](https://github.com/dagood/versions/blob/79fb01518f8ba7e0f780e306f854b3e73a3c1b16/build-info/dotnet/product/test-push-manifest/master/README.md)

There are a lot of broken Runtime links. Checksums are expected (we're waiting on generating those as a step of its own) but there are others that might need to be investigated--those would be caused by Core-Setup not pushing everything to the blob feed, or something like that. We might not want to include these in the readme for now (easily changed by modifying the template).

The templates will be checked into the DotNet-Orchestration-Utilities repo.

src/Microsoft.DotNet.Build.Tasks.Feed/BuildManifest/WriteOrchestratedBuildManifestSummaryToFile.cs
```diff
@@ -0,0 +1,85 @@
...
+            string contents = System.IO.File.ReadAllText(ManifestFile);
+            OrchestratedBuildModel model = OrchestratedBuildModel.Parse(XElement.Parse(contents));
+            EndpointModel blobFeed = model.Endpoints.First(e => e.IsOrchestratedBlobFeed);
+
+            string feedAssetsRoot = blobFeed.Url.Replace("/index.json", "/assets");
+
+            string sdkProductVersion = model.Builds.First(b => b.Name == "cli").BuildId;
+            string runtimeProductVersion = blobFeed.Artifacts.Blobs
+                .First(b => b.Id.StartsWith("Runtime/"))
+                .Id.Split('/')[1];
```
dagood: This logic that assumes a lot about the Core-Setup paths will be removed once we shuttle product version directly through the manifest. (See https://github.com/dotnet/buildtools/pull/1871.)

PR #1879, opened 2018-01-25
Changes required in order to build corefx on linux using CLI's MSBuild

joperezr:
cc: @weshaggard @ericstj 

These are the changes required in order to build corefx successfully on Linux using the MSBuild inside the CLI. I was able to run both `build.sh` and `build-tests.sh` successfully with these changes. These do not include the cleanup of the MSBuild that we restore on the Tools folder since I haven't verified if that is still needed on other repos. It is worth noting that in order to successfully build corefx with this, we require the SDK to be > 2.1. [Here are the changes](https://github.com/joperezr/corefx/commit/1e99cdb873495965d5d913433de6980c46100826) inside corefx that I needed to do in order to get a successful build.

Once this is merged and repos start ingesting this new version, we can take care of cleaning up all of the things that won't be required in buildtools any longer.

**Repos verified that work after consuming these changes**
 - [X] corefx (Green after [these changes](https://github.com/joperezr/corefx/commit/1e99cdb873495965d5d913433de6980c46100826))
 - [X] coreclr (Green after [these changes](https://github.com/joperezr/coreclr/commit/67ec57c6a852987d62e57896466a57e0d5468f66))
 - [x] standard (Green after [these changes](https://github.com/joperezr/standard/commit/3f0bd350d305bb305e9d14287f00fc76a932ac33))
 - [x] core-setup
 - [x] wcf

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -106,8 +106,13 @@
   <Import Project="IL.targets"
            Condition="'$(MSBuildProjectExtension)' == '.ilproj' AND '$(SkipImportILTargets)'!='true'" />
 
-  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets"
-          Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj'" />
+   <PropertyGroup Condition="'$(RunningOnCore)' == 'true'">
+     <CSharpCoreTargetsPath>$(MSBuildExtensionsPath32)\Roslyn\Microsoft.CSharp.Core.targets</CSharpCoreTargetsPath>
+     <VisualBasicCoreTargetsPath>$(MSBuildExtensionsPath32)\Roslyn\Microsoft.VisualBasic.Core.targets</VisualBasicCoreTargetsPath>
+   </PropertyGroup>
+ 
+ <!--   <Import Project="$(MSBuildThisFileDirectory)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets"
+           Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj' And '$(_IsDesktop)' != 'true'" /> -->
 
   <!--
       workaround file casing issue where it has different casing in different places which fails on linux builds
@@ -118,11 +123,22 @@
     <CSharpTargetsFile Condition="!Exists('$(CSharpTargetsFile)')">$(MSBuildToolsPath)\Microsoft.CSharp.Targets</CSharpTargetsFile>
   </PropertyGroup>
 
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft.CSharp.targets"
+           Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj' and '$(RunningOnCore)' == 'true'" />
+
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets"
+           Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj' and '$(RunningOnCore)' != 'true'" />
+ 
+
   <Import Project="$(CSharpTargetsFile)"
           Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj'" />
 
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft.VisualBasic.targets"
+           Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj' and '$(RunningOnCore)' == 'true'" />
+ 
+
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.VisualBasic.targets"
-          Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj'" />
+          Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj' and '$(RunningOnCore)' != 'true'" />
 
   <Import Project="$(MSBuildToolsPath)\Microsoft.VisualBasic.targets"
           Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj'" />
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -106,8 +106,13 @@
...
            Condition="'$(MSBuildProjectExtension)' == '.ilproj' AND '$(SkipImportILTargets)'!='true'" />
 
-  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets"
-          Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj'" />
+   <PropertyGroup Condition="'$(RunningOnCore)' == 'true'">
+     <CSharpCoreTargetsPath>$(MSBuildExtensionsPath32)\Roslyn\Microsoft.CSharp.Core.targets</CSharpCoreTargetsPath>
+     <VisualBasicCoreTargetsPath>$(MSBuildExtensionsPath32)\Roslyn\Microsoft.VisualBasic.Core.targets</VisualBasicCoreTargetsPath>
+   </PropertyGroup>
+ 
+ <!--   <Import Project="$(MSBuildThisFileDirectory)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets"
```
weshaggard: Just remove this don't just comment it out. By the way if we aren't using the portable targets any more we should probably clean up some of the other related stuff about .NET Portable above.

weshaggard: I see below that we are still using the .NETPortable stuff so we can clean that up as part of the move to the SDK projects.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -106,8 +106,13 @@
   <Import Project="IL.targets"
            Condition="'$(MSBuildProjectExtension)' == '.ilproj' AND '$(SkipImportILTargets)'!='true'" />
 
-  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets"
-          Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj'" />
+   <PropertyGroup Condition="'$(RunningOnCore)' == 'true'">
+     <CSharpCoreTargetsPath>$(MSBuildExtensionsPath32)\Roslyn\Microsoft.CSharp.Core.targets</CSharpCoreTargetsPath>
```
weshaggard: I assume these properties are used somewhere in the common targets?

weshaggard: Should they instead go in the roslyn targets/props file?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -15,6 +15,7 @@ __PACKAGES_DIR=${4:-$__TOOLRUNTIME_DIR}
 __TOOLS_DIR=$(cd "$(dirname "$0")"; pwd -P)
 __MICROBUILD_VERSION=0.2.0
 __PORTABLETARGETS_VERSION=0.1.1-dev
+__ROSLYNCOMPILER_VERSION=2.6.0-beta3-62316-02
```
weshaggard: This version number is still copied too many places. After your change is this the only one that is needed? Should we figure out a way to set this or allow overriding of this property in the repos? Ideally we would only need to set this in one place for all the scenarios.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
joperezr: Right we are still using that, but I'll remove the commented out part.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: Unfortunatelly it is not the only place, we still need it in init-tools.cmd as well since we restore the desktop version of the compilers in there as well. That said, we could use the BuildTools version approach and use a txt file to handle this version

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
joperezr: These properties will be used by the SDK targets since they will import these files. Build.Common.props or Roslyn.Common.props are both too early since if we set this value there, it will get overriden by the props file that are contained on the roslyn package. One place where we can place it other than here, is on Build.Common.targets, do you think that would be a better place for them?

weshaggard: I think this is better than the build.common.targets file.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
weshaggard: Now that we can use msbuild we can also create a project file to share between init-tools.sh/cmd and just call build on each and then we could manage the versions in msbuild files.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
joperezr: Actually looks like we were only setting the targetsfile here because we wanted to override the CoreCompile task to one that accepted the dotnet tool host when we add the /m to the non-Windows build, but this is not required anymore, I tried removing these lines entirely and build seems to be working by using the CSharp and Vb targets that come with the roslyn package.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
joperezr: That is a good point, would it be ok if we do that as part of the cleanup of files we won't need anymore like the tool-runtime?

weshaggard: Yes I'm fine with doing as part of the clean-up

joperezr:
BTW, I tried building with this on coreclr non-Windows and build seems to be fine after some simple changes. @weshaggard is the full list of the repos to check : corefx, coreclr, core-setup, standard? or should I also check other buildtools consumers like buildtools and wcf?

@shawnro after merging this we will force consuming repos to use cli >= 2.1.1. Is that ok?

weshaggard:
I would add wcf to the list of repo's to ensure still work as well. They all have auto-PR's for buildtools uptake.

joperezr:
Hey @weshaggard I have all the before mentioned consumers of buildtools building succesfully with this new package, do you have any additional feedback or is this good to go?

weshaggard:
>Hey @weshaggard I have all the before mentioned consumers of buildtools building succesfully with this new package, do you have any additional feedback or is this good to go?

Do they require updates in the repo to build with these changes?

joperezr:
yup, they all do, pretty much the same changes which are upgrading to a newer version of the CLI + some changes on init-tools.sh but not many.

weshaggard:
Is there any obvious error if someone updates the BuildTools package but not the CLI they are using?

joperezr:
What will happen is that if we don't update the CLI then we won't be able to use the new Roslyn compiler, which means that we will get errors saying that you can't use certain keywords, or attributes might be missing. We made a use of a bunch of these features across our repos beacause of the Span work. Also, these changes will force the use of the actual roslyn props files, which will cause errors on repos that used to use the custom made roslyn tasks  that contain additional properties, like for example:
```
The "OverrideToolHost" parameter is not supported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property.
```

weshaggard:
Alright. Lets do keep track of these changes because we may want to port them to 2.0 branches as well.

PR #1881, opened 2018-01-26
Add aspnet orchestrated build summary

dagood:
I noticed that I got lucky that the first blob this code saw was a core-setup one. I added `EndsWith` to distinguish the dotnet runtime vs. aspnet runtime.

@natemcmaster

PR #1882, opened 2018-01-29
Allow custom attributes in build manifest 'BuildIdentity'

dagood:
Also improve the required attribute mechanism to avoid redundant writes and verify dictionary contents during ToXml.

Core-Setup and CLI will use this to pass `ProductVersion` to the Copy to Latest post-build step (https://github.com/dotnet/buildtools/pull/1871). The repos need to take this change and add `ManifestBuildData="ProductVersion=$(...)"` on `PushToBlobFeed`.

I made `FetchOrchestratedBuildManifestInfo` put the extra info in MSBuild items so I can use it during final publish steps.

ASP.NET Core should pass this too: with all three repos, we can remove some code that derives product version from the blob artifact paths for README generation: https://github.com/dotnet/buildtools/pull/1876#discussion_r164016705, https://github.com/dotnet/buildtools/pull/1881. (I couldn't wait for this data to flow to get the READMEs generating. Copy to Latest is less time-sensitive.)

/cc @johnbeisner @natemcmaster

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BuildManifest/FetchOrchestratedBuildManifestInfo.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BuildManifest/FetchOrchestratedBuildManifestInfo.cs
@@ -40,6 +40,9 @@ namespace Microsoft.DotNet.Build.Tasks.Feed.BuildManifest
 
         [Output]
         public ITaskItem[] OrchestratedBlobFeedArtifacts { get; set; }
+        
+        [Output]
+        public ITaskItem[] OrchestratedBuilds { get; set; }
 
         public override bool Execute()
         {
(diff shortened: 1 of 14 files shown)


natemcmaster:
> ASP.NET Core should pass this too:

You lost me here. What changes do we need to make?

dagood:
Sorry, that's referring to:

> The repos need to take this change and add `ManifestBuildData="ProductVersion=$(...)"` on `PushToBlobFeed`.

It was `2.1.0-preview1-28181` for the last aspnet release/2.1 build, don't know which prop that is.

/cc @pranavkm

natemcmaster:
@dagood thanks. Looks like a simple change on our end

dagood:
@mmitche It's actually even more specific: only repos that need to have their blob assets copied to a "latest" virtual dir in the `dotnetcli`/`dotnetclichecksums` storage accounts. Only Core-Setup, CLI, and ASP.NET will be doing this.

Other repos may pass in ProductVersion to show up on the generated README.md, regardless what they produce, but there's generally no need since packages/blobs are self-descriptive.

PR #1884, opened 2018-01-30
Use project.json dependency versions in VersionTools package

dagood:
The library compiles against 4.3.0 NuGet, and the package should have the same dependency.

### DIFF
--- a/src/nuget/Microsoft.DotNet.VersionTools.nuspec
+++ b/src/nuget/Microsoft.DotNet.VersionTools.nuspec
@@ -18,8 +18,8 @@
       <group targetFramework="netstandard1.5">
         <dependency id="NETStandard.Library" version="1.6.0" />
         <dependency id="Newtonsoft.Json" version="9.0.1" />
-        <dependency id="NuGet.Packaging" version="4.3.0-preview2-4095" />
-        <dependency id="NuGet.Versioning" version="4.3.0-preview2-4095" />
+        <dependency id="NuGet.Packaging" version="4.3.0" />
+        <dependency id="NuGet.Versioning" version="4.3.0" />
         <dependency id="System.Diagnostics.Process" version="4.1.0" />
         <dependency id="System.Diagnostics.TraceSource" version="4.0.0" />
       </group>


src/nuget/Microsoft.DotNet.VersionTools.nuspec
```diff
@@ -18,8 +18,8 @@
       <group targetFramework="netstandard1.5">
         <dependency id="NETStandard.Library" version="1.6.0" />
         <dependency id="Newtonsoft.Json" version="9.0.1" />
-        <dependency id="NuGet.Packaging" version="4.3.0-preview2-4095" />
-        <dependency id="NuGet.Versioning" version="4.3.0-preview2-4095" />
+        <dependency id="NuGet.Packaging" version="4.3.0" />
+        <dependency id="NuGet.Versioning" version="4.3.0" />
```
eerhardt: I'm sure this would be a huge effort, but is there a better way to keep these in sync?  Like for example, using SDK-based `.csproj` and `dotnet pack`?

dagood: That's a goal for sure, would definitely work.

Short-term, we could set up auto-updaters so these stay in sync with https://github.com/dotnet/buildtools/blob/b76c4b6e773f85efa9566b7a0090a8bca98cb57c/dependencies.props#L86-L93. I think we'd have to add a new updater that works on nuspec elements.

dagood: Filed a couple of issues: https://github.com/dotnet/buildtools/issues/1885 (2.x CLI, surprisingly couldn't find this tracked) https://github.com/dotnet/buildtools/issues/1886 (auto-updater).

dagood:
Merging--`Innerloop Ubuntu14.04 Debug` has been waiting in the queue for a while, doesn't have any reason to be affected, and I want to get this unblocked.

src/nuget/Microsoft.DotNet.VersionTools.nuspec
eerhardt: another option is that .nuspec's allow for variables to be passed into them.  I don't know where or how we are building the package, but if we could flow an MSBuild property from dependencies.props into the nupkg packaging, that could work to keep these in sync.

dagood: Hah, I always forget about that. That's a good idea, should just be able to add `NuspecProperties` items in dependencies.props like this one:

https://github.com/dotnet/buildtools/blob/a65f13fbb73db6cf9ea4da276db44f6eb271deeb/src/packages.builds#L19-L21

PR #1888, opened 2018-01-30
Produce 2.2.0-prerelease-* packages

dagood:
I jumped to `2.2.0-prerelease` because BuildTools isn't currently on `preview1`-`preview2` naming, and it's next in the scheme after `2.1.x`, but we could go for that to match the Core repos.

### DIFF
--- a/src/packages.builds
+++ b/src/packages.builds
@@ -13,7 +13,7 @@
 
     <!-- Use the same package version for all output packages. -->
     <DoNotGeneratePackageVersion>true</DoNotGeneratePackageVersion>
-    <PackageVersion>2.1.0-prerelease-$(BuildNumberMajor)-$(BuildNumberMinor)</PackageVersion>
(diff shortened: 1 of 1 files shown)


eerhardt:
I think we should start treating `dotnet/buildtools` like we do the rest of the `dotnet/core*` repos.  The repos are so tied together that I think it makes sense to keep the version numbers in sync.

Can we change `dotnet/buildtoold/release/2.1` to produce `2.1.0-preview1-*` packages and `dotnet/buildtools/master` to produce `2.1.0-preview2-*` packages?  This would make it consistent with the packages we are producing from `dotnet/core*` repos, and I think it would make more sense to people.

Thoughts?

dagood:
I agree we should make it match, I'll submit PRs for this.

I don't see a need to make buildtools' branching strategy a special case. Calling it 2.2.0... implies we won't merge buildtools master back into release/2.1 for preview2, but when core\* merges master into release/2.1, we'll have to do that in buildtools too to resolve breaks.

PR #1894, opened 2018-02-02
Add idempotency options to publish tasks

dagood:
Added `PassIfExistingItemIdentical` to `UploadToAzure` and `PushToBlobFeed`. The `PushToBlobFeed` implementation works on asset and package pushes.

Push to MyGet is a little more complicated. I used the existing system to ignore error messages and added a new `ConditionalIdenticalOnFeed` metadata that takes a NuGet v2 feed endpoint. If it encounters the 409, it checks the contents of the package on the v2 feed. (v2 for now because it's dead simple. It avoids having to hop through the redirects of v3 or attempting to use the NuGet library packages.)

I tested each path with a new artifact publish, two publishes with the same artifacts, and modifying the artifact on the endpoint before a publish. Seems to work as intended. Haven't done any regression testing except accidentally when I didn't plumb the new properties through all the way.

https://github.com/dotnet/core-eng/issues/2449

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
@@ -228,8 +228,8 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                     MessageImportance.Low,
                     $"Downloading blob {destinationBlob} to check if identical.");
 
-                string urlGetBlob = AzureHelper.GetBlobRestUrl(accountName, containerName, destinationBlob);
-                var createRequest = AzureHelper.RequestMessage("GET", urlGetBlob, accountName, accountKey);
+                string blobUrl = AzureHelper.GetBlobRestUrl(accountName, containerName, destinationBlob);
+                var createRequest = AzureHelper.RequestMessage("GET", blobUrl, accountName, accountKey);
 
                 using (HttpResponseMessage response = await AzureHelper.RequestWithRetry(
                     log,
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
@@ -8,10 +8,12 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.DotNet.Build.CloudTestTasks
@@ -208,6 +210,57 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                 }
             }
         }
+
+        public async Task<bool> FileEqualsExistingBlobAsync(
+            string accountName,
+            string accountKey,
+            string containerName,
+            string filePath,
+            string destinationBlob,
+            int uploadTimeout)
+        {
+            using (var client = new HttpClient
+            {
+                Timeout = TimeSpan.FromMinutes(uploadTimeout)
+            })
+            {
+                log.LogMessage(
+                    MessageImportance.Low,
+                    $"Downloading blob {destinationBlob} to check if identical.");
+
+                string urlGetBlob = AzureHelper.GetBlobRestUrl(accountName, containerName, destinationBlob);
+                var createRequest = AzureHelper.RequestMessage("GET", urlGetBlob, accountName, accountKey);
+
+                using (HttpResponseMessage response = await AzureHelper.RequestWithRetry(
+                    log,
+                    client,
+                    createRequest))
+                {
+                    if (!response.IsSuccessStatusCode)
+                    {
+                        throw new HttpRequestException(
+                            $"Failed to retrieve existing blob {destinationBlob}, " +
+                            $"status code {response.StatusCode}.");
+                    }
+
+                    byte[] existingBytes = await response.Content.ReadAsByteArrayAsync();
+                    byte[] localBytes = File.ReadAllBytes(filePath);
+
+                    bool equal = localBytes.SequenceEqual(existingBytes);
+
+                    if (equal)
+                    {
+                        log.LogMessage(
+                            MessageImportance.Normal,
+                            "Item exists in blob storage, and is verified to be identical. " +
+                            $"File: '{filePath}' Blob: '{destinationBlob}'");
+                    }
+
+                    return equal;
+                }
+            }
+        }
+
         private string DetermineContentTypeBasedOnFileExtension(string filename)
         {
             if (Path.GetExtension(filename) == ".svg")
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
@@ -152,7 +152,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             if (!Overwrite && blobsPresent.Contains(relativeBlobPath))
             {
                 if (PassIfExistingItemIdentical &&
-                    await ItemEqualsExistingBlob(item, relativeBlobPath, uploadClient, clientThrottle))
+                    await ItemEqualsExistingBlobAsync(item, relativeBlobPath, uploadClient, clientThrottle))
                 {
                     return;
                 }
@@ -184,7 +184,7 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             }
         }
 
-        private async Task<bool> ItemEqualsExistingBlob(
+        private async Task<bool> ItemEqualsExistingBlobAsync(
             ITaskItem item,
             string relativeBlobPath,
             UploadClient client,
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
@@ -43,6 +43,16 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
         public bool Overwrite { get; set; } = false;
 
         /// <summary>
+        /// Enables idempotency when Overwrite is false.
+        /// 
+        /// false: (default) Attempting to upload an item that already exists fails.
+        /// 
+        /// true: When an item already exists, download the existing blob to check if it's
+        /// byte-for-byte identical to the one being uploaded. If so, pass. If not, fail.
+        /// </summary>
+        public bool PassIfExistingItemIdentical { get; set; }
+
+        /// <summary>
         /// Specifies the maximum number of clients to concurrently upload blobs to azure
         /// </summary>
         public int MaxClients { get; set; } = 8;
@@ -137,8 +147,18 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             if (!File.Exists(item.ItemSpec))
                 throw new Exception(string.Format("The file '{0}' does not exist.", item.ItemSpec));
 
+            UploadClient uploadClient = new UploadClient(Log);
+
             if (!Overwrite && blobsPresent.Contains(relativeBlobPath))
+            {
+                if (PassIfExistingItemIdentical &&
+                    await ItemEqualsExistingBlob(item, relativeBlobPath, uploadClient, clientThrottle))
+                {
+                    return;
+                }
+
                 throw new Exception(string.Format("The blob '{0}' already exists.", relativeBlobPath));
+            }
 
             string contentType = item.GetMetadata("ContentType");
 
@@ -147,7 +167,6 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
             try
             {
                 Log.LogMessage("Uploading {0} to {1}.", item.ItemSpec, ContainerName);
-                UploadClient uploadClient = new UploadClient(Log);
                 await
                     uploadClient.UploadBlockBlobAsync(
                         ct,
@@ -158,6 +177,29 @@ namespace Microsoft.DotNet.Build.CloudTestTasks
                         relativeBlobPath,
                         contentType,
                         UploadTimeoutInMinutes);
+            }
+            finally
+            {
+                clientThrottle.Release();
+            }
+        }
+
+        private async Task<bool> ItemEqualsExistingBlob(
+            ITaskItem item,
+            string relativeBlobPath,
+            UploadClient client,
+            SemaphoreSlim clientThrottle)
+        {
+            await clientThrottle.WaitAsync();
+            try
+            {
+                return await client.FileEqualsExistingBlobAsync(
+                    AccountName,
+                    AccountKey,
+                    ContainerName,
+                    item.ItemSpec,
+                    relativeBlobPath,
+                    UploadTimeoutInMinutes);
             }
             finally
             {
--- a/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
@@ -338,6 +338,8 @@ namespace Microsoft.DotNet.Build.Tasks.Feed
             AzureFileSystem fileSystem = GetAzureFileSystem();
             SleetLogger log = new SleetLogger(Log);
 
+            var packagesToPush = items.ToList();
+
             if (!allowOverwrite && passIfExistingItemIdentical)
             {
                 var context = new SleetContext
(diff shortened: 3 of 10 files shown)


src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -277,12 +329,80 @@ private AzureFileSystem GetAzureFileSystem()
...
+                {
+                    Log.LogMessage("After skipping idempotent uploads, no items need pushing.");
+                    return true;
+                }
+            }
+
+            return await PushCommand.RunAsync(
+                settings,
+                fileSystem,
+                items.ToList(),
```
jcagme: I would return packagesToPublish here instead of items to avoid the ToList() call. In that case you don't need line 390 and need to update 392

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -8,10 +8,12 @@
 using System.Collections.Generic;
```
jcagme: In general I'm a trying to understand the reason behind this PR. Looks like if a blob (name) already exists we check if the contents are identical? In what scenarios this is not the case? When we want to publish a package/asset that has changed but still we want to use the same version?

chcosta: I think that the overall motivation is to not consider a publish attempt a failure when overrite is false if an existing package has already been published and is identical.  ie, the package has already been published and is identical, so don't fail the publish attempt, just ignore it.

@dagood, where are you encountering this scenario?  In the past, effort has been made to ensure that scenario simply does not occur.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -208,6 +210,57 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
...
+            using (var client = new HttpClient
+            {
+                Timeout = TimeSpan.FromMinutes(uploadTimeout)
+            })
+            {
+                log.LogMessage(
+                    MessageImportance.Low,
+                    $"Downloading blob {destinationBlob} to check if identical.");
+
+                string urlGetBlob = AzureHelper.GetBlobRestUrl(accountName, containerName, destinationBlob);
```
chcosta: nit: I don't like the variable name because if I just read it somewhere I would assume that `urlGetBlob` represented a Url type.  Alternatively, putting "Url" as a suffix does not have the same connotation.  Maybe just 'blobUrl'.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -42,6 +42,16 @@ public class UploadToAzure : AzureConnectionStringBuildTask, ICancelableTask
...
 
+        /// <summary>
+        /// Enables idempotency when Overwrite is false.
+        /// 
+        /// false: (default) Attempting to upload an item that already exists fails.
+        /// 
+        /// true: When an item already exists, download the existing blob to check if it's
+        /// byte-for-byte identical to the one being uploaded. If so, pass. If not, fail.
+        /// </summary>
+        public bool PassIfExistingItemIdentical { get; set; }
```
chcosta: Any desire to just make this the default behavior?

src/Microsoft.DotNet.Build.Tasks/ExecWithRetriesForNuGetPush.cs
```diff
@@ -17,9 +20,18 @@ namespace Microsoft.DotNet.Build.Tasks
...
     public class ExecWithRetriesForNuGetPush : BuildTask
     {
+        private const int DownloadComparisonBufferBytes = 1024 * 1024;
+
         [Required]
         public string Command { get; set; }
 
+        /// <summary>
+        /// Package file that is pushed by the given command. Only used if an
+        /// IgnoredErrorMessagesWithConditional item is specified with ConditionalIfV2FeedIdentical
```
chcosta: This "feels" like a complicated way to enable things.  Perhaps it's just because I don't understand where this is being used.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
dagood: That's correct, what changed is that there's a specific scenario in product construction where this is the clearest solution. (I'll copy this to the core-eng issue, too--I thought there was context there but there really wasn't, sorry.)

### Now

When we hit a breaking change, e.g. in CoreCLR breaking CoreFX, prodcon fails and no publishing happens. Since this is in release/2.1, there are no official builds that devs can use to see if they fixed the break, and CI can't succeed because the break fix will actually break CI because it uses the checked-in dependencies.

### Desired

When we hit a breaking change, we want to publish what prodcon built so far. This lets devs easily validate their fix, and it can pass CI. After the fix is merged, we need to *resume* the build at the point of the break. Artifacts from these builds look like this:

| | coreclr | corefx | msbuild | cli  |
| --- | --- | --- | --- | --- |
| broken | y | | y | |
| retry after fix | y | y | y | y |

Publish is a final/post-build step that only sees the set of artifacts that were involved in/produced by a prodcon build. Without some significant new logic in PipeBuild, it has no way to tell if an artifact is new to a certain build, or if it was already published as part of a broken build.

This change fixes that by allowing final publish to just run on all artifacts without tracking which things have been uploaded before.

This is also a pragmatic fix so we can get this scenario running--we *could* track which artifacts have already been published, but we'd have to work where that's stored and what corner cases it has, which is much riskier IMO.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
```diff
@@ -164,5 +183,28 @@ public override bool Execute()
                 clientThrottle.Release();
             }
         }
+
+        private async Task<bool> ItemEqualsExistingBlob(
```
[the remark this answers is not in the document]
dagood: Good call, missed checking for this.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
dagood: Agreed, didn't notice this ToList. Fixing.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -208,6 +210,57 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
...
+
+        public async Task<bool> FileEqualsExistingBlobAsync(
+            string accountName,
+            string accountKey,
+            string containerName,
+            string filePath,
+            string destinationBlob,
+            int uploadTimeout)
+        {
+            using (var client = new HttpClient
```
weshaggard: I'm sure we already have existing code that does similar stuff, we should make sure we re-using it instead of making another copy of it.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
weshaggard: No I think that is a bad idea to make this the default behavior. I fully expect this to negatively impact performance so we should make folks that need it opt-in. I do think we might want to consider other names for the property though, not coming up with a good suggestion right now though.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
jcagme: Ok, now I understand the reason behind the change.  Thanks!

src/Microsoft.DotNet.Build.CloudTestTasks/UploadToAzure.cs
dagood: Even on a functional level, I think it should remain off: it would be a bug in normal cases since we expect everything to have unique versions.

src/Microsoft.DotNet.Build.Tasks/ExecWithRetriesForNuGetPush.cs
dagood: You might be right--it felt to me like a natural way to extend the existing functionality and avoid hard-coding the 409 error message, but I'll revisit this.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
dagood: I'll take another look. I didn't see anything first time around--it looked like the `AzureHelper` methods were the highest-level we had. I did copy those lines from somewhere, so I could just go ahead and centralize it now.

src/Microsoft.DotNet.Build.Tasks/ExecWithRetriesForNuGetPush.cs
dagood: The only thing I could think to do is remove the new metadata and add two new properties instead: the v2 feed to look at and the error message that triggers checking that feed. Barely different at all--maybe I'm mentally locked into thinking about this task this way. Any suggestions?

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
dagood: Here's an attempt to do some refactoring to use more centralized helpers: https://github.com/dagood/buildtools/commit/a52905777c02ff50ebbf01e53501375895dd24a4. Didn't include it in this PR because I'm not sure it's worth testing/reviewing in the context of this (time-sensitive) change. Also, on the backlog we plan to remove the custom Azure code anyway.

dagood:
@jcagme @chcosta @weshaggard PTAL, I fixed some stuff and replied to threads. Thanks for your comments.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
weshaggard: I'm OK not including it yet but I want to make sure we have an issue tracking it.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
```diff
@@ -208,6 +210,57 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
...
+                    {
+                        throw new HttpRequestException(
+                            $"Failed to retrieve existing blob {destinationBlob}, " +
+                            $"status code {response.StatusCode}.");
+                    }
+
+                    byte[] existingBytes = await response.Content.ReadAsByteArrayAsync();
+                    byte[] localBytes = File.ReadAllBytes(filePath);
+
+                    bool equal = localBytes.SequenceEqual(existingBytes);
```
weshaggard: The only time this will ever even have a chance of being byte-for-byte equal is if they have the exact build numbers passed it. Is that going to be the case for your scenarios? If so OK but if it is for a rebuild scenario this comparison will not work. It will only likely work for a republish of the same packages scenario.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
dagood: Thanks, created https://github.com/dotnet/buildtools/issues/1897.

src/Microsoft.DotNet.Build.CloudTestTasks/UploadClient.cs
dagood: Yeah, the only scenario I'm aiming at is package republish. Agreed this wouldn't work for anything else.

dagood:
@weshaggard @chcosta Any more feedback before I merge?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -104,7 +110,12 @@ public BlobFeedAction(string expectedFeedUrl, string accountKey, MSBuild.TaskLog
             return !Log.HasLoggedErrors;
         }
 
-        public async Task UploadAssets(ITaskItem item, SemaphoreSlim clientThrottle, int uploadTimeout, bool allowOverwrite = false)
+        public async Task UploadAssetsAsync(
```
chcosta: nit: should this be "UploadAssetAsync"?  it looks like it uploads a single item rather than multiple.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -277,12 +329,83 @@ private AzureFileSystem GetAzureFileSystem()
...
+                    packagesToPush.RemoveAt(i);
+
+                    if (identical == true)
+                    {
+                        Log.LogMessage(
+                            MessageImportance.Normal,
+                            "Package exists on the feed, and is verified to be identical. " +
+                            $"Skipping upload: '{item}'");
+                    }
+                    else if (identical == false)
```
chcosta: nit: just `else`

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
```diff
@@ -277,12 +329,83 @@ private AzureFileSystem GetAzureFileSystem()
             return fileSystem;
         }
 
-        private async Task<bool> PushAsync(IEnumerable<string> items, bool allowOverwrite)
+        private async Task<bool> PushAsync(
+            IEnumerable<string> items,
+            bool allowOverwrite,
+            bool passIfExistingItemIdentical)
```
chcosta: Not a blocker, but `allowOverwrite` and `passIfExistingItemIdentical` seem like ideal candidates for flags on some `PackageUploadOptions` object.

src/Microsoft.DotNet.Build.Tasks/ExecWithRetriesForNuGetPush.cs
chcosta: The presence of a single `PassIfIdenticalV2Feed` property seems like it would be sufficient.  ProdCon would just set this value.  That might be simpler than adding metadata to every item.  I think both properties is overkill.  Unless there are conditions where some packages want this condition and others do not.

dagood: With only that property and no "the error message that triggers checking that feed" property, we would have to hard-code the error message, right? I've been thinking we want to avoid that so BuildTools users can react more rapidly than taking a BuildTools fix if the error message changes for some reason.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
dagood: Yeah, I'll rename it while I'm in here.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
dagood: This is a `bool?`, so that would change things. Kind of the classic `True` `False` `FileNotFound` scenario. I'll change this to an enum to clear it up--`bool?` doesn't have enough meaning to make this readable.

src/Microsoft.DotNet.Build.Tasks/ExecWithRetriesForNuGetPush.cs
chcosta: I see.  I was thinking it would just fail, then check for identicalness, then pass if identical (regardless of what the error message was).  If the specific error message check is necessary, then that does complicate it.  Do you care about the error message if they're identical?

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
chcosta: `identical` can't be null at this point because of the `if (identical == null) { continue; }` line above.

dagood: Oh, right. :smile: Thanks.

src/Microsoft.DotNet.Build.Tasks/ExecWithRetriesForNuGetPush.cs
dagood: Hmm, the idea was to avoid taking the time to download when there's a different kind of error that seems unrelated. But it's true that if the feed has the package already, the error doesn't matter no matter what kind it is. I agree, I'll make this change.

src/Microsoft.DotNet.Build.Tasks.Feed/BlobFeedAction.cs
dagood: Agreed, created `PushOptions`. Not shared with `Microsoft.DotNet.Build.CloudTestTasks` `UploadToAzure`--there, these params are isolated to the task implementation so the grouping wouldn't have any benefit anyway.

dagood:
Thanks all for the reviews. I'll retest these changes and merge soon.

PR #1900, opened 2018-02-06
Checking Status code for downloading dumpling before executing

Anipik:
Current Error
``` 
 Finished running tests. End time=22:28:24. Return value was 0
  xUnit.net console test runner (64-bit .NET Core)
  Copyright (C) 2014 Outercurve Foundation.
  
    File "/Users/dotnet-bot/j/workspace/dotnet_corefx/master/osx10.12_release/bin/tests/System.Json.Tests/netcoreapp-OSX-Release-x64/dumpling.py", line 1
      <!DOCTYPE html>
      ^
  SyntaxError: invalid syntax
  /usr/bin/python: can't open file '/Users/dotnet-bot/j/workspace/dotnet_corefx/master/osx10.12_release/tempHome/.dumpling/dumpling.py': [Errno 2] No such file or directory
  Finished:    System.Net.Http.Unit.
```
Error After this
```
Dumpling cannot be installed due to Site Disabled and HTTP Status Code 403
```

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/DumplingHelper.py
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/DumplingHelper.py
@@ -1,6 +1,7 @@
 import os
 import platform
 import urllib
+import urllib2
 import glob
 import time
 import sys
@@ -16,10 +17,19 @@ def install_dumpling():
       url = "https://dumpling.azurewebsites.net/api/client/dumpling.py"
       scriptPath = os.path.dirname(os.path.realpath(__file__))
       downloadLocation = scriptPath + "/dumpling.py"
-      urllib.urlretrieve(url, downloadLocation)
-      subprocess.call([sys.executable, downloadLocation, "install", "--update"])
+      response = urllib2.urlopen(url)
+      if response.getcode() == 200:
+        with open(downloadLocation, 'w') as f:
+          f.write(response.read())
+        subprocess.call([sys.executable, downloadLocation, "install", "--update"])
+      else:
+        raise urllib2.URLError("HTTP Status Code" + str(result.getcode()))
 
     subprocess.call([sys.executable, dumplingPath, "install"])
+  except urllib2.HTTPError, e:
+    print("Dumpling cannot be installed due to " + e.reason)
+  except  urllib2.URLError, e:
+    print(e.reason)
   except:
     print("An unexpected error was encountered while installing dumpling.py: " + sys.exc_info()[0])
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/DumplingHelper.py
```diff
@@ -16,10 +17,19 @@ def install_dumpling():
...
+        with open(downloadLocation, 'w') as f:
+          f.write(response.read())
+        subprocess.call([sys.executable, downloadLocation, "install", "--update"])
+      else:
+        raise urllib2.URLError("HTTP Status Code" + str(result.getcode()))
 
     subprocess.call([sys.executable, dumplingPath, "install"])
+  except urllib2.HTTPError, e:
+    print("Dumpling cannot be installed due to " + e.reason + " and HTTP Status Code " + str(e.code))
+  except  urllib2.URLError, e:
```
danmosemsft:
```python
except (urllib2.HTTPError, urllib2.URLError) as e:
   print("Dumpling cannot be installed due to " + e.reason + " and HTTP Status Code " + str(e.code))
```

danmosemsft: I suggest "Dumpling cannot be retrieved from " + url + "  due to " + e.reason + " and HTTP Status Code " + str(e.code))

Anipik: URLError doesnot returns e.code

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/DumplingHelper.py
```diff
@@ -16,10 +17,19 @@ def install_dumpling():
       url = "https://dumpling.azurewebsites.net/api/client/dumpling.py"
       scriptPath = os.path.dirname(os.path.realpath(__file__))
       downloadLocation = scriptPath + "/dumpling.py"
-      urllib.urlretrieve(url, downloadLocation)
-      subprocess.call([sys.executable, downloadLocation, "install", "--update"])
+      response = urllib2.urlopen(url)
+      if response.getcode() == 200:
```
danmosemsft: I can't tell from docs, but it seems like urllib2.urlopen ought to throw when the code is not a success code (2xx). I guess it wasn't?
Anyway this seems fine as far as I can tell

Anipik: It should be the case but there was nothing in documentation about on which error code it urlopen raises an exception that's y I added status code check to make it more robust

PR #1909, opened 2018-02-11
Fix dumplinghelper to not log 'error format'

danmosemsft:
My previous fix solved the problem of HTTPError.reason not always being present by using `str()` instead. This caused another problem: to MSBuild, `Dumpling cannot be installed due to : HTTP error 403: Site Disabled` looks like standard error format, so absent other instructions, it fails the build.

Fix _that_ by removing any colon.


danmosemsft:
Merging to try to get build going.

PR #1922, opened 2018-02-21
Use ChecksumAlgorithm property

tmat:
csc task uses `ChecksumAlgorithm` property to set the checksum algorithm. No need for .rsp file.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -16,7 +16,7 @@
   <Import Project="$(MSBuildThisFileDirectory)RepoAPI.Mapping.props" />
 
   <PropertyGroup>
-    <CompilerResponseFile Condition="'$(CheckSumSHA256)'!='false'">$(MSBuildThisFileDirectory)checksum.rsp</CompilerResponseFile>
+    <ChecksumAlgorithm>SHA256</ChecksumAlgorithm>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


tmat:
@weshaggard

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -16,7 +16,7 @@
   <Import Project="$(MSBuildThisFileDirectory)RepoAPI.Mapping.props" />
 
   <PropertyGroup>
-    <CompilerResponseFile Condition="'$(CheckSumSHA256)'!='false'">$(MSBuildThisFileDirectory)checksum.rsp</CompilerResponseFile>
```
weshaggard: I'm not sure if anyone is setting the condition but could you add the same condition to this property?

weshaggard:
cc @joperezr

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
tmat: I'd rather not add a condition that no one uses. If there is any scenario that requires SHA1 then we need to know about it and make it not require SHA1. If anything gets broken `ChecksumAlgorithm` can be set directly to unblock such scenario temporarily.

joperezr: The previous task library we were using didn't support this argument, which is why we needed to pass in a response file instead. Now that we are using the updated task library we can make use of it.

weshaggard: >I'd rather not add a condition that no one uses. If there is any scenario that requires SHA1 then we need to know about it and make it not require SHA1. If anything gets broken ChecksumAlgorithm can be set directly to unblock such scenario temporarily.

Sounds good and your right there is a temp fallback if needed.

PR #1921, opened 2018-02-21
Allow generating resources cs code at design time

tarekgh:
Fixes https://github.com/dotnet/corefx/issues/26971

PR #1924, opened 2018-02-21
Revert "Allow generating resources cs code at design time"

tarekgh:
Reverts dotnet/buildtools#1921


tarekgh:
reverting the change for now as it broke the build and then will try to fix it otherway.

CC @weshaggard

PR #1926, opened 2018-02-22
Enable Design Time Resource Code Generation

tarekgh:


tarekgh:
@weshaggard could you please have one more look? I hope this will not break the official builds.

Issue #1886, opened 2018-01-30
Base VersionTools package dependencies on dependencies.props


dagood:
If we do this and make sure dependency validation stays on during builds, we would avoid mismatches between the VersionTools project.json and nuspec. (https://github.com/dotnet/buildtools/pull/1884)

This wouldn't take very long to do, but it might be preempted by https://github.com/dotnet/buildtools/issues/1885 eliminating the duplication.

dagood:
We should just flow the properties into the nuspec using `NuspecProperties`.

ViktorHofer:
// Auto-generated message

Going forward, the .NET team is using https://github.com/dotnet/arcade to develop the code and issues formerly in this repository. Feel free to reopen/move this issue if it still applies to servicing branches in this repository, or source code which now resides in the Arcade repository.

PR #1931, opened 2018-02-26
Upgrading NuGet mappings to contain UAP6.0 mapping to Netstandard2.0

joperezr:
cc: @weshaggard @ericstj 

FYI: @jcagme 

Updating nuget mappings on release 2.1 branch so that they have the UAP6.0 <-> NETStandard2.0 mapping.


joperezr:
These changes require some updates on corefx, since package validation will start failing because of this change. I have a branch and will send a PR soon that will fix this issues.

joperezr:
>Any reason you chose 4.4.0 over 4.5?

This one had the mapping I needed and it only had one breaking change that impacted us with a very easy codefix. I tried 4.5 as well but that one had several more breaking changes that would take longer to fix.

weshaggard:
>This one had the mapping I needed and it only had one breaking change

What is the breaking change? Can you point to your necessary changes in corefx?

joperezr:
The breaking change wasn't in corefx, it was in buildtools: https://github.com/dotnet/buildtools/pull/1931/files#diff-0bce2c6566c9b37dced17cae99908b8bR148

The corefx changes for this basically just involve adding uap10.0.16299 to the SupportedFramework metadata on the pkgprojs that have a ref that builds a netstandard configuration, which was expected.

ericstj:
This looks good to me.

We should file issues on those breaking changes in https://github.com/nuget/home.  We can absorb breaking changes ourselves since buildtools is effectively a closed system, but what happens for components that are sharing a NuGet DLL that gets upgraded underneath them (MSBuild, VS, etc).

src/Microsoft.DotNet.VersionTools/project.json
```diff
@@ -2,8 +2,8 @@
   "dependencies": {
     "NETStandard.Library": "1.6.0",
     "Newtonsoft.Json": "9.0.1",
-    "NuGet.Packaging": "4.3.0",
-    "NuGet.Versioning": "4.3.0",
+    "NuGet.Packaging": "4.4.0",
+    "NuGet.Versioning": "4.4.0",
```
dagood: This needs to be updated in the `Microsoft.DotNet.VersionTools` nuspec as well:
https://github.com/dotnet/buildtools/blob/c74c125e5939f8ed964e37f28302d4e3569a9890/src/nuget/Microsoft.DotNet.VersionTools.nuspec#L21-L22

https://github.com/dotnet/buildtools/issues/1886 tracks pulling these versions from `dependencies.props` for the nuspec.

joperezr: oh sorry I missed that one, I'll fix this.

PR #1939, opened 2018-03-05
Don't delete additional directory root

jcagme:
Currently we include directories like "temp", "LocalAppData/Nuget", etc folders in our list of things to be cleaned/removed. The root folder is deleted along with the cleanup and this might lead to build failures as in this issue: https://github.com/dotnet/core-eng/issues/2812

This recreates a "known" folder after the whole contents have been removed.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/CleanupVSTSAgent.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/CleanupVSTSAgent.cs
@@ -334,7 +334,7 @@ namespace Microsoft.DotNet.Build.Tasks
             cleanupTasks.Clear();
             foreach (string additionalCleanupDirectory in _additionalCleanupDirectories)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/CleanupVSTSAgent.cs
```diff
@@ -392,6 +392,13 @@ private DriveInfo ReportCommonDiskUsage(string dirType, string directory)
                     }
 #endif
                     Log.LogMessage("Success");
+
+                    // For additional directories we want only the contents deleted but not the root. The easiest is just to recreate the root folder
+                    // after the whole folder has been created. Since this just applies to additional ('known') folders we should not worry about long paths.
+                    if (isAdditionalDirectory && !Directory.Exists(directory))
```
weshaggard: While this might help with some of the issues I don't think this is going to really help the real race condition. I suspect there are tmp files being written that need to remain as well.

PR #1941, opened 2018-03-05
Add empty Directory.Build.props and targets files 

eerhardt:
so the tool-runtime project doesn't pick up the Directory.Build.props and targets from the calling repo.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/Directory.Build.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/Directory.Build.props
@@ -1,0 +1,6 @@
+<Project>
+  <!--
+  Empty Directory.Build.props to ensure that msbuild doesn't pick up the containing repo's
+  root Directory.Build.props.
+  -->
+</Project>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/Directory.Build.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/Directory.Build.targets
@@ -1,0 +1,6 @@
+<Project>
+  <!--
+  Empty Directory.Build.targets to ensure that msbuild doesn't pick up the containing repo's
+  root Directory.Build.targets.
+  -->
+</Project>


weshaggard:
>so the tool-runtime project doesn't pick up the Directory.Build.props and targets from the calling repo.

I actually think we want it to pick up the repo specific files, in particular the props files. I think we do this for controlling some settings.

eerhardt:
I’m not sure that is a good plan. The Tools should be initialized irrespective of how I set up my props for my repo.  This project shouldn’t just pick up these properties magically.

If you need specific properties set in this project, they should be set in init-Tools.cmd/sh.  

See my other PR that is failing because the version properties aren’t set up yet during init-Tools and is failing the build.  I shouldn’t have to code my repo’s props to workaround BuildTools initialization.

joperezr:
Can you point me to a repo that has these files?

joperezr:
I believe that this doesn't really matter today because we are not using SDK projects on corefx or on most of our repos, but there was an initial push on trying to get corefx libraries to be SDK projects and on that I did have to use the Directory.Build.* files, although that was just a prototype. If Wes is fine with this, we can merge this now and then re-think it once we move to SDK projects.

eerhardt:
@weshaggard - thoughts?

weshaggard:
It feels a little like we are fighting the system here. I can probably live with it but I'd like better understand the exact issue you are hitting. @eerhardt can you at least point out or document what issues you were hitting? Which PR were you working around this?

eerhardt:
I tagged you on the other PR.

Notice that the error is because I'm setting the assembly versions for all the projects being built in the repo.  The version is made up of properties that come from `$(ToolsDir)BuildVersion.targets`, which doesn't exist at the time that the `tool-runtime.csproj` is being published.  Thus an error occurs because the version is not formatted correctly.

`bin\obj\AnyCPU.Debug\project\netcoreapp2.0\project.AssemblyInfo.cs(13,59): error CS7035: The specified version string does not conform to the recommended format - major.minor.build.revision `

The repo setting properties/items/targets for its own projects' use shouldn't affect the `tool-runtime` at all IMO, which is why I proposed this change.

weshaggard:
OK that seems to be a side-effect that we have this tools project a csproj. I can live with this for now, as I expect us to change how we do this anyway.

4creators:
> Can you point me to a repo that has these files?

https://github.com/dotnet/coreclr/blob/master/Directory.Build.props

I am afraid this can break all performance improvements for coreclr native build and we will be hitting 14 GB memory requirement + 15 - 20% additional build time.

We can workaround this but if the actual problem can be solved in other way it would be great.

4creators:
cc @eerhardt @weshaggard

4creators:
@weshaggard Thanks for fast answer

Issue #1940, opened 2018-03-05
No Commit hash goes into InformationVersion in 2.0.0


wtgodbe:
https://github.com/dotnet/buildtools/blob/release/2.0.0/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props#L8 is not getting a commit hash inserted into it, because at the time that versioning.props is imported in versioning.targets (https://github.com/dotnet/buildtools/blob/release/2.0.0/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L27), no value has yet been set for SrcCodeRef (https://github.com/dotnet/buildtools/blob/release/2.0.0/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L295-L303). This appears to have already been fixed in master, we need a similar fix for 2.0.0, as without it, our assemblies are not being stamped with their corresponding commit hashes. @vancem are you already aware of this issue?

CC @eerhardt

vancem:
I was not previously aware of this issue, but I don't tend to look at old versions.   Thanks for the info.

eerhardt:
I don't think we should ship anymore `2.0.x` versions without this.  Adding `blocking release`.

wtgodbe:
https://github.com/dotnet/buildtools/pull/1946 should resolve

wtgodbe:
Resolved by https://github.com/dotnet/buildtools/pull/1946

PR #1946, opened 2018-03-06
Ensure commit hash is stamped on assemblies

wtgodbe:
For issue https://github.com/dotnet/buildtools/issues/1940

After this commit, we have the commit hash in "Product Version" for each .dll produced by a repo that consumes this version of buildtools. I can provide such a .dll (and the contents of buildversion.props) for inspection if desired.

CC @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
@@ -5,7 +5,7 @@
     <Copyright Condition="'$(Copyright)' == ''">%A9 Microsoft Corporation.  All rights reserved.</Copyright>
     <Description Condition="'$(Description)' == ''">$(AssemblyName)</Description>
     <FileVersion Condition="'$(FileVersion)' == '' and '$(AssemblyFileVersion)' != ''">$(AssemblyFileVersion)</FileVersion>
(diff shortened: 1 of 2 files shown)


vancem:
Is this just an undo of the commit?   That is what it looks like.

wtgodbe:
@vancem which commit are you referring to?

vancem:
I am just looking at the delta here, and it looks it is going 'back in time'.  I am just wondering where you got the delta for this PR.   

One option is to  undo the change (since it worked at one point and this is just a servicing branch, we can leave the code 'old').   The other possibility is to figure out why it is not working (since it works in the 2.1 branch).   I am OK with the first approach since it is easy (although you do loose some value, you it does not have the complete URL for the GitHub commit).

wtgodbe:
I got the delta by just copying what we do in master - There, we don't use SrcCodeRef at all, and instead fold its value into BuiltByString - https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props#L8
https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L367-L381

wtgodbe:
It appears what I have here is also the same as what is in the 2.1 branch.

vancem:
@wtgodbe - Great, it looks good to me.

wtgodbe:
I've just sent a shiproom request for this.

PR #1947, opened 2018-03-07
Add HelixAttempt (and MaxRetryCount)

ChadNedzlek:
Cherrypick from master


ChadNedzlek:
@MattGal or @dagood, any thoughts? This is literally just copying the master version of this file.

PR #1948, opened 2018-03-07
Fix dumping stack from test coredumps on Linux

danmosemsft:
- The dumps were in $EXECUTION_DIR. Pushd to that to find them.
- Corerun is someplace else. Use $RUNTIME_PATH/dotnet instead. It's what we actually ran, anyway.
- Add random suffix to keep more than one old dump

Tested locally by using `kill -SIGABRT` on running tests.

Fixes https://github.com/dotnet/corefx/issues/27818


stephentoub:
Initial results:
https://github.com/dotnet/corefx/pull/27899#issuecomment-371901222

Looks like we need gdb installed onto several of the images, and core dumps enabled on others.  But RedHat and CentOS got the expected results :)

PR #1950, opened 2018-03-09
Extract async task to avoid swallowing of TaskCanceledException

riarenas:
Extract async task to avoid swallowing of TaskCanceledException from HttpClient.GetStreamAsync() in DownloadFilesFromUrl


riarenas:
While I get my permissions propagated to assign reviewerd, CC @jcagme since he has context for the issue.

PR #1952, opened 2018-03-10
Fix partial facade PDB generation for AnyOS libraries

bartonjs:
The current check enables PDB generation only when the assembly is
being built as Windows-specific.  What it really means is that the
GenFacades PDB writer crashes on non-Windows systems, and thus that PDBs
cannot be generated when running on non-Windows.

This corrects the check to be about what OS the tool is executing on,
rather than what OS the binary is intended for.

AnyOS partial facades will now produce PDBs on Windows, allowing them
to run code coverage.


bartonjs:
I verified this by editing the copy in Tools on both Windows and Linux and built System.Security.Cryptography.Primitives on both.  Windows produced the PDB, Linux did not.  (And with a bad edit Linux failed the build due to GenFacades erroring out)

Issue #1954, opened 2018-03-13
Fix resources.pri problem in uwp6.1


StephenBonikowsky:
In the WCF repo we recently had to move our release/uwp6.1 branch to buildtools version "2.0.0-uwp61-02602-01", this version of build tools was forked of [a commit from back in July](https://github.com/dotnet/buildtools/commits/release/uwp6.1) which is shortly before a breaking change in CoreClr related to 'resource lookup fallback' necessitated a buildtools updated, see dotnet/wcf#2157.

I believe we need the fix made in buildtools master branch by @safern ported to the buildtools uwp6.1 branch in order to fix our WCF uwp6.1 branch. I'm not sure exactly which changes in buildtools are needed, there are several commits that appear to be related around the same time frame.

@safern could you assist with this?

FYI: @tarekgh @zhenlan @MattGal 

Related commits...
https://github.com/dotnet/buildtools/commit/c45406f85a56c903d02eab582db455afa98dce3d 
https://github.com/dotnet/buildtools/commit/8ecf1097f182c65de6122d729f27ab35d98dffaf 
https://github.com/dotnet/buildtools/commit/579bdc0a7cfb65f99b5e3221cc599c4cf4c1ba71 
https://github.com/dotnet/buildtools/commit/31b3b224a1ea47d5fb6a13384a42147d3d6fe3ea

safern:
I diffed the commits that you currently have in buildtools uwp6.1 branch and I see the following commits missing related to the resources extraction:

https://github.com/dotnet/buildtools/commit/34df12c76df040faa512dd8cc47bd996ad246454
https://github.com/dotnet/buildtools/commit/bbbede8fef122f43180b343ca93d1032904a6cb7#diff-dccabf83e08d3281efb7ebb5798da47a
https://github.com/dotnet/buildtools/commit/1f57d19b4a19c972d7e94050cd839a4b992f39fc#diff-dccabf83e08d3281efb7ebb5798da47a

The other commits that you point out are already in that branch.

Do you need help porting this changes? Could you please confirm this fixes your issue?

StephenBonikowsky:
We know this is the failure we are currently hitting when we use buildtools version "2.0.0-uwp61-02602-01".
I would prefer that you do the ports since this package is being used by CoreClr and Corefx in their uwp6.1 branches and I don't want to do something unintended that breaks them.

StephenBonikowsky:
Fixed by #1955

PR #1955, opened 2018-03-14
Add task to get the resources and merge them into the common pri for …

StephenBonikowsky:
…external dependencies (#1630)

* Add task to get the resources and merge them into the common pri for external dependencies

* PR Feedback

* PR Feedback, remove System.Windows.Forms dependency, and change target Outputs

* Make RewResourceWriter IDisposable

* Fix assembly filtering when it doesn't need resources to be extracted

* Change to NS implementation using MetadataReader instead of Assembly.LoadFrom

* Add option to disable external resources extraction and use runner's default pri file if common pri doesn't exist

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/ExtractResWResourcesFromAssemblies.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/ExtractResWResourcesFromAssemblies.cs
@@ -1,0 +1,328 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Collections;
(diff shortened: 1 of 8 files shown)


StephenBonikowsky:
I don't have permissions to add reviewers. @safern could you review and add folks as necessary?

This fixes #1954 

I built buildtools locally with this change and used it with our WCF uwp6.0 and uwp6.1 branches and verified that it fixes the problem.

safern:
Thanks @StephenBonikowsky for porting this. I've added @tarekgh and @MattGal

safern:
You need to port https://github.com/dotnet/buildtools/commit/bbbede8fef122f43180b343ca93d1032904a6cb7 and https://github.com/dotnet/buildtools/commit/1f57d19b4a19c972d7e94050cd839a4b992f39fc as well because these commits where done specifically for WCF and they impact the resource extractor.

StephenBonikowsky:
Ok, will add them.

PR #1961, opened 2018-03-15
Produce empty dispose methods and not create finalizers when creating a dummy assembly

safern:
Fixes: https://github.com/dotnet/corefx/issues/25605

### DIFF
--- a/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
+++ b/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
@@ -333,6 +333,30 @@ namespace Microsoft.Cci.Extensions.CSharp
             return false;
         }
 
+        public static bool IsDispose(this IMethodDefinition methodDefinition)
+        {
+            if (methodDefinition.Name.Value != "Dispose")
+            {
+                return false;
+            }
+
+            if (methodDefinition.ParameterCount == 0 && methodDefinition.Visibility == TypeMemberVisibility.Public &&
+                TypeHelper.TypesAreEquivalent(methodDefinition.Type, methodDefinition.ContainingTypeDefinition.PlatformType.SystemVoid))
+            {
+                // Dispose() should be public void.
+                return true;
+            }
+
+            if (methodDefinition.ParameterCount == 1 && methodDefinition.Visibility == TypeMemberVisibility.Family && methodDefinition.IsVirtual &&
+                TypeHelper.TypesAreEquivalent(methodDefinition.Parameters.First().Type, methodDefinition.ContainingTypeDefinition.PlatformType.SystemBoolean))
+            {
+                // Dispose(Boolean) should be protected virtual and its only parameter a boolean.
+                return true;
+            }
+
+            return false;
+        }
+
         public static bool IsAssembly(this ITypeDefinitionMember member)
         {
             return member.Visibility == TypeMemberVisibility.FamilyAndAssembly ||
--- a/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
+++ b/src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
@@ -335,26 +335,19 @@ namespace Microsoft.Cci.Extensions.CSharp
 
         public static bool IsDispose(this IMethodDefinition methodDefinition)
         {
-            if (methodDefinition.Name.Value != "Dispose")
+            if ((methodDefinition.Name.Value != "Dispose" && methodDefinition.Name.Value != "System.IDisposable.Dispose") || methodDefinition.ParameterCount > 1 ||
+                !TypeHelper.TypesAreEquivalent(methodDefinition.Type, methodDefinition.ContainingTypeDefinition.PlatformType.SystemVoid))
             {
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
```diff
@@ -333,6 +333,30 @@ public static bool IsDestructor(this IMethodDefinition methodDefinition)
...
+        {
+            if (methodDefinition.Name.Value != "Dispose")
+            {
+                return false;
+            }
+
+            if (methodDefinition.ParameterCount == 0 && methodDefinition.Visibility == TypeMemberVisibility.Public &&
+                TypeHelper.TypesAreEquivalent(methodDefinition.Type, methodDefinition.ContainingTypeDefinition.PlatformType.SystemVoid))
+            {
+                // Dispose() should be public void.
```
weshaggard: This isn't necessarily true. I'm guessing we have some explicit implementations of IDisposable which means the method will not be public nor will it be named Dispose it will be named "IDisposable.Dispose" and private.

weshaggard: I think I would lossen the check to make sure the name is Dispose or IDisposable.Dispose and returns void. That will allow other things it but for our usage should be fine.

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
```diff
@@ -333,6 +333,30 @@ public static bool IsDestructor(this IMethodDefinition methodDefinition)
...
         }
 
+        public static bool IsDispose(this IMethodDefinition methodDefinition)
+        {
+            if (methodDefinition.Name.Value != "Dispose")
+            {
+                return false;
+            }
+
+            if (methodDefinition.ParameterCount == 0 && methodDefinition.Visibility == TypeMemberVisibility.Public &&
```
tarekgh: >f (methodDefinit [](start = 13, length = 16)

Is there anyway (in CCI) to detect if the method is part of IDisposable interface implementation?

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
safern: So you don't think it is worth checking for parameterCount and if ParameterCount == 1 check that the parameter type is boolean either?

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
safern: @weshaggard and I talked about this and we said it was not worth the effort since we would need to traverse the whole tree as it could come from a base class, not necessarily direct implementation.

safern: I think there is a way through the Type object:

https://stackoverflow.com/questions/503263/how-to-determine-if-a-type-implements-a-specific-generic-interface-type

what do you guys think?

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
weshaggard: probably not worth checking the parameters or visibility of the method.

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
tarekgh: I think this should not be much to implement. is it possible someone will have Dispose and not implementing IDisposable?

safern: You can have it, but for our usages I don't think we would have it like that. I think for our convention, having a Dispose method should mean being IDisposable

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
weshaggard: I'm not opposed to checking for the bool parameter if you want to keep it but you should fix name check to support explicit interface implementations and that will require you to remove the visibility check.

safern: Ok.

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
weshaggard: Yes it is possible but for our scenarios I think it is OK to just ensure the method is called Dispose and returns void.

src/Microsoft.Cci.Extensions/Extensions/CSharp/CSharpCciExtensions.cs
weshaggard: thinking about this I think the explicit implementation name is "System.IDisposable.Dispose" but please verify.

safern: Yes, you are right.

safern: Fixed.

PR #1963, opened 2018-03-16
Produce 2.1.0-preview3-* packages

dagood:
This matches what the Core repos are doing in `master` to make room for `release/2.1` starting preview2 builds.

### DIFF
--- a/src/packages.builds
+++ b/src/packages.builds
@@ -13,7 +13,7 @@
 
     <!-- Use the same package version for all output packages. -->
     <DoNotGeneratePackageVersion>true</DoNotGeneratePackageVersion>
-    <PackageVersion>2.1.0-preview2-$(BuildNumberMajor)-$(BuildNumberMinor)</PackageVersion>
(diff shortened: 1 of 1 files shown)


weshaggard:
2.1 merged with https://github.com/dotnet/buildtools/pull/1965 so we can merge this now.

dagood:
This really should have been merged before #1965 though, right? In case there was a different PR merged into `master` causing a build with a version that collides with `release/2.1`'s? Not important in this case because the branches are identical and no merges were done, but for the future.

weshaggard:
It just needs to be merged before we build, which in the case of BuildTools is triggered by commits so I think your correct that we should have merged this into master before merging the release/2.1 PR. For most core repo's that isn't an issue because builds are triggered manually. 

Thanks for this PR and sorry about my duplicated effort. I was just doing the core repo's so I wanted to be sure we got these changes in before they started breaking.

dagood:
I see, and yeah, the difference makes sense.

No problem, I just noticed BT hadn't merged over yet when I was starting some work that I think we'll need in both branches, wanted to check if the merge would be clean, and I ended up with a PR. :smile:

PR #1967, opened 2018-03-19
Add an override so that packages can treat frameworks as out of box.

StephenBonikowsky:
@ericstj This is a port of 1657 attempting to unblock dotnet/wcf#2689.

Discussion in dotnet/wcf#2687 on whether this is the correct or complete solution.


StephenBonikowsky:
Sorry for the delay, we were having a discussion to make sure we were going on the right direction in WCF. Based on that discussion we do need this fix.

StephenBonikowsky:
@ericstj @dagood @MattGal could one of you merge this for me?

PR #1968, opened 2018-03-19
Update submodules from ProdCon orchestrated build manifests

dagood:
This will be used by source-build to move the submodules to the hashes that were successful in the latest orchestrated build.

Setting this up in source-build: https://github.com/dagood/source-build/commit/47918930761a74b6b92f9f432cec8b7839a52596. To avoid duplicating info outside the project files, I used the MSBuild task and Target output flow @eerhardt suggested a while back.

https://github.com/dotnet/source-build/issues/342

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets
@@ -56,7 +56,7 @@
     'TargetSubmodule' property to the name/path of the submodule.
   -->
   <Target Name="CreateDefaultSubmoduleUpdateSteps"
-          Condition="'$(GitModulesPath)' != ''">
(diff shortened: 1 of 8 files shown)


dagood:
Wes had a comment on the source-build commit that I'm going to take a stab at before merging this PR: https://github.com/dagood/source-build/commit/47918930761a74b6b92f9f432cec8b7839a52596#r28179476. (Including an updater for the checked-in commit depths in the project files.)

dagood:
Eric pinged me about that too offline, it's tracked by https://github.com/dotnet/arcade/issues/32.

dagood:
I split off a new issue for the project file hard-coded metadata updater (commit hash, depth): https://github.com/dotnet/source-build/issues/370. The API I wanted isn't available so I think what I end up with will deserve its own PR/review.

Thanks, merging.

PR #1970, opened 2018-03-19
Update GenFacades targets to support all options

weshaggard:
Our genfacades task targets file was missing some options we support
this exposes those as properties.

This change also includes a fix in GenFacades tool to force all
assembly references to 0.0.0.0 when the option is set. Currently it
forces all the seed assemblies to 0.0.0.0 but not any pre-existing
references from our partial facade inputs. This change will force those
to 0.0.0.0 as well.

cc @ericstj

### DIFF
--- a/src/GenFacades/GenFacades.Core/GenFacades.cs
+++ b/src/GenFacades/GenFacades.Core/GenFacades.cs
@@ -119,7 +119,9 @@ namespace GenFacades
                                     partialFacadeAssembly.AssemblyIdentity));
                         }
 
-                        Assembly filledPartialFacade = facadeGenerator.GenerateFacade(contractAssembly, seedCoreAssemblyRef, ignoreMissingTypes, overrideContractAssembly: partialFacadeAssembly);
+                        Assembly filledPartialFacade = facadeGenerator.GenerateFacade(contractAssembly, seedCoreAssemblyRef, ignoreMissingTypes, 
+                            overrideContractAssembly: partialFacadeAssembly, 
+                            forceAssemblyReferenceVersionsToZero: forceZeroVersionSeeds);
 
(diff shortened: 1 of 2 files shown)


weshaggard:
This change is needed to fix https://github.com/dotnet/corefx/issues/28139.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
```diff
@@ -68,8 +68,14 @@
       FacadePath="$(GenFacadesOutputPath.TrimEnd('/'))"
       SeedTypePreferencesUnsplit="@(SeedTypePreference)"
       ProducePdb="$(ProducePdb)"
+      ClearBuildAndRevision="$(GenFacadesClearBuildAndRevisition)"
```
danmosemsft: Typo in property name

src/GenFacades/GenFacades.Core/GenFacades.cs
```diff
@@ -119,7 +119,7 @@ public class Generator
                                     partialFacadeAssembly.AssemblyIdentity));
                         }
 
-                        Assembly filledPartialFacade = facadeGenerator.GenerateFacade(contractAssembly, seedCoreAssemblyRef, ignoreMissingTypes, overrideContractAssembly: partialFacadeAssembly);
+                        Assembly filledPartialFacade = facadeGenerator.GenerateFacade(contractAssembly, seedCoreAssemblyRef, ignoreMissingTypes, overrideContractAssembly: partialFacadeAssembly, forceAssemblyReferenceVersionsToZero: forceZeroVersionSeeds);
```
danmosemsft: Line width?

Issue #1956, opened 2018-03-14
[Helix OS X test failure] ls: /cores: No such file or directory


markwilkie:
@joshfree commented on [Wed Mar 14 2018](https://github.com/dotnet/core-eng/issues/2904)

Infra Execution Failures as part of PR https://github.com/dotnet/corefx/pull/28034

https://mc.dot.net/#/user/weshaggard/pr~2Fjenkins~2Fdotnet~2Fcorefx~2Fmaster~2F/test~2Ffunctional~2Fcli~2F/6813af03e761741b57d0961e4b71df429e6e7c09/workItem/System.Reflection.Metadata.Tests/wilogs

System.Reflection.Metadata.Tests
```
Executed on dci-macpro-07.local
2018-03-14 10:02:10,226: INFO: scriptrunner(33): _main: BuildTools Helix Script Runner v0.1 starting
2018-03-14 10:02:10,226: INFO: helix_test_execution(24): __init__: Copying execution payload files from /Users/dotnet-bot/dotnetbuild/work/b14239dc-1363-462f-938c-356083db3454/Work/e57db0f2-146a-4b12-ab16-d83fac8c7403/Unzip to /Users/dotnet-bot/dotnetbuild/work/b14239dc-1363-462f-938c-356083db3454/Work/e57db0f2-146a-4b12-ab16-d83fac8c7403/Exec/execution
2018-03-14 10:02:10,227: INFO: io(27): copy_tree_to: Copying '/Users/dotnet-bot/dotnetbuild/work/b14239dc-1363-462f-938c-356083db3454/Work/e57db0f2-146a-4b12-ab16-d83fac8c7403/Unzip/DumplingHelper.py' to '/Users/dotnet-bot/dotnetbuild/work/b14239dc-1363-462f-938c-356083db3454/Work/e57db0f2-146a-4b12-ab16-d83fac8c7403/Exec/execution/DumplingHelper.py'
2018-03-14 10:02:10,228: INFO: io(27): copy_tree_to: Copying '/Users/dotnet-bot/dotnetbuild/work/b14239dc-1363-462f-938c-356083db3454/Work/e57db0f2-146a-4b12-ab16-d83fac8c7403/Unzip/RunTests.sh' to '/Users/dotnet-bot/dotnetbuild/work/b14239dc-1363-462f-938c-356083db3454/Work/e57db0f2-146a-4b12-ab16-d83fac8c7403/Exec/execution/RunTests.sh'
[...]

PR #1971, opened 2018-03-20
directory exist check added

Anipik:
Fixes https://github.com/dotnet/buildtools/issues/1956

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
@@ -50,10 +50,10 @@ if [ "$(uname -s)" == "Darwin" ]; then
   # On OS X, we will enable core dump generation only if there are no core 
   # files already in /cores/ at this point. This is being done to prevent
   # inadvertently flooding the CI machines with dumps.
-  if [ -d "/cores" ]; then
-    if [ ! "$(ls -A /cores)" ]; then
-      ulimit -c unlimited
-    fi
+  if [ ! -d "/cores" ]; then
+    ulimit -c unlimited
+  elif [ ! "$(ls -A /cores)" ]; then
+    ulimit -c unlimited
   fi
 elif [ "$(uname -s)" == "Linux" ]; then
   # On Linux, we'll enable core file generation unconditionally, and if a dump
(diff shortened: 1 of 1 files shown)


Anipik:
@danmosemsft can you take a look at this also ?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
```diff
@@ -50,8 +50,10 @@ if [ "$(uname -s)" == "Darwin" ]; then
   # On OS X, we will enable core dump generation only if there are no core 
   # files already in /cores/ at this point. This is being done to prevent
   # inadvertently flooding the CI machines with dumps.
-  if [ ! "$(ls -A /cores)" ]; then 
-    ulimit -c unlimited
+  if [ -d "/cores" ]; then
```
danmosemsft: The original code was checking there are no dump files already. If /core does not exist then there are no dump files already. So it should run ulimit.
As you have wrutten it it will NOT run ulimit if /cores does not exist?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Unix.txt
```diff
@@ -50,10 +50,10 @@ if [ "$(uname -s)" == "Darwin" ]; then
   # On OS X, we will enable core dump generation only if there are no core 
   # files already in /cores/ at this point. This is being done to prevent
   # inadvertently flooding the CI machines with dumps.
-  if [ -d "/cores" ]; then
-    if [ ! "$(ls -A /cores)" ]; then
-      ulimit -c unlimited
-    fi
+  if [ ! -d "/cores" ]; then
+    ulimit -c unlimited
+  elif [ ! "$(ls -A /cores)" ]; then
```
Anipik: The or operator checks the second statement even if the first statement is true so I added else if

danmosemsft:
> The or operator checks the second statement even if the first statement is true so I added else if

I believe you need `[[ ... ]]` to shortcircuit.

Anipik:
yeah u r right if we use ```-o``` as or operator then second statement is checked but if we ```[[ ||  ]]``` then it works. I will have a PR in buildtools repo soon for uploading the dumps. I will change it with that

PR #1977, opened 2018-03-22
Make finding the branch from GIT more robust.

vancem:
Previously I tried to include the GIT branch name in the versioning information.   I used the 
rev-parse command , which worked locally but does not work if you don't have a local branch (which 
is what we do when we build in our automation).   The result is we get a unhelpful 'HEAD' as the 
name of the commit.

Changed this to use 'git describe -all' which should use all the remote branches as well.   I also changed
the logic so that if we don't get something reasonably sane out of 'git describe' I give up and don't put
any branch information in.  

I have tested this locally, but it still may need tweeks.   Still at worst it doesn't include branch information.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
@@ -264,8 +264,8 @@
       Condition="'$(SkipVersionGeneration)' != 'true' AND '$(VersionPropsImported)' != 'true'">
 
     <!-- ############################### -->
-    <Exec Command="git rev-parse --abbrev-ref HEAD 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true" Condition="'$(GitBranchName)' == ''">
-      <Output TaskParameter="ConsoleOutput" PropertyName="GitBranchName" />
+    <Exec Command="git describe --all HEAD 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true" Condition="'$(RawGitBranchName)' == ''">
+      <Output TaskParameter="ConsoleOutput" PropertyName="RawGitBranchName" />
     </Exec>
 
     <!-- ############################### -->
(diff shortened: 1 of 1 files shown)


vancem:
cc: @dagood

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -284,6 +284,17 @@
       <LatestCommit>N/A</LatestCommit>
     </PropertyGroup>
 
+
```
dagood: \<nit> Extra newline, should just be one between sections.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -284,6 +284,17 @@
...
+    <!-- If things match expected patterns, create $(GitBranchName) from $(RawGitBranchName) -->
+    <Message Text="GIT Branch Name as seen by git describe: '$(RawGitBranchName)'" /> 
+    <PropertyGroup Condition="$(RawGitBranchName.StartsWith('he:ads/'))">
```
dagood: Is the `:` in `'he:ads/'` a typo?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -284,6 +284,17 @@
...
+      <GitBranchName>$(RawGitBranchName.SubString(6))</GitBranchName>
+    </PropertyGroup>
+    <PropertyGroup Condition="$(RawGitBranchName.StartsWith('remotes/origin/'))">
```
dagood: Do you think it would be worth it to make this generic rather than assuming the remote's called `origin`?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
vancem: For what it was worth, the first two issues were because I staged the edit early and forgot to stage it again (thus pushed stale things).  Anyway it is fixed now.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
vancem: MSBuild does have search and substitute, so it is not that big of a deal to do this, but I do wonder whether it is interesting.   Build servers are likely to be the simplest case (clone some other repo in the simplest way possible), and thus you will get 'origin'   Given that if we are wrong it is not fatal (no worse then what we have now) I am OK with reacting as needed.  If we know it is going to be an issue, I can do it now...  Do you think we will have a problem?

dagood: That's true, I agree this almost definitely won't be a problem for official builds. It can be fixed for dev builds if somebody cares, but I think that's unlikely.

PR #1978, opened 2018-03-22
Revert "Filter out empty dependencies from promotion consideration"

weshaggard:
This reverts commit e8f73c0d8d8891fddacc344ddda1e702015d5dbd.

This commit causes other issues like https://github.com/dotnet/corefx/issues/26390
we need to revert and find another solution to the original issue
https://github.com/dotnet/corefx/issues/26129

cc @ericstj @joperezr


weshaggard:
>Thanks for fixing. Did we figure out what will we do for the netstandard.library case? 

CoreFX PR incoming.

weshaggard:
https://github.com/dotnet/corefx/pull/28387 is the alternative fix.

PR #1984, opened 2018-03-27
Update RoslynDependencies.OptimizationData to version 2.0.0-rc-61101-17

tmat:


tmat:
@weshaggard

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
```diff
@@ -3,7 +3,7 @@
   <UsingTask TaskName="ZipFileExtractToDirectory" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
-    <OptimizationDataVersion Condition="'$(OptimizationDataVersion)'==''">2.0.0-rc-61101-16</OptimizationDataVersion>
+    <OptimizationDataVersion Condition="'$(OptimizationDataVersion)'==''">2.0.0-rc-61101-17</OptimizationDataVersion>
```
weshaggard: While we can update the version here I think we probably set this version in corefx instead of BuildTools.

tmat: oh.

tmat: Perhaps we should remove this from BuildTools then? The optimization data should only be used by CoreFX repo, rigght?

weshaggard: The idea was to us the infra in other repo's as well but the data would be unique to the library/repo. However to this point no other repo's have used it yet.

tmat: Where would be the best place to set it in CoreFX repo?

weshaggard: I think dependencies.props https://github.com/dotnet/corefx/blob/master/dependencies.props is the best place to list this package id and version.

PR #1892, opened 2018-01-31
Fixing crossgen on core-setup by forcing the implicit NuGet fallback folder

joperezr:
FYI: @weshaggard @eerhardt

PR #1986, opened 2018-03-27
Fix NoWarn warning in builds

BruceForstall:
```
Tools/crossgen.sh: line 21: NoWarn: command not found
```

Simply escape the dollar sign so the shell doesn't interpret it.


BruceForstall:
Introduced with https://github.com/dotnet/buildtools/pull/1892

Another option: don't include the `$(NoWarn)` at all, if that is "more correct".

BruceForstall:
@weshaggard

Issue #1980, opened 2018-03-24
Add graceful error handling to Git branch fetch


dagood:
If the current directory isn't a Git repo (such as during source-build's tarball build), this code ends up putting the Git output's newlines in a C++ string, which gives CoreCLR compile errors:

https://github.com/dotnet/buildtools/blob/2f0dec8eba53942a758910ea7b4cc79e4a7d2c36/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L266-L269

It should handle error exit codes like it does when getting the commit:

https://github.com/dotnet/buildtools/blob/2f0dec8eba53942a758910ea7b4cc79e4a7d2c36/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L271-L285

This will also let us pass in the branch name from source-build if we know it.

weshaggard:
As part of this fix we should change the message importance from high to normal.

vancem:
I don't understand why the original code was causing problems.  The reason for the 'RawGitBranchName was that I only use it as a branch name if it looks like output I expect (namely if it begins with a 'heads' or 'remotes/origin/'    
```
    <!-- If things match expected patterns, create $(GitBranchName) from $(RawGitBranchName) -->
    <Message Importance="high" Text="GIT Branch Name as seen by git describe: '$(RawGitBranchName)'" /> 
    <PropertyGroup Condition="$(RawGitBranchName.StartsWith('heads/'))">
      <GitBranchName>$(RawGitBranchName.SubString(6))</GitBranchName>
    </PropertyGroup>
    <PropertyGroup Condition="$(RawGitBranchName.StartsWith('remotes/origin/'))">
      <GitBranchName>$(RawGitBranchName.SubString(15))</GitBranchName>
    </PropertyGroup>
    <Message Importance="high" Text="GIT Branch Name: '$(GitBranchName)'" /> 
```
On errors, I would have expected it to not machine, and thus leave the GitBranchName empty (which means it should not be shown.

Presumably this did not happen, but I don't understand why my logic above is wrong...

dagood:
I think you're actually right, and I misread the code... I hit this problem before your changes when the git command was getting `GitBranchName` directly, and I was thinking of the matching patterns you added as "extra" ways to try to get a meaningful branch name, rather than filters.

I tried a source-build tarball build of CoreCLR with a BuildTools that contains your change (and doesn't contain my workaround) and it works fine. Sorry about that. :smile:

[...]

PR #1988, opened 2018-03-28
[release/2.1] Merge master into release/2.1

weshaggard:
Reference release/2.1 branch from master for source-build effort.

We minimally need the partial facade fix https://github.com/dotnet/buildtools/commit/24ef27105b36fe11ab6ac1583867b2a3c28c6742 to address source-build issue in corefx. 

cc @dagood


src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -264,8 +264,8 @@
       Condition="'$(SkipVersionGeneration)' != 'true' AND '$(VersionPropsImported)' != 'true'">
 
     <!-- ############################### -->
-    <Exec Command="git rev-parse --abbrev-ref HEAD 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true" Condition="'$(GitBranchName)' == ''">
-      <Output TaskParameter="ConsoleOutput" PropertyName="GitBranchName" />
+    <Exec Command="git describe --all HEAD 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true" Condition="'$(RawGitBranchName)' == ''">
```
dagood: We'll have to use this new variable name in the source-build workaround: [repos/dir.targets#L7-L8](https://github.com/dotnet/source-build/blob/b8d175fa98bb4b52f61a212ee99bfec9327d828b/repos/dir.targets#L7-L8). (BuildTools-side fix tracked by https://github.com/dotnet/buildtools/issues/1980.)

src/packages.builds
```diff
@@ -13,7 +13,7 @@
 
     <!-- Use the same package version for all output packages. -->
     <DoNotGeneratePackageVersion>true</DoNotGeneratePackageVersion>
-    <PackageVersion>2.1.0-preview2-$(BuildNumberMajor)-$(BuildNumberMinor)</PackageVersion>
+    <PackageVersion>2.1.0-preview3-$(BuildNumberMajor)-$(BuildNumberMinor)</PackageVersion>
```
dagood: Would be nice to wait to merge `master` into `release/2.1` again until we start doing preview3 so this change works naturally. For now I guess this has to be reverted.

weshaggard: We plan to do a full branch merge in the core repo's on Friday but still for preview2 so I will revert this one change.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
weshaggard: Thanks for the pointer. Should we just rename the property back to the old name in BuildTools?

weshaggard: Nevermind I think I will just need to update the workaround in source-build.

dagood: Sounds good, I think if you're going to make BuildTools-side changes, might as well fix https://github.com/dotnet/buildtools/issues/1980 fully. (Shouldn't be a very complicated change.)

PR #1989, opened 2018-03-28
Logging added to Init-tools

Anipik:


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
```diff
@@ -56,10 +58,15 @@ if ($packageVersionPropsUrl)
...
+            Write-Host "Failed to download '$packageVersionPropsPath'. Trying again..."
         }
     } while ($success -eq $false);
 
     Write-Host "Downloaded package version props:"
     Get-Content $packageVersionPropsPath
 }
+else
+{
+    Write-Host "'$packageVersionPropsUrl' url does not exist."
```
danmosemsft: if this is false ` if ($packageVersionPropsUrl)` then is it safe to write it out? I assume it is false because it doesn't exist or possibly is whitespace.
I suggest just logging "Environment variable PACKAGEVERSIONPROPSURL is not defined."

Also as a matter of style whenever you have a pattern liek
```
if (something)
{
.... big block of actual work ...
}
else
{
... small block returning early...
}
```

it is often easier to read if you change to
```
if (!something)
{
     ... small block return early...
}
... big block ...
```

This reduces indenting and makes it easier to follow what is going on when you get to the bottom of the code. Basically if you can early return, it is usually better. Rarely is there a reason not to (long ago, it was important to return only at the end of the function to help the compiler)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.ps1
```diff
@@ -56,10 +58,15 @@ if ($packageVersionPropsUrl)
                 $proxyCredentialsRequired = ($we -ne $null -and ([Net.HttpWebResponse]$we.Response).StatusCode -eq [Net.HttpStatusCode]::ProxyAuthenticationRequired)
                 Start-Sleep -Seconds (5 * $retryCount)
                 $retryCount++
-            } 
+            }
+            Write-Host "Failed to download '$packageVersionPropsPath'. Trying again..."
         }
     } while ($success -eq $false);
 
     Write-Host "Downloaded package version props:"
```
danmosemsft: "Successfully downloaded ... etc"

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -34,6 +34,8 @@ __PUBLISH_TFM=netcoreapp2.0
 __INIT_TOOLS_RESTORE_ARGS="--source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json ${__INIT_TOOLS_RESTORE_ARGS:-}"
 __TOOLRUNTIME_RESTORE_ARGS="--source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json ${__INIT_TOOLS_RESTORE_ARGS}"
 
+echo "Checking if the passed arguments are valid paths."
```
danmosemsft: I don't think this adds anything

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -49,6 +51,8 @@ if [ -z "$__TOOLRUNTIME_DIR" ]; then
     exit 1
 fi
 
+echo "The passed arguments are valid paths."
```
danmosemsft: Nor this

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -119,9 +123,19 @@ if [ "$__PACKAGE_VERSION_PROPS_URL" ]; then
...
         wget -q -O "$__PACKAGE_VERSION_PROPS_PATH" "$__PACKAGE_VERSION_PROPS_URL"
+        exit_Code=$?
+        download_Method="wget"
+    fi
+
+    if [ $ret -ne 0 ];then
+        echo "$download_Method returned exit code $exit_Code"
     fi
 
     echo "Downloaded package version props:"
```
danmosemsft: "Successfully downloaded ... etc"

danmosemsft:
fyi @mmitche

danmosemsft:
LGTM

weshaggard:
Does this add a bunch more logging to the console during normal operations?

PR #1992, opened 2018-03-29
detect long running tests and print out warning on console

wfurt:
This is meant to improve our ability to triage hanging tests in CI runs. 

It is also somewhat temporary fix. The whole Visitor class is obsolete in current version on Xunit (2.3.x) and it will need to be re-written when we upgrade. 
In mean time, this should provide some more insight. It won't fore-fail tests (yet) but it should provide more visibility. 

I tested this using private tests with sleep as well as I followed instructions for dog-fooding and I modified one of HTTP tests to have incorrect async await. When I run the test, after 5 minutes I see following console output:

> Discovering: System.Net.Http.Functional.Tests
> Discovered:  System.Net.Http.Functional.Tests
> Starting:    System.Net.Http.Functional.Tests
>    System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientMiniStress.ManyClients_ManyGets [SKIP]
>       Condition(s) not met: \"IsStressModeEnabled\"
>    System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientMiniStress.CreateAndDestroyManyClients [SKIP]
>       Condition(s) not met: **\"IsStressModeEnabled\"
> ...
>   System.Net.Http.Functional.Tests.PlatformHandler_HttpClientHandlerTest.Ctor_ExpectedDefaultPropertyValues_UapPlatform [SKIP]
>       Condition(s) not met: \"IsUap\"
>    System.Net.Http.Functional.Tests.PlatformHandler_HttpClientHandlerTest.SendAsync_RequestVersion20_ResponseVersion20 [SKIP]
>       Condition(s) not met: \"IsWindows10Version1607OrGreater\"
> **WARNING: System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientHandlerTest.ProxyAuth_Digest_Succeeds is running for 348s.**
> WARNING: System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientHandlerTest.ProxyAuth_Digest_Succeeds is running for 408s.
> WARNING: System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientHandlerTest.ProxyAuth_Digest_Succeeds is running for 468s.
> WARNING: System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientHandlerTest.ProxyAuth_Digest_Succeeds is running for 528s.
> 

Note this did not make the time configurable. XUnit 2.3 has some parameters and we may want to watch them.  Since CI kills tests  in ~ 10 minutes, I use 5 delay and than 1 minute probes.

### DIFF
--- a/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
+++ b/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
@@ -73,12 +73,6 @@ namespace Xunit.ConsoleClient
                     Errors = Errors
                 });
 
-            if (watcher != null)
-            {
-            //    watcher.Interrupt();
-                //watcher.Abort();
-            //    watcher = null;
-            }
             runningTests = null;
             clock.Stop();
             return result;
--- a/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
+++ b/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
@@ -17,10 +17,10 @@ namespace Xunit.ConsoleClient
         readonly string defaultDirectory;
         readonly bool showProgress;
         readonly Stopwatch clock;
-        private ConcurrentDictionary<String, long> runningTests;
+        private ConcurrentDictionary<string, long> runningTests;
         private Thread watcher;
-        readonly int longTestMaxMiliseconds = 1_000 * 60 * 5;
-        readonly int longTestCheckMiliseconds = 1_000 * 60;
+        readonly int longTestMaxMilliseconds = 1_000 * 60 * 5;
+        readonly int longTestCheckMilliseconds = 1_000 * 60;
 
 
         public StandardOutputVisitor(object consoleLock,
@@ -37,7 +37,7 @@ namespace Xunit.ConsoleClient
             this.showProgress = showProgress;
 
             this.clock =  new Stopwatch();
-            this.runningTests = new ConcurrentDictionary<String , long>();
+            this.runningTests = new ConcurrentDictionary<string, long>();
         }
 
         protected override bool Visit(ITestAssemblyStarting assemblyStarting)
@@ -48,7 +48,7 @@ namespace Xunit.ConsoleClient
                 Console.WriteLine("Starting:    {0}", Path.GetFileNameWithoutExtension(assemblyFileName));
 
             clock.Start();
-            watcher = new Thread(new ThreadStart(testWatcher));
+            watcher = new Thread(new ThreadStart(TestWatcher));
             watcher.IsBackground = true;
             watcher.Start();
 
@@ -124,7 +124,6 @@ namespace Xunit.ConsoleClient
                 }
             }
 
-
             if (!runningTests.TryAdd(testStarting.Test.DisplayName, clock.ElapsedMilliseconds))
             {
                 lock (consoleLock)
@@ -152,7 +151,7 @@ namespace Xunit.ConsoleClient
                     Console.WriteLine("ERROR: Failed to find {0} in running test set.", testFinished.Test.DisplayName);
                 }
             }
-            if (elapsed > longTestMaxMiliseconds)
+            if (elapsed > longTestMaxMilliseconds)
             {
                 lock (consoleLock)
                 {
@@ -241,13 +240,13 @@ namespace Xunit.ConsoleClient
             }
         }
 
-        private void testWatcher()
+        private void TestWatcher()
         {
             try
             {
                 while (runningTests != null)
                 {
-                    Thread.Sleep(longTestCheckMiliseconds);
+                    Thread.Sleep(longTestCheckMilliseconds);
 
                     if (runningTests == null)
                     {
@@ -255,9 +254,9 @@ namespace Xunit.ConsoleClient
                     }
 
                     long  now = clock.ElapsedMilliseconds;
-                    foreach (KeyValuePair<String, long> pair in runningTests)
+                    foreach (KeyValuePair<string, long> pair in runningTests)
                     {
-                        if (( now - pair.Value) > longTestMaxMiliseconds)
+                        if (( now - pair.Value) > longTestMaxMilliseconds)
                         {
                             lock (consoleLock)
                             {
--- a/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
+++ b/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
@@ -267,7 +267,7 @@ namespace Xunit.ConsoleClient
                         {
                             lock (consoleLock)
                             {
-                                Console.WriteLine("WARNING: {0} is running for {1}ms.", pair.Key, now - pair.Value);
+                                Console.WriteLine("WARNING: {0} is running for {1}s.", pair.Key, (now - pair.Value) / 1000);
                             }
                         }
                     }
--- a/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
+++ b/src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
@@ -1,6 +1,9 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
+using System.Threading;
 using System.Xml.Linq;
 using Xunit.Abstractions;
 
@@ -13,6 +16,12 @@ namespace Xunit.ConsoleClient
         readonly ConcurrentDictionary<string, ExecutionSummary> completionMessages;
         readonly string defaultDirectory;
         readonly bool showProgress;
+        readonly Stopwatch clock;
+        private ConcurrentDictionary<String, long> runningTests;
+        private Thread watcher;
+        readonly int longTestMaxMiliseconds = 1_000 * 60 * 5;
+        readonly int longTestCheckMiliseconds = 1_000 * 60;
+
 
         public StandardOutputVisitor(object consoleLock,
                                      string defaultDirectory,
@@ -26,6 +35,9 @@ namespace Xunit.ConsoleClient
             this.defaultDirectory = defaultDirectory;
             this.completionMessages = completionMessages;
             this.showProgress = showProgress;
+
+            this.clock =  new Stopwatch();
+            this.runningTests = new ConcurrentDictionary<String , long>();
         }
 
         protected override bool Visit(ITestAssemblyStarting assemblyStarting)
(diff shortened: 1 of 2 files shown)


stephentoub:
Thanks, @wfurt.  This is helpful and it's great to see us trying to improve the system, but newer versions of xunit already support doing this implicitly. @danmosemsft, @karelz, where are we on getting off our modified xunit runner and just getting on to the real xunit runner?

karelz:
@stephentoub when we have someone to make the time :( - @wfurt expressed interest, but it proved to be too costly and other priorities (Networking) were higher.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -13,6 +16,12 @@ public class StandardOutputVisitor : XmlTestExecutionVisitor
         readonly ConcurrentDictionary<string, ExecutionSummary> completionMessages;
         readonly string defaultDirectory;
         readonly bool showProgress;
+        readonly Stopwatch clock;
+        private ConcurrentDictionary<String, long> runningTests;
+        private Thread watcher;
+        readonly int longTestMaxMiliseconds = 1_000 * 60 * 5;
```
danmosemsft: both this and below should be const (readonly means "modified only in constructor")

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -13,6 +16,12 @@ public class StandardOutputVisitor : XmlTestExecutionVisitor
...
+        readonly int longTestCheckMiliseconds = 1_000 * 60;
```
danmosemsft: Miliseconds -> Milliseconds

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -104,6 +123,16 @@ protected override bool Visit(ITestStarting testStarting)
                     Console.WriteLine("   {0} [STARTING]", XmlEscape(testStarting.Test.DisplayName));
                 }
             }
+
+
```
danmosemsft: Nit double blank lines

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -116,6 +145,22 @@ protected override bool Visit(ITestFinished testFinished)
                     Console.WriteLine("   {0} [FINISHED] Time: {1}s", XmlEscape(testFinished.Test.DisplayName), testFinished.ExecutionTime);
                 }
             }
+            if (!runningTests.TryRemove(testFinished.Test.DisplayName, out long elapsed))
+            {
+                lock (consoleLock)
```
danmosemsft: Console is internally thread safe for reads/writes I believe. You should not need any lock -- unless you want to write two consecutive lines, and even then you would need all of xunit to participate in the lock.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -195,5 +240,34 @@ void WriteStackTrace(string stackTrace)
                 Console.Error.WriteLine("         {0}", StackFrameTransformer.TransformFrame(stackFrame, defaultDirectory));
             }
         }
+
+        private void testWatcher()
```
danmosemsft: Please use PascalCasing for types, methods, and namespaces

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -195,5 +240,34 @@ void WriteStackTrace(string stackTrace)
...
+                while (runningTests != null)
+                {
+                    Thread.Sleep(longTestCheckMiliseconds);
+
+                    if (runningTests == null)
+                    {
+                        break;
+                    }
+
+                    long  now = clock.ElapsedMilliseconds;
```
danmosemsft: Nit - spacing. also on 260

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -26,6 +35,9 @@ public class StandardOutputVisitor : XmlTestExecutionVisitor
             this.defaultDirectory = defaultDirectory;
             this.completionMessages = completionMessages;
             this.showProgress = showProgress;
+
+            this.clock =  new Stopwatch();
+            this.runningTests = new ConcurrentDictionary<String , long>();
```
danmosemsft: nit spacing

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -195,5 +240,34 @@ void WriteStackTrace(string stackTrace)
...
+                        {
+                            lock (consoleLock)
+                            {
+                                Console.WriteLine("WARNING: {0} is running for {1}s.", pair.Key, (now - pair.Value) / 1000);
+                            }
+                        }
+                    }
+                }
+            }
+            catch { };
```
danmosemsft: If you are writing errors to console you could write exception messages here.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -35,6 +47,11 @@ protected override bool Visit(ITestAssemblyStarting assemblyStarting)
             lock (consoleLock)
                 Console.WriteLine("Starting:    {0}", Path.GetFileNameWithoutExtension(assemblyFileName));
 
+            clock.Start();
+            watcher = new Thread(new ThreadStart(testWatcher));
```
danmosemsft: The "modern" way would probably be to use a Task, with TaskCreationOptions.LongRunning so it didn't use the thread pool. But this is fine.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -13,6 +16,12 @@ public class StandardOutputVisitor : XmlTestExecutionVisitor
         readonly ConcurrentDictionary<string, ExecutionSummary> completionMessages;
         readonly string defaultDirectory;
         readonly bool showProgress;
+        readonly Stopwatch clock;
+        private ConcurrentDictionary<String, long> runningTests;
```
danmosemsft: nit we use lower case for string used as a type

danmosemsft:
BTW CI timeout for the entire test job is 20 mins. Most libraries finish in 1-2 mins max. So a long running test would have to be nearly 20 mins to get cut off.

Did any get logged - without your test change?

wfurt:
What I see in CI Console logs are just skips and than nothing until whole set get killed. 
My long term hope would be ability to fail tests directly as well as possibly expose this to test sets somehow. 
Also note on the Task note: I did not want to use task or thread pool. With speculations about us hitting limits or locking there, I want something 100% independent. As I mentioned, XUnit upgrade will give us chance to re-factor if useful.

weshaggard:
I share a similar concern to @stephentoub that we should instead be focusing energy on moving off our custom runner and instead using the one from xunit itself. While I'm not opposed to moving forward with this PR I do think it is better we spend our time moving off this runner.

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
```diff
@@ -13,6 +16,12 @@ public class StandardOutputVisitor : XmlTestExecutionVisitor
         readonly ConcurrentDictionary<string, ExecutionSummary> completionMessages;
         readonly string defaultDirectory;
         readonly bool showProgress;
+        readonly Stopwatch clock;
+        private ConcurrentDictionary<string, long> runningTests;
+        private Thread watcher;
+        readonly int longTestMaxMilliseconds = 1_000 * 60 * 5;
```
weshaggard: This change appears to be breaking BuildTools. Looks like the compiler doesn't like this line. 

https://devdiv.visualstudio.com/DevDiv/Net%20CoreFX/_build/index?buildId=1577238&_a=summary

```
2018-04-10T23:28:56.6798516Z Visitors\StandardOutputVisitor.cs(22,49): error CS1002: ; expected [E:\A\_work\3365\s\src\xunit.console.netcore\xunit.console.netcore.csproj]
2018-04-10T23:28:56.6799210Z Visitors\StandardOutputVisitor.cs(22,56): error CS1519: Invalid token '60' in class, struct, or interface member declaration [E:\A\_work\3365\s\src\xunit.console.netcore\xunit.console.netcore.csproj]
2018-04-10T23:28:56.6799797Z Visitors\StandardOutputVisitor.cs(23,51): error CS1002: ; expected [E:\A\_work\3365\s\src\xunit.console.netcore\xunit.console.netcore.csproj]
2018-04-10T23:28:56.6800539Z Visitors\StandardOutputVisitor.cs(23,58): error CS1519: Invalid token '60' in class, struct, or interface member declaration [E:\A\_work\3365\s\src\xunit.console.netcore\xunit.console.netcore.csproj]
2018-04-10T23:28:56.6801176Z Visitors\StandardOutputVisitor.cs(147,76): error CS1525: Invalid expression term 'long' [E:\A\_work\3365\s\src\xunit.console.netcore\xunit.console.netcore.csproj]
2018-04-10T23:28:56.6801755Z Visitors\StandardOutputVisitor.cs(147,81): error CS1003: Syntax error, ',' expected [E:\A\_work\3365\s\src\xunit.console.netcore\xunit.console.netcore.csproj]
```

I suspect it has to do with `_`, do you know if that is a new compiler feature, as I've never seen constants using that.

kasper3: can the compiler be updated on official build machine? c#7 is over one year old now :)

weshaggard: @dotnet/dnceng can we look into getting build tools on a machine pool that supports VS2017?

ViktorHofer:
When migrating to xunit.console we will use the in-built feature `longRunningTestSeconds`: https://xunit.github.io/docs/configuring-with-json

```
{
  "longRunningTestSeconds": "300"
}
```

wfurt:
you will also need to bump up debug level AFAIK. We should make sure this is enabled by default in our runs.

ViktorHofer:
Right, I forgot to add that switch in the config above. In fact they are diagnostic messages: https://github.com/xunit/xunit/issues/1380#issuecomment-317822206

Thanks!

src/xunit.console.netcore/Visitors/StandardOutputVisitor.cs
ViktorHofer: @dotnet/dnceng @MattGal for RS4 runner work we definitely need a machine pool with VS2017 and Windows 10 RS4 SDK 17134 installed. Is that feasible?

PR #1993, opened 2018-03-30
Reduce priority of logging message

vancem:
This is a side request from #1980


vancem:
cc: @dagood  @weshaggard

PR #1999, opened 2018-04-03
Allow versioning.props to be imported directly.

eerhardt:
When repo's try to use pieces of buildtools, they get warnings because the versioning.props file is already imported.  This change checks the versioning.props file is already imported, and not importing it again.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <BuildToolsVersioningPropsImported>true</BuildToolsVersioningPropsImported>
+
     <Company Condition="'$(Company)' == ''">Microsoft Corporation</Company>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.props
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <BuildToolsVersioningPropsImported>true</BuildToolsVersioningPropsImported>
```
chcosta: I think internal only properties are generally prefixed by '_' in BuildTools, but it's certainly not consistent throughout.

eerhardt: I was kind of thinking of this as public, and some external repo could use this to check whether it was already imported, and not import it again....  thoughts?

chcosta: Makes sense

eerhardt: Similar to this

https://github.com/Microsoft/msbuild/blob/f72bd2eaa37476819ab7647f63d445df8055dab6/src/Tasks/Microsoft.Common.props#L140

Maybe I should rename it to the same name format?

Issue #1995, opened 2018-03-31
lldb output is garbled


danmoseley:
https://mc.dot.net/#/user/sywhang/pr~2Fjenkins~2Fdotnet~2Fcorefx~2Fmaster~2F/test~2Ffunctional~2Fcli~2F/49c17d3e8819acd03d904489a1490392761c8694/workItem/System.Runtime.Extensions.Tests/wilogs

After it tries to do `pe -nested` and fails, there are innumerable lines like
`2018-03-31 02:12:10,196: INFO: proc(54): run_and_log_output: Output: ^D`

`^D` means "end of file"

Removing the `^D` lines we get a sort of broken native stack. Then I guess it runs out of space in the log.

I think we need to get hold of one of these dumps and ask @mikem8361 to look at it.

@Anipik 

```
2018-03-31 02:12:10,074: INFO: proc(54): run_and_log_output: Output: ----- start =============== lldb Output =====================================================
2018-03-31 02:12:10,074: INFO: proc(54): run_and_log_output: Output: Printing managed exceptions and managed call stack.
2018-03-31 02:12:10,128: INFO: proc(54): run_and_log_output: Output: Core file '/home/helixbot/dotnetbuild/work/75f48688-e531-40fa-843d-0939604b5288/Work/c8a597ba-ad4c-434d-af5f-fe86283754e4/Unzip/core.72991' (x86_64) was loaded.
2018-03-31 02:12:10,135: INFO: proc(54): run_and_log_output: Output: Aborting after_file command execution, command: 'pe -nested' failed.
2018-03-31 02:12:10,197: INFO: proc(54): run_and_log_output: Output: 
[KProcess 0 stopped
2018-03-31 02:12:10,197: INFO: proc(54): run_and_log_output: Output: * thread #1: tid = 0, 0x00007fd3739081f7, name = 'dotnet', stop reason = signal SIGABRT
2018-03-31 02:12:10,198: INFO: proc(54): run_and_log_output: Output:     frame #0: 0x00007fd3739081f7
2018-03-31 02:12:10,198: INFO: proc(54): run_and_log_output: Output: -> 0x7fd3739081f7:  addb   %al, (%rax)
2018-03-31 02:12:10,198: INFO: proc(54): run_and_log_output: Output:    0x7fd3739081f9:  addb   %al, (%rax)
2018-03-31 02:12:10,198: INFO: proc(54): run_and_log_output: Output:    0x7fd3739081fb:  addb   %al, (%rax)
[...]

PR #2000, opened 2018-04-04
Disabling lldb

Anipik:
Related to https://github.com/dotnet/buildtools/issues/1995
Running lldb output only when the task fails


Anipik:
@danmosemsft can you take a look ?

PR #2001, opened 2018-04-04
Generate resources code.refactor

ericstj:
I needed the ability to generate an SR file that only has constant IDs in it, ala the old desktop source.  I added this as an option on GenerateResourcesCode.  It's opt-in on a per-project basis so it does nothing as is.  Projects that opt-in will need to call SR.GetResourceString themselves in order to get the actual resource string.

While I was at it I refactored the code here a bit to be more readable and removed the DebugOnly property which doesn't appear to be used by anyone.

I've validated that my changes produce white-space-only diff in the generated SR files corefx.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
@@ -143,7 +143,7 @@ namespace Microsoft.DotNet.Build.Tasks
             for (var i = 0; i < original.Length; i++)
             {
                 // duplicate '"' for VB and C#
-                if (original[i] == '\"' && (_targetLanguage == TargetLanguage.VB || _targetLanguage == TargetLanguage.CSharp))
+                if (original[i] == '\"')
                 {
                     stringLiteral.Append("\"");
                 }
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
@@ -18,8 +18,6 @@ namespace Microsoft.DotNet.Build.Tasks
     {
         private TargetLanguage _targetLanguage = TargetLanguage.CSharp;
         private StreamWriter _targetStream;
-        private StringBuilder _debugCode = new StringBuilder();
-        private Dictionary<string, int> _keys;
         private String _resourcesName;
 
         [Required]
@@ -30,8 +28,6 @@ namespace Microsoft.DotNet.Build.Tasks
 
         [Required]
         public string AssemblyName { get; set; }
-
-        public bool DebugOnly { get; set; }
 
         public override bool Execute()
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -92,91 +88,116 @@ private void WriteClassHeader()
...
-                StoreValues((string)pair.Key, (string)pair.Value);
+                foreach (var resourcePair in resources)
+                {
+                    WriteResourceConstant((string)resourcePair.Key);
+                }
+            }
+            else
+            {
+                _targetStream.WriteLine(_targetLanguage == TargetLanguage.CSharp ?
+                    "#if !DEBUGRESOURCES" :
```
tarekgh: >DEBUGRESOURCES [](start = 26, length = 14)

I am not sure if we still need this either. This would be useful for .Net Native only and I don't think it is used any more as you have removed the Debug property anyway.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -92,91 +88,116 @@ private void WriteClassHeader()
...
             }
-            _keys[leftPart] = 0;
-            StringBuilder sb = new StringBuilder(rightPart.Length);
-            for (var i = 0; i < rightPart.Length; i++)
+            stringLiteral.Append('\"');
+            for (var i = 0; i < original.Length; i++)
             {
                 // duplicate '"' for VB and C#
-                if (rightPart[i] == '\"' && (_targetLanguage == TargetLanguage.VB || _targetLanguage == TargetLanguage.CSharp))
+                if (original[i] == '\"' && (_targetLanguage == TargetLanguage.VB || _targetLanguage == TargetLanguage.CSharp))
```
tarekgh: >_targetLanguage == TargetLanguage.VB || _targetLanguage == TargetLanguage.CSharp [](start = 44, length = 80)

I think we support either CSharp or VB, so this condition looks not needed at least for now.

tarekgh:
other than my comments, LGTM. Thank you.

src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
ericstj: This is definitely still used in corefx. It keeps the literal resource strings out of the il in ret builds.

tarekgh: What I mean we always produce the code without literal resource strings. so we'll not produce the parts under DEBUGRESOURCES.

tarekgh: it is ok to keep for now though so you don't have to block on that.

tarekgh: I confirmed with Morgan than we don't need to generate the literal strings anymore. this was used before with .Net Native but it is not needed anymore.

tarekgh: if you remove it, you will need to remove the line 

https://github.com/dotnet/buildtools/blob/63db91f2fe1a7aec78ff190e7307dbdc0b62757d/src/Microsoft.DotNet.Build.Tasks/PackageFiles/resources.targets#L16

too. Thanks

ericstj: I'm not willing to bet on that just yet, so I'll hold off on that cleanup.

PR #2003, opened 2018-04-07
Update the ref assembly generation rules and logic for structs.

ahsonkhan:
Part of https://github.com/dotnet/corefx/issues/28870

Related PR in corefx: https://github.com/dotnet/corefx/pull/28906

cc @weshaggard, @nguerrera, @joshfree

### DIFF
--- a/src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
+++ b/src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
@@ -164,7 +164,7 @@ namespace Microsoft.Cci.Writers
                         newFields.Add(genericField);
 
                     // For definiteassignment checks the compiler needs to know there is a private field
-                    // that has not be initialized so if there are any we need to add a dummy private
+                    // that has not been initialized so if there are any we need to add a dummy private
                     // field to help the compiler do its job and error about uninitialized structs
                     bool hasRefPrivateField = excludedFields.Any(f => f.Type.IsOrContainsReferenceType());
 
@@ -173,14 +173,16 @@ namespace Microsoft.Cci.Writers
                     // taking pointers to this struct because the GC will not track updating those references
                     if (hasRefPrivateField)
                     {
-                        DummyFieldWriterHelper(parentType, excludedFields, newFields, parentType.PlatformType.SystemObject);
+                        IFieldDefinition fieldType = DummyFieldWriterHelper(parentType, excludedFields, parentType.PlatformType.SystemObject);
+                        newFields.Add(fieldType);
                     }
 
-                    bool hasValueTypePrivateField = excludedFields.Any(f => f.Type.IsValueType);
+                    bool hasValueTypePrivateField = excludedFields.Any(f => !f.Type.IsOrContainsReferenceType());
 
                     if (hasValueTypePrivateField)
                     {
-                        DummyFieldWriterHelper(parentType, excludedFields, newFields, parentType.PlatformType.SystemInt32, "_dummyInt");
+                        IFieldDefinition fieldType = DummyFieldWriterHelper(parentType, excludedFields, parentType.PlatformType.SystemInt32, "_dummyPrimitive");
+                        newFields.Add(fieldType);
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
```diff
@@ -161,32 +163,24 @@ public override void Visit(ITypeDefinition parentType, IEnumerable<IFieldDefinit
...
+                    // that has not be initialized so if there are any we need to add a dummy private
+                    // field to help the compiler do its job and error about uninitialized structs
+                    bool hasRefPrivateField = excludedFields.Any(f => f.Type.IsOrContainsReferenceType());
+
+                    // If at least one of the private fields contains a reference type then we need to
+                    // set this field type to object or reference field to inform the compiler to block
+                    // taking pointers to this struct because the GC will not track updating those references
+                    if (hasRefPrivateField)
+                    {
+                        DummyFieldWriterHelper(parentType, excludedFields, newFields, parentType.PlatformType.SystemObject);
```
weshaggard: nit: I would probably have the DummyFieldWriterHelper return the field instead of taking in the list and adding it to the list explicitly.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
```diff
@@ -161,32 +163,24 @@ public override void Visit(ITypeDefinition parentType, IEnumerable<IFieldDefinit
                     foreach (var genericField in genericTypedFields)
                         newFields.Add(genericField);
 
-                    if (!genericTypedFields.Any())
+                    // For definiteassignment checks the compiler needs to know there is a private field
+                    // that has not be initialized so if there are any we need to add a dummy private
```
weshaggard: "has not been initialized"

src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
```diff
@@ -161,32 +163,24 @@ public override void Visit(ITypeDefinition parentType, IEnumerable<IFieldDefinit
...
+
+                    // If at least one of the private fields contains a reference type then we need to
+                    // set this field type to object or reference field to inform the compiler to block
+                    // taking pointers to this struct because the GC will not track updating those references
+                    if (hasRefPrivateField)
+                    {
+                        DummyFieldWriterHelper(parentType, excludedFields, newFields, parentType.PlatformType.SystemObject);
+                    }
+
+                    bool hasValueTypePrivateField = excludedFields.Any(f => f.Type.IsValueType);
```
weshaggard: I think this `IsValueType` check is what caused IntPtr/UIntPtr to loose their dummy fields. I think we should instead simply negate the reference type check instead, or simply filter the list down and anything left consider them to be valuetype fields.

src/Microsoft.Cci.Extensions/Writers/CSharp/CSharpWriter.cs
```diff
@@ -173,14 +173,16 @@ public override void Visit(ITypeDefinition parentType, IEnumerable<IFieldDefinit
                     // taking pointers to this struct because the GC will not track updating those references
                     if (hasRefPrivateField)
                     {
-                        DummyFieldWriterHelper(parentType, excludedFields, newFields, parentType.PlatformType.SystemObject);
+                        IFieldDefinition fieldType = DummyFieldWriterHelper(parentType, excludedFields, parentType.PlatformType.SystemObject);
+                        newFields.Add(fieldType);
                     }
 
-                    bool hasValueTypePrivateField = excludedFields.Any(f => f.Type.IsValueType);
+                    bool hasValueTypePrivateField = excludedFields.Any(f => !f.Type.IsOrContainsReferenceType());
```
weshaggard: I'm not sure how expensive this is but I wonder if instead you just filter the list while checking for the private ref fields and this will just be what remains.

ahsonkhan: I don't know how that would by faster while still using linq expressions and IEnumerable. We still have to first check if we have any private ref fields. So what exactly would we be filtering during the first pass?

I could do something like this, but I doubt it would be much faster:
```C#
bool hasRefPrivateField = excludedFields.Any(f => f.Type.IsOrContainsReferenceType());

var fieldsLeft = excludedFields.Where(f => !f.Type.IsOrContainsReferenceType());
if (fieldsLeft.Count() > 0)
{
    // ...
}
```

ahsonkhan: How about something like this:
```C#
int count = 0;
foreach (IFieldDefinition field in excludedFields)
{
    if (field.Type.IsOrContainsReferenceType())
    {
        hasRefPrivateField = true;
        break;
    }
    count++;
}

var leftoverFields = excludedFields.Skip(count);
if (count > 0)
{
    // ...
}
```

weshaggard: Don't worry about it for now. I don't think it is going to matter as the number of fields is pretty small.

PR #2005, opened 2018-04-10
Allow further customization of SR class

ericstj:
This enables projects to define separate types for SR and whatever type contains the
resource keys.

While doing this I further simplified the code generation.


src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -62,33 +84,47 @@ public override bool Execute()
...
-
-                _targetStream.WriteLine("    internal static partial class SR");
+                _targetStream.WriteLine("");
+                _targetStream.WriteLine($"    internal static partial class {_srClassName}");
                 _targetStream.WriteLine("    {");
-
-                _targetStream.WriteLine("#pragma warning disable 0414");
-                _targetStream.WriteLine($"        private const string s_resourcesName = \"{_resourcesName}.SR\";");
-                _targetStream.WriteLine("#pragma warning restore 0414");
+                _targetStream.WriteLine($"        internal static Type ResourceType {{ get; }} = typeof({_resourcesName}.SR); ");
```
tarekgh: >ResourceType [](start = 71, length = 12)

are we going to remove this property from the common SR.cs file?

PR #2010, opened 2018-04-17
Don't preemptively omit inbox frameworks when creating trimmed dependencies

ericstj:
When creating trimmed dependency groups we cannot know if a group is redundant
during the first pass (since another group might be created). Always create the group
and only trim in the second pass where we have the full set of dependency groups and
can know for certain if an addition is redundant.

Here's the diff after making this change: ericstj/scratch@1db38c5

It impacted a single project, the one with this issue: dotnet/corefx#29149

This is a release/2.1 port of https://github.com/dotnet/buildtools/pull/2009


weshaggard:
@ericstj I included your change in my 2.1 merge at https://github.com/dotnet/buildtools/pull/2011

PR #2015, opened 2018-04-19
Remove legacy RIDs from testing

ericstj:
fedora.23-x64, opensuse.13.2-x64, and ubuntu.16.10-x64 are no longer supported RIDs

Don't validate for them (even if they would work on portable RID packages).

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -938,22 +938,31 @@
 
   <!-- BEGIN validation and output -->
   <ItemGroup>
+    <NETCoreApp10RIDs Condition="'@(NETCoreApp10RIDs)' == ''" Include="win7-x86;win7-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;linuxmint.17-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64" />
     <!-- Default validation frameworks : frameworks which we test for support / not support in all packages -->
     <DefaultValidateFramework Include="netcoreapp1.0">
-      <RuntimeIDs>win7-x86;win7-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;linuxmint.17-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64</RuntimeIDs>
+      <RuntimeIDs>@(NETCoreApp10RIDs)</RuntimeIDs>
     </DefaultValidateFramework>
+    <NETCoreApp11RIDs Condition="'@(NETCoreApp11RIDs)' == ''" Include="@(NETCoreApp10RIDs);osx.10.12-x64;fedora.24-x64;opensuse.42.1-x64;rhel.7-x64" />
     <DefaultValidateFramework Include="netcoreapp1.1">
-      <RuntimeIDs>win7-x86;win7-x64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.24-x64;linuxmint.17-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64</RuntimeIDs>
+      <RuntimeIDs>@(NETCoreApp11RIDs)</RuntimeIDs>
     </DefaultValidateFramework>
+    <NETCoreApp20RIDs Condition="'@(NETCoreApp20RIDs)' == ''" Include="@(NETCoreApp11RIDs)" />
     <DefaultValidateFramework Include="netcoreapp2.0">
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -938,18 +938,18 @@
...
     <DefaultValidateFramework Include="netcoreapp1.1">
-      <RuntimeIDs>win7-x86;win7-x64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.23-x64;fedora.24-x64;linuxmint.17-x64;opensuse.13.2-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64</RuntimeIDs>
+      <RuntimeIDs>win7-x86;win7-x64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.24-x64;linuxmint.17-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64</RuntimeIDs>
     </DefaultValidateFramework>
     <DefaultValidateFramework Include="netcoreapp2.0">
-      <RuntimeIDs>win7-x86;win7-x64;win10-arm64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.23-x64;fedora.24-x64;linuxmint.17-x64;opensuse.13.2-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64</RuntimeIDs>
+      <RuntimeIDs>win7-x86;win7-x64;win10-arm64;osx.10.10-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;fedora.24-x64;linuxmint.17-x64;opensuse.42.1-x64;rhel.7-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64</RuntimeIDs>
     </DefaultValidateFramework>
     <DefaultValidateFramework Include="netcoreapp2.1">
```
weshaggard: Would you mind adding netcoreap2.2 while your in here?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -938,18 +938,18 @@
 
   <!-- BEGIN validation and output -->
   <ItemGroup>
-    <!-- Default validation frameworks : every framework that supports contracts -->
+    <!-- Default validation frameworks : frameworks which we test for support / not support in all packages -->
     <DefaultValidateFramework Include="netcoreapp1.0">
-      <RuntimeIDs>win7-x86;win7-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;linuxmint.17-x64;opensuse.13.2-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64</RuntimeIDs>
+      <RuntimeIDs>win7-x86;win7-x64;osx.10.11-x64;centos.7-x64;debian.8-x64;linuxmint.17-x64;rhel.7.2-x64;ubuntu.14.04-x64;ubuntu.16.04-x64</RuntimeIDs>
```
weshaggard: I suspect it is worth adding an item group with these distro's so we can add remove from them in cases were we need to workaround issues.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: OK.  Any new RIDs to include?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: Ok, not a bad idea.  For each TFM I'll have an itemgroup that can be overridden.

weshaggard: Are they not all the same right now?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
weshaggard: I'm sure there are some new ones but it would require cross referencing this list with the list in the notes I sent you. I don't see alpine listed in either list currently.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: no, each TFM has different sets

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
ericstj: That list is a bit too much different than this for me to bite off taking that change at the moment.  We can consider doing as a future update.

weshaggard: sounds good. Can you at least file an issue on it?

PR #2006, opened 2018-04-11
Update roslyn version to latest available (2.8.0-beta4-62810-06)

ahsonkhan:
See https://github.com/dotnet/corefx/pull/29003#issuecomment-380290429

cc: @joperezr, @weshaggard, @VSadov

PR #2019, opened 2018-04-23
Update Roslyn to 2.8.0-beta4

jkotas:


jkotas:
Reopening #2006. Now that master is open for 2.2 we do not need to worry about risk of updating the compiler close to release.

PR #2021, opened 2018-04-24
Add support for named operands in ApiCompat

ericstj:
In the case we run reverse API compat we swap the names of the operands so that it
makes sense.

### DIFF
--- a/src/ApiCompat/Program.cs
+++ b/src/ApiCompat/Program.cs
@@ -12,6 +12,7 @@ using System.Linq;
 using Microsoft.Cci;
 using Microsoft.Cci.Comparers;
 using Microsoft.Cci.Differs;
+using Microsoft.Cci.Differs.Rules;
 using Microsoft.Cci.Extensions;
 using Microsoft.Cci.Filters;
 using Microsoft.Cci.Mappings;
@@ -25,13 +26,19 @@ namespace ApiCompat
     public class ExportCciSettings
     {
         public static IEqualityComparer<ITypeReference> StaticSettings { get; set; }
+        public static IDifferenceOperands  StaticOperands { get; set; }
+
         public ExportCciSettings()
         {
             Settings = StaticSettings;
+            Operands = StaticOperands;
         }
 
         [Export(typeof(IEqualityComparer<ITypeReference>))]
-        public IEqualityComparer<ITypeReference> Settings { get; set; }
+        public IEqualityComparer<ITypeReference> Settings { get; }
+
+        [Export(typeof(IDifferenceOperands))]
+        public IDifferenceOperands Operands { get; }
     }
 
     public class Program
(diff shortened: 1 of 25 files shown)


ericstj:
This requires corefx changes to swap implementation and contract in all MatchingRefApiCompatBaseline files.

src/ApiCompat/Rules/Compat/DifferenceRule.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Composition;
+
+namespace Microsoft.Cci.Differs.Rules
+{
+    internal abstract class DifferenceRule : Microsoft.Cci.Differs.DifferenceRule
```
weshaggard: LOL. Too scared to update the base rule and just make it optional?

src/ApiCompat/Rules/Compat/DifferenceRule.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        [Import]
+        public IDifferenceOperands Operands { get; set; }
+
+        public string Left => Operands?.Left ?? "contract";
```
weshaggard: I wonder if calling this Contract and Implementation may be easier to understand given all the rules are still written in that method? Or do you plan to do a pace to rename all the parameters as well :)

src/ApiCompat/Rules/Compat/CannotSealType.cs
```diff
@@ -18,7 +18,7 @@ public override DifferenceType Diff(IDifferences differences, ITypeDefinition im
             if (impl.IsEffectivelySealed() && !contract.IsEffectivelySealed())
             {
                 differences.AddIncompatibleDifference(this,
-                    "Type '{0}' is sealed in the implementation but not sealed in the contract.", impl.FullName());
+                    $"Type '{impl.FullName()}' is sealed in the {Right} but not sealed in the {Left}.");
```
weshaggard: I suspect I will have to keep looking whether Left/Right refers to contract/Implementation while working on the rules code.

src/ApiCompat/Rules/Compat/DifferenceRule.cs
ericstj: Yeah I was thinking the same.  I think left and right is better if we were starting from scratch but maybe I should just stick to contract/implementation for now.

src/ApiCompat/Program.cs
```diff
@@ -276,6 +288,10 @@ private static void ParseCommandLine(string[] args)
                 parser.DefineOptionalQualifier("mdil", ref s_mdil, "Enforce MDIL servicing rules in addition to IL rules.");
                 parser.DefineAliases("excludeNonBrowsable", "enb");
                 parser.DefineOptionalQualifier("excludeNonBrowsable", ref s_excludeNonBrowsable, "When MDIL servicing rules are not being enforced, exclude validation on types that are marked with EditorBrowsable(EditorBrowsableState.Never).");
+                parser.DefineAliases("leftOperand", "lhs");
+                parser.DefineOptionalQualifier("leftOperand", ref s_contractOperand, "Name for left operand in comparison, default is 'contract'.");
```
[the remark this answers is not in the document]
weshaggard: I'm fine with this being left and right.

src/ApiCompat/Rules/Compat/DifferenceRule.cs
weshaggard: I don't really like playing games and using the same type name in a different namespace as it just causes confusion. I'm OK with it staying in ApiCompat but can you change the name of the type to "CompatDifferenceRule" or something along those lines (yes I realize you have to then update all the base types of the rules but you are already touching those files anyway).

src/ApiCompat/NamespaceRemappingComparer.cs
```diff
@@ -20,7 +20,7 @@ public NamespaceRemappingComparers(string mappingFile)
         {
             if (!string.IsNullOrEmpty(mappingFile))
                 ParseConfig(mappingFile);
-            Contract.Assert(_mappings.Count > 0, "NamespaceRemappingComparers has no namespace remappings.");
+            Debug.Assert(_mappings.Count > 0, "NamespaceRemappingComparers has no namespace remappings.");
```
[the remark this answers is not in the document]
weshaggard: I'm good with changing these to Debug.Assert.

src/ApiCompat/Rules/Compat/DifferenceRule.cs
ericstj: Ok, I had done that first, but you know how much I hate naming...  I'll use CompatDifferenceRule.

PR #2026, opened 2018-04-30
Test Category Attribute Added To Xunit

Anipik:
Tested Locally with corefx

### DIFF
--- a/src/xunit.netcore.extensions/Attributes/TestCategoryAttribute.cs
+++ b/src/xunit.netcore.extensions/Attributes/TestCategoryAttribute.cs
@@ -1,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    /// <summary>
+    /// Apply this attribute to your test method to specify the test category.
+    /// </summary>
+    [TraitDiscoverer("Xunit.NetCore.Extensions.TestCategoryDiscoverer", "Xunit.NetCore.Extensions")]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true)]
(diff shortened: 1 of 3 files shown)


src/xunit.netcore.extensions/Attributes/TestCategoryAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    /// <summary>
+    /// Apply this attribute to your test method to specify the test category.
+    /// </summary>
+    [TraitDiscoverer("Xunit.NetCore.Extensions.TestCategoryDiscoverer", "Xunit.NetCore.Extensions")]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true)]
+    public class TestCategoryAttribute : Attribute, ITraitAttribute
```
weshaggard: How is this different from https://github.com/nunit/docs/wiki/Category-Attribute?

danmosemsft: this is xunit that's nunit?

Anipik: Its the meant to give the same behavior as testcategory in Nunit. We currently don't have anything like that in Xunit.

weshaggard: Sorry I copied the wrong link. I was looking at http://www.brendanconnolly.net/organizing-tests-with-xunit-traits/. Do we need a something beyond the Trait attribute?

Anipik: It will avoid repittion of using ```[Trait("Category" ,"X"]```
The other testing platforms like Nunit and mstest gives this feature.

weshaggard: Sounds good thanks.

Anipik:
@danmosemsft can you sign off this PR ?

PR #2030, opened 2018-05-03
Add triage step to dumpling upload

DrewScoggins:
This adds in the steps that we need to ensure that we are able to do
triage of the dumps before they are uploaded so that we can get failing
stack information.


DrewScoggins:
PTAL @Anipik @danmosemsft

DrewScoggins:
@jorive PTAL

Issue #1221, opened 2016-11-11
Remove OpenSourceSign from Microsoft.DotNet.Build.Tasks


mellinoe:
Ported from https://github.com/dotnet/corefx/issues/6759 (cc @tmat)

> Csc now supports OSS signing via /publicsign switch. The build task is not needed anymore and can be removed and CoreFX build simplified.

https://github.com/dotnet/buildtools/blob/e7768f0b40786251c787daaac43533f06fee8b26/src/Microsoft.DotNet.Build.Tasks/PackageFiles/sign.targets#L9

https://github.com/dotnet/buildtools/blob/78260239e4db68ad5c2df774a27d5f9a54fd92e5/src/Microsoft.DotNet.Build.Tasks/OpenSourceSign.cs

weshaggard:
Yes please

weshaggard:
I was going to switch to doing this but there is one issue with switching to let the compiler do this work for us. Namely some projects have assembly rewriting that they do after compile (i.e. GenFacades, ILLinker, etc) and for those projects they will not be correctly signed after rewriting if we do it during the compile phase. I didn't fully verify that we don't already have this issue but I wanted to call it out for anyone that looks at doing this.

weshaggard:
Another thought here if we cannot easily remove the rewriting tools (which I suspect we cannot) then we can just augment them to round-trip the open/public signing bit in the binary (assuming it isn't already, I've not verified).

ViktorHofer:
// Auto-generated message

Going forward, the .NET team is using https://github.com/dotnet/arcade to develop the code and issues formerly in this repository. Feel free to reopen/move this issue if it still applies to servicing branches in this repository, or source code which now resides in the Arcade repository.

PR #2036, opened 2018-05-12
Fix 'File is already strong-name signed'

danmosemsft:
Fixes https://github.com/dotnet/corefx/issues/29495

The OpenSourceSign task modifies the file in place. It also has a test when it runs that the file has not been already signed. Its execution is gated by comparing the assembly timestamp against a fake file with the extension ".oss_signed" which is written after the task executes.

Modifying a file in place is not good practice for builds because it is hard to fail atomically. Eg., if the build is interrupted between the signing and the writing of the ".oss_signed" file then the next build will attempt to sign again and get the error. There are likely other failure modes.

Options to fix include 
(1) signing to a new location (note - signing to a temporary location then renaming back does not count.)
(2) stop erroring when file is already signed, just re-sign it

Since (2) is simple and makes things no worse, this does (2).

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/OpenSourceSign.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/OpenSourceSign.cs
@@ -30,6 +30,17 @@ namespace Microsoft.DotNet.Build.Tasks
         }
 
         /// <summary>
+        /// Whether the task should fail in the case the file is already signed. If not, it will re-sign.
+        /// By default it does not fail, because this condition can occur simply because the build was 
+        /// interrupted at an inopportune moment.
+        /// </summary>
+        public bool FailIfAlreadySigned
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
         /// The number of bytes from the start of the <see cref="CorHeader"/> to its <see cref="CorFlags"/>.
         /// </summary>
         private const int OffsetFromStartOfCorHeaderToFlags =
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/OpenSourceSign.cs
```diff
@@ -29,6 +29,17 @@ public string AssemblyPath
             set;
         }
 
+        /// <summary>
+        /// Whether the task should fail in the case the file is already signed. If not, it will re-sign.
+        /// By default it does not fail, because this condition can occur simply because the build was 
+        /// interrupted at an inopportune moment.
+        /// </summary>
+        public bool FailIfAlreadySigned
```
joperezr: This might be a dumb question, but what is the default for non-required bool MSBuild Task properties? Is it the same as any regular property?

weshaggard:
FWIW I think we should eliminate this task and move the compiler feature, see https://github.com/dotnet/buildtools/issues/1221.

Personally I wouldn't re-sign but instead I would just skip and perhaps just make this a warning.

src/Microsoft.DotNet.Build.Tasks/OpenSourceSign.cs
danmosemsft: Yes...false

danmosemsft:
I'm going to submit this as a bandaid to unlock folks and leave the issue open.

PR #2043, opened 2018-05-22
Add error message for invalid input in ConditionalClass

caesar1995:
Follow up PR for: https://github.com/dotnet/buildtools/pull/2040#discussion_r190044557

### DIFF
--- a/src/xunit.netcore.extensions/Discoverers/ConditionalClassDiscoverer.cs
+++ b/src/xunit.netcore.extensions/Discoverers/ConditionalClassDiscoverer.cs
@@ -53,12 +53,7 @@ namespace Xunit.NetCore.Extensions
                 if (string.IsNullOrWhiteSpace(entry)) continue;
 
                 MethodInfo conditionMethodInfo = ConditionalTestDiscoverer.LookupConditionalMethod(calleeType, entry);
-                if (conditionMethodInfo == null)
-                {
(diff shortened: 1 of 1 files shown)


src/xunit.netcore.extensions/Discoverers/ConditionalClassDiscoverer.cs
```diff
@@ -53,12 +53,7 @@ internal static bool EvaluateParameterHelper(IAttributeInfo traitAttribute)
                 if (string.IsNullOrWhiteSpace(entry)) continue;
 
                 MethodInfo conditionMethodInfo = ConditionalTestDiscoverer.LookupConditionalMethod(calleeType, entry);
-                if (conditionMethodInfo == null)
-                {
-                    // Unable to get MethodInfo. It's caused by bad user input.
-                    // We need to report some error here. For now, just don't run the test.
-                    return false;
-                }
+                Debug.Assert(conditionMethodInfo != null, $"Unable to get MethodInfo, please check input for {entry}.");
```
safern: I don't know if this will work as we might be using the release build of xunit extensions.

caesar1995: Ok, I searched through Xunit repo and not able to find a proper API which we can use to report an error. We can throw a Exception here.

PR #2045, opened 2018-05-23
Workaround new errors from genfacades

danmosemsft:
Previously some detailed errors/warnings/messages were not reaching the build log at all. Recent change fixed that but there are some errors in CoreFX. Temporarily changing errors to warnings while the cause is fixed.

https://github.com/dotnet/corefx/issues/29861


ahsonkhan:
cc @safern, @weshaggard

PR #2048, opened 2018-05-23
Remove msbuild dependency from buildtools

wtgodbe:
Part of https://github.com/dotnet/corefx/projects/3

CC @weshaggard @joperezr @eerhardt @safern

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -10,7 +10,6 @@ if [%PACKAGES_DIR%] == [] set PACKAGES_DIR=%TOOLRUNTIME_DIR%
 set PACKAGES_DIR=%PACKAGES_DIR:"=%
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
 set MICROBUILD_VERSION=0.2.0
-set PORTABLETARGETS_VERSION=0.1.1-dev
 set ROSLYNCOMPILERS_VERSION=2.8.0-beta4
 
 :: Default to x64 native tools if nothing was specified.
@@ -92,7 +91,6 @@ if not [%RESTORE_PORTABLETARGETS_ERROR_LEVEL%]==[0] (
   echo ERROR: An error ocurred when running: '"%DOTNET_CMD%" restore "%PORTABLETARGETS_PROJECT%"'. Please check above for more details.
   exit /b %RESTORE_PORTABLETARGETS_ERROR_LEVEL%
 )
-Robocopy "%PACKAGES_DIR%\Microsoft.Portable.Targets\%PORTABLETARGETS_VERSION%\contentFiles\any\any\Extensions." "%TOOLRUNTIME_DIR%\." /E
 Robocopy "%PACKAGES_DIR%\MicroBuild.Core\%MICROBUILD_VERSION%\build\." "%TOOLRUNTIME_DIR%\." /E
 
 :: Copy Roslyn Compilers Over to ToolRuntime
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -26,7 +26,6 @@ set MSBUILD_PROJECT_CONTENT= ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
     ^^^<PackageReference Include=^"Microsoft.Net.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
-    ^^^<PackageReference Include=^"Microsoft.NetCore.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
   ^^^</ItemGroup^^^> ^
  ^^^</Project^^^>
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -26,8 +26,6 @@ set MSBUILD_PROJECT_CONTENT= ^
   ^^^</PropertyGroup^^^> ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
-    ^^^<PackageReference Include=^"Microsoft.Portable.Targets^" Version=^"%PORTABLETARGETS_VERSION%^" /^^^> ^
-    ^^^<PackageReference Include=^"Microsoft.Net.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
   ^^^</ItemGroup^^^> ^
  ^^^</Project^^^>
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -26,6 +26,7 @@ set MSBUILD_PROJECT_CONTENT= ^
   ^^^</PropertyGroup^^^> ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
+    ^^^<PackageReference Include=^"Microsoft.Net.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
   ^^^</ItemGroup^^^> ^
  ^^^</Project^^^>
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -26,6 +26,7 @@ set MSBUILD_PROJECT_CONTENT= ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
     ^^^<PackageReference Include=^"Microsoft.Net.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
+    ^^^<PackageReference Include=^"Microsoft.NetCore.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
   ^^^</ItemGroup^^^> ^
  ^^^</Project^^^>
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -73,10 +73,6 @@ $__DOTNET_CMD restore "${__PORTABLETARGETS_PROJECT}" $__INIT_TOOLS_RESTORE_ARGS 
 # Copy MicroBuild targets from packages, allowing for lowercased package IDs.
 cp -R "${__PACKAGES_DIR}"/[Mm]icro[Bb]uild.[Cc]ore/"${__MICROBUILD_VERSION}/build/." "$__TOOLRUNTIME_DIR/."
 
-# Temporary Hacks to fix couple of issues in the msbuild and roslyn nuget packages
-# https://github.com/dotnet/buildtools/issues/1464
-[ -e "$__TOOLRUNTIME_DIR/Microsoft.CSharp.Targets" ] || mv "$__TOOLRUNTIME_DIR/Microsoft.CSharp.targets" "$__TOOLRUNTIME_DIR/Microsoft.CSharp.Targets"
-
 # Copy some roslyn files over
 cp $__TOOLRUNTIME_DIR/runtimes/any/native/* $__TOOLRUNTIME_DIR/
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -25,8 +25,6 @@ __PORTABLETARGETS_PROJECT_CONTENT="
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include=\"MicroBuild.Core\" Version=\"$__MICROBUILD_VERSION\" />
-    <PackageReference Include=\"Microsoft.Portable.Targets\" Version=\"$__PORTABLETARGETS_VERSION\" />
-    <PackageReference Include=\"Microsoft.NETCore.Compilers\" Version=\"$__ROSLYNCOMPILER_VERSION\" />
   </ItemGroup>
 </Project>"
 __PUBLISH_TFM=netcoreapp2.0
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -25,6 +25,7 @@ __PORTABLETARGETS_PROJECT_CONTENT="
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include=\"MicroBuild.Core\" Version=\"$__MICROBUILD_VERSION\" />
+    <PackageReference Include=\"Microsoft.NETCore.Compilers\" Version=\"$__ROSLYNCOMPILER_VERSION\" />
   </ItemGroup>
 </Project>"
 __PUBLISH_TFM=netcoreapp2.0
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -14,7 +14,6 @@ __TOOLRUNTIME_DIR=${3:-}
 __PACKAGES_DIR=${4:-$__TOOLRUNTIME_DIR}
 __TOOLS_DIR=$(cd "$(dirname "$0")"; pwd -P)
 __MICROBUILD_VERSION=0.2.0
-__PORTABLETARGETS_VERSION=0.1.1-dev
 __ROSLYNCOMPILER_VERSION=2.8.0-beta4
 
 __PORTABLETARGETS_PROJECT_CONTENT="
@@ -71,8 +70,7 @@ echo $__PORTABLETARGETS_PROJECT_CONTENT > "${__PORTABLETARGETS_PROJECT}"
 echo "Running: \"$__DOTNET_CMD\" restore \"${__PORTABLETARGETS_PROJECT}\" $__INIT_TOOLS_RESTORE_ARGS --packages \"${__PACKAGES_DIR}/.\""
 $__DOTNET_CMD restore "${__PORTABLETARGETS_PROJECT}" $__INIT_TOOLS_RESTORE_ARGS --packages "${__PACKAGES_DIR}/."
 
-# Copy portable and MicroBuild targets from packages, allowing for lowercased package IDs.
-cp -R "${__PACKAGES_DIR}"/[Mm]icrosoft.[Pp]ortable.[Tt]argets/"${__PORTABLETARGETS_VERSION}/contentFiles/any/any/Extensions/." "$__TOOLRUNTIME_DIR/."
+# Copy MicroBuild targets from packages, allowing for lowercased package IDs.
 cp -R "${__PACKAGES_DIR}"/[Mm]icro[Bb]uild.[Cc]ore/"${__MICROBUILD_VERSION}/build/." "$__TOOLRUNTIME_DIR/."
 
 # Temporary Hacks to fix couple of issues in the msbuild and roslyn nuget packages
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
@@ -16,16 +16,8 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
-    <PackageReference Include="Microsoft.Build.Framework" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build.Runtime" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build" Version="15.3.409" />
-    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.0.0" />
     <PackageReference Include="Microsoft.Net.Compilers.netcore" Version="2.6.0-beta3-62316-02" />
-    <PackageReference Include="Microsoft.Net.Compilers.Targets.NetCore" Version="0.1.5-dev" />
     <PackageReference Include="Microsoft.Cci" Version="4.0.0-rc3-24214-00" />
-    <PackageReference Include="System.Composition" Version="1.1.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
   </ItemGroup>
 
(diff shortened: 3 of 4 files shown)


wtgodbe:
Building CoreFx with this changes gives the following error:

> System.Memory.cs(244,43): error CS8339: The parameter modifier 'in' cannot be used after the modifier 'this' [E:\code\c
orefx\src\System.Memory\ref\System.Memory.csproj]

https://github.com/dotnet/corefx/blob/master/src/System.Memory/ref/System.Memory.cs#L244-L246

Looks like the roslyn compiler doesn't like this syntax.

weshaggard:
cc @eerhardt 

I suspect that is because we need a newer version of the rolsyn compiler to make this all work.

wtgodbe:
It appears the issue may have been fixed in January: https://github.com/dotnet/roslyn/pull/23643, https://github.com/dotnet/roslyn/issues/23895

wtgodbe:
Although it appears we most recently updated the roslyn compiler version a month ago, to a version published on April 23: https://github.com/dotnet/buildtools/pull/2019

joperezr:
I believe what is happening is that you are not using the right compiler. you must be using the one coming from VS 2017. Check the diag logs and make sure that we are in fact using the one we are restoring.

wtgodbe:
@joperezr is right, the build is using C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\15.0\bin\Roslyn\csc.exe

wtgodbe:
@weshaggard under what circumstances would we not want to use the roslyn compiler? Looks like today we only set `CscToolPath` in the event that we're using a `RoslynIncompatibleMsbuildVersion`. My build has MsBuild version 15.0, which doesn't trigger that property.
https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L30

https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props#L45-L50

wtgodbe:
However when I run `Tools\dotnetcli\dotnet.exe msbuild` on System.Memory.csproj, the build uses the Roslyn compiler & the build succeeds.

joperezr:
yeah that is because the SDK inside the CLI would now be dictating which version of the compiler to use. What you need is a pp log in order to see who is responsible of picking which compiler to use, and make sure that it points to the right place. Most likely what you will need to fix here, will be to fix which CSharp and VB targets get imported since they are the ones that set compiler paths.

eerhardt:
@wtgodbe - make sure you test coreclr too.  :)  Just to be safe.

wtgodbe:
Looks like the branch point is here: https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets#L118

When I build with `dotnet.exe msbuild`, `MSBuildExtensionsPath32` resolves to the SDK directory in Tools (`Tools\dotnetcli\sdk\2.1.300-rc1-008673`). When I run with just msbuild, it resolves to `C:\Program Files (x86)\MSBuild\Microsoft\Portable\v4.5`

wtgodbe:
Looks like it's because of `RoslynTargetsPath`.

Old: 
> Property reassignment: $(RoslynTargetsPath)="E:\code\scratch\corefx\Tools

New: 
> Property reassignment: $(RoslynTargetsPath)="C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\15.0\bin/roslyn"

So https://github.com/dotnet/buildtools/pull/2048/files#diff-8f98589c8326995136a136dd8f2635f2R6 is not the right change. I need to find another property that points to the SDK dir, other than `MSBuildToolsPath`

joperezr:
where is roslyn targets path used? I would think that it should point to the place where you are restoring the new roslyn package, the one we will use for building.

wtgodbe:
I've fixed the wrong-CSC error with my last commit, but now I have:

> E:\code\corefx\Tools\ApiCompat.targets(72,5): error : ApiCompat failed for 'E:\
code\corefx\bin\Windows_NT.AnyCPU.Debug\System.Net.NameResolution\netcoreapp\Sy
stem.Net.NameResolution.dll' [E:\code\corefx\src\System.Net.NameResolution\src\
System.Net.NameResolution.csproj]

wtgodbe:
Running "msbuild" from the command line on a .csproj does work, however.

wtgodbe:
Looks like https://github.com/dotnet/buildtools/blob/0eaa521c6c414f04620350ce8aaffbf8a55fc594/src/ApiCompat/Program.cs#L8 is the culprit here, so I'm thinking https://github.com/dotnet/buildtools/pull/2048/files#diff-4e041e6755f6b11ef1de3599b2a087eeL28 was a bad removal. Will try adding that back in & rebuilding

wtgodbe:
I must have built non-clean the first time, I just re-built this from scratch & ingested it in CoreFx & the build succeeded, so pending your guys' approval this should be good to merge. CC @weshaggard @joperezr

eerhardt:
> so pending your guys' approval this should be good to merge

Can you remove the `[WIP]` from the title and the `NoMerge` label?

weshaggard:
@wtgodbe have you done a diff between the tools directory before and after your change? Can you given an estimate of what changed?

wtgodbe:
@weshaggard here's the diff file: https://github.com/wtgodbe/scratch/blob/master/diffTools.txt

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
```diff
@@ -16,14 +16,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
-    <PackageReference Include="Microsoft.Build.Framework" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build.Runtime" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build" Version="15.3.409" />
-    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.0.0" />
     <PackageReference Include="Microsoft.Net.Compilers.netcore" Version="2.6.0-beta3-62316-02" />
```
weshaggard: @joperezr shouldn't Microsoft.Net.compilers.netcore also be removed now?

wtgodbe:
@weshaggard it's actually https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets#L121 that gets hit, and we import C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\Microsoft\Portable\v4.5\Microsoft.Portable.CSharp.targets

wtgodbe:
`RunningOnCore` doesn't have any value set in my log when I msbuild some .csproj

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
joperezr: Yes, we should be using `Microsoft.NETCore.Compilers which we are restoring on the init-tools.cmd and init-tools.sh now.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -26,7 +26,6 @@ set MSBUILD_PROJECT_CONTENT= ^
   ^^^</PropertyGroup^^^> ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
-    ^^^<PackageReference Include=^"Microsoft.Portable.Targets^" Version=^"%PORTABLETARGETS_VERSION%^" /^^^> ^
```
joperezr: You should also remove the copying of the portable targets bellow on this file.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -25,7 +25,6 @@ __PORTABLETARGETS_PROJECT_CONTENT="
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include=\"MicroBuild.Core\" Version=\"$__MICROBUILD_VERSION\" />
-    <PackageReference Include=\"Microsoft.Portable.Targets\" Version=\"$__PORTABLETARGETS_VERSION\" />
```
joperezr: Same here, I believe they get copied somewhere so we should remove that piece of code as well

joperezr:
> __PORTABLETARGETS_VERSION=0.1.1-dev

You should also remove the portable targets version variable, and also remove it from the cmd script

---
Refers to: src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh:17 in c751219. [](commit_id = c751219cae35a002932efd9ba4512fe2d38da04b, deletion_comment = False)

joperezr:
I assume you also tried building corefx with this in a non-Windows machine? In theory that should just work because we use the CLI's msbuild, but I would double check to make sure we didn't delete a dependency of one of our tools or something with this.

joperezr:
> RunningOnCore doesn't have any value set in my log when I msbuild some .csproj

I believe this will only be defined if you build using `dotnet msbuild` which only happens on Linux

wtgodbe:
Updated to address feedback. I got a green build on my Ubuntu VM with the most recent changes.

wtgodbe:
@joperezr is this good to merge?

joperezr:
I don't think you need any of the following either:

https://github.com/dotnet/buildtools/blob/6f2e0975bfb892384007ac1df4f32c0496a6a9c5/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh#L76-L84

But yeah if you can build corefx and coreclr at least in both windows and linux with these changes then this would be a good time to merge this in.

wtgodbe:
I get the following error building CoreClr on Windows:

> E:\code\dotnet\coreclr\Tools\dotnetcli\sdk\2.1.300-rtm-008820\NuGet.targets(114
,5): error : The supplied user buffer is not valid for the requested operation
[E:\code\dotnet\coreclr\tests\scripts\scripts.csproj] [E:\code\dotnet\coreclr\t
ests\build.proj]
E:\code\dotnet\coreclr\tests\build.proj(55,5): error MSB3073: The command ""E:\
code\dotnet\coreclr\tests\..\Tools\dotnetcli\dotnet.exe" restore --packages "E:
\code\dotnet\coreclr\tests\..\packages" --source https://dotnet.myget.org/F/dot
net-corefxlab/api/v3/index.json --source https://dotnetfeed.blob.core.windows.n
et/dotnet-core/index.json --source https://dotnet.myget.org/F/dotnet-core/api/v
3/index.json --source https://api.nuget.org/v3/index.json E:\code\dotnet\corecl
r\tests\scripts\scripts.csproj " exited with code 1.

scripts.csproj is the first restore project specified, so I don't think it's anything specific about that project. The block in NuGet.targets is:

>     <RestoreTask
      RestoreGraphItems="@(_RestoreGraphEntryFiltered)"
      RestoreDisableParallel="$(RestoreDisableParallel)"
      RestoreNoCache="$(RestoreNoCache)"
      RestoreIgnoreFailedSources="$(RestoreIgnoreFailedSources)"
      RestoreRecursive="$(RestoreRecursive)"
      RestoreForce="$(RestoreForce)"
      HideWarningsAndErrors="$(HideWarningsAndErrors)"/>
  </Target>

Where _RestoreGraphEntryFiltered is a duplicate-stripped version of `_RestoreGraphEntry`. Trying to figure out what the invalid buffer is referring to.

wtgodbe:
The CoreClr Linux build failed due to the ILLink.tasks package no longer being restored: https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/CreateRuntimeRootILLinkDescriptorFile.targets#L21

wtgodbe:
Never mind on the Linux failure, the ILLink package is a direct dependency of CoreClr that gets restored during init-tools, but didn't get restored in my build because I overwrote the restore source to be the buildtools dir on my box.

wtgodbe:
CoreClr Linux build is green. Still trying to figure out https://github.com/dotnet/buildtools/pull/2048#issuecomment-394848783

wtgodbe:
I am able to successfully run `"E:\code\dotnet\coreclr\Tools\dotnetcli\dotnet.exe" restore --packages "E:\code\dotnet\coreclr\packages" --source https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json --source https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://api.nuget.org/v3/index.json E:\code\dotnet\coreclr\tests\scripts\scripts.csproj` from the command line. Not sure why it's failing during the build.

wtgodbe:
Looks like the restore bug was likely due to a configuration error on my part - I'm hitting the following in the product build now:

> C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\Common7\IDE\V
C\VCTargets\Microsoft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited
 with code 3. [E:\code\coreclr\bin\obj\Windows_NT.x64.Debug\src\dlls\mscoree\co
reclr\inject_debug_resources.vcxproj]

The surrounding messages aren't very helpful. Will try to get a more detailed log & see if that helps

wtgodbe:
Looks like I had a configuration error last time. Hitting this error now, which looks legit:

> E:\code\coreclr\Tools\dotnetcli\sdk\2.1.300-rtm-008820\Microsoft.CSharp.CurrentVersion.targets(322,5): error MSB4019: The imported project E:\code\coreclr\Tools\Microsoft.CSharp.Core.targets was not found

The import in question looks like this:

> <CSharpCoreTargetsPath Condition="'$(CSharpCoreTargetsPath)' == ''">$(RoslynTargetsPath)\Microsoft.CSharp.Core.targets</CSharpCoreTargetsPath>

Looks like `RoslynTargetsPath` is evaluating to the Tools directory, as per https://github.com/dotnet/buildtools/blob/4aa764b049d9df84e57f89d0450951be99934593/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L6.

wtgodbe:
Entire error message:

> Running: E:\code\coreclr\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /nodeReuse:false  /l:BinClashLogger,Tools/net46/Microsoft.DotNet.Build.Tasks.dll;LogFile=binclash.log  E:\code\coreclr\build.proj /p:__BuildType=Debug /p:__BuildArch=x64 /p:__BuildOS=Windows_NT  /maxcpucount /p:EnableProfileGuidedOptimization=true /p:RestoreDefaultOptimizationDataPackage=false /p:UsePartialNGENOptimization=false /p:OptimizationDataDir=E:\code\coreclr\packages/optimization.Windows_NT-x64.IBC.CoreCLR/E:\code\coreclr\Tools\dotnetcli\sdk\2.1.300-rtm-008820\Microsoft.CSharp.CurrentVersion.targets(322,5): error MSB4019: The imported project E:\code\coreclr\Tools\Microsoft.CSharp.Core.targets was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk. [E:\code\coreclr\src\.nuget\optdata\optdata.csproj]/data/ /p:PortableBuild=true  /flp:Verbosity=normal;LogFile=E:\code\coreclr\bin\Logs\System.Private.CoreLib_Windows_NT__x64__Debug.log  /flp1:WarningsOnly;LogFile=E:\code\coreclr\bin\Logs\System.Private.CoreLib_Windows_NT__x64__Debug.wrn  /flp2:ErrorsOnly;LogFile=E:\code\coreclr\bin\Logs\System.Private.CoreLib_Windows_NT__x64__Debug.err /p:BuildNugetPackage=true

wtgodbe:
From the `pp` log, I can't tell why we'd be importing the wrong version of Microsoft.CSharp.Core.targets:

`<CSharpCoreTargetsPath>$(MSBuildThisFileDirectory)..\tools\Microsoft.CSharp.Core.targets</CSharpCoreTargetsPath>`

This property gets set in E:\code\coreclr\Tools\net46\roslyn\build\Microsoft.Net.Compilers.props, so it should resolve to the right version of `Microsoft.CSharp.Core.targets`. I also can't tell where we're importing `Microsoft.CSharp.CurrentVersion.targets`, as it doesn't show up in the `pp` log.

wtgodbe:
Actually, the error is coming from the invocation of `dotnet.exe` at https://github.com/dotnet/coreclr/blob/master/build.proj#L47. getting a log now.

wtgodbe:
- Running `dotnet msbuild init.csproj /pp:out.txt` gives me an empty log for some reason. 
- Running `dotnet msbuild init.csproj` gives me `coreclr\Tools\Microsoft.CSharp.Core.targets was not found`, from `coreclr\Tools\dotnetcli\sdk\2.1.300-rtm-008820\Microsoft.CSharp.CurrentVersion.targets(322,5)`
- Running `msbuild init.csproj /pp:out.txt` gives me a pp log that imports `C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\15.0\bin\Microsoft.CSharp.CurrentVersion.targets`, which then succesfully imports `coreclr\Tools\net46\roslyn\tools\Microsoft.CSharp.Core.targets`.
- Running `msbuild init.csproj` gives the same error as `dotnet msbuild init.csproj` - that is, `coreclr\Tools\Microsoft.CSharp.Core.targets was not found`, from `coreclr\Tools\dotnetcli\sdk\2.1.300-rtm-008820\Microsoft.CSharp.CurrentVersion.targets(322,5)`

I'm not sure why a `pp` log indicates something different from the actual build. A diag log isn't very helpful either, as it shows no value being set for `CSharpCoreTargetsPath`, and only contains the pre-set properties, an indication it's building `init.csproj`, and the error message.

wtgodbe:
Only thing I can think of now is that we're not importing `Microsoft.Net.Compilers.props` for some reason while running `dotnet msbuild` (this is the file that defines CSharpCoreTargetsPath - if it's not defined before we hit `Microsoft.CSharp.CurrentVersion.targets`, that would trigger us importing the wrong `Microsoft.CSharp.Core.targets`. Here's where we import it: https://github.com/dotnet/coreclr/blob/master/dir.props#L197

`dotnet msbuild -version` gives `15.7.177.53362`, which shouldn't trigger `RoslynIncompatibleMsbuildVersion`, and I don't know why `UseRoslynCompilers` would be false. The file does exist on my machine, so I'm not sure why we wouldn't be importing it...

wtgodbe:
According to my /bl build, we don't import `Microsoft.Net.Compilers.props`. So that's the problem, but I don't know why we wouldn't be importing it.

wtgodbe:
The issue is that `RoslynPropsFile` is evaluating to ` Exists('E:\code\coreclr\packages\/microsoft.netcore.compilers/2.8.0-beta4/build/Microsoft.NETCore.Compilers.props`, which is wrong. That gets computed here: https://github.com/dotnet/buildtools/blob/4aa764b049d9df84e57f89d0450951be99934593/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L19

@joperezr should this be pointing at the "Microsoft.Net.Compilers" package instead? This is present in buildtools package dir, but not "Microsoft.NetCore.Compilers"

wtgodbe:
Neither package is present in Coreclr\packages after 'init-tools.cmd', yet we're looking for Microsoft.NetCore.Compilers relative to a package. Do we expect CoreClr itself to restore this package? looks to me like it should happen in buildtools' init-tools.cmd

wtgodbe:
Alright, so we're hitting this block: https://github.com/dotnet/buildtools/blob/4aa764b049d9df84e57f89d0450951be99934593/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L16

Which is causing the wrong `Microsoft.NETCore.Compilers.props` to get imported - it should be this one:
https://github.com/dotnet/buildtools/blob/4aa764b049d9df84e57f89d0450951be99934593/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L10

The issue is that `RunningOnCore` is being set to `true` (on Windows) here: https://github.com/dotnet/coreclr/blob/master/dir.props#L37

When it should be `false`. Investigating what `MsBuildRuntimeType` is & why it's set to `Core` on my Windows build.

wtgodbe:
Running `dotnet msbuild Project.proj`, where Project.proj contains:

`<Message Text="$(MsBuildRuntimeType)" Importance="High" />`

shows that the property is indeed set to `Core`...

eerhardt:
> Running dotnet msbuild Project.proj, where Project.proj contains:

Correct.  When you use the desktop `msbuild.exe`, it won't be set to `Core`.  But when you use the .NET Core SDK `dotnet build|msbuild|etc`, it will be set to `Core`, because you the build is running on the .NET Core runtime.

wtgodbe:
The issue occurs because, even if we run `msbuild init.csproj`, that winds up invoking the following command:

`"E:\code\coreclr\Tools\dotnetcli\dotnet.exe" restore --packages "E:\code\coreclr\packages" --source https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://api.nuget.org/v3/index.json E:\code\coreclr\src\.nuget\init\init.csproj /p:TargetGroup= /p:ConfigurationGroup= /p:ArchGroup=x64 /p:OSGroup= /p:TargetFramework=`

Which causes `RunningOnCore` to be set to `True`, causing us to look for the wrong version of `Microsoft.NETCore.Compilers.props`

wtgodbe:
https://github.com/dotnet/buildtools/blob/8bc74f90524b14ea0feb178dd5d5352503168aeb/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets#L48-L49 is where this gets invoked. `_DnuRestoreCommandFull` is set to the command I posted in the comment above.

eerhardt:
> causing us to look for the wrong version of Microsoft.NETCore.Compilers.props

Why do we need this `.props` file during restore?  We aren't going to be compiling code....

wtgodbe:
We don't, but it happens to get imported during Restore. When we run with `dotnet.exe`, we try to import it from a location that doesn't exist, causing the build to error out.

wtgodbe:
As per offline discussion with @eerhardt, the new plan is as follows:

- Start importing both `Microsoft.NETCore.Compilers` & `Microsoft.NET.Compilers` in msbuild's init-tools.cmd
- pass the Packages dir from CoreClr/CoreFx's init-tools.cmd to buildtools' init-tools.cmd, so that packages go into the local packages dir, instead of the Tools dir (This matches what we do on Unix) - https://github.com/dotnet/corefx/blob/master/init-tools.cmd#L80

Hopefully that should "just work". Will give it a try after my meeting at 1.

weshaggard:
>Start importing both Microsoft.NETCore.Compilers & Microsoft.NET.Compilers in buildtools' init-tools.cmd

I'm not sure I understand this. Why do you need to import both?

wtgodbe:
@weshaggard because of https://github.com/dotnet/buildtools/blob/4aa764b049d9df84e57f89d0450951be99934593/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L9-L19. When we're running on Core, we want Microsoft.NetCore.Compilers, but when not running on Core, we want Microsoft.Net.Compilers. Windows builds can do both, since we run both msbuild & dotnet.exe during the build. Therefore we need access to both - `RunningOnCore` used to mean the same thing as "not running on windows", but it seems that's no longer the case. When it was, restoring `NetCore` on Unix & `Net` on windows worked fine.

joperezr:
I would ideally suggest this PR to just address the fact of removing the extra msbuild and things we don't use anymore in corefx. From your last comment sounds like you are also doing work to add support for building with dotnet.exe msbuild on windows which is a different workitem. I would try to keep this simple and thus not add the restore of the M.NetCore.compilers to the windows build yet.

wtgodbe:
@joperezr the issue isn't `dotnet msbuild`, but `dotnet restore`, which CoreClr runs here: https://github.com/dotnet/buildtools/blob/8bc74f90524b14ea0feb178dd5d5352503168aeb/src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets#L48-L49

See https://github.com/dotnet/buildtools/pull/2048#issuecomment-395531844

wtgodbe:
As of my last change, we no longer import Microsoft.NetCore.Compilers in init-tools.cmd, and instead stop setting `RoslynTargetsPath` in Roslyn.Common.Props. This allows the SDK to set its own path, and find Microsoft.CSharp.Core.Targets in the correct location. With this change, CoreClr & CoreFx both build without any changes.

@joperezr @eerhardt @weshaggard any other concerns, or should I go ahead and merge?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tool-runtime/project.csproj
weshaggard: So this is still needed because we are using our own custom Roslyn? If so I'm fine with keeping it until we clean that up.

weshaggard:
@wtgodbe I think the term "Importing" was throwing me off as I was expect we were importing both props files but you seem to be using that to mean we are restoring the package in init-tools. I understand now why that is necessary. We will try to clean that up when we work on removing our custom Roslyn, but for this work I'm fine with that.

Issue #2050, opened 2018-05-25
Can crossgen.sh be made to accept RID as argument?


am11:
It seems like the calling script of https://github.com/dotnet/buildtools/blob/d9ae0d7/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh is always aware of the RID under question. Yet this script once again try to resolve RID and does not provide all variants of supported RIDs.

When building CoreRT on Alpine Linux, where RID is `linux-musl-x64`, it throws lots of errors of the following form:

```
Unable to successfully compile /corert/Tools/csc.dll
Unable to successfully compile /corert/Tools/project.dll
Tools/crossgen.sh: line 56: /corert/Tools/../packages/runtime.linux-x64.microsoft.netcore.app/2.1.0-rc1/tools/crossgen: No such file or directory
```

Can RID be passed to crossgen.sh instead of (or in addition to) recalculating it from the call site?

am11:
@weshaggard, @janvorli, this is my steps to repro docker which results in the above error:

```dockerfile
FROM alpine
RUN cat /etc/*-release

RUN apk update
RUN apk add --no-cache \
        autoconf bash build-base clang clang-dev \
        cmake curl-dev gcc gettext-dev git icu-dev libtool \
        linux-headers libunwind-dev llvm make openssl zlib-dev \
        util-linux-dev

# note that the testing URL will be changed once lldb-dev is moved to
# aports' (more prominent) "main" repo.
# See https://github.com/alpinelinux/aports/pull/2342#issuecomment-330894890
RUN apk add --no-cache \
       -X http://dl-cdn.alpinelinux.org/alpine/edge/testing \
        lldb-dev

RUN rm /usr/bin/c++ /usr/bin/g++   # quick hack to select clang++ instead of g++

RUN git clone https://github.com/dotnet/corert -b master --single-branch \
    && cd corert && ./build.sh
```

weshaggard:
I'd be fine with crossgen.sh taking in a rid option or updating to logic to detect the rid better. @am11 are you interested in submitting a PR?

weshaggard:
Fixed with https://github.com/dotnet/buildtools/pull/2051.

PR #2051, opened 2018-05-25
Update crossgen.sh to accept RID arg

am11:
Addresses #2050

@weshaggard, this is backwards compatible change. We can probably remove the `case` block once/if the usage is updated everywhere.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
@@ -86,23 +86,25 @@ __mncaFolder=$__toolsDir/dotnetcli/shared/Microsoft.NETCore.App
 __sharedFxVersion=`ls $__mncaFolder | sed 'r/\([0-9]\+\).*/\1/g' | sort -n | tail -1`
 __sharedFxDir=$__toolsDir/dotnetcli/shared/Microsoft.NETCore.App/$__sharedFxVersion/
 
-# TODO: once all call-sites start passing <rid> argument to this script
-#       remove case block and replace with error and exit statements
 if [ -z "$__packageRid" ]; then
     case $(uname -s) in
         Darwin)
             __packageRid=osx-x64
-            __libraryExtension=dylib
             ;;
         Linux)
             __packageRid=linux-x64
-            __libraryExtension=so
             ;;
         *)
             echo "Unsupported OS $(uname -s) detected. Skipping crossgen of the toolset."
             exit 0
             ;;
     esac
+fi
+
+if [ "$__packageRid" == "osx-x64" ]; then
+    __libraryExtension=dylib
+else
+    __libraryExtension=so
 fi
 
 restore_crossgen
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 # Restores crossgen and runs it on all tools components.
 usage()
 {
-    echo "crossgen.sh <directory>"
+    echo "crossgen.sh <directory> <rid>"
     echo "    Restores crossgen and runs it on all assemblies in <directory>."
     exit 0
 }
@@ -77,6 +77,7 @@ fi
 
 __MyGetFeed=${BUILDTOOLS_CROSSGEN_FEED:-https://dotnet.myget.org/F/dotnet-core/api/v3/index.json}
 __targetDir=$1
+__packageRid=$2
 __scriptpath=$(cd "$(dirname "$0")"; pwd -P)
 __toolsDir=$__scriptpath/../Tools
 __dotnet=$__toolsDir/dotnetcli/dotnet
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -85,20 +86,24 @@ __mncaFolder=$__toolsDir/dotnetcli/shared/Microsoft.NETCore.App
...
+        Linux)
+            __packageRid=linux-x64
+            __libraryExtension=so
+            ;;
+        *)
+            echo "Unsupported OS $(uname -s) detected. Skipping crossgen of the toolset."
+            exit 0
+            ;;
+    esac
+fi
```
weshaggard: You need to also initialize the `__libraryExtension` property. Just put it after this step and set it to be `dylib` for if `__packageRid` is `osx-x64` otherwise set it to `so`.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -85,20 +86,24 @@ __mncaFolder=$__toolsDir/dotnetcli/shared/Microsoft.NETCore.App
...
-    Linux)
-        __packageRid=linux-x64
-        __libraryExtension=so
-        ;;
-    *)
-        echo "Unsupported OS $(uname -s) detected. Skipping crossgen of the toolset."
-        exit 0
-        ;;
-esac
+# TODO: once all call-sites start passing <rid> argument to this script
```
weshaggard: I would just remove the TODO. It is OK to keep this as a default fallback for anyone using the script that doesn't pass a rid.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
am11: Done, set `_libraryExtension` to `so`  by default that will be overwritten in case of OSX.

weshaggard: If someone passes in osx-x64 as the rid we still need to set the extension to `dylib` so I think it should be done outside of the check here.

am11: Didn't thought of that case. Fixed. :)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -85,20 +86,27 @@ __mncaFolder=$__toolsDir/dotnetcli/shared/Microsoft.NETCore.App
...
-    Linux)
-        __packageRid=linux-x64
-        __libraryExtension=so
-        ;;
-    *)
-        echo "Unsupported OS $(uname -s) detected. Skipping crossgen of the toolset."
-        exit 0
-        ;;
-esac
+__libraryExtension-so
```
weshaggard: typo: `-` -> `=`. You can probably just remove this now that you have it below.

weshaggard:
@am11 thanks for the update. Is there anyway you can test this locally? Just copy over the changes in your local buildtools directory and give it a try.

am11:
@weshaggard, i have tested it with this patch https://github.com/am11/corert/commit/a6e5bbec7819b04278c7824123bbd02c19dc4f9d, changed the docker script (https://github.com/dotnet/buildtools/issues/2050#issuecomment-392153052) a bit to point to fork branch. The crossgen error due to wrong RID is fixed, However, the next error is about missing package:

```
The crossgen executable could not be found at /corert/Tools/../packages/runtime.linux-musl.microsoft.netcore.app/2.1.0-rc1/tools/crossgen. Aborting crossgen.sh.
```

Is this CoreRT specific issue?

am11:
My bad, RID was missing host architecture in CoreRT script. With https://github.com/am11/corert/commit/ef56f8d59b68, crossgen is working. However, later down in the process, `ilc.csproj` throws:

```
/corert/src/ILCompiler/netcoreapp/ilc.csproj : error : Project is targeting runtime 'linux-musl-x64' but did not resolve any runtime-specific packages. This runtime may not be supported by the target framework.
```

which is unrelated issue that I will try to fix in CoreRT, once these changes are propagated.

@weshaggard, I also ran the linter on this script (https://www.shellcheck.net/) and no issues in the lines I am touching. However, there are some good suggestions in other areas of this script.

weshaggard:
@am11 thanks for the contribution.

weshaggard:
>@weshaggard, I also ran the linter on this script (https://www.shellcheck.net/) and no issues in the lines I am touching. However, there are some good suggestions in other areas of this script.

If you feel they are worth fixing please feel free to file and issue and/or submit a PR. Thanks.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
```diff
@@ -77,6 +77,7 @@ fi
 
 __MyGetFeed=${BUILDTOOLS_CROSSGEN_FEED:-https://dotnet.myget.org/F/dotnet-core/api/v3/index.json}
 __targetDir=$1
+__packageRid=$2
```
dagood: This causes an error when only one parameter is passed: `$2` is unbound and this script has `set -euo pipefail`. I think you need `${2:-}`.

/cc @safern

am11: @dagood, I missed pipefaul and I didn’t knew its function, will send a fix. Sorry for the troubles. :(

dagood: Thanks! FWIW, it's the `u` that's the direct cause. (That command is equivalent to `set -e -u -o pipefail`.) http://wiki.bash-hackers.org/commands/builtin/set has some info about this and I've been finding that wiki a pretty good reference in general.

safern: I can add the fix to my PR itself if that is quicker.

safern: Sorry, I didn't read the location of the file affected :neutral_face:, I thought it was in this repo. Anyway I can make one to fix this :smile:

dagood: I see--just to clarify then: this thread is on a BuildTools PR, so it is this repo. 😄

I think it's up to you and @am11 who should make the PR. Since it's so small, maybe just whoever gets to it first.

safern: > I see--just to clarify then: this thread is on a BuildTools PR, so it is this repo. 😄

haha I know, just miss-reading and commenting without looking carefully at where/what was affected 😆

PR #2054, opened 2018-06-01
Include dumpasync command when evaluating core dumps

stephentoub:
https://github.com/dotnet/coreclr/pull/18160 and https://github.com/dotnet/coreclr/pull/18213 added a new `dumpasync` command to sos that will dump out data on async methods currently in use.  Including this in the core dump analysis has similar value to dumping out all call stacks, in that it'll give us more details into what was going on in the process.

cc: @danmosemsft, @Anipik


danmosemsft:
fyi @ahsonkhan who has taken over fixing this experience.

danmosemsft:
I assume your SOS changes arrive for free in CoreFX with CoreCLR (at least right now)?

stephentoub:
> I assume your SOS changes arrive for free in CoreFX with CoreCLR (at least right now)?

Yeah.  It's not there yet, but it should be in the next coreclr update consumed into corefx.

PR #2056, opened 2018-06-01
GenAPI features

ericstj:
I'm adding a few features to GenAPI that I needed for contract assembly refactoring.

See individual commits for details.  These are all guarded by new parameters.


weshaggard:
Looks reasonable.

PR #2061, opened 2018-06-09
Set default value of RID to empty string

am11:
Based on https://github.com/dotnet/buildtools/pull/2051#discussion_r194077549

@dagood, TIL, http://redsymbol.net/articles/unofficial-bash-strict-mode/. Thanks for your pointers too! :)

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/crossgen.sh
@@ -77,7 +77,7 @@ fi
 
 __MyGetFeed=${BUILDTOOLS_CROSSGEN_FEED:-https://dotnet.myget.org/F/dotnet-core/api/v3/index.json}
 __targetDir=$1
-__packageRid=$2
+__packageRid=${2:-}
 __scriptpath=$(cd "$(dirname "$0")"; pwd -P)
 __toolsDir=$__scriptpath/../Tools
 __dotnet=$__toolsDir/dotnetcli/dotnet


am11:
On a hindsight, should we (also) pursue to update the call-sites of this script and pass RID, as all of those places should already know what is the target RID? The issue is the RID detection in this script is naïve and obsolete. CoreCLR, CoreFX and CoreRT have the most advanced way of determining the RID for the target, then this script downgrades all that to old-style Linux/Mac-only RIDs,

dagood:
Good call on looking that up, I didn't think to point at why we're using it. The IFS one is interesting too.

> On a hindsight, should we (also) pursue to update the call-sites of this script and pass RID, as all of those places should already know what is the target RID?

Hearing this, I'm more inclined to say that the logic should be centralized somehow... but since we don't have a great mechanism for that, removing this single outdated duplicate seems like a reasonable goal. I'm not so sure it's worthwhile to make the effort, though, since https://github.com/dotnet/arcade will be taking over the initialization process at some point.

dagood:
By the way, thanks for the fix. 😃 

/cc @safern @weshaggard

safern:
Thanks for the fix!

am11:
@dagood, dotnet/arcade looks promising, will the .NET Core repos switch over to it in near future?

I was looking at a recent Linux-musl build in CoreFX (<a href="https://ci3.dot.net/job/dotnet_corefx/job/master/job/linux-musl-TGroup_netcoreapp%2BCGroup_Debug%2BAGroup_x64%2BTestOuter_false_prtest/1581/consoleText">plaintext</a>), if we search "crossgen" on the page, we will find lots of `crossgen: No such file or directory`. Would it make sense to simply pass RID to crossgen in corefx to fix it meanwhile? 🤓

dagood:
Yep, that's the plan.

Passing in RID in CoreFX seems fine to me. (I haven't worked much in CoreFX recently though, FWIW.) My thought is mainly that it doesn't seem worth it to try to *remove* the logic that already exists here, since it's hard to know for sure it isn't relied on somewhere unexpected. Passing in RIDs in places where it makes sense (even across *all* the .NET Core Runtime repos) sounds good, then at some point I expect Arcade to centralize platform detection, the crossgen script, and the tool init code that calls it all.

weshaggard:
I think it is fine to incrementally pass in the RID in places we already have it to pass in but I would still like to keep the fallback RID detection in this script for now.

PR #2066, opened 2018-06-13
Replaces CoreLib with respective IL when doing coverage runs

pjanotti:
This is another part of https://github.com/dotnet/corefx/issues/23588

This one replaces CoreLib crossgen in coverage runs that specified it as one of the included assemblies. It tries to restore the crossgen version at the end of the coverage run, but also whenever starting any test to avoid running against the IL version unexpectedly.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -16,7 +16,7 @@
     <CodeCoverageEnabled Condition="'$(SkipTests)' != 'true' and '$(RunningOnUnix)' != 'true' and '$(Coverage)' == 'true' and '$(Performance)' != 'true'">true</CodeCoverageEnabled>
     <CoverageReportDir Condition="'$(CoverageReportDir)' == ''">$(TestWorkingDir)coverage\</CoverageReportDir>
 
-    <RestoreCrossgenWithHardlinks Condition="'$(RestoreCrossgenWithHardlinks)' == ''">true</RestoreCrossgenWithHardlinks>
+    <RestoreIlPdbsWithHardlinks Condition="'$(RestoreIlPdbsWithHardlinks)' == ''">true</RestoreIlPdbsWithHardlinks>
 
     <!-- This targets file has two modes one for individual projects and one for all -->
     <GenerateCodeCoverageReportForAll Condition="'$(GenerateCodeCoverageReportForAll)'==''">false</GenerateCodeCoverageReportForAll>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -50,6 +52,47 @@
...
+        Condition="Exists('$(NETCoreAppTestSharedFrameworkPath)/il/%(Identity).dll') AND Exists('$(NETCoreAppTestSharedFrameworkPath)/ni/%(Identity).dll')">
+          <NativeImagePath>$(NETCoreAppTestSharedFrameworkPath)%(Identity).dll</NativeImagePath>
+      </SelectedIlAssemblies>
+    </ItemGroup>
+
+    <!-- Delete the files under the test host to break any existing hardlinks -->
+    <Delete Files="@(SelectedIlAssemblies->'%(NativeImagePath)')" />
+
+    <!-- Put IL versions in place -->
+    <Copy SourceFiles="@(SelectedIlAssemblies)"
```
weshaggard: I worry about all this copying back and forth as it is likely to cause us to get into weird states. It only takes one locked dll because of a test not shutting down correctly to corrupt this output. I wonder if it might be better to create a parallel output folder to do the the code coverage run on.

weshaggard:
adding @stephentoub @danmosemsft who do more with code coverage.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
pjanotti: I agree with the sentiment, I will do some tries copying everything to a separate folder to see how it goes.

pjanotti: https://github.com/dotnet/coreclr/pull/11040#issuecomment-297902475 ... I wish I had found this earlier, but, anyway I opened an issue with CoreClr to see if this is something that can be fixed on the profiler side. I will put this on the back burner until I hear back in https://github.com/dotnet/coreclr/issues/18488

pjanotti: (See https://github.com/dotnet/corefx/issues/23588#issuecomment-397832843 for some background info)

For tools doing IL rewrite it is basically required to use a separate folder, but for tools using the Profiler API that shouldn't be necessary. That said for the version of OpenCover (4.6.519) being used in CoreFx we need some workaround since it can't match the proper PDB for native images that follow the convention of shipping .ni.pdb (matching native image) and .pdb (matching IL image) - even if PDBs are moved to a different folder.

In the short run I want to reserve the separate folder for tools doing IL rewrite and workaround OpenCover issues by moving PDB files around: there still be some risky for tools depending on symbols but much smaller than using the assemblies directly. After that I will try to get a change to OpenCover in which it can properly find PDBs for native images, so we can stop moving PDBs around.

Notice that, as before, there is an attempt to ensure expected state for non-coverage runs.

pjanotti:
Switching pdbs instead of dlls to reduce the risk of side effects. Notice the targets try to ensure the state of pdbs for non-coverage runs.

weshaggard:
If we aren't going to need the IL file any longer in the package then why wouldn't we just update our tooling to download the ni pdb instead of copying it from the coreclr package? It would seem we just need to update the code that pulls in the PDB instead of copying them around like this.

pjanotti:
> If we aren't going to need the IL file any longer in the package then why wouldn't we just update our tooling to download the ni pdb instead of copying it from the coreclr package?

We will need the IL file because we are going to use IL rewrite to get coverage on Unix (that will be a copy folder). Per conversation [here](https://github.com/dotnet/corefx/issues/23588#issuecomment-395611134) I should not change `ni.pdb` and `.pdb` for the affected assemblies in general.

pjanotti:
Observation: I will try to get a fix in OpenCover so we don't depend on this workaround, the idea is to eventually remove this copy of files but this gets coverage for CoreLib in CI back in the short run.

PR #2067, opened 2018-06-14
Don't mark ReferenceAssemblies with APTCA

ericstj:
We've stopped putting all security annotations on our assemblies so this should be removed as well.


ericstj:
Can you elaborate on those scenarios where lying about APTCA is beneficial?

I actually found otherwise.  Some desktop code that I was porting that still had attributes wouldn’t compile against these refs because some tools were complaining about ref-assembly defined members being transparent rather than critical.  I can elaborate more in person.

I think this attribute was only valuable when we still had security attributes in the references.

PR #2068, opened 2018-06-14
Support using the Microsoft.NET.Sdk with dotnet/buildtools

eerhardt:
- Don't import Microsoft.$(Language).targets in buildtools - the SDK will do that.
- Use the Sdk's GenerateAssemblyInfo target when possible

Still support non-SDK projects for backwards compatiblity.

Note: I removed support for @(AssemblyInfoSource) during GenerateAssemblyInfo because it is no longer used as of dotnet/corefx/pull/30400.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -104,14 +104,14 @@
           Condition="'$(MSBuildProjectExtension)' == '.depproj'" />
 
   <Import Project="IL.targets"
-           Condition="'$(MSBuildProjectExtension)' == '.ilproj' AND '$(SkipImportILTargets)'!='true'" />
+           Condition="'$(MSBuildProjectExtension)' == '.ilproj' AND '$(SkipImportILTargets)'!='true' and '$(UsingMicrosoftNETSdk)' != 'true'" />
 
-  <!--
-      workaround file casing issue where it has different casing in different places which fails on linux builds
-      https://github.com/dotnet/buildtools/issues/1464
-  -->
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true'">
     <CSharpTargetsFile>$(MSBuildToolsPath)\Microsoft.CSharp.targets</CSharpTargetsFile>
+    <!--
+        workaround file casing issue where it has different casing in different places which fails on linux builds
+        https://github.com/dotnet/buildtools/issues/1464
+    -->
     <CSharpTargetsFile Condition="!Exists('$(CSharpTargetsFile)')">$(MSBuildToolsPath)\Microsoft.CSharp.Targets</CSharpTargetsFile>
   </PropertyGroup>
 
@@ -123,7 +123,7 @@
  
 
   <Import Project="$(CSharpTargetsFile)"
-          Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj'" />
+          Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj' and '$(UsingMicrosoftNETSdk)' != 'true'" />
 
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft.VisualBasic.targets"
            Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj' and '$(RunningOnCore)' == 'true'" />
@@ -133,7 +133,7 @@
           Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj' and '$(RunningOnCore)' != 'true'" />
 
   <Import Project="$(MSBuildToolsPath)\Microsoft.VisualBasic.targets"
-          Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj'" />
+          Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.vbproj' and '$(UsingMicrosoftNETSdk)' != 'true'" />
 
   <PropertyGroup>
     <!--
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
@@ -1,0 +1,66 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <!-- BuildTools never generated this attribute before, so don't let the SDK generate it. -->
+    <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
+
+    <!--
+    When using the Microsoft.NET.Sdk, Company and Product are defaulted to $(AssemblyName)
+    in Microsoft.NET.Sdk.BeforeCommon.targets, which means we would have to default them in
+    a props file. However, the versioning.props file has to be imported during versioning.targets
+    in order to get all the version properties set correctly.
+    So overwrite these two properties here, so they are set correctly.
+    -->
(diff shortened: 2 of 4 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -123,7 +123,7 @@
  
 
   <Import Project="$(CSharpTargetsFile)"
-          Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj'" />
+          Condition="'$(TargetFrameworkIdentifier)' != '.NETPortable' and '$(MSBuildProjectExtension)' == '.csproj' and '$(UsingMicrosoftNETSdk)' != 'true'" />
```
joperezr: Just wondering, do we have any project we build that targets .NETPortable? In case we don't then it seems like the above and bellow imports are dead code. If we do have .NETPortable projs, wouldn't we want to add the condition `'$(UsingMicrosoftNETSdk)' != 'true'`

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -41,8 +41,7 @@
   </PropertyGroup>
 
   <Target Name="DecideIfWeNeedToIncludeDllSafeSearchPathAttribute" 
-    Condition="'$(IsDotNetFrameworkProductAssembly)' == 'true' AND '$(IsTestProject)' != 'true'"
-    BeforeTargets="GenerateAssemblyInfo">
+    Condition="'$(IsDotNetFrameworkProductAssembly)' == 'true' AND '$(IsTestProject)' != 'true'">
```
joperezr: Do we not need this guy any longer? If not then we should probably remove it since I believe that with your change we are basically removing it from the build.

eerhardt: We do still need it.  I just switched how it is invoked.  Instead of being `BeforeTargets`, I made the other targets Depend on this.  With the new SDK, it doesn't make sense for this to be `BeforeTargets=GenerateAssemblyInfo`.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
eerhardt: I'm not sure, and I'm not completely sure how to check. Do you or @weshaggard know?  I can remove all the .NETPortable stuff, if we are no longer using it.

AFAIK, the NET.Sdk doesn't support .NETPortable at all, which is why this PR was made, but never merged: https://github.com/dotnet/sdk/pull/889.  So if we still still need .NETPortable, we probably can't use the NET.Sdk without doing some work ourselves.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -104,14 +104,14 @@
           Condition="'$(MSBuildProjectExtension)' == '.depproj'" />
 
   <Import Project="IL.targets"
-           Condition="'$(MSBuildProjectExtension)' == '.ilproj' AND '$(SkipImportILTargets)'!='true'" />
```
weshaggard: I assume this means our ilproj's stop working? If they don't work as sdk projects then we don't need this condition but if they do then we do still need the import.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
weshaggard: My goal will be to eventually clean out most if not all of this targets file. Lets keep that separate from the initial switching to sdk projects though.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
```diff
@@ -0,0 +1,66 @@
...
+    <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
+
+    <!--
+    When using the Microsoft.NET.Sdk, Company and Product are defaulted to $(AssemblyName)
+    in Microsoft.NET.Sdk.BeforeCommon.targets, which means we would have to default them in
+    a props file. However, the versioning.props file has to be imported during versioning.targets
+    in order to get all the version properties set correctly.
+    So overwrite these two properties here, so they are set correctly.
+    -->
+    <Company>Microsoft Corporation</Company>
```
weshaggard: We may need a way to override these at some point.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
```diff
@@ -0,0 +1,66 @@
...
+    <Product>Microsoft%AE .NET Framework</Product>
+  </PropertyGroup>
+
+  <Target Name="_ComputeBuildToolsAssemblyInfoAttributes"
+          DependsOnTargets="CreateVersionFileDuringBuild;DecideIfWeNeedToIncludeDllSafeSearchPathAttribute"
+          BeforeTargets="GetAssemblyAttributes">
+    <ItemGroup>
+      <AssemblyAttribute Include="System.Reflection.AssemblyDefaultAliasAttribute">
```
weshaggard: Is it worth sharing more logic between the sdk and nonsdk version of these files and essentially just condition an extra set of item groups?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
eerhardt: Yea, `ilproj` is a bit tricky, unfortunately.

We DO use the SDK style project for it.  However, the SDK doesn't have support for the `IL` language.  It only ships with support for C#, VB and F#. In order to make other languages work correctly, there is a property called `$(LanguageTargets)` which gets imported at the correct place.
See https://github.com/dotnet/sdk/blob/f8542e15d2dc54bc7fcce32bdf5046cde4483d9a/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.targets#L28-L41.

So in the one `ilproj` we have in dotnet/corefx, I'm setting `<LanguageTargets>$(ToolsDir)\IL.targets<LanguageTargets>`, which allows it to be imported correctly with the SDK.  See https://github.com/dotnet/corefx/pull/29831/commits/be5ceb5379738fab0a071f98e63f18496bc0388a#diff-89d3c36d89fc54d71aa52b9438ac028cR8 for the corefx change.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
eerhardt: My hope was the nonsdk version just goes away and we pull the SDK version back into versioning.targets.

I'm not seeing how more logic can be shared.  Can you explain more?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
eerhardt: Do you want me to add a property that can disable overwriting these properties?

My thought was, if you aren't a Microsoft .NET Framework assembly, you probably shouldn't be using buildtools.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
weshaggard: The writing of the assmeblyinfolines into a file is all shared. In once case you simply need to add more items to the itemgroup then the other case. At any rate I agree with you that I only want the one at the end of the day so I'm not that worried about it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/GenerateAssemblyInfo.Sdk.targets
eerhardt: You can override them with this change.  You just need to do it after the buildtools versioning.targets file is imported.

weshaggard: I'm fine with what we have for now. I can potentially imagine scenarios were we want a mix of at least the product name for different assemblies.

PR #2069, opened 2018-06-15
Allow auto-updating from on-disk ProdCon manifest; add SkipIfNoReplacementFound

dagood:
On-disk manifest allows you to download or generate a manifest, then use it to update without uploading it to GitHub (dotnet/versions) first.

`SkipIfNoReplacementFound` reduces the maintenance burden in source-build. Right now we have to disable updaters when ProdCon stops building particular repos. More importantly, we have to remember to enable them again later or we start missing updates. With this attribute, we can leave the updaters enabled all the time.

If this looks ok to merge, I'm porting to `release/2.1` immediately.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
@@ -7,6 +7,7 @@ using Microsoft.DotNet.VersionTools;
 using Microsoft.DotNet.VersionTools.Automation;
 using Microsoft.DotNet.VersionTools.Automation.GitHubApi;
 using Microsoft.DotNet.VersionTools.BuildManifest;
+using Microsoft.DotNet.VersionTools.BuildManifest.Model;
 using Microsoft.DotNet.VersionTools.Dependencies;
 using Microsoft.DotNet.VersionTools.Dependencies.BuildManifest;
 using Microsoft.DotNet.VersionTools.Dependencies.BuildOutput;
@@ -15,8 +16,10 @@ using Microsoft.DotNet.VersionTools.Dependencies.Submodule;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
+using System.Xml.Linq;
 
 namespace Microsoft.DotNet.Build.Tasks.VersionTools
 {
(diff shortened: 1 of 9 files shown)


dagood:
Seems weird to me that I'm hitting it, even though `XElement` is used a decent amount elsewhere in the same project.

> BaseDependenciesTask.cs(216,29): error CS0012: The type 'XElement' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Xml.XDocument, Version=4.0.10.0 [...] src/Microsoft.DotNet.Build.Tasks.net45/Microsoft.DotNet.Build.Tasks.net45.csproj

Pulling a container to try to repro in.

dagood:
I brought forward the part of https://github.com/dotnet/buildtools/commit/f8444e2bfdd036a2de0fcc7041b12ccc22c09168 that fixes the problem.

src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -51,7 +51,6 @@
     <file src="Microsoft.DotNet.Build.Tasks.Packaging.Desktop\Microsoft.DotNet.Build.Tasks.Packaging.dll" target="lib\net46" />
     <file src="Microsoft.DotNet.Build.Tasks.Packaging.Desktop\NuGet.*.dll" target="lib\net46" />
     <file src="Microsoft.DotNet.VersionTools\Microsoft.DotNet.VersionTools.dll" target="lib" />
-    <file src="Microsoft.DotNet.VersionTools.net45\Microsoft.DotNet.VersionTools.dll" target="lib\net46" />
```
eerhardt: What's the story here?

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -240,6 +254,16 @@ private FileRegexUpdater CreateXmlUpdater(ITaskItem step)
             return updater;
         }
 
+        private FileUpdater ConfigureFileUpdater(FileUpdater updater, ITaskItem step)
```
eerhardt: I'm curious why FileRegexUpdater doesn't inherit from FileUpdater....

dagood: Yeah, structure isn't great there. `FileRegexUpdater` came first, and it never got a significant refactor. For this and the last PR I touched these I didn't want to take up the effort to avoid breaking something. (No automated testing to speak of.)

src/nuget/Microsoft.DotNet.BuildTools.nuspec
dagood: Looking at it deeper, it shouldn't be necessary. Reverting it and the `<Private>False</Private>` removal. All I really needed to fix the XML DLL load error was the `$(VersionToolsProjectReferenceSuffix)`.

It's originally from https://github.com/dotnet/buildtools/pull/1935. I wanted to be able to load the built build task DLL during the official build to use it to publish to dotnet/versions, but it wouldn't load because the VersionTools DLL couldn't be found. I removed `<Private>false</Private>` so VersionTools would be in the same dir and the load worked. Then this nuspec line was doing a duplicate copy so I removed it.

PR #2077, opened 2018-06-22
Create a runtime dedicated to coverage using IL only

pjanotti:
This allows OpenCover to get source information for CoreLib (perhaps an issue to be investigated separately, since in theory it should have worked with the ni image and respective PDB). It also makes easier to experiment with tools that perform IL rewrite instead of relying on Profiler API.

Related to https://github.com/dotnet/corefx/issues/23588

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -17,6 +17,7 @@
     <CoverageReportDir Condition="'$(CoverageReportDir)' == ''">$(TestWorkingDir)coverage\</CoverageReportDir>
 
     <RestoreIlPdbsWithHardlinks Condition="'$(RestoreIlPdbsWithHardlinks)' == ''">true</RestoreIlPdbsWithHardlinks>
+    <UseCoverageDedicatedRuntime Condition="'$(UseCoverageDedicatedRuntime)' == ''">true</UseCoverageDedicatedRuntime>
 
     <!-- This targets file has two modes one for individual projects and one for all -->
     <GenerateCodeCoverageReportForAll Condition="'$(GenerateCodeCoverageReportForAll)'==''">false</GenerateCodeCoverageReportForAll>
@@ -53,7 +54,9 @@
   </Target>
 
   <!-- Workaround OpenCover symbol matching behavior by moving pdb files for coverage runs -->
-  <Target Name="ReplaceIlPdbsOnTestHost"  BeforeTargets="GenerateTestExecutionScripts" Condition="'$(CodeCoverageEnabled)'=='true'">
+  <Target Name="ReplaceIlPdbsOnTestHost"
+    BeforeTargets="GenerateTestExecutionScripts"
+    Condition="'$(CodeCoverageEnabled)'=='true' and '$(UseCoverageDedicatedRuntime)'!='true'">
     <Message Importance="high" Text="Replacing IL pdbs with respective pdbs for crossgen images (workaround OpenCover issue)." />
     <ItemGroup>
       <SelectedNiPdbFiles Include="@(_CodeCoverageAssemblies -> '$(NETCoreAppTestSharedFrameworkPath)ni/%(Identity).ni.pdb')"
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -131,6 +91,38 @@
   </Target>
   <!-- *********************************************************************************************** -->
 
+  <Target Name="CreateCoverageDedicatedRuntime"
+    BeforeTargets="GenerateTestExecutionScripts"
+    DependsOnTargets="GenerateWindowsPdbsForAssemblyBeingTested"
+    Condition="'$(CodeCoverageEnabled)'=='true' and '$(UseCoverageDedicatedRuntime)'=='true'">
+    <PropertyGroup>
+      <CoverageDedicatedRuntimeDir>$(TestHostRootPath)shared/Microsoft.NETCore.App/10.10.10</CoverageDedicatedRuntimeDir>
```
tarekgh: >10.10.10 [](start = 83, length = 8)

do we'll need to update this version every while?

tarekgh: how this will work with the private runs? I mean when someone uses /p:CoreClrOverridePath?

---
In reply to: [197517955](https://github.com/dotnet/buildtools/pull/2077#discussion_r197517955) [](ancestors = 197517955)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -16,7 +16,7 @@
     <CodeCoverageEnabled Condition="'$(SkipTests)' != 'true' and '$(RunningOnUnix)' != 'true' and '$(Coverage)' == 'true' and '$(Performance)' != 'true'">true</CodeCoverageEnabled>
     <CoverageReportDir Condition="'$(CoverageReportDir)' == ''">$(TestWorkingDir)coverage\</CoverageReportDir>
 
-    <RestoreIlPdbsWithHardlinks Condition="'$(RestoreIlPdbsWithHardlinks)' == ''">true</RestoreIlPdbsWithHardlinks>
+    <UseCoverageDedicatedRuntime Condition="'$(UseCoverageDedicatedRuntime)' == ''">true</UseCoverageDedicatedRuntime>
```
tarekgh: >UseCoverageDedicatedRuntime [](start = 5, length = 27)

can this be just the Microsoft.NETCore.App version instead of true/false value? and if the version set, we use it, otherwise we'll skip it.

pjanotti: I think this will be confusing: it is not clear what is the side effect of each version, having as a bool make clear of the behavior here.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
tarekgh: Is there anyway we can detect the version of the used corelib automatically? instead of hardcoding version values? I am not expert here but I can imagine we can get what we need from the runtime folder.

pjanotti:
FYI: Even after reviewed I will hold off merge until I'm able to fully test this in CoreFx (it takes a bit of time to run it when coverage is enabled).

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
tarekgh: I don't have a strong feeling here but I can imagine anyone choose specific version will guarantee it is a good version to work with.

tarekgh:
@pjanotti don't block on my comments if @weshaggard and @vancem are ok with the changes. thanks for fixing this :-)

pjanotti:
Dev box runs for coverage are looking good.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
pjanotti: This is the behavior of runtimeconfig.json when we use dotnet.exe, AFAIK we have to specify it in the file and then it will look in the respective folder.

pjanotti:
@weshaggard, please, take a look when you have a chance.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -131,6 +91,38 @@
...
+    Condition="'$(CodeCoverageEnabled)'=='true' and '$(UseCoverageDedicatedRuntime)'=='true'">
+    <PropertyGroup>
+      <CoverageDedicatedRuntimeDir>$(TestHostRootPath)shared/Microsoft.NETCore.App/10.10.10</CoverageDedicatedRuntimeDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <RuntimeFiles Include="$(NETCoreAppTestSharedFrameworkPath)/*.*" />
+      <RuntimeFiles Include="$(NETCoreAppTestSharedFrameworkPath)il/*.*" />
+      <RuntimeFilesAmd64 Include="$(NETCoreAppTestSharedFrameworkPath)amd64/*.*" />
+      <RuntimeFilesx86 Include="$(NETCoreAppTestSharedFrameworkPath)x86/*.*" />
+      <RuntimeWindowsPdbs Include="$(NETCoreAppTestSharedFrameworkPath)WindowsPDB/*.*" />
```
weshaggard: What is putting the WindowsPDB directory here? If it is only being created for coverage can we start putting it somewhere else instead?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -131,6 +91,38 @@
...
+    </ItemGroup>
+    <Copy SourceFiles="@(RuntimeFiles)"
+          DestinationFolder="$(CoverageDedicatedRuntimeDir)"
+          SkipUnchangedFiles="true"
+          UseHardlinksIfPossible="false" />
+    <Copy SourceFiles="@(RuntimeFilesAmd64)"
```
weshaggard: Is there any expected overwriting happening? If so it is probably worth considering producing the filelist in an item group to merge them and remove duplicates before doing the copy to avoid overwriting.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/coverage/xunit.console.netcore.runtimeconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+	"runtimeOptions": {
+		"framework": {
+			"name": "Microsoft.NETCore.App",
+			"version": "10.10.10"
```
weshaggard: Is there are reason you chose 10.10.10? I don't have any particular concerns just curious about the choice.

weshaggard:
I asked a couple questions but the general approach looks good.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
pjanotti: That's a good point: we could start to have it only on the coverage folder. Let me see what I can do about it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
pjanotti: I expect that to happen just to System.Private.CoreLib all others should not cause any overwrite.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/coverage/xunit.console.netcore.runtimeconfig.json
pjanotti: I just needed it to be bigger than 2.* something so the runtime didn't complain, since 9.9.9 already existed I took the next one.

pjanotti:
I had to add some explicit copying/deleting for the runtimeconfig.json since SupplementalTestData by default use hard links (we don't want to overwrite the one using the "default" shared test runtime). Planning to merge as soon as tests are passing.

PR #2058, opened 2018-06-04
Update MicrosoftDiaSymReaderConverterPackageVersion to latest 1.1.0-beta version

ahsonkhan:
Fixes https://github.com/dotnet/corefx/issues/29861

cc @danmosemsft, @tmat, @stephentoub

PR #2079, opened 2018-06-25
Package Microsoft.DiaSymReader assemblies in buildtools nuget package.

eerhardt:
This allows GenFacades (and other tasks that depend on DiaSymReader) to use DiaSymReader on .NET Core's MSBuild.

This is needed in order to get corefx's build on Windows to use .NET Core MSBuild.

### DIFF
--- a/src/nuget/Microsoft.DotNet.BuildTools.nuspec
+++ b/src/nuget/Microsoft.DotNet.BuildTools.nuspec
@@ -26,6 +26,7 @@
     <file src="Microsoft.DotNet.Build.CloudTestTasks\RunnerScripts\**\*.*" target="lib\RunnerScripts" />
     <file src="Microsoft.DotNet.Build.Tasks\PackageFiles\**\*.*" target="lib" />
     <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DotNet.Build.Tasks.dll" target="lib" />
+    <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DiaSymReader*.dll" target="lib" />
     <file src="Run\run.exe" target="lib" />
     <file src="Run\run.runtimeconfig.json" target="lib" />
     <file src="BclRewriter\BclRewriter.exe" target="lib" />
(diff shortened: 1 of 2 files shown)


src/nuget/Microsoft.DotNet.BuildTools.nuspec
```diff
@@ -26,6 +26,7 @@
     <file src="Microsoft.DotNet.Build.CloudTestTasks\RunnerScripts\**\*.*" target="lib\RunnerScripts" />
     <file src="Microsoft.DotNet.Build.Tasks\PackageFiles\**\*.*" target="lib" />
     <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DotNet.Build.Tasks.dll" target="lib" />
+    <file src="Microsoft.DotNet.Build.Tasks\Microsoft.DiaSymReader*.dll" target="lib" />
```
joperezr: Aren't the Native parts of DiaSymReader platform specific(x86 vs amd64 vs arm)? If so then how are you taking into account which one to copy here if the task library builds as anyCPU? Also, I believe that the DiaSymReader package we depend on today is very old, do we ever plan in updating it?

eerhardt: > Aren't the Native parts of DiaSymReader platform specific(x86 vs amd64 vs arm)?

Yes.

> If so then how are you taking into account which one to copy here if the task library builds as anyCPU? 

The native assemblies aren't being copied here. The native assembly is part of the shared fx, which is sort of part of the problem I've been having. Because it is in the shared fx, it isn't possible to override which version is loaded from an MSBuild task.

I could change this line to explicitly add the 2 necessary assemblies to the nupkg:

```xml
<file src="Microsoft.DotNet.Build.Tasks\Microsoft.DiaSymReader.dll" target="lib" />
<file src="Microsoft.DotNet.Build.Tasks\Microsoft.DiaSymReader.Converter.dll" target="lib" />
```

Thoughts?

> Also, I believe that the DiaSymReader package we depend on today is very old, do we ever plan in updating it?

We just updated it a couple weeks ago #2058.

PR #2080, opened 2018-06-27
Update MSBuild references to `15.1.548`.

eerhardt:
This fixes the build errors https://github.com/dotnet/corefx/pull/30675. As it exists currently, the old MSBuild packages pull in references to System.Resources.ReaderWriter, which no longer exists. Thus when building on .NET Core's MSBuild, our tasks fail because that assembly can't be found.
Updating to an officially supported MSBuild package version updates the transitive references to System.Resources.Reader, which exists in the shared framework.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/project.json
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/project.json
@@ -1,7 +1,7 @@
 {
   "dependencies": {
-    "Microsoft.Build.Framework": "0.1.0-preview-00022",
-    "Microsoft.Build.Utilities.Core": "0.1.0-preview-00022",
+    "Microsoft.Build.Framework": "15.1.548",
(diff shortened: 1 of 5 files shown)


joperezr:
One thing to consider here (which most likely doesn't matter) is that now our task libraries that build for core and desktop (like buildtools task library or packaging) will now be building against completely different msbuild assemblies. I don't expect that to be a problem but just wanted to mention it

eerhardt:
> will now be building against completely different msbuild assemblies

It was always that way, wasn't it?  The .NET Core project was using `project.json` and this old `0.1.0` version of MSBuild assemblies.  The .NET Framework project is using `TargetingPackReference` to reference the MSBuild assemblies.

> assuming corefx and friends build fine with this.

I built corefx locally both `master` and with my https://github.com/dotnet/corefx/pull/30675 change. They both succeeded locally.

eerhardt:
Merging preemptively to unblock my corefx PR.

PR #2081, opened 2018-06-29
Revert "Create a runtime dedicated to coverage using IL only"

pjanotti:
Reverts dotnet/buildtools#2077

Reverting until I can solve the issue with the perf runs.


pjanotti:
Not needed anymore issue with VS2015 tools fixed.

PR #2083, opened 2018-06-29
Add BuildTools support for SDK-style projects in VS.

eerhardt:
1. Turn off FillPartialFacadeUsingTask during design-time builds. These builds don't produced output assemblies, and the task fails if there is no output assembly.
2. The design-time builds run targets in a different order than with the legacy .csproj format. Hook our targets into different locations to make VS design-time builds work correctly.
3. Update UpdateVSConfigurations to write out .sln and .csproj files for SDK-style projects.

See https://github.com/dotnet/project-system/blob/master/docs/configurations.md for more information on the VS configurations change for the new project system.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -64,7 +64,7 @@
 
   <Target Name="UpdateReferenceItems"
           DependsOnTargets="SetupDefaultReferences"
-          BeforeTargets="BeforeResolveReferences"
+          BeforeTargets="BeforeResolveReferences;ResolveAssemblyReferences"
   >
     <ItemGroup>
       <Reference Include="@(DefaultReference)" />
@@ -146,6 +146,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <PrepareProjectReferencesDependsOn>
+      AddProjectReferencesDynamically;
+      $(PrepareProjectReferencesDependsOn);
+    </PrepareProjectReferencesDependsOn>
     <ResolveReferencesDependsOn>
       AddProjectReferencesDynamically;
       $(ResolveReferencesDependsOn);
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -49,6 +49,22 @@
     <NuGetTargetMoniker Condition="'$(NuGetTargetMoniker)' == ''">.NETCoreApp,Version=v1.0</NuGetTargetMoniker>
   </PropertyGroup>
 
+  <!--
+    Between the legacy C# project system and the new managed C# project system, some Targets run in different
+    orders. A normal "Build" and the legacy C# project system will execute 'Compile', which depends on
+    'ResolveReferences', which then depends on 'BeforeResolveReferences', so hooking before 
+    'BeforeResolveReferences' allows @(Reference) items to be added before 'ResolveAssemblyReferences'
+    gets executed.
+
+    However, in the new managed C# project system, 'ResolveProjectReferencesDesignTime' gets executed first,
+    which depends on 'ResolveAssemblyReferences'. This means 'ResolveReferences' doesn't get executed before
+    'ResolveAssemblyReferences' when doing a design-time build in the managed C# project system.
+    Work around this by BeforeTargets both 'BeforeResolveReferences' (to preserve any legacy ordering) and
+    'ResolveAssemblyReferences' (to work in the managed C# project system).
+  -->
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -64,7 +64,7 @@
 
   <Target Name="UpdateReferenceItems"
           DependsOnTargets="SetupDefaultReferences"
-          BeforeTargets="BeforeResolveReferences"
+          BeforeTargets="BeforeResolveReferences;ResolveAssemblyReferences"
```
ericstj: Why two here?  Usually having multiple entries for BeforeTargets is a bad idea because it means you'll run before one or the other, whichever comes first.

src/Microsoft.DotNet.Build.Tasks/UpdateVSConfigurations.cs
```diff
@@ -19,7 +19,7 @@ public class UpdateVSConfigurations : BuildTask
 
         private const string ConfigurationPropsFilename = "Configurations.props";
         private static Regex s_configurationConditionRegex = new Regex(@"'\$\(Configuration\)\|\$\(Platform\)' ?== ?'(?<config>.*)'");
-        private static string[] s_configurationSuffixes = new [] { "Debug|AnyCPU", "Release|AnyCPU" };
+        private static string[] s_configurationSuffixes = new [] { "Debug", "Release" };
```
ericstj: Just to confirm: every consumer of this is doing the switch so you don't need to keep the old format?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
eerhardt: The issue is that the new VS C# project system runs targets in a different order than they are run during a normal `Build`.

I don't want to break the existing/legacy .csprojs by moving this target around (thus why I left `BeforeTargets=BeforeResolveReferences`).

However, when doing design-time builds in the new project system, `BeforeResolveReferences` is running AFTER `ResolveAssemblyReferences`.  Thus `UpdateReferenceItems` is happening too late (after RAR), and so the References we are adding here aren't getting picked up.

I'll add a comment here to describe why we are doing it before both targets (in order to support the legacy format, normal builds, and the new C# project system).

weshaggard: Should we create a AddReferencesDynamically target similar to AddProjectReferencesDynamically to be a placeholder for how to setup the references?

eerhardt: I can if you want.  Do you think it will be valuable going forward?

weshaggard: I suspect it will be valuable going forward as we will likely still do some similar default referencing.

src/Microsoft.DotNet.Build.Tasks/UpdateVSConfigurations.cs
weshaggard: I think it is OK to move this forward. If we need the old task for some reason we can fork it later or use a older branch.

src/Microsoft.DotNet.Build.Tasks/UpdateVSConfigurations.cs
```diff
@@ -40,14 +40,14 @@ public override bool Execute()
                 Log.LogMessage($"Updating {projectFile}");
 
                 var project = ProjectRootElement.Open(projectFile);
-                ICollection<ProjectPropertyGroupElement> propertyGroups;
-                var actualConfigurations = GetConfigurationFromPropertyGroups(project, out propertyGroups);
+                ICollection<ProjectPropertyGroupElement> propertyGroups = GetPropertyGroupsToRemove(project);
+                var actualConfigurations = GetConfigurationsFromProperty(project);
 
                 bool addedGuid = EnsureProjectGuid(project);
```
weshaggard: Does this mean we are still going to keep project guid's in the project? Is that still our plan to keep them stable?

eerhardt: I debated on removing them now or not.  My thinking is to leave them for now, in the hopes of getting the first wave of SDK-style changes merged - have to draw the line somewhere.

My plan for removing them, and keeping the GUIDs stable, is to read the GUIDs from the `.sln` file, and associate that GUID with the project.  However, parsing the full `.sln` file was something I didn't want to try to tackle right now. Thus why I left them in.

weshaggard: Sounds good. We could also hash the project name to get a stable guid.

eerhardt:
Any more feedback before this gets merged?

PR #2084, opened 2018-06-29
Fix OptionalTooling so it works on .NET Core

eerhardt:
Need to add support for MSBuild projects instead of project.json files, so builds running on .NET Core can restore optional tooling.

Partial change for https://github.com/dotnet/corefx/issues/30758

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
@@ -71,7 +71,8 @@
     </PropertyGroup>
 
     <GenerateEncryptedNuGetConfig ConfigPath="$(OptionalRestoreConfigPath)"
-                                  Sources="@(OptionalRestoreSource)" />
+                                  Sources="@(OptionalRestoreSource)"
+                                  PackagesDir="$(PackagesDir)" />
 
     <!-- restore the project.json file, if it is being used -->
     <EncryptedConfigNuGetRestore Condition="'$(OptionalToolingProjectJsonPath)' != ''"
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
```diff
@@ -68,33 +73,16 @@
...
-      <OptionalRestoreCommand>$(OptionalRestoreCommand) --configfile $(OptionalRestoreConfigPath)</OptionalRestoreCommand>
-    </PropertyGroup>
-
-    <Error Text="Restoring optional tooling with in-memory credentials is not supported on Windows. Use the desktop framework for the encrypted credential flow."
-           Condition="'$(RunningOnUnix)'!='true'" />
-
-    <Error Text="Restoring optional tooling with in-memory credentials is not yet implemented. No cross-platform optional tooling exists yet." />
+  
+    <!-- restore the MSBuild project file, if it is being used -->
+    <Exec Condition="Exists('$(OptionalToolingProjectPath)')"
```
dagood: I usually prefer != rather than Exists in places like that so that if it doesn't exist, I get an error at the first detectable point to make it obvious. Is there a situation where we legitimately get here without `OptionalToolingProjectPath` existing?

eerhardt:
I still am working on the corefx change. The piece that is left is the auto-update work since I had to move from a `project.json` to `.csproj` file in order to get `dotnet restore` working correctly on .NET Core.

eerhardt:
One piece that still isn't working right with this change is that the restored packages aren't going to `corefx/packages`.  They are going to the user profile.  I think the reason is that we are using `--configfile` and `--packages`, which don't jive together.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
eerhardt: Good point.  Switching to `!= ''`.  If the property is set, we should try using it.  It's an error case if the property is set to a non-existent file.

PR #2086, opened 2018-07-02
Add support for generating resource IDs as an enum

ericstj:
/cc @tarekgh 

One remaining mode to address porting of desktop code.  Some libraries used an Enum to represent resource IDs.  I ran into a case where one library was mapping a range of that enum to a range of error codes and relying on a numeric conversion.

EG: resourceId = (ResourceEnumType) (ResourceEnumType.FirstErrrorId + (errorCode - firstErrorCode))

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
@@ -57,8 +57,18 @@ namespace Microsoft.DotNet.Build.Tasks
         /// </summary>
         public bool AsConstants { get; set; }
 
+        /// <summary>
+        /// Emit enum of Resource IDs instead of a class with properties that retrieve values.
+        /// </summary>
+        public bool AsEnum { get; set; }
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/GenerateResourcesCode.cs
```diff
@@ -136,12 +146,14 @@ private void WriteResources()
...
 
+
             if (_targetLanguage == TargetLanguage.CSharp)
             {
                 _targetStream.WriteLine($"namespace {accessorNamespace}");
                 _targetStream.WriteLine("{");
                 _targetStream.WriteLine("");
-                _targetStream.WriteLine($"    internal static partial class {accessorClassName}");
+                var accessor = AsEnum ? "enum" : "partial class";
+                _targetStream.WriteLine($"    internal static {accessor} {accessorClassName}");
```
eerhardt: Is ‘static enum’ valid syntax?  As written this will produce ‘internal static enum foo’ when AsEnum is true.

PR #2088, opened 2018-07-05
Support simple auto-update value substitutions

dagood:
Add `ReplacementSubstituteOld` / `ReplacementSubstituteNew`, which can be used to replace a part of the new value being updated to.

---

This continues from https://github.com/dotnet/buildtools/pull/2069 "Allow auto-updating from on-disk ProdCon manifest". The on-disk manifest might not have the exact values the repo should auto-update to.

Originally I was going to modify the manifest itself (https://devdiv.visualstudio.com/DevDiv/_git/DotNet-Orchestration-Utilities/pullrequest/128309) but here the logic is more flexible and might be useful for other things.

Used for https://github.com/dotnet/core-eng/issues/3815.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
@@ -261,6 +261,14 @@ namespace Microsoft.DotNet.Build.Tasks.VersionTools
                 "true",
                 StringComparison.OrdinalIgnoreCase);
 
+            string oldValue = step.GetMetadata("ReplacementSubstituteOld");
+            string newValue = step.GetMetadata("ReplacementSubstituteNew");
+
+            if (!string.IsNullOrEmpty(oldValue) && !string.IsNullOrEmpty(newValue))
+            {
+                updater.ReplacementTransform = v => v.Replace(oldValue, newValue);
+            }
+
             return updater;
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -261,6 +261,14 @@ private FileUpdater ConfigureFileUpdater(FileUpdater updater, ITaskItem step)
                 "true",
                 StringComparison.OrdinalIgnoreCase);
 
+            string oldValue = step.GetMetadata("ReplacementSubstituteOld");
+            string newValue = step.GetMetadata("ReplacementSubstituteNew");
```
eerhardt: Should we log a warning/error if only one is specified and not the other?

dagood: Yeah, adding. Thanks.

dagood: `GetMetadata` actually returns empty string rather than null if the value isn't specified, so I'm just adding an error case for `newValue` but no `oldValue`. `oldValue` with no `newValue` means replacing something with the empty string.

dagood:
@dotnet-bot test Innerloop Ubuntu14.04 Debug

PR #2090, opened 2018-07-05
Remove code that places source files in packages

wtgodbe:
This is a replacement for https://github.com/dotnet/corefx/pull/30850 - exclude source code files from packaging if the consuming repo sets `ExcludeSourceFiles=true`. I tested this locally, and the built packages in CoreFx wound up not including source files.

@weshaggard @dagood @ericstj @MattGal PTAL

Once this is in I'll repurpose https://github.com/dotnet/corefx/pull/30850 to set `ExcludeSourceFiles` & take a buildtools update.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
@@ -182,12 +182,6 @@
         <TargetPath>%(TargetPath)%(FilesToPackage.SubFolder)</TargetPath>
       </FilesToPackage>
 
-      <!-- Include source files for symbol packages. -->
-      <FilesToPackage Include="@(Compile->'%(FullPath)')">
-        <TargetPath>src</TargetPath>
-        <TargetPath Condition="$([System.String]::new('%(FullPath)').StartsWith('$(ProjectRoot)'))">$(SourceFileTargetPathPrefix)$([System.String]::new('%(FullPath)').Substring('$(ProjectRootLength)'))</TargetPath>
-        <IsSourceCodeFile>true</IsSourceCodeFile>
-      </FilesToPackage>
     </ItemGroup>
   </Target>
 </Project>
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -493,15 +493,6 @@
         <PackageVersion>$(PackageVersion)</PackageVersion>
       </PackageFile>
 
-      <PackageSymbolFiles Include="@(File)" Condition="'%(File.IsSymbolFile)'=='true'" />
-      <!-- Add a placeholder source file if there are symbols but no sources, so that MyGet will treat symbol packages as valid -->
-      <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
-        <IsSourceCodeFile>true</IsSourceCodeFile>
-        <TargetPath>src</TargetPath>
-      </PackageSources>
-      <PackageFile Include="@(PackageSources)" />
-
-
       <!-- Nuget will treat TargetPath as a directory if the extensions dont match,
            however we need to package files without an extension (Unix exectuables).
            As such nuget will always consider TargetPath to be a file path for these
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -493,6 +493,14 @@
         <PackageVersion>$(PackageVersion)</PackageVersion>
       </PackageFile>
 
+      <!-- Add a placeholder source file if there are symbols but no sources, so that MyGet will treat symbol packages as valid -->
+      <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
+        <IsSourceCodeFile>true</IsSourceCodeFile>
+        <TargetPath>src</TargetPath>
+      </PackageSources>
+      <PackageFile Include="@(PackageSources)" />
+
+
       <!-- Nuget will treat TargetPath as a directory if the extensions dont match,
            however we need to package files without an extension (Unix exectuables).
            As such nuget will always consider TargetPath to be a file path for these
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -493,19 +493,6 @@
         <PackageVersion>$(PackageVersion)</PackageVersion>
       </PackageFile>
 
-      <!-- Include Sources. Deduplicate so the nuspec doesn't contain multiple entries for each source file. -->
-      <PackageSources Include="@(File)"
-                      KeepMetadata="TargetPath;IsSourceCodeFile"
-                      KeepDuplicates="false"
-                      Condition="'%(File.IsSourceCodeFile)'=='true'" />
-      <!-- Add a placeholder source file if there are symbols but no sources. -->
-      <PackageSymbolFiles Include="@(File)" Condition="'%(File.IsSymbolFile)'=='true'" />
-      <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
-        <IsSourceCodeFile>true</IsSourceCodeFile>
-        <TargetPath>src</TargetPath>
-      </PackageSources>
-      <PackageFile Include="@(PackageSources)" />
-
       <!-- Nuget will treat TargetPath as a directory if the extensions dont match,
            however we need to package files without an extension (Unix exectuables).
            As such nuget will always consider TargetPath to be a file path for these
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -493,6 +493,7 @@
         <PackageVersion>$(PackageVersion)</PackageVersion>
       </PackageFile>
 
+      <PackageSymbolFiles Include="@(File)" Condition="'%(File.IsSymbolFile)'=='true'" />
       <!-- Add a placeholder source file if there are symbols but no sources, so that MyGet will treat symbol packages as valid -->
       <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
         <IsSourceCodeFile>true</IsSourceCodeFile>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
@@ -331,13 +331,6 @@
       </_docFiles>
       <_itemsToSave Include="@(_docFiles)"/>
 
-      <!-- Include source files. -->
-      <!-- Here we use "sources" rather than "src" because MyGet treats packages with "src" as symbol packages  -->
-      <_itemsToSave Condition="'@(Compile)' != ''" Include="@(Compile->'%(FullPath)')">
-        <TargetPath>sources</TargetPath>
-        <TargetPath Condition="$([System.String]::Copy('%(FullPath)').StartsWith('$(ProjectDir)'))">sources/$([System.String]::Copy('%(FullPath)').Substring($(_projectDirLength)).Replace('\', '/'))</TargetPath>
-        <IsSourceCodeFile>true</IsSourceCodeFile>
-      </_itemsToSave>
     </ItemGroup>
 
     <Message Importance="low" Text="PackageFileDir: @(PackageFileDir)" />


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
```diff
@@ -183,7 +183,7 @@
       </FilesToPackage>
 
       <!-- Include source files for symbol packages. -->
-      <FilesToPackage Include="@(Compile->'%(FullPath)')">
+      <FilesToPackage Include="@(Compile->'%(FullPath)')" Condition="'$(ExcludeSourceFiles)' != 'true'">
```
weshaggard: Lets please reverse the default and make the name more specific like "IncludeSourceFilesFromPackages".

wtgodbe: So exclude source files by default, unless the consuming repo opts in?

weshaggard: Actually you can default set it to true if UseSourceLink=false.

weshaggard: If we are going to use the UseSourceLink property you can keep this as "ExcludeSourceFilesInPackages"

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -497,10 +497,10 @@
       <PackageSources Include="@(File)"
                       KeepMetadata="TargetPath;IsSourceCodeFile"
                       KeepDuplicates="false"
-                      Condition="'%(File.IsSourceCodeFile)'=='true'" />
+                      Condition="'%(File.IsSourceCodeFile)'=='true' AND '$(ExcludeSourceFiles)' != 'true'" />
       <!-- Add a placeholder source file if there are symbols but no sources. -->
       <PackageSymbolFiles Include="@(File)" Condition="'%(File.IsSymbolFile)'=='true'" />
-      <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
+      <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'=='' AND '$(ExcludeSourceFiles)' != 'true'">
```
dagood: Would it be reasonable to put the condition on the `<PackageFile Include="@(PackageSources)" />` below? Would be slightly different if the repo supplied its own `PackageSources` items and I think preferable. (Although I don't know if repos do provide `PackageSources` items in practice.)

ericstj:
Curious if we have a good reason for keeping the source file feature around.  I'd prefer to just delete it if we're sure we're not going to use it.  Disabled features have a way of atrophying.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets
wtgodbe: So to be clear - set `ExcludeSourceFilesInPackages=$(UseSourceLink)`, then continue using the same logic that's already in this PR? Also, do you know if there's a common place that both `FrameworkTargeting.targets` and this file import? I can't find one (in buildtools), which would mean I'd have to set a default value for `ExcludeSourceFilesInPackages` in both places.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
wtgodbe: That seems reasonable. I'll test out that change along w/ Wes' suggestions

wtgodbe:
@ericstj I don't know of any reason we'd need to keep Source Files around - as far as I can tell it's only used by CoreFx, and we're disabling it there. @weshaggard @dagood do you guys know of any reason we can't delete the concept entirely?

weshaggard:
>Curious if we have a good reason for keeping the source file feature around. I'd prefer to just delete it if we're sure we're not going to use it. Disabled features have a way of atrophying.

I'm not aware of anyone that will use this, I only suggested a condition as a risk mitigation but we can take a stronger stance and just remove the code. @wtgodbe lets try to just remove it.

wtgodbe:
Testing out a changeset locally that removes the concept

wtgodbe:
Updated to delete source file inclusion entirely. I tested this locally and it worked. @ericstj @weshaggard @dagood

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -493,19 +493,6 @@
         <PackageVersion>$(PackageVersion)</PackageVersion>
       </PackageFile>
 
-      <!-- Include Sources. Deduplicate so the nuspec doesn't contain multiple entries for each source file. -->
-      <PackageSources Include="@(File)"
-                      KeepMetadata="TargetPath;IsSourceCodeFile"
-                      KeepDuplicates="false"
-                      Condition="'%(File.IsSourceCodeFile)'=='true'" />
-      <!-- Add a placeholder source file if there are symbols but no sources. -->
```
ericstj: Interesting, that this part was no longer needed.  I guess NuGet fixed this bug @dagood.  @wtgodbe can you double check that we still get the symbols?

ericstj: Actually looks like this was an assumption here: https://github.com/dotnet/buildtools/blob/3306ca2481414624a56f22e0ee28f7c08b93037e/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs#L264.  I'm a little surprised if that no longer holds.

weshaggard: We have decided that our symbol packages are only used as a transport mechanism so they don't need to be "valid" by that sense.

wtgodbe: I just compared the symbol packages before and after this change, other than the lack of source files, they look identical. I can share a new version of the packages if you guys would like to inspect them yourselves.

dagood: Good spot, this actually is probably still needed. The impact is MyGet: symbol packages are only barely a NuGet feature. I don't think this is a bug, it's just how MyGet classifies the packages. Depending on how the build pushes symbol packages to MyGet not getting this right can cause issues like symbol packages showing up as the "main" package for a certain id/version. (There is a symbol-package-specific feed but I don't think it's reliably used.) The opposite is also true, like in https://github.com/dotnet/coreclr/issues/13798, where something intended as a "normal" package is treated like a symbol package.

weshaggard: @dagood didn't we stop pushing nuget packages to myget?

dagood: We don't use MyGet as the *only* way to transport nupkgs from one of our builds to another, but it's still a maintained public endpoint and it's still used by a bunch of builds in the dotnet org.

I would like to simply go through and fix this, but I don't think it's best to start by removing this placeholder source file. The problems it can cause on the MyGet feed aren't easy to detect, and can have downstream effects that may be flaky and I don't think would be easy to diagnose.

weshaggard: What is suggestion then @dagood? Should we abandon this PR?

dagood: I think it's mainly fine, we just need to keep making the placeholder file in some way. Maybe just keep:

```xml
<PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
  <IsSourceCodeFile>true</IsSourceCodeFile>
  <TargetPath>src</TargetPath>
</PackageSources>
<PackageFile Include="@(PackageSources)" />
```

Or reimplement in `NuGetPack.cs` maybe, seems like that might be cleaner.

ericstj: I'm ok either way.  Moving it into NuGetPack seems like a separate issue.

weshaggard: OK yeah for this change lets just keep the placeholder with an updated comment.

wtgodbe: I've just updated this to add the placeholder back

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -493,19 +493,14 @@
...
-      <PackageSymbolFiles Include="@(File)" Condition="'%(File.IsSymbolFile)'=='true'" />
+      <!-- Add a placeholder source file if there are symbols but no sources, so that MyGet will treat symbol packages as valid -->
       <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
         <IsSourceCodeFile>true</IsSourceCodeFile>
```
weshaggard: @ericstj @dagood Do we still need the IsSourceCodeFile metadata?

dagood: I imagine it might be used in some later target to avoid adding sources to the main package, but I don't know the sequence well enough to know without poking around. I think @wtgodbe might be in the best position to answer that with actual build results.

wtgodbe: I would think we do need that metaData for the sake of excluding source code files, for example https://github.com/dotnet/buildtools/pull/2090/files#diff-03c502c8604ae1b4ae2efeb1f3d0c45eL490

weshaggard: Sounds good I thought we removed all usages of the metadata. I'm good keeping it though.

wtgodbe:
@weshaggard @dagood @ericstj is this alright to merge now?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -493,19 +493,14 @@
...
 
-      <!-- Include Sources. Deduplicate so the nuspec doesn't contain multiple entries for each source file. -->
-      <PackageSources Include="@(File)"
-                      KeepMetadata="TargetPath;IsSourceCodeFile"
-                      KeepDuplicates="false"
-                      Condition="'%(File.IsSourceCodeFile)'=='true'" />
-      <!-- Add a placeholder source file if there are symbols but no sources. -->
-      <PackageSymbolFiles Include="@(File)" Condition="'%(File.IsSymbolFile)'=='true'" />
+      <!-- Add a placeholder source file if there are symbols but no sources, so that MyGet will treat symbol packages as valid -->
       <PackageSources Include="$(PlaceholderFile)" Condition="'@(PackageSymbolFiles)'!='' AND '@(PackageSources)'==''">
```
dagood: I just realized that since this will only add the placeholder with `'@(PackageSymbolFiles)'!=''`, we probably also need to keep `<PackageSymbolFiles Include="@(File)" Condition="'%(File.IsSymbolFile)'=='true'" />`.

Have you tried this out in a build? What we're looking for is the nupkg having no `src/_._` placeholder, and the symbols.nupkg having some `**/*.pdb` file(s) and `src/_._`.

wtgodbe: My last build of this looked right for packages, but there is no `src/_._` in the symbol packages. I'll try adding that line back in & rebuilding.

wtgodbe: @dagood there is now a placeholder in both normal & symbol packages. Should I undo https://github.com/dotnet/buildtools/pull/2090/commits/47007205f0abddd29e96dd4d1f7bdc7cd90a9010?

dagood: I don't see how that would fix it, but if it does, go for it. I'd have to do some stuff myself with binlogs, diag logs, `/pp`, etc. to figure out exactly what to do.

dagood: Can also use the LaunchDebugger buildtools task to figure out why NuGetPack is/isn't filtering certain things out.

wtgodbe: As far as I can tell from the binlog/playing around with the targets, we can either have the placeholder in both places, or in neither (without making more significant changes). Is there a real harm to having a placeholder file in the normal package?

dagood: The harm is potential flakiness for transport packages (since they include PDBs) on MyGet. This is what's happening in https://github.com/dotnet/coreclr/issues/13798.

Without this PR, are you also seeing that behavior? It doesn't seem right to give up and regress now. 😕

wtgodbe: Without this PR, both normal packages & symbol packages contain the full set of sources. I haven't yet been able to make a change to ones `src` directory without making the same change to the other's. I can keep poking around for a way though, if it will be a problem to have a placeholder in the non-symbol packages.

ericstj: Note that the transport packages are also ["special"](https://github.com/dotnet/corefx/blob/0b0f63d47951773cc46393312d80b40176d3a413/pkg/Microsoft.Private.CoreFx.NETCoreApp/Microsoft.Private.CoreFx.NETCoreApp.pkgproj#L16) so they may be a bit confusing to look at.  For those ones I expect symbols and source in both sets of packages.  Make sure you look at a regular package as well.

wtgodbe: I've been building CoreFx so I haven't seen a transport package yet. I'll try building CoreClr.

dagood: Ah, I think you're looking at transport packages (rightly) and I'm thinking more about e.g. `System.IO.Ports` which doesn't have sources (or symbols) https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/system.io.ports/4.6.0-preview1-26712-04/system.io.ports.4.6.0-preview1-26712-04.nupkg

I think this is happening because of `NuGetPack`'s `IncludeSymbolsInPackage`, which I think is being used to make the transport package. It's what filters the items down and excludes the placeholder, so I think it makes sense to change the task code itself (and probably delete this item).

ericstj: @wtgodbe we call those "private" packages transport packages, so you have seen them.

dagood: (Missed last two comments.) CoreFX's, transport packages are the `runtime.win10-x64.Microsoft.Private.CoreFx.UAP` ones--"transport package" just means it only delivers bits between repos and isn't published to the nuget.org gallery, kinda confusing that coreclr's have the `transport.` prefix and CoreFX's don't.

wtgodbe: Oh, to be clear, I've been looking at `Microsoft.Private.CoreFx.NETCore.App`.

If I look at `Microsoft.NETCore.Targets`, neither the normal package nor the symbol package have a placeholder file before or after my change

ericstj: That package has no dlls (so no src or symbols) only a runtime.json, use a library package.  Something System.*

dagood: https://github.com/dotnet/buildtools/pull/1369 is a good place to understand the changes to NuGetPack to enable transport/private packages. Then I think something like this should be added (maybe not in the same if statement) that adds to `builder` if it needs to add a placeholder:

https://github.com/dotnet/buildtools/blob/185e36b4c1372d815af1b5a728249c0afdde90c0/src/Microsoft.DotNet.Build.Tasks.Packaging/src/NuGetPack.cs#L269-L273

---

Stepping back... we could just remove all src placeholders, and sweep through the `master` builds and make sure they all publish to the symbol-package-specific MyGet endpoint (if they publish to MyGet) so we don't need to care about the MyGet symbol package detection anymore. This is kind of risky because it's hard to tell immediately by browsing MyGet when things go wrong, but seems possible.

Package push endpoint: `https://dotnet.myget.org/F/dotnet-core/api/v2/package`
Symbol package push endpoint: `https://dotnet.myget.org/F/dotnet-core/symbols/api/v2/package`

wtgodbe: @ericstj I built System.Security.Cryptography.Xml & neither its package nor its symbol package had a `src` folder

@dagood so the 2 options are to modify NugetPack to inject a placeholder into symbol packages, or to not bother with a placeholder at all, and change how we publish to myget - but both options entail removing placeholders from this .targets file entirely?

dagood: > so the 2 options are [...] both options entail removing placeholders from this .targets file entirely?

Yep. I can imagine other options that don't involve removing the placeholder MSBuild code (I imagine it requires more substantial changes to the `NuGetPack` logic to make sure our desired scenarios are covered) but those two options seem like the simplest to me.

wtgodbe: @dagood it looks like we already segregate symbol packages & normal packages into the symbol package/normal package myget endpoints (at least in master for the 3 core repos), so easiest would probably be to just merge this as-is (after removing this placeholder code)

dagood: WCF doesn't: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/results?buildId=1857075. But apart from dotnet/standard (which doesn't even push symbol packages to MyGet), that's everyone using BuildTools, so I agree.

wtgodbe: That should be easy enough to change, they already have the package publish & symbol package publish divided into 2 steps. @StephenBonikowsky is it alright if I modify the symbol publish step to push to the myget symbols endpoint (`https://dotnet.myget.org/F/dotnet-core/symbols/api/v2/package`) instead of the package endpoint `https://dotnet.myget.org/F/dotnet-core/api/v2/package`)?

StephenBonikowsky: We haven't been updating our Master branch since 2.1 released. I have a work item to base our Master on dependencies from 2.1 such as [buildtools/release/2.1](https://github.com/dotnet/buildtools/tree/release/2.1). So you can go ahead and make this change without updating our build definition.

wtgodbe: Sounds good. In that case, it sounds like all that needs to be done is to merge this PR. @dagood does that sound right to you?

dagood: Yep. Thanks for the info @StephenBonikowsky!

PR #2091, opened 2018-07-06
Need to re-default output properties after writing $(OutputPath).

eerhardt:
When using SDK-style projects, Microsoft.Common.targets are getting imported before buildtools' targets. Thus when we overwrite $(OutputPath), we need to also reset the other output properties.

Fixes the issue described here https://github.com/dotnet/corefx/pull/29831#discussion_r200525701

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
@@ -24,6 +24,18 @@
     <NuGetDeploySourceItem>Reference</NuGetDeploySourceItem>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsReferenceAssembly)'=='true' AND '$(UsingMicrosoftNETSdk)'=='true'">
+    <!-- When not using Sdk-style projects, Microsoft.Common.CurrentVersion.targets,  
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
```diff
@@ -24,6 +24,18 @@
     <NuGetDeploySourceItem>Reference</NuGetDeploySourceItem>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsReferenceAssembly)'=='true' AND '$(UsingMicrosoftNETSdk)'=='true'">
+    <!-- When not using Sdk-style projects, Microsoft.Common.CurrentVersion.targets,
```
weshaggard: Hopefully no hasn't already used these property so resetting them will work. Do you think it is possible to move these in the ReferenceAssembly.props instead which should be before the Microsoft.Common.CurrentVersion.targets?

eerhardt: I tried moving them to the `.props` file, but I couldn't because `$(AssemblyVersion)` isn't set by that time.

weshaggard: :( OK

PR #2092, opened 2018-07-08
add recognition for FreeBSD

wfurt:
With this, corefx can build on FreeBSD without having to pass -OSGroup=FreeBSD every time.

Note that I was puzzled for a while as even with OSGroup set, tests would still target Linux. 
That is caused by:

```
 tests.targets:    <TargetOS Condition="'$(TargetOS)' == ''">$(DefaultOSGroup)</TargetOS>
```

using DefaultOSGroup instead of OSGroup seems wrong here. 

However since my fix fixes DefaultOSGroup, it will also fix test script generation. 
So the risk is for next OS.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -26,6 +26,7 @@
   <!-- Platform detection -->
   <PropertyGroup>
     <DefaultOSGroup Condition="'$(OS)'=='Unix' AND Exists('/Applications')">OSX</DefaultOSGroup>
+    <DefaultOSGroup Condition="'$(DefaultOSGroup)'=='' AND '$(OS)'=='Unix' AND $([System.Runtime.InteropServices.RuntimeInformation]::OSDescription.Contains('FreeBSD'))">FreeBSD</DefaultOSGroup>
     <DefaultOSGroup Condition="'$(DefaultOSGroup)'=='' AND '$(OS)'=='Unix'">Linux</DefaultOSGroup>
     <DefaultOSGroup Condition="'$(DefaultOSGroup)'==''">$(OS)</DefaultOSGroup>
     <RunningOnUnix Condition="'$(OS)'!='Windows_NT'">true</RunningOnUnix>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -26,6 +26,7 @@
   <!-- Platform detection -->
   <PropertyGroup>
     <DefaultOSGroup Condition="'$(OS)'=='Unix' AND Exists('/Applications')">OSX</DefaultOSGroup>
+    <DefaultOSGroup Condition="'$(DefaultOSGroup)'=='' AND '$(OS)'=='Unix' AND $([System.Runtime.InteropServices.RuntimeInformation]::OSDescription.Contains('FreeBSD'))">FreeBSD</DefaultOSGroup>
```
eerhardt: I think you can use `$([MSBuild]::IsOSPlatform('FREEBSD'))` here instead.

See https://github.com/Microsoft/msbuild/blob/1e05f35b10755cc31311dc4b6ea256700331804c/Directory.Build.props#L13

wfurt: possibly. Is that much better? I sort of liked the OSDescription as it could provide more flexibility in future. But I'm happy to change as you suggested as it should work for FreeBSD now. I wish we could somehow access  "uname".

Also any thoughts on DefaultOSGroup and TargetOS for testing?

eerhardt: I think it is better than relying on OSDescription, as OSDescription can change since it is just a "description".  In general, using descriptions for detection code feels wrong.

> Also any thoughts on DefaultOSGroup and TargetOS for testing?

I agree it seems wrong to use `DefaultOSGroup` here.  Should probably use `OSGroup`.

weshaggard: I also don't like the fragility of using OSDescription as it was never designed for this use case. I do agree we need some APIs for detection and there are various issues tracking different designs.

PR #2094, opened 2018-07-10
Fix python path for coverage runs

pjanotti:
Coverage runs are without dumplings due to the dependency on HELIX_PYTHONPATH env var. 

On my local box I noticed the following:

1. DumplingHelper.py requires Python 2.7 (no problem, just noticing)
2. My runs on local box report:
```
  Traceback (most recent call last):
    File "C:\s\github\pjanotti\corefx\bin\tests\System.Net.Mail.Unit.Tests\netcoreapp-Windows_NT-Debug-x64/dumpling.py", line 18, in <module>

  ImportError: No module named requests
  C:\Python27\python.exe: can't open file 'C:\Users\pauloja/.dumpling/dumpling.py': [Errno 2] No such file or directory
  C:\Python27\python.exe: can't open file 'C:\Users\pauloja/.dumpling/dumpling.py': [Errno 2] No such file or directory
```
not sure if this is something to pursue or not.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
@@ -25,9 +25,16 @@
     <DumplingIncPaths Condition="'$(TargetOS)'=='OSX'" Include="/usr/lib/libSystem.B.dylib" />
     <DumplingIncPaths Condition="'$(TargetOS)'=='OSX'" Include="/usr/lib/libc++.1.dylib" />
   </ItemGroup>
+  <PropertyGroup Condition="'$(HELIX_PYTHONPATH)'==''">
+    <!-- env var not defined, assume that python is already on the path, e.g.: Jenkins -->
+    <PythonPath>python</PythonPath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(HELIX_PYTHONPATH)'!=''">
+    <PythonPath>$(HELIX_PYTHONPATH)</PythonPath>
+  </PropertyGroup>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/Dumpling.targets
```diff
@@ -25,9 +25,16 @@
     <DumplingIncPaths Condition="'$(TargetOS)'=='OSX'" Include="/usr/lib/libSystem.B.dylib" />
     <DumplingIncPaths Condition="'$(TargetOS)'=='OSX'" Include="/usr/lib/libc++.1.dylib" />
   </ItemGroup>
+  <PropertyGroup Condition="'$(HELIX_PYTHONPATH)'==''">
```
danmosemsft: You can write this much simpler
```xml
  <PropertyGroup>
    <PythonPath>$(HELIX_PYTHONPATH)</PythonPath>
    <PythonPath Condition="'$(PythonPath)' == ''">python</PythonPath>
 </PropertyGroup>
```

danmosemsft: Also I think that would be idiomatic

pjanotti: One of these days I should learn msbuild... :smile: currently I think I am writing more msbuild than c#...

pjanotti:
Unfortunately this breaks dumpling on some runs (see [here](https://github.com/dotnet/corefx/pull/30973#issuecomment-404180554)). It seems that the env var is defined *after* the test scripts are created.

PR #2100, opened 2018-07-17
Updating Packaging.targets to support specifying the PackageType

tannergooding:
FYI. @weshaggard, this is the buildtools change required to support moving the ILProj SDK into CoreCLR (see https://github.com/dotnet/arcade/pull/317 for more details).

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
@@ -69,6 +69,8 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
 
         public string Tags { get; set; }
 
+        public string[] PackageTypes { get; set; }
+
         public ITaskItem[] Dependencies { get; set; }
 
         public ITaskItem[] References { get; set; }
@@ -183,6 +185,7 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
             manifestMetadata.UpdateMember(x => x.Tags, Tags);
             manifestMetadata.UpdateMember(x => x.Title, Title);
             manifestMetadata.UpdateMember(x => x.Version, Version != null ? new NuGetVersion(Version) : null);
+            manifestMetadata.UpdateMember(x => x.PackageTypes, GetPackageTypes());
             manifestMetadata.Serviceable |= Serviceable;
 
             manifest.AddRangeToMember(x => x.Files, GetManifestFiles());
@@ -282,6 +285,29 @@ namespace Microsoft.DotNet.Build.Tasks.Packaging
                                        select reference.File
                                        )
                     ).ToList();
+        }
+
+        private List<PackageType> GetPackageTypes()
+        {
+            var listOfPackageTypes = new List<PackageType>();
+
+            // Copied and slightly modified from ParsePackageTypes():
+            // https://github.com/NuGet/NuGet.Client/blob/50af5271b98ac5cb2896a707569bc4cd1e87a017/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTaskLogic.cs#L338
+
+            foreach (var packageType in PackageTypes.TrimAndExcludeNullOrEmpty())
+            {
+                string[] packageTypeSplitInPart = packageType.Split(new char[] { ',' });
+                string packageTypeName = packageTypeSplitInPart[0].Trim();
+                var version = PackageType.EmptyVersion;
+                if (packageTypeSplitInPart.Length > 1)
+                {
+                    string versionString = packageTypeSplitInPart[1];
+                    System.Version.TryParse(versionString, out version);
+                }
+                listOfPackageTypes.Add(new PackageType(packageTypeName, version));
+            }
+
+            return listOfPackageTypes;
         }
 
         private static VersionRange AggregateVersions(VersionRange aggregate, VersionRange next)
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -1209,7 +1209,8 @@
                     References="@(Reference)"
                     FrameworkReferences="@(FrameworkReference)"
                     Files="@(_packageFile)"
-                    Serviceable="$(Serviceable)"/>
+                    Serviceable="$(Serviceable)"
+                    PackageTypes="$(PackageType)"/>
```
tannergooding: The actual NuGet `PackTask` does this (instead of `@(PackageType)`) as well: https://github.com/NuGet/NuGet.Client/blob/50af5271b98ac5cb2896a707569bc4cd1e87a017/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L224

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
```diff
@@ -284,6 +287,13 @@ select reference.File
                     ).ToList();
         }
 
+        private List<PackageType> GetPackageTypes()
+        {
+            var packageTypes = from p in PackageTypes.NullAsEmpty()
+                               select new PackageType(p.ItemSpec, PackageType.EmptyVersion);
```
tannergooding: We shouldn't need to set the PackageType version for anything right now (and it would probably be better to move to the real pack task, if that support was needed, anyways)

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
```diff
@@ -69,6 +69,8 @@ public class GenerateNuSpec : Task
 
         public string Tags { get; set; }
 
+        public ITaskItem[] PackageTypes { get; set; }
```
tannergooding: The real pack task takes a `string[]` instead: https://github.com/NuGet/NuGet.Client/blob/50af5271b98ac5cb2896a707569bc4cd1e87a017/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTask.cs#L22

weshaggard:
@ericstj could you please take a look?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
ericstj: Ok, so why aren’t you doing that if you don’t need items?

tannergooding: The real pack task takes the raw item strings and does custom parsing on it. Using `ITaskItem[]` for our own code makes it simpler without needing to duplicate the `ParsePackageTypes` method in `PackTaskLogic.cs`. If we actually need the more complex logic later on, our time would probably be better invested in just moving to use the official task.

tannergooding:
What is the `Pre-Merge signing tool` job and do I need to wait for it before merging?

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
ericstj: How is our code made simpler by taking `ITaskItem[]`?  I see how it is made simpler by not supporting the encoded version in the identity string, but that doesn't have anything to do with accepting `ITaskItem[]` vs `string[]`.  Seems to me the code is made *slightly* more complex by having to get the `ItemSpec` out instead of just using the value.

Typically (though I know we don't follow this guideline everywhere) it is better/cheaper/simpler to accept `string` if you have no need for metadata, path semantics, etc provided by `ITaskItem`.

ericstj:
> Pre-Merge signing tool

Changes in buildtools must be signed by the contributor.  We should make sure you're on the list for folks with permission to do that.  Can you try again (it appears to be up for me)?  If it still is not working then @markwilkie or @mmitche may be able to help get you on the right list.

src/Microsoft.DotNet.Build.Tasks.Packaging/src/GenerateNuSpec.cs
tannergooding: @ericstj stopped by and we chatted. I was under the impression that using `ITaskItem[]` was the general/recommended approach for simplicitly/cost. However, that may not always be the case.

In either case, I took another look at the way NuGet expects the MSBuild items to be encoded, it appears I was mistaken.

Unlike the `nuspec`, which represents these as `<packageType name="Dependency" version="Version" />`, MSBuild is expected to encode it as `<PackageType>Dependency1,Version2;Dependency2,Version2</PackageType>` (`<PackageType Include="Dependency" Version="Version" />` is not supported). So we ultimately need to just take a `string[]` anyways.

tannergooding: For reference: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets and https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D

tannergooding:
Updated the code to take a `string[]` and just copied the ~12 lines of code from `ParsePackageTypes`: https://github.com/NuGet/NuGet.Client/blob/50af5271b98ac5cb2896a707569bc4cd1e87a017/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTaskLogic.cs#L338, as that is probably better then silently doing the wrong thing or failing. -- See https://github.com/dotnet/buildtools/pull/2100#discussion_r203451046 for more details.

tannergooding:
Fixed to not depend on `Array.Empty<T>` so that the desktop build works.

ericstj:
Test Innerloop Ubuntu14.04 Debug

PR #2101, opened 2018-07-18
Option to use coverlet.msbuild for coverage (allow coverage in Unix)

pjanotti:
[EDIT: Updated instructions after validation on macOS 10.13]

/cc @tarekgh @tonerdo @stephentoub 

This change allows one to experiment with coverlet.msbuild to generate coverage information on the CoreFx repo. Some notes:

* There are some issues to consume dotnet/cli global tools on CoreFx repo, it is possible to workaround them but I opted to use the msbuild due to the large set of exclusion files, making easier for me to debug/tweak things via a msbuild task.
* There is a related PR in CoreFx to bring the coverlet.msbuild package, see https://github.com/dotnet/corefx/pull/31171.
* Once this reach the CoreFx repo I will update the wiki with the instructions below - please, review them too.

# Cross-platform Coverage 
As of 07/2018 CoreFx is only able to get coverage information on Windows. To correct this we are experimenting with [coverlet](https://github.com/tonerdo/coverlet).

## Know Issues ##

1. Instrumenting "System.Private.CoreLib" is causing test to crash (both on Windows and Unix).

## Windows Instructions ##
On Windows by default coverage runs will use OpenCover instead of coverlet, use the build property `UseCoverlet` to change this default. Currently the use of `dotnet msbuild` is required to avoid a problem with one of the coverlet dependencies. Here is the command:

```
dotnet msbuild /t:RebuildAndTest /p:Coverage=True /p:UseCoverlet=True
``` 

## Unix Instructions ##
On Unix just specifying `/p:Coverage=True` triggers the usage of coverlet. However, in order to generate the html report a few setup steps are needed.

1. Install the [dotnet/cli global tool reportgenerator](https://www.nuget.org/packages/dotnet-reportgenerator-globaltool), add it to the PATH if not automatically added by dotnet/cli (it can be only for the current session)
2. On Linux install the Arial font required for report generation: `sudo apt-get install ttf-mscorefonts-installer`

After that you request a coverage run using either `dotnet msbuild` or the `msbuild.sh` from a test folder, e.g.:

```
dotnet msbuild /t:RebuildAndTest /p:Coverage=True
```

Open index.htm located at bin/tests/coverage to see the results of the coverage run.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -14,7 +14,7 @@
   <!-- Coverage options -->
   <PropertyGroup>
     <CodeCoverageEnabled>false</CodeCoverageEnabled>
-    <CodeCoverageEnabled Condition="'$(SkipTests)' != 'true' and '$(RunningOnUnix)' != 'true' and '$(Coverage)' == 'true' and '$(Performance)' != 'true'">true</CodeCoverageEnabled>
+    <CodeCoverageEnabled Condition="'$(SkipTests)' != 'true' and '$(Coverage)' == 'true' and '$(Performance)' != 'true'">true</CodeCoverageEnabled>
     <UseCoverlet Condition="'$(Coverage)' == 'true' and ('$(RunningOnUnix)' == 'true' or '$(UseCoverlet)' == 'true')">true</UseCoverlet>
     <CoverageReportDir Condition="'$(CoverageReportDir)' == ''">$(TestWorkingDir)coverage\</CoverageReportDir>
 
@@ -169,7 +169,6 @@
       <_AllDlls Include="$(CoverageRuntimeDir)/*.dll"/>
       <_AllAssemblies Include="@(_AllDlls -> '%(Filename)')" />
       <_AllAssemblies Remove="@(_CodeCoverageAssemblies)" />
-      <_CoverletExcludedAssemblies Include="@(_AllAssemblies)" />
     </ItemGroup>
     <PropertyGroup>
       <CoverletExclusionFilter>@(_AllAssemblies->'[%(Identity)]*', ',')</CoverletExclusionFilter>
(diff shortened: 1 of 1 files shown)


tarekgh:
CC @weshaggard

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -39,6 +41,35 @@
...
+    <ExcludeByFile Condition="$(ExcludeByFile) == ''"></ExcludeByFile>
+    <Threshold Condition="$(Threshold) == ''">0</Threshold>
+    <ThresholdType Condition="$(ThresholdType) == ''">line,branch,method</ThresholdType>
+    <CollectCoverage Condition="$(CollectCoverage) == ''">$(CodeCoverageEnabled)</CollectCoverage>
+  </PropertyGroup>
+
+  <UsingTask Condition="'$(UseCoverlet)'=='true'" TaskName="Coverlet.MSbuild.Tasks.InstrumentationTask" AssemblyFile="$(PackagesDir)coverlet.msbuild/$(CoverletMsbuildVersion)/build/netstandard2.0/coverlet.msbuild.tasks.dll" />
+  <UsingTask Condition="'$(UseCoverlet)'=='true'" TaskName="Coverlet.MSbuild.Tasks.CoverageResultTask" AssemblyFile="$(PackagesDir)coverlet.msbuild/$(CoverletMsbuildVersion)/build/netstandard2.0/coverlet.msbuild.tasks.dll" />
+ 
+  <Target Condition="'$(UseCoverlet)'=='true'" Name="GenerateCoverageResult" AfterTargets="$(GenerateIndividualCoverageReportAfterTargets)">
```
tarekgh: Can you give some name for this Target? it will be useful when looking at the logs.

tarekgh: ignore the comment. somehow I  was expecting the name to be the first property :-)

tonerdo:
> Instrumenting "System.Private.CoreLib" is causing test to crash (both on Windows and Unix).

@tarekgh what error message do you get when the tests crash? Possibly coverlet isn't properly instrumenting the module, I wanna look into it.

tarekgh:
@pjanotti could you please send @tonerdo the error you are getting when instrumenting the corelib? I guess corelib can be tricky as it doesn't reference any assembly but I am not sure if this is the case. @pjanotti should have more info here.

PR #2104, opened 2018-07-19
Update xunit version to 2.4.0

ViktorHofer:
Not sure if we should target NS2.0 instead and also get rid of the project.json?

Related https://github.com/Microsoft/xunit-performance/pull/269


safern:
> Not sure if we should target NS2.0 instead and also get rid of the project.json?

I think that would require some more changes in the repo and probably not worth now that we're migrating to arcade.

safern:
You're getting error: 

`/mnt/j/workspace/dotnet_buildtools/master/Ubuntu14.04_prtest/src/xunit.netcore.extensions/Discoverers/ConditionalTestDiscoverer.cs(37,13): error CS0246: The type or namespace name 'TestMethodDisplayOptions' could not be found (are you missing a using directive or an assembly reference?) [/mnt/j/workspace/dotnet_buildtools/master/Ubuntu14.04_prtest/src/xunit.netcore.extensions.46/xunit.netcore.extensions.net46.csproj]`

When cross compiling xunit extensions for 46

PR #2095, opened 2018-07-11
Decouple AoT and UAP testing

morganbr:
As part of adding the netcoreappaot flavor in https://github.com/dotnet/corefx/pull/30980, this change decouples AoT and UAP testing by switching to using the UseDotNetNativeToolchain build variable. It also enables testing with the ```-win32``` ILC flag, which disables WinRT support and matches the netcoreappaot profile.

The CoreFX and Buildtools changes can be merged in either order (it just won't be possible to test the CoreFX change until the Buildtools change is merged).

PR #2106, opened 2018-07-22
Use official xunit runner

ViktorHofer:
Fixes https://github.com/dotnet/corefx/issues/21253
Fixes https://github.com/dotnet/corefx/issues/31723
Fixes https://github.com/dotnet/corefx/issues/10567

Use the supplied runner from the nuget package which we copy into our testhost and runtime folder. Also include the test adapter to enable Visual Studio Test Explorer integration.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -129,7 +129,8 @@
     </ItemGroup>
     <ItemGroup Condition="'$(BuildingNETFxVertical)' != 'true'">
       <SupplementalTestData Condition="'$(SkipXunitRuntimeConfigCopying)' != 'true' and '$(UsingCoverageDedicatedRuntime)' != 'true'" Include="$(XunitRuntimeConfig)" />
-      <SupplementalTestData Include="$(RuntimePath)xunit.console.netcore.exe" />
+      <SupplementalTestData Include="$(RuntimePath)xunit.console.dll" />
+      <SupplementalTestData Include="$(RuntimePath)xunit.runner.utility.netcoreapp10.dll" />
```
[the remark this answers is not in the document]
ViktorHofer: Because the xunit team decided to name it so ;)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -40,8 +40,8 @@
     <TestHostExecutablePath Condition="'$(RunningOnUnix)'!='true' AND '$(TestHostExecutablePath)' == '' AND '$(BuildingUAPAOTVertical)' != 'true'">$(TestRuntimeEnvVar)dotnet.exe</TestHostExecutablePath>
     <TestHostExecutablePath Condition="'$(RunningOnUnix)'=='true' AND '$(TestHostExecutablePath)' == '' AND '$(BuildingUAPAOTVertical)' != 'true'">$(TestRuntimeEnvVar)dotnet</TestHostExecutablePath>
 
-    <XunitExecutable Condition="'$(XunitExecutable)' == ''">xunit.console.netcore.exe</XunitExecutable>
-    <XunitRuntimeConfigFile>xunit.console.netcore.runtimeconfig.json</XunitRuntimeConfigFile>
+    <XunitExecutable Condition="'$(XunitExecutable)' == ''">xunit.console.dll</XunitExecutable>
```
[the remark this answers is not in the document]
ViktorHofer: yes

ViktorHofer:
Thanks. Not merging yet as before I want PRs in all related repos set up. @safern coreclr/corefx/wcf, anything else I forgot?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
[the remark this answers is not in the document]
safern: That makes it confusing 😆

safern:
>  @safern coreclr/corefx/wcf, anything else I forgot?

I know those are the repos use our tests.targets to run tests, but might be more, @weshaggard should know better.

ViktorHofer:
These should now be all the necessary changes. Please take one more look. The commits will be squashed together.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -145,24 +171,64 @@
...
+    </ItemGroup>
+
+    <!-- uap -->
+    <ItemGroup Condition="'$(IncludeVSTestReferences)' == 'true' AND ('$(BuildingUAPVertical)' == 'true' OR '$(BuildingUAPAOTVertical)' == 'true')">
+      <ReferenceFromRuntime Include="testhost" />
+      <ReferenceFromRuntime Include="xunit.runner.visualstudio.uwp.testadapter" />
+    </ItemGroup>
+
+    <!-- netfx -->
+    <ItemGroup Condition="'$(IncludePerformanceTestReferences)' == 'true' AND '$(BuildingNETFxVertical)' == 'true'">
```
safern: This can be unified with the netcoreapp one. I understand your intention of having them separate -- but that brings duplication.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -326,10 +392,14 @@
...
-      <RunnerScriptName Condition="'$(TargetOS)'!='Windows_NT'" >RunTests.sh</RunnerScriptName>
-      <RunnerTemplateName Condition="'$(TargetOS)'!='Windows_NT'" >RunnerTemplate.Unix.txt</RunnerTemplateName>
+      <RunnerScriptName Condition="'$(TargetOS)' == 'Windows_NT'">RunTests.cmd</RunnerScriptName>
+      <RunnerScriptName Condition="'$(TargetOS)' != 'Windows_NT'">RunTests.sh</RunnerScriptName>
+
+      <!-- In UAP we don't want the repro steps printed -->
+      <RunnerTemplateName Condition="'$(TargetOS)' == 'Windows_NT' AND '$(BuildingUAPVertical)' != 'true'">RunnerTemplate.Windows.txt</RunnerTemplateName>
+      <RunnerTemplateName Condition="'$(TargetOS)' == 'Windows_NT' AND ('$(BuildingUAPVertical)' == 'true' OR '$(UseDotNetNativeToolchain)' == 'true')">RunnerTemplate.Quiet.Windows.txt</RunnerTemplateName>
+      <RunnerTemplateName Condition="'$(TargetOS)' != 'Windows_NT' AND '$(BuildingUAPVertical)' != 'true'">RunnerTemplate.Unix.txt</RunnerTemplateName>
+      <RunnerTemplateName Condition="'$(TargetOS)' != 'Windows_NT' AND ('$(BuildingUAPVertical)' == 'true' OR '$(UseDotNetNativeToolchain)' == 'true')">RunnerTemplate.Quiet.Unix.txt</RunnerTemplateName>
```
safern: I don't think UAPVertical will be true when in Unix.

ViktorHofer: right, there is no BuildingUAPVertical or .NET Native Toolchain on Unix right? Then I can just the Unix line.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -115,7 +115,7 @@
       <NativeVersionLines Include="#endif" />
     </ItemGroup>
 
-    <MakeDir Condition="!Exists('$(NativeVersionFileDir)')" Directories="$(NativeVersionFileDir)" />
```
safern: Why are you removing this condition?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
ViktorHofer: or I'll just keep it for in case?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
ViktorHofer: MakeDir doesn't throw if it already exists. The condition isn't necessary.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
safern: UseDotNetNativeToolchain I think will be enabled for netcoreapp https://github.com/dotnet/buildtools/pull/2095

ViktorHofer: ok then I'll just leave it as it is. Not worth changing that. Maybe some day there will be targeting pack for Unix for UWP :D

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -37,6 +37,7 @@
 
   <!-- Common BuildTools properties -->
   <PropertyGroup>
+    <RootIntermediateOutputPath Condition="$(RootIntermediateOutputPath) == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), init-tools.msbuild))/bin/obj/</RootIntermediateOutputPath>
```
eerhardt: I don't think we want to rely on `init-tools.msbuild` being around.  Can we just assume this file is 1 level below the "RepoRoot"?  And if that doesn't work for folks, they can just set RootIntermediateOutputPath themselves.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
```diff
@@ -20,6 +20,12 @@
     <CLSCompliant Condition="'$(CLSCompliant)'=='' and '$(IsTestProject)'=='true'">false</CLSCompliant>
     <CLSCompliant Condition="'$(CLSCompliant)'==''">true</CLSCompliant>
 
+    <!--
+      Find projects in the tests folder which are not executable to distinguish
+      from accompanying support libraries.
+    -->
+    <IsRealTestProject Condition="'$(IsTestProject)' == 'true' AND '$(OutputType.ToLower())' != 'exe'">true</IsRealTestProject>
```
eerhardt: I'd use a different word than "real" here.  `IsExecutableTestProject` maybe?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/FrameworkTargeting.targets
```diff
@@ -70,6 +65,12 @@
     <IncludeDefaultReferences Condition="'$(MSBuildProjectExtension)' == '.vbproj'">true</IncludeDefaultReferences>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsTestProject)' == 'true'">
+    <IncludeTestReferences Condition="'$(IncludeTestReferences)' == ''">true</IncludeTestReferences>
+    <IncludePerformanceTestReferences Condition="'$(IncludePerformanceTestReferences)' == ''">true</IncludePerformanceTestReferences>
```
eerhardt: This seems wrong.  ALL test projects are going to get perf test references?

ViktorHofer: This is the current behavior but no, I'm right now this moment changing this.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
```diff
@@ -61,6 +61,11 @@
       <GenFacadesBuildPartialReferenceFacade>true</GenFacadesBuildPartialReferenceFacade>
     </PropertyGroup>
 
+    <ItemGroup>
```
eerhardt: Why'd you move these around?  Typically, `<FileWrites>` happen AFTER the task that writes the file.

ViktorHofer: When the task below fails these files never get cleaned up. I hit this is error multiple times locally.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
ViktorHofer: It's exactly the opposite. It's not executable by default. `IsRunnableTestProject` or `IsLibTestProject`?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets
```diff
@@ -242,7 +241,6 @@
     <!-- ******* SourceLink Support, activated by setting $(UseSourceLink) ********* -->
 
     <!-- Generate the SourceLink file that will be passed to the C# compiler (using ProjectDir and GitHubRepositoryUrl as data) -->
-    <MakeDir Directories="$(BaseIntermediateOutputPath)" />
```
eerhardt: You're going to still want this MakeDir for the RootIntermediateOutputPath.  There are official build scripts that directly call the `CreateOrUpdateCurrentVersionFile` target, and the directoy may not be created yet.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
ViktorHofer: right but with what I had before I had that really ugly path: "C:\git\corefx\Tools\..\bin/....". I wanted to avoid that and currently all projects to have a init-tools.msbuild

eerhardt:
> #!/usr/bin/env bash

These "RunnerTemplate" files look very close to duplicated with the existing "RunnerTemplate" files.  Is it possible to refactor this so there is less duplication?

---
Refers to: src/Microsoft.DotNet.Build.Tasks/PackageFiles/RunnerTemplate.Quiet.Unix.txt:1 in fa23084. [](commit_id = fa230845688e83ece56aeb17b8f1721df3477b41, deletion_comment = False)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets
ViktorHofer: Ok I'll go with `IsInvokableTestProject`?

eerhardt: Good enough for me.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/partialfacades.task.targets
ViktorHofer: And just to clarify, these intermediate files already exist before the task is executed because of operations above, i.e. Move

ViktorHofer:
> These "RunnerTemplate" files look very close to duplicated with the existing "RunnerTemplate" files. Is it possible to refactor this so there is less duplication?

I will keep the duplication for now (as I don't want to touch the templating engine code). We can address that with a follow-up PR.

PR #2108, opened 2018-07-25
Make BaseIntermediateOutputPath project specific

ViktorHofer:
Contributes towards https://github.com/dotnet/corefx/pull/31364

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -37,6 +37,7 @@
 
   <!-- Common BuildTools properties -->
   <PropertyGroup>
+    <RootIntermediateOutputPath>$(MSBuildThisFileDirectory)../bin/obj/</RootIntermediateOutputPath>
     <ToolsDir Condition="'$(ToolsDir)'==''">$(MSBuildThisFileDirectory)</ToolsDir>
     <ToolRuntimePath Condition="'$(ToolRuntimePath)'==''">$(ToolsDir)</ToolRuntimePath>
     <DotnetCliPath Condition="'$(DotnetCliPath)'==''">$(ToolRuntimePath)dotnetcli/</DotnetCliPath>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
@@ -37,7 +37,7 @@
 
   <!-- Common BuildTools properties -->
   <PropertyGroup>
-    <RootIntermediateOutputPath>$(MSBuildThisFileDirectory)../bin/obj/</RootIntermediateOutputPath>
+    <RootIntermediateOutputPath Condition="$(RootIntermediateOutputPath) == ''">$(MSBuildThisFileDirectory)../bin/obj/</RootIntermediateOutputPath>
     <ToolsDir Condition="'$(ToolsDir)'==''">$(MSBuildThisFileDirectory)</ToolsDir>
     <ToolRuntimePath Condition="'$(ToolRuntimePath)'==''">$(ToolsDir)</ToolRuntimePath>
     <DotnetCliPath Condition="'$(DotnetCliPath)'==''">$(ToolRuntimePath)dotnetcli/</DotnetCliPath>
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
@@ -2,7 +2,7 @@
   
   <PropertyGroup>
     <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
-    <BuildVersionFilePath>$(RootBaseIntermediateOutputPath)</BuildVersionFilePath>
+    <BuildVersionFilePath>$(ObjDir)</BuildVersionFilePath>
     <BuildVersionFile Condition="'$(BuildVersionFile)'==''">$(BuildVersionFilePath)BuildVersion-$(TodayTimeStamp).props</BuildVersionFile>
   </PropertyGroup>
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
@@ -2,7 +2,7 @@
   
   <PropertyGroup>
     <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
-    <BuildVersionFilePath>$(ObjDir)</BuildVersionFilePath>
+    <BuildVersionFilePath>$(RootIntermediateOutputPath)</BuildVersionFilePath>
     <BuildVersionFile Condition="'$(BuildVersionFile)'==''">$(BuildVersionFilePath)BuildVersion-$(TodayTimeStamp).props</BuildVersionFile>
   </PropertyGroup>
 
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
@@ -2,7 +2,7 @@
   
   <PropertyGroup>
     <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
-    <BuildVersionFilePath>$(BaseIntermediateOutputPath)</BuildVersionFilePath>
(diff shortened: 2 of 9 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets
```diff
@@ -2,7 +2,7 @@
   
   <PropertyGroup>
     <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
-    <BuildVersionFilePath>$(BaseIntermediateOutputPath)</BuildVersionFilePath>
+    <BuildVersionFilePath>$(RootBaseIntermediateOutputPath)</BuildVersionFilePath>
```
eerhardt: Who is going to set this new property in the other repos that use buildtools?  coreclr, wcf, core-setup?

Also, would it be a terrible idea to use `$(ObjDir)` instead of a new property `$(RootBaseIntermediateOutputPath)`?  Maybe I'm just against the name.... "RootBase" feels redundant/confusing.

ViktorHofer: I don't have a strong opinion here. I  prefer `RootIntermediateOutputPath` (the base in it is a mistake) over `ObjDir` because it's more abstract and does not contain the name of the actual folder in it.

Regards setting the value, I expect follow-up work to be necessary, as in corefx.

ViktorHofer: Giving it a second thought I will revert to ObjDir to minimize necessary changes. It doesn't make sense to refactor now before we move to arcade.

eerhardt: That sounds good to me.  Thanks!

src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
```diff
@@ -66,7 +66,7 @@
           DependsOnTargets="CreateOptionalRestoreFeedItems;
                             PrepareOptionalToolProject">
     <PropertyGroup>
-      <GeneratedNuGetConfigDir>$(ObjDir)generatedNuGetConfig/</GeneratedNuGetConfigDir>
+      <GeneratedNuGetConfigDir>$(BaseIntermediateOutputPath)generatedNuGetConfig/</GeneratedNuGetConfigDir>
```
weshaggard: Why did this one get changed from ObjDir to BaseIntermediateOutputPath?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
```diff
@@ -7,8 +7,6 @@
 
   <PropertyGroup>
     <ProjectJson Condition="'$(ProjectJson)' == '' and '$(ContainsPackageReferences)' == 'true'">$(MSBuildProjectFullPath)</ProjectJson>
-    <RestoreOutputPath Condition="'$(RestoreOutputPath)' == ''">$(MSBuildProjectDirectory)</RestoreOutputPath>
-    <RestoreOutputPath Condition="'$(ContainsPackageReferences)' != 'true' And '$(ProjectJson)' != ''">$([System.IO.Path]::GetDirectoryName('$(ProjectJson)'))/obj</RestoreOutputPath>
```
weshaggard: Please ensure you test this to make sure there aren't any cases that depend on this being set. I doubt there is but please verify by testing.

weshaggard:
What is BaseIntermeidateOutputPath going to be set to in cases like https://github.com/dotnet/buildtools/blob/bc9cdc2d67c66327a62be62716c6022b334a6e07/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets#L12 and https://github.com/dotnet/buildtools/blob/d334d93614b88989563ecf3daf3efce083cd3801/src/Microsoft.DotNet.Build.Tasks.Feed/PackageFiles/Microsoft.DotNet.Build.Tasks.Feed.targets#L134. Assuming it is at least a valid path I'm OK but I just want to make sure that the projectname doesn't break these.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
ViktorHofer: This is a temporary location for creating the nuget config and then using it which doesn't need to be the intermediate root. Correct me if I'm wrong.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/packageresolve.targets
ViktorHofer: Yes, I did A LOT of tests already, basically testing the whole day all possible projects in corefx and restoring the whole time. Everything works fine.

ViktorHofer:
> I don't particularly like using ObjDir in these cases but I think I can live with it for now until we moved to the newer directory structure for Arcade.

Me neither. Coreclr currently does not have an ObjDir property, I haven't yet looked into wcf and core-setup but if I need to touch all the repos maybe it's worth to introduce a new property called  `RootIntermediateOutputPath`?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
weshaggard: No it doesn't have to be in the root it just needs to be in the a place where the project will pick it up and we seem to be explicitly passing it to the restore command so I think the change is OK.

weshaggard:
>Me neither. Coreclr currently does not have an ObjDir property, I haven't yet looked into wcf and core-setup but if I need to touch all the repos maybe it's worth to introduce a new property called RootIntermediateOutputPath?

Yeah you could introduce a new property and define/default it in https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props.

ViktorHofer:
@weshaggard I would like to get rid of the `IntermediateOutputPath` override in Symbols.targets but a comment states that its path is used by MicroBuild: https://github.com/dotnet/buildtools/blob/bc9cdc2d67c66327a62be62716c6022b334a6e07/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets#L329

I don't know what MicroBuild is but can we change the path it uses for the symbol catalogue here?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props
```diff
@@ -37,6 +37,7 @@
 
   <!-- Common BuildTools properties -->
   <PropertyGroup>
+    <RootIntermediateOutputPath>$(MSBuildThisFileDirectory)../bin/obj/</RootIntermediateOutputPath>
```
eerhardt: You should have a condition here, so someone can default it first:

`Condition="'$(RootIntermediateOutputPath)'==''`

This is typically good practice in MSBuild.  If someone else set this property, don't overwrite what they set.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
```diff
@@ -16,7 +16,8 @@
     <AssemblyName Condition="'$(AssemblyName)' == ''">$(MSBuildProjectName)</AssemblyName>
     <TargetOutputRelPath Condition="'$(IsCompatAssembly)'=='true'">$(TargetOutputRelPath)Compat/</TargetOutputRelPath>
     <OutputPath>$(ReferenceAssemblyOutputPath)$(AssemblyName)/$(AssemblyVersion)/$(TargetOutputRelPath)</OutputPath>
-    <IntermediateOutputPath>$(BaseIntermediateOutputPath)ref/$(AssemblyName)/$(AssemblyVersion)/$(TargetOutputRelPath)</IntermediateOutputPath>
+    <IntermediateOutputPath>$(RootIntermediateOutputPath)ref/$(AssemblyName)/$(AssemblyVersion)/$(TargetOutputRelPath)</IntermediateOutputPath>
+    <RestoreOutputPath>$(IntermediateOutputPath)</RestoreOutputPath>
```
eerhardt: Why are we setting `RestoreOutputPath` here? I didn't think `ref` projects got restored....

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Symbols.targets
```diff
@@ -333,6 +333,7 @@
       <!-- always overwrite IntermediateOutputPath so we know we can sign the catalogs.
            some repos do weird things with BaseIntermediateOutputPath -->
       <IntermediateOutputPath>$(SymbolCatalogIntermediateDir)</IntermediateOutputPath>
+      <RestoreOutputPath>$(IntermediateOutputPath)</RestoreOutputPath>
```
eerhardt: Why is this necessary?

ViktorHofer:
Merged into https://github.com/dotnet/buildtools/pull/2106

PR #2109, opened 2018-07-25
Ensure global prefix handles explicit interface implementations

ericstj:
Also enable use of -global for not-supported assemblies

@luqunl hit this when trying to enable System.Runtime.WindowsRuntime's not-supported assembly.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets
```diff
@@ -42,6 +42,7 @@
       <GenAPIArgs>$(GenAPIArgs) -libPath:"@(_referencePathDirectories)"</GenAPIArgs>
       <GenAPIArgs>$(GenAPIArgs) -out:"$(NotSupportedSourceFile)"</GenAPIArgs>
       <GenAPIArgs Condition="'$(GeneratePlatformNotSupportedAssembly)' == 'true' OR '$(GeneratePlatformNotSupportedAssemblyMessage)' != ''">$(GenAPIArgs) -throw:"$(GeneratePlatformNotSupportedAssemblyMessage)"</GenAPIArgs>
+      <GenAPIArgs Condition="'$(GeneratePlatformNotSupportedAssemblyWithGlobalPrefix)' == 'true'">$(GenAPIArgs) -global</GenAPIArgs>
```
ericstj: We could just make this the default for not-supported assemblies since its just an intermediate source file.  Thoughts?

ericstj: ok, that's fair.

luqunl:
Thanks a lot for this fix.

PR #2111, opened 2018-07-29
Fix VB solution guid in solution generation

danmosemsft:
My VB project did not end with VisualBasic.vbproj so it got the C# guid.


danmosemsft:
For my reference
```
vb
        public const string ProjectTypeGuid = "778DAE3C-4631-46EA-AA77-85C1314464D9";
        public const string LegacyProjectTypeGuid = "F184B08F-C81C-45F6-A57F-5ABD9991F28F";

c#
        public const string ProjectTypeGuid = "9A19103F-16F7-4668-BE54-9A1E7A4F7556";
        public const string LegacyProjectTypeGuid = "FAE04EC0-301F-11d3-BF4B-00C04F79EFBC";
```

danmosemsft:
This assumes that in the various repoes that use buildtools, there are no old-style VBproj. There are none in coresetup/corefx/coreclr.

PR #2115, opened 2018-08-01
Updating CannotSealType to report a better error message when the type is sealed due to a private constructor

tannergooding:
CC. @ericstj who helped me debug this issue.


tannergooding:
@ericstj, do the new error messages look good to you?

src/ApiCompat/Rules/Compat/CannotSealType.cs
```diff
@@ -18,7 +18,7 @@ public override DifferenceType Diff(IDifferences differences, ITypeDefinition im
             if (impl.IsEffectivelySealed() && !contract.IsEffectivelySealed())
             {
                 differences.AddIncompatibleDifference(this,
-                    $"Type '{impl.FullName()}' is sealed in the {Implementation} but not sealed in the {Contract}.");
+                        $"Type '{impl.FullName()}' is {(impl.IsSealed ? "effectively (has a private constructor)" : "actually (has the sealed modifier)")} sealed in the {Implementation} but not sealed in the {Contract}.");
```
ericstj: I think this is wrong.  It looks to me like when impl is actually sealed it will log the "effectively" sealed message.

tannergooding: Doh, will fix.

PR #2117, opened 2018-08-07
Adding an MSBuildSdkUpdater to the BaseDependenciesTask

tannergooding:
This will allow the the ILProj SDK to be automatically updated. See https://github.com/dotnet/corefx/pull/31512 for more details.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
+++ b/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
@@ -82,6 +82,9 @@ namespace Microsoft.DotNet.Build.Tasks.VersionTools
         protected Regex CreateXmlUpdateRegex(string elementName, string contentGroupName) =>
             new Regex($@"<{elementName}>(?<{contentGroupName}>.*)</{elementName}>");
 
+        protected Regex CreateMSBuildSdkUpdateRegex(string msbuildSdkName, string contentGroupName) =>
+            new Regex($@"""{msbuildSdkName}""\s*:\s*""(?<{contentGroupName}>.*)""");
+
         protected IEnumerable<IDependencyUpdater> CreateUpdaters()
         {
             if (ProjectJsonFiles != null && ProjectJsonFiles.Any())
@@ -96,6 +99,10 @@ namespace Microsoft.DotNet.Build.Tasks.VersionTools
                 {
                     case "Xml":
                         yield return CreateXmlUpdater(step);
+                        break;
+
+                    case "MSBuildSdk":
+                        yield return CreateMSBuildSdkUpdater(step);
                         break;
 
                     case "File":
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -298,6 +319,11 @@ private FileRegexUpdater ConfigureFileRegexUpdater(FileRegexUpdater updater, ITa
                 updater.Regex = new Regex(manualRegex);
                 updater.VersionGroupName = GetRequiredMetadata(step, "VersionGroupName");
             }
+            else if (!string.IsNullOrEmpty(msbuildSdkName))
+            {
+                updater.Regex = CreateMSBuildSdkUpdateRegex(Regex.Escape(msbuildSdkName), nameof(msbuildSdkName));
```
tannergooding: Escaping since the name will almost always have `.`, which needs to become `\.`

MSBuildSdkName and PackageId will also always be the same, but there isn't an easy/existing way to switch on the different types of updaters right now

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -298,6 +319,11 @@ private FileRegexUpdater ConfigureFileRegexUpdater(FileRegexUpdater updater, ITa
...
+                updater.VersionGroupName = nameof(msbuildSdkName);
+            }
             else
             {
                 throw new ArgumentException(
```
dagood: Could you add `MSBuildSdkName` to the exception message?

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
dagood: > there isn't an easy/existing way to switch on the different types of updaters right now

Yeah, I can imagine it being better with more general refactoring, but we have no automated tests (and manual tests are tedious without some extra scripts on top). Looks good for now.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
```diff
@@ -82,6 +82,9 @@ public override bool Execute()
         protected Regex CreateXmlUpdateRegex(string elementName, string contentGroupName) =>
             new Regex($@"<{elementName}>(?<{contentGroupName}>.*)</{elementName}>");
 
+        protected Regex CreateMSBuildSdkUpdateRegex(string msbuildSdkName, string contentGroupName) =>
+            new Regex($@"""{msbuildSdkName}"": ""(?<{contentGroupName}>.*)""");
```
dagood: For the `": "` match, consider being friendly to bad formatting, like `":"` or `" : "`. A manual bad change might get past review and it'd be nice for auto-update to keep working without a manual fix.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
tannergooding: Fixed.

src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs
tannergooding: Fixed. Now looks for 0+ whitespace characters.

PR #2119, opened 2018-08-08
Updating init-tools to always use PACKAGES_DIR

tannergooding:
This ensures that all packages get restored to `PACKAGES_DIR`, rather than a mix of `PACKAGES_DIR` and the NuGet default.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -5,9 +5,6 @@ set PROJECT_DIR=%~1
 set DOTNET_CMD=%~2
 set TOOLRUNTIME_DIR=%~3
 set PACKAGES_DIR=%4
-if [%PACKAGES_DIR%] == [] set PACKAGES_DIR=%TOOLRUNTIME_DIR%
-:: Remove quotes to the packages directory
-set PACKAGES_DIR=%PACKAGES_DIR:"=%
 set BUILDTOOLS_PACKAGE_DIR=%~dp0
 set MICROBUILD_VERSION=0.2.0
 set ROSLYNCOMPILERS_VERSION=2.8.0-beta4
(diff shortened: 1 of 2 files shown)


tannergooding:
FYI. @eerhardt based on the discussion we had earlier

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -80,11 +81,11 @@ Robocopy "%TOOLRUNTIME_DIR%\runtimes\any\native" "%TOOLRUNTIME_DIR%\."
 Robocopy "%BUILDTOOLS_PACKAGE_DIR%\." "%TOOLRUNTIME_DIR%\." "MSBuild.runtimeconfig.json"
 
 :: Copy Portable Targets Over to ToolRuntime
-if not exist "%PACKAGES_DIR%\generated" mkdir "%PACKAGES_DIR%\generated"
-set PORTABLETARGETS_PROJECT=%PACKAGES_DIR%\generated\project.csproj
+if not exist "%TOOLRUNTIME_DIR%\generated" mkdir "%TOOLRUNTIME_DIR%\generated"
```
tannergooding: Changed this, because it doesn't seem correct for the `generated` folder to exist inside the packages directory.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -31,7 +31,8 @@ set MSBUILD_PROJECT_CONTENT= ^
 
 set PUBLISH_TFM=netcoreapp2.0
 
-set INIT_TOOLS_RESTORE_ARGS=--source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json %INIT_TOOLS_RESTORE_ARGS%
+set DEFAULT_RESTORE_ARGS=--no-cache --packages "%PACKAGES_DIR%"
```
tannergooding: Specified `--no-cache` because that seems to be the default for the CoreFX restores done outside this init-tools.cmd

eerhardt:
> if [%PACKAGES_DIR%] == [] set PACKAGES_DIR=%TOOLRUNTIME_DIR%

If PACKAGES_DIR isn't passed in (which it doesn't look like it is in corefx), doesn't this just use the ToolRuntime?

---
Refers to: src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd:8 in dd82278. [](commit_id = dd822787ed1616ed393a96c4d633eca7ad53f5b9, deletion_comment = False)

tannergooding:
@eerhardt, yes, but that was already the "existing" behavior. 

If we have some better place to use, I'd be happy to change it to that instead (we could also just "require" that `PACKAGES_DIR` be specified)

eerhardt:
I think requiring `PACKAGES_DIR` would be a good route.  As it stands today, it doesn't get passed on Windows, which means we will be restoring packages to the `Tools` directory, which isn't correct.

tannergooding:
@eerhardt, updated both `init-tools.cmd` and `init-tools.sh` to error if `PACKAGES_DIR` or `TOOLRUNTIME_DIR` isn't passed in.

PR #2123, opened 2018-08-15
Adding ILDasm restore and resources switches in order to be able to produce resources for System.Runtime.CompilerServices.Unsafe.dll

joperezr:
This is in order to fix this issue: https://github.com/dotnet/corefx/issues/24012

S.R.CS.Unsafe is built as an ilproj where we call ilasm directly. This causes that the write version headers are not built into the managed library. With these changes (plus a PR comming up in corefx) we will be able to produce a resource file and then pass it down to ilasm so that the dll will have the right headers on it.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
@@ -43,6 +43,7 @@
       <_IlasmSwitches Condition="'$(DebugType)' == 'Impl'">$(_IlasmSwitches) -DEBUG=IMPL</_IlasmSwitches>
       <_IlasmSwitches Condition="'$(DebugType)' == 'PdbOnly'">$(_IlasmSwitches) -DEBUG=OPT</_IlasmSwitches>
       <_IlasmSwitches Condition="'$(Optimize)' == 'True'">$(_IlasmSwitches) -OPTIMIZE</_IlasmSwitches>
+      <_IlasmSwitches Condition="'$(IlasmResourceFile)' != ''">$(_IlasmSwitches) -RESOURCES=$(IlasmResourceFile)</_IlasmSwitches>
     </PropertyGroup>
 
     <Exec Command="$(IlasmToolPath) $(_IlasmSwitches) $(_OutputTypeArgument) $(IlasmFlags) -OUTPUT=@(IntermediateAssembly) $(_KeyFileArgument) @(Compile, ' ')">
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
@@ -20,6 +20,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NETCore.ILAsm" Version="$(ILAsmPackageVersion)" />
+    <PackageReference Include="Microsoft.NETCore.ILDAsm" Version="$(ILAsmPackageVersion)" />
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
```diff
@@ -20,6 +20,7 @@
 
   <ItemGroup>
```
ericstj: You might consider renaming this project.

joperezr: I thought about that, but didn't come up with a good name. `AssemblyTools.depproj` perhaps? Any good suggestion?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
```diff
@@ -43,6 +43,7 @@
       <_IlasmSwitches Condition="'$(DebugType)' == 'Impl'">$(_IlasmSwitches) -DEBUG=IMPL</_IlasmSwitches>
       <_IlasmSwitches Condition="'$(DebugType)' == 'PdbOnly'">$(_IlasmSwitches) -DEBUG=OPT</_IlasmSwitches>
       <_IlasmSwitches Condition="'$(Optimize)' == 'True'">$(_IlasmSwitches) -OPTIMIZE</_IlasmSwitches>
+      <_IlasmSwitches Condition="'$(IlasmResourceFile)' != ''">$(_IlasmSwitches) -RESOURCES=$(IlasmResourceFile)</_IlasmSwitches>
```
tannergooding: I would add the same switch to https://github.com/dotnet/coreclr/blob/master/src/.nuget/Microsoft.NET.Sdk.IL/targets/Microsoft.NET.Sdk.IL.targets#L79

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
joperezr: One thing though is that changing the project name might break some repos since the restore of this depproj happens on each individual repo, so we would have to react to the change on each one. Obviously, it's doable, but just think it's worth mentioning it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/IL.targets
joperezr: Sure we can do that. Out of curiosity, which managed assembly on coreclr gets built using ilasm? Just asking since I presume they would have the same issue where version headers are not built in.

tannergooding: CoreCLR has some 4k test projects, which haven't been migrated to use the `ilproj msbuild-sdk` yet.

CoreFX has `S.R.CS.Unsafe`, which is being ported to use the linked targets in https://github.com/dotnet/corefx/pull/31512

joperezr: I see, that makes sense. I'll add this change there as well then.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/ilasm/ilasm.depproj
joperezr: I was checking with @tannergooding 's PR that is changing to the sdk and looks like this project will soon go away anyways, so I'll keep the name for now just so that we don't have to change consuming repos twice. That said let me know if you still want me to do it so that I change it on a subsequent PR.

PR #2070, opened 2018-06-15
Port 'Add blob prefix support' to 1.1.0 buildtools

wtgodbe:
Port of https://github.com/dotnet/buildtools/pull/1494 for the sake of https://github.com/dotnet/core-setup/pull/4241

@weshaggard PTAL

PR #2125, opened 2018-08-15
Port 'Add blob prefix support' to 1.0.0 buildtools

wtgodbe:
Port of https://github.com/dotnet/buildtools/pull/2070 for 1.0.0. We need this so that we can download packages from a specific folder in blob storage in core-setup 1.x.x. Without it, we're currently restoring 2 different versions of buildtools, and only using the second one for downloading stuff from Azure.

@dagood PTAL.

### DIFF
--- a/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
+++ b/src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
@@ -12,8 +12,8 @@
     <Error Condition="'$(PublishPattern)' == ''" Text="Please specify a value for PublishPattern using standard msbuild 'include' syntax." />
 
     <PropertyGroup>
-      <ReleativeBlobPathBase>$(BlobNamePrefix)</ReleativeBlobPathBase>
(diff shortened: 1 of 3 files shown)


wtgodbe:
Part of https://github.com/dotnet/core-setup/issues/4248

src/Microsoft.DotNet.Build.CloudTestTasks/PackageFiles/PublishContent.targets
```diff
@@ -10,13 +10,18 @@
   <!-- gathers the items to be published -->
   <Target Name="GatherItemsForPattern">
     <Error Condition="'$(PublishPattern)' == ''" Text="Please specify a value for PublishPattern using standard msbuild 'include' syntax." />
+
+    <PropertyGroup>
+      <ReleativeBlobPathBase>$(BlobNamePrefix)</ReleativeBlobPathBase>
```
dagood: It would be nice to also include a port of the other commit in https://github.com/dotnet/buildtools/pull/1494 that fixes this, if there's no big reason not to. It wasn't fixed in `release/1.1.0` either, though, so... I'd say this is a non-blocking comment.

wtgodbe: Agreed & done

PR #2131, opened 2018-08-22
Remove echo line breaks on Windows

morganbr:
Port the fix for echoes triggering commands to be re-executed to the Windows scripts. This correctly prevents running ILC a second time. This should be a complete fix for https://github.com/dotnet/corefx/issues/31723 (#2106 worked around it by not echoing for uapaot, but that doesn't handle netcoreappaot or any other case we might accidentally be hitting this today).


morganbr:
Thanks for the reviews!

PR #2132, opened 2018-08-22
Fix test runs not executed

ViktorHofer:
Test runs were not executed on helix as -skiptests is usually passed to the build-tests script which then builds each test assembly, compresses it and uploads it to helix. There was a wrong condition that didn't generate the execution script and compressed the test directory, which is now fixed.

**I will merge this as soon as its green as fixing this is critical. I tested the different scenarios offline and now everything behaves as expected. I will react to feedback as it comes with a follow-up PR.**

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -1,17 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
-    <TestDisabled Condition="'$(IsTestProject)'!='true' Or '$(SkipTests)'=='true' Or ('$(IsPerformanceTestProject)' == 'true' AND '$(Performance)' == 'false')">true</TestDisabled>
-  </PropertyGroup>
 
   <!-- This is the target that copies the test assets to the test output -->
-  <Import Condition="'$(TestDisabled)' != 'true'" Project="$(MSBuildThisFileDirectory)publishtest.targets" />
-  <Import Condition="'$(TestDisabled)' != 'true' AND '$(IsPerformanceTestProject)' == 'true'" Project="$(MSBuildThisFileDirectory)PerfTesting.targets" />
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -1,17 +1,16 @@
...
-  <Import Condition="'$(TestDisabled)' != 'true' AND '$(IsPerformanceTestProject)' == 'true'" Project="$(MSBuildThisFileDirectory)PerfTesting.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)publishtest.targets" />
+  <Import Condition="'$(IsPerformanceTestProject)' == 'true' AND '$(Performance)' != 'false'" Project="$(MSBuildThisFileDirectory)PerfTesting.targets" />
   <UsingTask TaskName="GenerateTestExecutionScripts" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask TaskName="GetTargetMachineInfo" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
   <UsingTask Condition="'$(BuildingUAPVertical)' == 'true'" TaskName="ExtractResWResourcesFromAssemblies" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
+    <SkipTestRun Condition="'$(IsTestProject)' != 'true' OR ('$(IsPerformanceTestProject)' AND '$(Performance)' == 'false')">true</SkipTestRun>
+    <TestDisabled Condition="'$(SkipTests)'=='true'">true</TestDisabled>
```
eerhardt: What is the difference between `TestDisabled` and `SkipTests`?

ViktorHofer: SkipTests => Do not run the tests but still invoke all the targets in tests.targets, i.e. GenerateTestExecutionScripts so that they can be run on helix
TestDisabled => SkipTests + lazy evaluation later for test configs (https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets#L628)

PR #2133, opened 2018-08-22
Exclude Microsoft TestPlatform assemblies and resw

ViktorHofer:
Excluding Microsoft.TestPlatform and Microsoft.VisualStudio.TestPlatform resw and dlls from `CreateReswFilesForExternalDependencies` computation.


ViktorHofer:
ok will do. thx

PR #2135, opened 2018-08-23
Add minimal VSTS VersionTools auto-PR client

dagood:
I suggest turning on ignore whitespace for the `PullRequestCreator.cs` changes--I got rid of an indentation level. (GitHub's file review screen has this option accessible now!)

This client allows submitting a PR to a VSTS repo using the `VersionTools.targets` flow by setting:

* VstsPat
* VstsInstanceName (ex `devdiv`)
* ProjectRepoOwner (ex `DevDiv`. For `dotnet` instance projects, the options are `public` and `internal`)
* ProjectRepoName (ex `DotNet-SourceBuild-Trusted`)
* ProjectRepoBranch (ex `release/2.2`)
* GitHubAuthor
* VstsCommitterEmail

I'm pretty sure I left the GitHub stuff alone, but I'm going to do some basic testing.

My approach was to copy GitHubClient and force it to use VSTS instead, stuffing some VSTS info a bit awkwardly into the GitHub model classes. This was just to avoid making more significant structural changes. I referenced Darc's VSTS client and another one based on it, but kept VersionTools' way of doing things in terms of factoring.


src/Microsoft.DotNet.Build.Tasks/VersionTools/SubmitPullRequest.cs
```diff
@@ -5,20 +5,36 @@
...
-        [Required]
         public string GitHubAuthToken { get; set; }
-        [Required]
         public string GitHubUser { get; set; }
-        [Required]
         public string GitHubEmail { get; set; }
 
+        /// <summary>
+        /// A VSTS PAT. Setting this makes VstsInstanceName and GitHubAuthor required. Setting this
+        /// property indicates that the PR will be sent to VSTS (regardless whether GitHubAuthToken
```
eerhardt: What do you think about erroring out here instead if they are both set?

src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/IGitHubClient.cs
```diff
@@ -2,11 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.VersionTools.Automation.GitHubApi
 {
-    public interface IGitHubClient
+    public interface IGitHubClient : IDisposable
```
eerhardt: Should this be renamed?  `IGitClient` sounds better...  Or `IRemoteGitClient`?

src/Microsoft.DotNet.Build.Tasks/VersionTools/SubmitPullRequest.cs
dagood: From the point of view of this task, yeah, having both doesn't make sense. I had it that way at first. But when I started running the code, I saw my GitHub rate limit tick down because the upgrade itself has to fetch from dotnet/versions. Since anonymous rate limit is much less than authenticated, I changed the task to allow both even though this task doesn't use them. I have an explanation a few lines down:

> VstsPat and GitHubAuthToken may both be set: this is useful to get a larger rate limit for GitHub requests during the earlier tasks in the update process and submit the final PR to VSTS.

Looking at it again, though (and mulling over this comment)... I can see value in actually having a `PullRequestTargetService` param that lets the user specify which service they want to use very intentionally. Then error messages would be very specific, and you can't have a mistake like accidentally submitting to GitHub because VSTS creds aren't flowed properly.

src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/IGitHubClient.cs
dagood: I don't think so, at least until I can refactor the models to not be specific to GitHub API responses. Ideally what makes the most sense to me is actually layering over both of them to avoid the gaps between how GitHub and VSTS model things.

dagood:
Tested the new change successfully on GitHub too, following how the CoreFX master Maestro subscription call goes.

@eerhardt mind taking a look at the new commit?

src/Microsoft.DotNet.Build.Tasks/VersionTools/SubmitPullRequest.cs
```diff
@@ -78,19 +108,79 @@ private void TraceListenedExecute()
...
+                        VstsInstanceName,
+                        VstsApiVersionOverride))
                     {
-                        ForceCreate = AlwaysCreateNewPullRequest,
-                        MaintainersCanModify = MaintainersCanModifyPullRequest,
-                        TrackDiscardedCommits = TrackDiscardedCommits
-                    }).Wait();
+                        VstsProfile profile = idClient.GetMyProfileAsync().Result;
+
+                        PullRequestAuthor = PullRequestUser;
```
eerhardt: Why are we overwriting `PullRequestAuthor` here?

dagood: Ah, yeah, this is because normally Author defaults to User based on the user in `GitHubAuth`, but for VSTS we set the `GitHubAuth` user to the user's VSTS GUID because that's used later for the PR search.

Looking at it again, I don't like how this odd bit of logic is in the build task. I think I can move it into the `PullRequestCreator` to make VSTS more consistent to use from the .NET lib.

dagood:
This is available in 2.2.0-preview1-03124-03 (Microsoft.DotNet.BuildTools, Microsoft.DotNet.VersionTools)

Anipik:
Thanks @dagood I updated the pr with the latest version

PR #2138, opened 2018-08-24
Update to latest S.R.Metadata and S.C.Immutable

weshaggard:
Newer versions of VS/msbuild have higher versions of these
and so our tasks need to align with the higher versions in order
to unify with those higher versions.

Addresses https://github.com/dotnet/corefx/issues/31925


weshaggard:
@mmitche @dagood doesn't look like we have setup CI for 2.2 branch.

dagood:
Submitted https://github.com/dotnet/dotnet-ci/pull/1047 to add that.

PR #2141, opened 2018-08-28
Only construct HttpClient once in the DownloadFromAzure task

trylek:
This change modifies the DownloadFromAzure task by only constructing
the HttpClient once in ExecuteAsync and passing it to DownloadItem
instead of allocating it over and over again for each file; after
I checked in the compiled binary of CloudTestTasks with this fix
into ProjectN about 3 weeks back, we stopped seeing the occasional
failure when accessing the Azure storage:

System.Net.Http.HttpRequestException: An error occurred while
sending the request. ---> System.Net.WebException: Unable to connect
to the remote server ---> System.Net.Sockets.SocketException:
Only one usage of each socket address (protocol/network address/port)
is normally permitted 13.88.144.240:443

Thanks

Tomas

P.S. Sadly the diff looks extensive due to honoring whitespace
that naturally changed after I removed one nesting level from
DownloadItem. Ignoring whitespace reduces the diff to about 4 lines.


trylek:
Merged - thanks a lot Matt and Anton for your quick reviews and for tons of help with this issue in the first place!

PR #2149, opened 2018-08-29
Removing the archaic two spaces after peroid in our copyright.

wtgodbe:
The `<Copyright>` value has two spaces after the period. This breaks compliance checks with nuget.org, since they only expect a single space.

Part of https://github.com/dotnet/buildtools/issues/2144

CC @weshaggard @eerhardt


wtgodbe:
@weshaggard this will be ported to 2.2 automatically, right?

PR #2151, opened 2018-08-29
Add support for netcoreapp3.0 to packaging targets

weshaggard:
Add the 3.0 supported tfm/rid combinations.

Also add a SupportedFramework filter to exclude unknown versions
from the set. The unknown version only comes from package harvesting
where we cannot figure out the version number from the package, in
those cases we don't have a great way to get the expected version so
we should just filter them out from validation.

cc @ericstj @joperezr

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
+++ b/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
@@ -35,14 +35,14 @@
     <IdPrefix Condition="'$(PackageTargetRuntime)' != ''">$(IdPrefix)$(PackageTargetRuntime).</IdPrefix>
     <IdPrefix Condition="'$(PackageTargetFramework)' != ''">$(IdPrefix)$(PackageTargetFramework).</IdPrefix>
   </PropertyGroup>
-   
-  <!-- If Signing is enabled, we'll always want to write *.nupkg_requires_signing files --> 
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets
```diff
@@ -1121,6 +1125,15 @@
...
         <Version>$(_AssemblyVersion)</Version>
       </SupportedFramework>
+
+      <!--
+        Harvesting supported versions can sometimes lead to unknown versions when harvesting a package
+        which has placeholders for older frameworks like net45. We should just exclude those from validation
+        since we cannot easily determine what they are without looking at each specific platform itself.
+      -->
+      <ItemGroup>
+        <SupportedFramework Remove="@(SupportedFramework)" Condition="'%(SupportedFramework.Version)' == 'unknown'" />
```
joperezr: I didn't look but I assume that you updated your corefx PR to remove this from there since it will live here now.

weshaggard: I didn't update my PR yet. Both of these will work with them being in both places. I will either update my PR or remove them from corefx once this gets merged whichever happens first.

weshaggard:
>I guess that as a rule of thumb we should try to edit the files on VS when possible so that we don't get this annoying diffs.

Depends on what I'm doing. If I'm working in master I like to let sublime text normalize these line endings but if I'm in a release branch I tend to try and keep the changes minimal.

PR #2152, opened 2018-08-29
Exclude System.Collections.Immutable

weshaggard:
System.Collections.Immutable is included with msbuild from at
least 15.7+ and between 15.7 and 15.8 they updated the version they
include which conflicts with our usages across both. To eliminate
this mismatch we instead stop carrying the version with our tasks
and depend on the one they carry with them.

cc @ericstj 

FYI @tmat

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks.net45/project.json
+++ b/src/Microsoft.DotNet.Build.Tasks.net45/project.json
@@ -8,8 +8,8 @@
     "NuGet.Packaging": "4.4.0",
     "NuGet.ProjectModel": "4.4.0",
     "NuGet.Versioning": "4.4.0",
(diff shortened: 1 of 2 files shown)


weshaggard:
This is a second attempt to fix dotnet/corefx#31925.

tmat:
Makes sense.

ericstj:
This change broke a build that was still using an older MSBuild without System.Collections.Immutable.dll.  

> To eliminate this mismatch we instead stop carrying the version with our tasks and depend on the one they carry with them.

I'm not sure I understand your reasoning.  Why is the mismatch an issue?   Tasks are self-contained and we don't exchange types from Collections.Immutable with MSBuild.  Now you've removed a dependency our tasks require and they're failing when running on older VS installations.

PR #2165, opened 2018-09-19
Reenable benchview upload target

ViktorHofer:
The `UploadToBenchview` target wasn't executed because I wrongly conditioned importing the PerfTesting.targets. The target wasn't queued correctly after the `TestAllProjects` target because at that time the PerfTesting.targets project wasn't imported.


jorive:
Have you tested it on CoreFx?

ViktorHofer:
I ran it offline but did get some unrelated errors. I think this change is safe but I will do another run later.

jorive:
I think some of the errors were fixed yesterday. Could you please give it a try?

ViktorHofer:
Running now, will let you know when it's done.

ViktorHofer:
Hmm I got other unrelated local failures probably related to my configuration. As I think the change is safe let me merge it and in case something is wrong we can simply revert it.

PR #2166, opened 2018-09-19
Enable xunit diagnostics for test runs

ViktorHofer:
cc @danmosemsft @stephentoub 

Proposed by steve in an offline discussion.


src/Microsoft.DotNet.Build.Tasks/PackageFiles/xunit.runner.json
```diff
@@ -0,0 +1,4 @@
+{
+    "diagnosticMessages": true,
+    "longRunningTestSeconds": 30
```
stephentoub: Might want to bump this up to something longer, like a minute or two... we're only going to pay attention to it generally in a log if the whole suite hangs in a run, at which point it's better to avoid false positive noise. But we can also tweak it later as needed.

PR #2168, opened 2018-09-21
Generate launchSettings and enable netfx debugging

ViktorHofer:
Fixes https://github.com/dotnet/corefx/issues/32385

Moving the launchSettings generation to buildtools and add netfx debugging.
Refactor tests.targets to reduce complexity and to clean up code.

Tested:
- [x] netcoreapp
- [x] netfx
- [x] coverage
- [x] performance
- [x] uap
- [x] uapaot (building)

cc @danmosemsft @safern @eerhardt


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -72,7 +70,6 @@
     <_XunitOptions Condition="'$(XunitOptions)' != ''">$(_XunitOptions) $(XunitOptions)</_XunitOptions>
     <_XunitOptions Condition="'$(XunitMethodName)' != ''">$(_XunitOptions) -method $(XunitMethodName)</_XunitOptions>
     <_XunitOptions Condition="'$(XunitClassName)' != ''">$(_XunitOptions) -class $(XunitClassName)</_XunitOptions>
-    <_XunitOptions Condition="'$(XunitShowProgress)' == 'true'">$(_XunitOptions) -showprogress</_XunitOptions>
```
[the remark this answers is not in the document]
ViktorHofer: There is no showprogress flag in xunit. It is on by default when you use the verbose logger: https://github.com/xunit/xunit/pull/1725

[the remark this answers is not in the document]
safern: Then instead of removing this parameter could we somehow enable the verbose logger is XunitShowProgress is set to true?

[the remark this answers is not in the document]
safern: Then instead of removing this parameter could we somehow enable the verbose logger is XunitShowProgress is set to true?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -637,10 +603,30 @@ robocopy /S /NP %EXECUTION_DIR%int\%(Identity)\ %EXECUTION_DIR%native\
...
+    </PrepareForRunDependsOn>
+  </PropertyGroup>
+
+  <Target Name="GenerateLaunchSettingsFile"
+          Condition="'$(OmitLaunchSettings)' != 'true'"
+          Inputs="$(LaunchSettingsInputFileFullPath)"
+          Outputs="$(LaunchSettingsOutputFileFullPath)">
+
+    <Copy SourceFiles="$(LaunchSettingsInputFileFullPath)"
+          DestinationFolder="$(AppDesignerfolder)" />
```
danmosemsft: Maybe instead use DestinationFile with `$( LaunchSettingsOutputFileFullPath)` then it matches inputs /outputs and if you change the destination from AppDesignerFolder you don't need to change it here too

danmosemsft: If the file exists does it overwrite without complaint?

ViktorHofer: thanks that was an oversight, changed it to `DestinationFiles`. Yes overwrite is the default for the copy task.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -637,10 +603,30 @@ robocopy /S /NP %EXECUTION_DIR%int\%(Identity)\ %EXECUTION_DIR%native\
 
   <Target Name="SetupTestProperties" DependsOnTargets="CheckTestPlatforms;CheckTestCategories" />
 
+  <PropertyGroup Condition="$(IsTestProject) == 'true' AND '$(OmitLaunchSettings)' != 'true'">
```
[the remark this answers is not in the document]
ViktorHofer: I would say it's 50:50, most newer changes in corefx use the capital ones. I'll be consistent with myself and stick with them :)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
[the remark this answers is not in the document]
ViktorHofer: https://github.com/xunit/xunit/issues/1818

src/Microsoft.DotNet.Build.Tasks/PackageFiles/launchSettings.json
```diff
@@ -0,0 +1,19 @@
+{
+    "profiles": {
+        ".NET Core xUnit Console": {
+            "commandName": "Executable",
+            "executablePath": "$(BinDir)testhost/netcoreapp-$(_bc_OSGroup)-$(ConfigurationGroup)-$(ArchGroup)/dotnet.exe",
```
eerhardt: If we can use MSBuild properties here, why can't we just have 1 property for `$(TestExecutablePath)` and set it here? Similar for workingDirectory and DEVPATH. That way we don't have to duplicate the folder structures here. The structures can just be in one place - in MSBuild.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/xunit.runner.json
```diff
@@ -1,4 +1,5 @@
 {
     "diagnosticMessages": true,
-    "longRunningTestSeconds": 120
+    "longRunningTestSeconds": 120,
+    "shadowCopy": false
```
eerhardt: I thought this setting was ignored for .NET Core?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -16,6 +16,19 @@
     <TestPath Condition="'$(TestPath)'==''">$(OutDir)</TestPath>
   </PropertyGroup>
 
+  <!--
+    Copied from Microsoft.Net.Test.Sdk.targets which currently skips netstandard projects as 
+    they are not intended to be executable. We override this as we launch the test assemblies
+    directly with the runner. This is needed for VS debugging support.
+    TODO: Fix VB test project reference to Test SDK.
+  -->
+  <PropertyGroup Condition="'$(IncludeVSTestReferences)' == 'true' AND '$(BuildingNETCoreAppVertical)' == 'true' AND '$(Language)' != 'VB'">
```
eerhardt: `BuildingNETCoreAppVertical` is a very `corefx` specific concept. Does it really belong in buildtools? It definitely shouldn't go into arcade.

/cc @weshaggard

src/Microsoft.DotNet.Build.Tasks/PackageFiles/xunit.runner.json
ViktorHofer: Right, but we use it for netfx as we also use that config for the runs there.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/launchSettings.json
ViktorHofer: right if you look at commit 1 I had it like that but changed it as with that it was configuration dependent, means that if you selected the wrong configuration you couldn't launch the profile. IMO launchSettings should be build configuration independent. The duplication of the folder structure is a small price to pay in comparison to the confusion of users.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
ViktorHofer: right now the whole tests.targets file is mainly used by corefx and wcf. It's also imported by the coreclr tests but I'm not sure to what extent they are using it. As we planned to refactor the test framework and move it to arcade I would keep it in mind as a follow-up item.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/launchSettings.json
```diff
@@ -0,0 +1,19 @@
...
+            "commandLineArgs": "xunit.console.dll $(XunitArguments) -wait -parallel none",
+            "workingDirectory": "$(TestWorkingDir)$(MSBuildProjectName)/netcoreapp-$(_bc_OSGroup)-$(ConfigurationGroup)-$(ArchGroup)/"
+        },
+        ".NET Framework xUnit Console": {
+            "commandName": "Executable",
+            "executablePath": "$(TestWorkingDir)$(MSBuildProjectName)/netfx-Windows_NT-$(ConfigurationGroup)-$(ArchGroup)/xunit.console.exe",
+            "commandLineArgs": "$(XunitArguments) -wait -parallel none",
+            "workingDirectory": "$(TestWorkingDir)$(MSBuildProjectName)/netfx-Windows_NT-$(ConfigurationGroup)-$(ArchGroup)/",
+            "environmentVariables": {
+                "DEVPATH": "$(BinDir)runtime/netfx-Windows_NT-$(ConfigurationGroup)-$(ArchGroup)/"
```
safern: Doesn't matter much since in netfx we binplace the same binaries in bin/runtime and bin/testhost, but to be consistent should we use testhost instead? That is what is used by Helix and RunTests.cmd since %RUNTIME_PATH% is set to that as that is the first argument passed to the script.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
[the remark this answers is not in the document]
safern: Thanks!

ViktorHofer:
The race that currently happens in master will be fixed with this and with the corresponding change in corefx: https://github.com/dotnet/corefx/pull/32380. PTAL @weshaggard @joperezr

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -16,6 +16,19 @@
     <TestPath Condition="'$(TestPath)'==''">$(OutDir)</TestPath>
   </PropertyGroup>
 
+  <!--
```
ViktorHofer: I just moved that code block, I didn't touch any of the code.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -637,10 +599,41 @@ robocopy /S /NP %EXECUTION_DIR%int\%(Identity)\ %EXECUTION_DIR%native\
...
+  <PropertyGroup>
+    <LaunchSettingsOutputFileFullPath>$(MSBuildProjectDirectory)/$(AppDesignerFolder)/launchSettings.json</LaunchSettingsOutputFileFullPath>
+    <LaunchSettingsInputFileFullPath>$(MSBuildThisFileDirectory)launchSettings.json</LaunchSettingsInputFileFullPath>
+  </PropertyGroup>
+
+  <!--
+    Invoked either manually by the build.proj to generate or update the launch settings file
+    or by the build if it doesn't exist yet.
+  -->
+  <Target Name="GenerateLaunchSettingsFile" Condition="$(IsInvokableTestProject) == 'true' AND '$(GenerateLaunchSettings)' == 'true'">
```
weshaggard: You should have an input/output here to avoid copying this every time.

weshaggard: Where is GenerateLaunchSettings expected to get set?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
weshaggard: I agree with @ViktorHofer that this is pretty specific to corefx testing and would take a while to clean-up this for a general repo to use. There are other such usages in this targets file already. 

For the time being when we move this to arcade it will be corefx specific for a while until we have some other team interested in using it then we can generalize it.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/launchSettings.json
```diff
@@ -0,0 +1,19 @@
+{
+    "profiles": {
+        ".NET Core xUnit Console": {
+            "commandName": "Executable",
+            "executablePath": "$(BinDir)testhost/netcoreapp-$(_bc_OSGroup)-$(ConfigurationGroup)-$(ArchGroup)/dotnet.exe",
+            "commandLineArgs": "xunit.console.dll $(XunitArguments) -wait -parallel none",
+            "workingDirectory": "$(TestWorkingDir)$(MSBuildProjectName)/netcoreapp-$(_bc_OSGroup)-$(ConfigurationGroup)-$(ArchGroup)/"
+        },
+        ".NET Framework xUnit Console": {
```
weshaggard: Do we want to blindly always have .NET Framework as a configuration even if the test project doesn't support it?

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
ViktorHofer: > You should have an input/output here to avoid copying this every time.

That target is only hooked to build if the file doesn't already exist: https://github.com/dotnet/buildtools/pull/2168/files/d2d2905092f1931b1c35614addecd5bbf52d0a66#diff-425cb0dcf83ca9680e4a736c021dabe8R619. This is only to support the scenario when someone adds a new project or accidentally deleted his launchSettings.json file.

The strategy changed from generating the launchSettings.json during the test project build to hooking it into the repo build: https://github.com/dotnet/corefx/pull/32380/files#diff-3ad318a39efd856540e8cf813f20ec09R53. I talked with @joperezr about that and he also recommended that approach. This solves the race as we only once create the files when calling build and not for each project for each configuration (which caused the race in the first place).

> Where is GenerateLaunchSettings expected to get set?

It's set in the Directory.Build.props in corefx with my accompanying PR. I decided to make it an opt-in as the coreclr tests are also using the tests.targets somehow and I don't want to drop random files there.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/launchSettings.json
ViktorHofer: For now I think that's the best way but I created https://github.com/dotnet/project-system/issues/4051 earlier today to remove the copying and be able to attach the launch profiles to certain configurations.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
weshaggard: OK I think I can accept although it will require a full build for anyone adding new test projects but I suspect we can live with that. I would put all the conditions in one or the other place and not have them split between the target and the PropertyGroup that will make it clearer as to what is controlling it.

ViktorHofer: ok cool. just one question, why do you think it requires a full build when adding a new test project? the target is hooked into the test assembly build if the file doesn't yet exist. or do you mean with full build the the test assemblies build and not the root build?

ViktorHofer: > I would put all the conditions in one or the other place and not have them split between the target and the PropertyGroup that will make it clearer as to what is controlling it.

The condition on the PropertyGroup is only for the scenario after the build, if the file doesn't already exist to create it --> new project added. The conditions on the target also apply here.

The condition on the target applies to both the build and to the manual invocation from corefx. We don't want the file check here as it should overwrite already existing files when a new buildtools version comes in --> when building from root.

weshaggard: Unless I missed something you currently only plan to run this target once up front and not during the test assembly build any longer, or did I miss-understand you on that? If it runs during the build as well as upfront then I agree it will work for both scenarios.

weshaggard: >The condition on the target applies to both the build and to the manual invocation from corefx. We don't want the file check here as it should overwrite already existing files when a new buildtools version comes in --> when building from root.

That is not very clear at all based on how things are setup. If that is the goal then we should add a comment to that affect.

ViktorHofer: right, we also run during the build but only if the file doesn't already exist (i.e. new project or the file was deleted).

ViktorHofer: ok will do.

PR #2171, opened 2018-09-23
Cleanup runtimeconfigs and set randomized hashing for netfx

ViktorHofer:
Set randomized hashing globally so that we don't need to enable it at places where we assert on it in our corefx tests. Other work is fixing an issue with the runtime copying when measuring code coverage on netfx (theoretical fix, we aren't doing that currently).

Fixes the currently failing netfx tests in corefx.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/DesktopRunnerConfigFile.config
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/DesktopRunnerConfigFile.config
@@ -2,5 +2,6 @@
 <configuration>
   <runtime>
     <developmentMode developerInstallation="true" />
+    <UseRandomizedStringHashAlgorithm enabled="1" />
   </runtime>
 </configuration>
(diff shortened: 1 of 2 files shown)


ViktorHofer:
@stephentoub added you as a reviewer I saw you online. This together with https://github.com/dotnet/corefx/pull/32422 fixes failing tests in the netfx leg in corefx.

ViktorHofer:
@dotnet/dnceng build definition "DotNetCore-BuildTools-Trusted" is hanging in the "publish to Azure feed" step with the following message: "Waiting to obtain an exclusive lock on the feed.". 

https://devdiv.visualstudio.com/DefaultCollection/DevDiv/DevDiv%20Team/_build/results?buildId=2056311&view=logs

ViktorHofer:
Right, I pinged you as 30-45 minutes hanging in that state seems pretty long to me. Maybe a lock was taken but not released? Nevertheless I'll wait.

dagood:
It's normal for ProdCon blob feed publish in particular to take over an hour, locking out other builds from being able to publish. A bit of info at https://github.com/dotnet/arcade/issues/660.

PR #2174, opened 2018-09-26
Allow explicit coverage target

pjanotti:
Preparing to make CoreLib coverage to be explicit (see https://github.com/dotnet/corefx/issues/32420), opportunistically fixing report filter that was not correctly set when targeting all projects.


pjanotti:
Merging this so we can have a chance of running it on next code coverage CI run - I'll take any feedback later.

PR #2175, opened 2018-09-30
Add property to disable code coverage for project

pjanotti:
Related to https://github.com/dotnet/corefx/pull/32537#issuecomment-425685002

For now as WIP just to track needed changes.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -30,16 +30,30 @@
     <GenerateFullCoverageReport Condition="'$(GenerateCodeCoverageReportForAll)'=='true' and '$(CodeCoverageEnabled)'=='true' and '$(SkipFullCoverageReport)' != 'true'">true</GenerateFullCoverageReport>
     <GenerateVisitedMethodsReport Condition="'$(GenerateCodeCoverageReportForAll)'=='true' and '$(CodeCoverageEnabled)'=='true' and '$(GenerateVisitedReport)' == 'true'">true</GenerateVisitedMethodsReport>
 
-    <!-- We need to filter the data to only the assembly being tested. Otherwise we will gather tons of data about other assemblies. -->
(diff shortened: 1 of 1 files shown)


pjanotti:
Removing WIP status, this is useful in general to allow finer control on code coverage CI runs (see related issue https://github.com/dotnet/corefx/pull/32537#issuecomment-425685002)

src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -14,7 +14,7 @@
   <!-- Coverage options -->
   <PropertyGroup>
     <CodeCoverageEnabled>false</CodeCoverageEnabled>
-    <CodeCoverageEnabled Condition="'$(SkipTests)' != 'true' and '$(Coverage)' == 'true' and '$(IsPerformanceTestProject)' != 'true'">true</CodeCoverageEnabled>
+    <CodeCoverageEnabled Condition="'$(SkipTests)' != 'true' and '$(Coverage)' == 'true' and '$(IsPerformanceTestProject)' != 'true' and '$(DisableCodeCoverage)' != 'true'">true</CodeCoverageEnabled>
```
ViktorHofer: Why do we need to introduce another property here? Can't we just reuse the existing `CodeCoverageEnabled´ one? i.e.

```
<CodeCoverageEnabled Condition="'$(CodeCoverageEnabled)' == '' and '$(SkipTests)' != 'true' and '$(Coverage)' == 'true' and '$(IsPerformanceTestProject)' != 'true' and '$(DisableCodeCoverage)' != 'true'">true</CodeCoverageEnabled>
<CodeCoverageEnabled Condition="'$(CodeCoverageEnabled)' == ''">false</CodeCoverageEnabled>
```

weshaggard: I tend to agree we should try and use the same property.

pjanotti: Doh! Agreed.

PR #2180, opened 2018-10-07
Support non SDK projects in targeting packs

ViktorHofer:
Fixing coreclr build breaks with newer buildtools versions. That allows coreclr to use the 3.0 versions.


ViktorHofer:
cc @tannergooding

ViktorHofer:
Merging for now to unblock coreclr. Will address feedback later.

PR #2181, opened 2018-10-07
Automatically copy runtime config files if executable

ViktorHofer:
Relates to https://github.com/dotnet/corefx/issues/20109

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -107,9 +107,23 @@
   <PropertyGroup>
     <RuntimeConfigName Condition="'$(BuildingNETCoreAppVertical)' == 'true'">netcoreapp.runtimeconfig.json</RuntimeConfigName>
     <RuntimeConfigName Condition="'$(BuildingNETFxVertical)' == 'true'">DesktopRunnerConfigFile.config</RuntimeConfigName>
+    <RuntimeConfigExtension Condition="'$(BuildingNETCoreAppVertical)' == 'true'">.runtimeconfig.json</RuntimeConfigExtension>
+    <RuntimeConfigExtension Condition="'$(BuildingNETFxVertical)' == 'true'">.exe.config</RuntimeConfigExtension>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -107,10 +107,24 @@
...
+    
+    <!-- Copy the app.config / runtimeSettings.json file into the OutDir if the project is an executable test project. -->
+    <CopyConfigurationFiles Condition="'$(CopyConfigurationFiles)' == ''
+                                  AND ('$(BuildingNETCoreAppVertical)' == 'true' OR '$(BuildingNETFxVertical)' == 'true')
+                                  AND '$(IsTestProject)' == 'true'
+                                  AND '$(IsInvokableTestProject)' != 'true'">true</CopyConfigurationFiles>
   </PropertyGroup>
 
+  <Target Name="CopyConfigurationFilesTarget" Condition="'$(CopyConfigurationFiles)' == 'true'">
+    <Copy SourceFiles="$(RuntimeConfigPath)"
```
safern: Why use copy and not add it SuplementalTestData? I think it is safer to use the current binplace system rather than copying it manually to avoid race conditions. 

You could potentially make CopySupplementalTestData depend on this target.

ViktorHofer: I tried different approaches with SupplementalTestData but it didn't work out. In fact it should just be part of the CopyToOutputDirectory items collection as we did included the configuration files manually in the respective projects in corefx. Added a commit to address that.

ViktorHofer: The SupplementalTestData group only works for the invokable test projects. If you use that for non invokable test projects like the RemoteExecutorConsoleApp then you end up putting the test data into its output directory which then not gets binplaced into the invokable test project directory as we only binplace the executable and pdb files. I added it to the CopyToOutputItems as they will correctly be copied to the invokable test project folder.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -344,7 +353,7 @@
            directory.
        2.  Runs the tests. -->
   <Target Name="GenerateTestExecutionScripts"
-          DependsOnTargets="DiscoverTestInputs;SetupTestProperties;MakeTestSpecificResourcesPriFile;MakeCommonResourcesPriFile">
```
safern: Why do we don't depend on all the supplemental data that is added in DiscoverTestInputs target?

ViktorHofer: The supplemental test data and the configuration files are copied as part of the build target. To run the test target the assembly needs to be built. No need to have the depends on here as as that target is part of the build.

PR #2182, opened 2018-10-08
Diff attributes applied to classes between contract and implementation

ericstj:
This makes ApiCompat diff attributes applied to classes between contract and implementation.

By default all attributes are included and attributes can be opted out via an exclude list, similar
to how we exclude the same attributes from being written by GenAPI.


weshaggard:
FYI: corefx actually doesn't consume these bits any longer as they have moved to arcade. So you will need to make the same change over there to get the changes into corefx.

PR #2184, opened 2018-10-08
WIP: Make Coverlet.Console Default Code Coverage Tool

pjanotti:
/cc @ViktorHofer @danmosemsft @AlexGhiondea @tonerdo 

Eventually when https://github.com/tonerdo/coverlet/pull/209 is published in a official package we should make Coverlet the official code coverage tool for CoreFx. This PR simply switches from Coverlet.MSBuild to Coverlet.Console (a command-line tool) that allows us to keep the same model already in code coverage runs in CI.

Marked as WIP because it can't be merged until 2 things happen:

1. https://github.com/tonerdo/coverlet/pull/209 is published in an official package;
2. The official package is added as a dotnet CLI global tool, see https://github.com/dotnet/arcade/blob/master/eng/common/init-tools.ps1#L211-L222 for example on how to do that;

Per linked init-tools it seems that the preference is to add it to the global path, so this PR assumes that coverlet is available on the path but we can also use a full explicit path if needed.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
@@ -6,7 +6,6 @@
   -->
   <PropertyGroup>
     <OpenCoverVersion>4.6.519</OpenCoverVersion>
-    <CoverletMsbuildVersion>2.3.0</CoverletMsbuildVersion>
     <ReportGeneratorVersion>3.0.1</ReportGeneratorVersion>
     <CoverallsUploaderVersion>1.4</CoverallsUploaderVersion>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/CodeCoverage.targets
```diff
@@ -6,7 +6,6 @@
   -->
   <PropertyGroup>
     <OpenCoverVersion>4.6.519</OpenCoverVersion>
-    <CoverletMsbuildVersion>2.3.0</CoverletMsbuildVersion>
```
pjanotti: There is a corresponding change in CoreFx, removing the properties at: https://github.com/dotnet/corefx/blob/d6646295b889daf9cf46caae1f9097b719873439/external/test-runtime/XUnit.Runtime.depproj#L22

pjanotti:
When adding Coverlet.Console to dotnet CLI we should also add the reportgenerator per Code Coverage instructions.

ViktorHofer:
Why do we want to use Coverlet.Console instead of Coverlet.MsBuild? What is the advantage of it?

pjanotti:
It allows us to separate the steps of building and running code coverage, the two tasks for coverlet.msbuild share information and as such need to be executed in conjunction. This in the end means that you build one test suite and run coverage for it, which is not ideal for CI in which we want to stop if we fail to build any test module.

tonerdo:
@pjanotti @tarekgh tonerdo/coverlet#209 has been merged in and is available in release: https://www.nuget.org/packages/coverlet.msbuild/2.3.1

PR #2189, opened 2018-10-16
Fix wrong condition in BinPlace

ViktorHofer:
cc @weshaggard @stephentoub 

This is the final commit. The previous ones didn't quite work but this one now builds fine in corefx. I probably discovered at least one msbuild issue which I'll log.

Again merging this one now to unblock the maestro PRs in all repos and gonna apply feedback later.


ViktorHofer:
ok still some more work required here.

PR #2191, opened 2018-10-19
Refactor SupplementalTestData

ViktorHofer:
Refactor SupplementalTestData code and avoid the custom ItemDefinitionGroup which currently causes issues in coreclr.

(This is a backport from the arcade test framework effort)

Tested in corefx.


ViktorHofer:
I'm using a msbuild 15 feature here, is that safe to use on our CI machines?

ViktorHofer:
Merging because of https://github.com/dotnet/corefx/issues/24320

PR #2196, opened 2018-10-26
Update codeOptimization targets

DrewScoggins:
This change has two parts to it. The first is fixing a breaking change
that was introduced with the Arcade work in CoreFX. The restore target
that we had for brining down the IBC data was relying on the sync
target which was removed. It now follows the restore target.

The second piece is work that allows us to specify the IBC data package
that we will pull down depending on whether we are doin application for
Linux or Windows.


DrewScoggins:
PTAL @weshaggard @adiaaida @maririos

adiaaida:
LGTM

src/Microsoft.DotNet.Build.Tasks/PackageFiles/codeOptimization.targets
```diff
@@ -84,7 +85,7 @@
 
   <!-- We need the OptimizationData package in order to be able to optimize the assembly -->
   <Target Name="RestoreOptimizationDataPackage"
-          BeforeTargets="Sync"
+          BeforeTargets="Restore"
```
weshaggard: Please just remove this we should avoid this "magic" hook-up because we will easily break it if we don't look for it.

adiaaida: How would you suggest these targets get called?

weshaggard: Actually you should be careful about changing this at all. If there are other repo's still hooking this up with Sync either changing it or removing will break those.

weshaggard: >How would you suggest these targets get called?

It should be explicitly pulled in like I did in https://github.com/dotnet/corefx/pull/33076.

DrewScoggins:
I decided to just remove the dependency on Sync, since we are calling this explicitly . @weshaggard

PR #2205, opened 2018-11-20
Update MSBuildToolsVersion check to support "Current"

stephentoub:
I'm not sure if this _usingRoslyn check is still meaningful (don't we always build with Roslyn now?), but recent Visual Studio builds include a tools version coming back as "Current", which then causes the failure:
```
error MSB4086: A numeric comparison was attempted on "$(MSBuildToolsVersion)" that evaluates to "Current" instead of a number, in condition "'$(MSBuildToolsVersion)' == 'dogfood' OR '$(MSBuildToolsVersion)' >= '14.0'"
```
when building Corelib, so I'm adding "Current" to the list of versions that means we're using Roslyn.

Fixes https://github.com/dotnet/coreclr/issues/21069
cc: @danmosemsft, @jkotas

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
@@ -6,7 +6,7 @@
          the third part is for bugfixes, the fourth is for hotfixes/securtity-updates -->
     <APIVersion>$([System.Version]::Parse('$(AssemblyVersion)').ToString(2))</APIVersion>
 
-    <_usingRoslyn Condition="'$(MSBuildToolsVersion)' == 'dogfood' OR '$(MSBuildToolsVersion)' &gt;= '14.0'">true</_usingRoslyn>
+    <_usingRoslyn Condition="'$(MSBuildToolsVersion)' == 'dogfood' OR Condition="'$(MSBuildToolsVersion)' == 'Current' OR '$(MSBuildToolsVersion)' &gt;= '14.0'">true</_usingRoslyn>
 
     <!-- Create a common root output directory for all reference assemblies -->
     <ReferenceAssemblyOutputPath Condition="'$(ReferenceAssemblyOutputPath)' == ''">$(BaseOutputPath)ref/</ReferenceAssemblyOutputPath>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/ReferenceAssemblies.targets
```diff
@@ -6,7 +6,7 @@
          the third part is for bugfixes, the fourth is for hotfixes/securtity-updates -->
     <APIVersion>$([System.Version]::Parse('$(AssemblyVersion)').ToString(2))</APIVersion>
 
-    <_usingRoslyn Condition="'$(MSBuildToolsVersion)' == 'dogfood' OR '$(MSBuildToolsVersion)' &gt;= '14.0'">true</_usingRoslyn>
+    <_usingRoslyn Condition="'$(MSBuildToolsVersion)' == 'dogfood' OR '$(MSBuildToolsVersion)' == 'Current' OR '$(MSBuildToolsVersion)' &gt;= '14.0'">true</_usingRoslyn>
```
jkotas: I think you can delete this whole line and use of this property at the end of this file. I do not see `_usingRoslyn` used anywhere else.

stephentoub: Yup.  I'll leave it as-is for now, as I've already tested this change and was hoping to just keep things as minimal as possible to get it working again.  But I'll open an issue to review all of this, including whether it's even necessary to do these Roslyn checks at all anymore.

dagood:
@stephentoub Any plans to port this to `release/2.1` and `release/2.2` too? (After figuring out if the whole property should just be removed?) I hit this trying to get source-build working on Windows.

(I've uninstalled the new version of VS from my machine for now, which seems to work.)

danmosemsft:
@dagood  if you need this change in 2.1 or 2.2 to make source build work, probably you can just go ahead and port it as part of your larger changes, since that effort is presumably okayed by shiproom.

dagood:
I don't think we're going to drive any action here--it doesn't seem like a blocker, and even so, we don't care much about building on Windows. Just wondering about status.

PR #2207, opened 2018-11-27
Launch xunit.console.dll using 'dotnet' in CoreRT testing

trylek:
This change modifies the test execution script generator so that,
in CoreRT, it uses the 'dotnet' command to execute xunit.console.dll;
when we launch the app using 'call' from the VS 2019 developer
prompt, it crashes due to being unable to find System.Runtime
4.2.0.0.

Thanks

Tomas

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
@@ -87,11 +87,11 @@
   </PropertyGroup>
 
   <!-- netcoreapp (not aot) uses the dotnet host. -->
-  <PropertyGroup Condition="'$(BuildingNETCoreAppVertical)' == 'true' AND '$(UseDotNetNativeToolchain)' != 'true'">
+  <PropertyGroup Condition="'$(UseDotNetNativeToolchain)' != 'true'">
     <TestHostExecutablePath Condition="'$(TestHostExecutablePath)' == '' AND '$(RunningOnUnix)' != 'true'">%RUNTIME_PATH%\dotnet.exe</TestHostExecutablePath>
     <TestHostExecutablePath Condition="'$(TestHostExecutablePath)' == '' AND '$(RunningOnUnix)' == 'true'">$RUNTIME_PATH/dotnet</TestHostExecutablePath>
-    <TestProgram Condition="'$(BuildingNETCoreAppVertical)' == 'true'">$(TestHostExecutablePath)</TestProgram>
-    <TestArguments Condition="'$(BuildingNETCoreAppVertical)' == 'true'">$(XunitExecutable) $(XunitArguments)</TestArguments>
+    <TestProgram>$(TestHostExecutablePath)</TestProgram>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
```diff
@@ -87,11 +87,11 @@
   </PropertyGroup>
 
   <!-- netcoreapp (not aot) uses the dotnet host. -->
-  <PropertyGroup Condition="'$(BuildingNETCoreAppVertical)' == 'true' AND '$(UseDotNetNativeToolchain)' != 'true'">
+  <PropertyGroup Condition="'$(UseDotNetNativeToolchain)' != 'true'">
```
jkotas: The original condition was added by @ViktorHofer in https://github.com/dotnet/buildtools/pull/2168/files#diff-425cb0dcf83ca9680e4a736c021dabe8R89 . Does everything in corefx still work without this condition?

ViktorHofer:
which repository benfits from that? if corefx then this is the wrong place to make the change as we will enable the new testing framework based on arcade probably later this week. This is the new place where the test targets live: https://github.com/dotnet/arcade/pull/1318

src/Microsoft.DotNet.Build.Tasks/PackageFiles/tests.targets
trylek: That's what I'm seeing in the CoreRT repo - it uses the MSIL "xunit.console.dll" app to run the AOT tests, it was just executing it using "call", not "dotnet"; this doesn't work well in VS2019 - according to JanK, we would need binding redirects as "call" launches the xunit.console.dll using the desktop .NET Framework, not .NET Core.

trylek:
@ViktorHofer As I mentioned in the PR description, this is intended to be used in the CoreRT repo. I suspect that, once we switch the repo over to use Arcade, we'll need to double-check the version of tests.targets in Arcade as it will might require the same fix.

ViktorHofer:
> this is intended to be used in the CoreRT repo

I see. Still would be nice to make sure this doesn't break anything in corefx but I guess we will see.

trylek:
Hmm, have you got any suggestions as to how to test this change in corefx context?

ViktorHofer:
Locally by changing the lines in the Tools folder in corefx and then running a test project with TargetGroup uapaot.

ViktorHofer:
Just to let you know, corefx isn't using any of the test related files anymore in buildtools since yesterday. So I'm fine with whatever changes you need here but watch out to not break others like coreclr.

trylek:
Thanks for the heads-up Viktor. Frankly speaking I have temporarily dropped the VS2019 effort as I have more pressing tasks on my plate right now. I intend to return to this over the holiday period and hopefully figure something out. Based on your offline help I have modified the CoreRT scripts such that they no longer require a BuildTools change so I hope the concerns you pointed out should be hopefully moot.

ericstj:
Closing as this code is no longer used in master.

PR #2209, opened 2018-11-30
Enable opt-out of binplacing targets

ericstj:


ViktorHofer:
Hard to see in the diff, I assume you didn't change anything besides the Condition?

ericstj:
Correct, just factored it into a separate file.  I could do it in two commits if it made it more clear.

ViktorHofer:
No that's fine for me. Thanks

PR #2212, opened 2019-01-02
Fix build for paths with spaces

ViktorHofer:
Contributes to https://github.com/dotnet/corefx/pull/34315

Fix build errors when paths contain empty spaces in the init-tools and optional tools setup.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
@@ -33,15 +33,15 @@
   -->
   <Target Name="GetOptionalToolingPaths">
     <PropertyGroup>
-      <OptionalToolingDir>$(ToolsDir)optional-tool-runtime\</OptionalToolingDir>
+      <OptionalToolingDir>$(ToolsDir)optional-tool-runtime</OptionalToolingDir>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/OptionalTooling.targets
```diff
@@ -33,15 +33,15 @@
   -->
   <Target Name="GetOptionalToolingPaths">
     <PropertyGroup>
-      <OptionalToolingDir>$(ToolsDir)optional-tool-runtime\</OptionalToolingDir>
+      <OptionalToolingDir>$(ToolsDir)optional-tool-runtime</OptionalToolingDir>
```
stephentoub: Why did you move the `\` from here to all of the places it's used?

ViktorHofer: Because a quote follows the backslash further down which causes parsing errors: https://github.com/dotnet/buildtools/pull/2212/files#diff-27ba4b34f549889e007e06d384e699ecR85

stephentoub: I see.

ViktorHofer:
Thanks!

PR #2213, opened 2019-01-17
Use NullValueHandling.Ignore in GitHub client

dagood:
GitHub seems to have changed to no longer handle null in create tree calls. The API returns a 422 Unprocessable Entity error "Must supply tree.sha or tree.content" when we specify tree.sha as null *and* tree.content as the text content we want. Omit the tree.sha property completely to fix this.

For example, this doesn't work anymore:

```js
{
  "base_tree": "1abcea798bad23c4ff88e89ebaf2a3ff51733c48",
  "tree": [
    {
      "path": "build-info/dotnet/core-setup/master/Last_Build_Packages.txt",
      "mode": "100644",
      "type": "blob",
      "sha": null,
      "content": "Microsoft.DotNet.PlatformAbstractions 3.0.0-preview-27316-0..."
    },
    // ...
  ]
}
```

This does:

```js
{
  "base_tree": "1abcea798bad23c4ff88e89ebaf2a3ff51733c48",
  "tree": [
    {
      "path": "build-info/dotnet/core-setup/master/Last_Build_Packages.txt",
      "mode": "100644",
      "type": "blob",
      "content": "Microsoft.DotNet.PlatformAbstractions 3.0.0-preview-27316-0..."
    },
    // ...
  ]
}
```

I sent a GitHub support request asking about this, too.

I'm getting ramped up to test it out locally, haven't run dev build buildtools in a while.

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/GitHubClient.cs
@@ -29,7 +29,13 @@ namespace Microsoft.DotNet.VersionTools.Automation.GitHubApi
 
         private static JsonSerializerSettings s_jsonSettings = new JsonSerializerSettings
         {
-            ContractResolver = new CamelCasePropertyNamesContractResolver()
+            ContractResolver = new CamelCasePropertyNamesContractResolver(),
+
+            // GitHub seems to have changed to no longer handle null in create tree calls. The API
+            // returns a 422 Unprocessable Entity error "Must supply tree.sha or tree.content" when
+            // we specify tree.sha as null *and* tree.content as the text content we want. Omit the
+            // tree.sha property completely to fix this.
+            NullValueHandling = NullValueHandling.Ignore
         };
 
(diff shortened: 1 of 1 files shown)


dagood:
Copying `Microsoft.DotNet.VersionTools.dll` over makes my CoreFX repro start working, I'm satisfied with that. `UpdateToRemoteDependencies` also still works as expected.

dagood:
GitHub support has confirmed this is a intentional breaking API change. It is part of "some improvements to our validation" so it sounds unlikely to be rolled back.

FYI @wtgodbe @MichaelSimons @dotnet/dnceng: the error shows up during versions repo update/publish as a `422 Unprocessable Entity` error, with this GitHub response:

```json
{
  "message":"Must supply tree.sha or tree.content",
  "documentation_url":"https://developer.github.com/v3/git/trees/#create-a-tree"
}
```

BuildTools/VersionTools need to be updated by anyone using it to publish to dotnet/versions.

---

I've followed up asking how we can know what other things may have changed that we just haven't hit yet, and how we can know about this kind of change in the future.

Issue #2219, opened 2019-01-29
Upgrade to C#8


benaadams:
Since preview 2 is out and it includes C#8 preview; the version of C# used should be upgraded so it can be dogfooded before release.

Currently it fails with  https://github.com/dotnet/coreclr/pull/22272
```
CSC : error CS1617: Invalid option '8.0' for /langversion. 
Use '/langversion:?' to list supported values.
 [D:\j\workspace\x64_release_w---8f801950\src\System.Private.CoreLib\System.Private.CoreLib.csproj]
```

PR #2220, opened 2019-01-31
Upgrading Roslyn to 3.0.0-beta2-final

tannergooding:
This moves Roslyn to the C# 8.0 Preview 2 compiler


tannergooding:
Fixes #2219 

CC. @jaredpar, @agocke, @jkotas, @danmosemsft, @eerhardt, @safern, @benaadams

jaredpar:
I thought buildtools was deprecated. why upgrade it instead of removing it from the repositories that need a new C#? @tmat

eerhardt:
FYI - @dleeapho @dseefeld for awareness in source-build.

tannergooding:
> I thought buildtools was deprecated. why upgrade it instead of removing it from the repositories that need a new C#

If that is the case, that would probably be better. We should probably also indicate that somewhere on the repo homepage (or maybe archive it).

What is the current mechanism to do this?

tannergooding:
Ping. @tmat?

Looks like CoreFX might be using the new mechanism, as the binlogs show it using `D:\repos\corefx\.dotnet\dotnet.exe D:\repos\corefx\.dotnet\sdk\2.1.401\Roslyn\bincore\csc.dll`
CoreCLR, however, looks to still be using BuildTools, as the binlogs show it using `D:\repos\coreclr\Tools\net46\roslyn\tools\csc.exe`

tmat:
@tannergooding I don't have much knowledge of the current usage of buildtools or what version of the compiler should CoreFX be using. @ericstj

ericstj:
We don't rely use this anymore in master.  We should use the arcade mechanism.  I can point you to the change to make, in a moment.

tannergooding:
@ericstj, what about CoreCLR which looks to still be using BuildTools?

tannergooding:
@ericstj, looks like https://github.com/dotnet/buildtools/pull/2220#issuecomment-460748607 isn't sufficient.

It looks like we override the restore targets and completely skip the normal arcade package restore in favor of our external\dir.proj restore. This prevents the package from getting restored in the first place.

Getting it working looks to require us to add the appropriate logic to our custom restore targets and then additional replicate the nuget.g.props logic to ensure that the `Build.props`/`Build.targets` files that are part of `Microsoft.Net/NetCore.Compilers` are included by each project at the appropriate point.

jkotas:
> Getting it working looks to require us to add the appropriate logic to our custom restore targets

This sounds like a discussion for the Arcade clone of this PR.

ericstj:
> looks like #2220 (comment) isn't sufficient.

We still do the restore of tools.  No need for arcade changes. Have a look at eng/tools.props and the props/targets in our root.  You may just need to set the compiler properties in a different props file.  I can help if you point me to a PR where you’ve tried.

PR #2222, opened 2019-02-13
Updating init-tools.cmd to also restore Microsoft.NETCore.Compilers

tannergooding:
This updates init-tools.cmd to also restore Microsoft.NETCore.Compilers which allows CoreCLR to be compiled using `dotnet msbuild` on Windows.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
@@ -26,6 +26,7 @@ set MSBUILD_PROJECT_CONTENT= ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
     ^^^<PackageReference Include=^"Microsoft.Net.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
+    ^^^<PackageReference Include=^"Microsoft.NETCore.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
   ^^^</ItemGroup^^^> ^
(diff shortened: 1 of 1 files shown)


tannergooding:
CC. @jkotas, @sbomer, @jashook

tannergooding:
As per https://github.com/dotnet/coreclr/pull/22452#issuecomment-462950553

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -26,6 +26,7 @@ set MSBUILD_PROJECT_CONTENT= ^
   ^^^<ItemGroup^^^> ^
     ^^^<PackageReference Include=^"MicroBuild.Core^" Version=^"%MICROBUILD_VERSION%^" /^^^> ^
     ^^^<PackageReference Include=^"Microsoft.Net.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
+    ^^^<PackageReference Include=^"Microsoft.NETCore.Compilers^" Version=^"%ROSLYNCOMPILERS_VERSION%^" /^^^> ^
```
jkotas: Do we also need to copy this over to somewhere?  (look for "Copy Roslyn Compilers Over to ToolRuntime" later in this file)

tannergooding: No, we can just reference it directly from the packages folder as is done on Unix systems.

tannergooding: I'll double check I didnt miss something, however.

tannergooding: Yes. Looks like this is fine. Unix (`init-tools.sh`) does no such copying and everything looks to work correctly locally.

I imagine it would be possible to not copy for the desktop case (and just reference the files directly) as well (it might make some logic simpler), but that would be a separate PR.

PR #2223, opened 2019-02-15
Updating the Roslyn Toolset to 3.0.0-beta3-final

tannergooding:
CC. @jaredpar, @agocke, @jkotas, @danmosemsft, @eerhardt, @safern

The beta3 package was published yesterday.


tannergooding:
> I guess this is for coreclr, right?

Correct. We should be able to stop updating this whenever CoreCLR moves to arcade (and hopefully get it on the same auto-updating scheme that we get CoreFX on, whatever that is).

Issue #2224, opened 2019-02-15
Using eval dotnet command instead of dotnet command in buildscripts


Anipik:
There has been some recent changes in how dotnet cli parser is parsing the arguments. https://github.com/dotnet/cli/pull/9892

Earlier they were directly passing the  command line arguments to the MSBuild command line as `/p:RestorePackagesPath="/git/packages"`; thus, MSBuild will strip the quotes out when setting the property because it parses its own command line.

But now they are trying to treat `"/git/packages"` as the path to expand before passing the full path to MSBuild. since quotes are valid directory characters, this would expand this path parameter to `<cwd>/"/git/packages"`.

One of the proposed solution
https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh#L82

is to convert these lines 
`$__DOTNET_CMD restore "${__PORTABLETARGETS_PROJECT}" $__INIT_TOOLS_RESTORE_ARGS`
to
`eval "$__DOTNET_CMD restore \"${__PORTABLETARGETS_PROJECT}\" $__INIT_TOOLS_RESTORE_ARGS"`

Creating an issue to discuss about the pros and cons of using this solution 

Related issue https://github.com/dotnet/cli/issues/10802

cc @danmosemsft @eerhardt @safern

eerhardt:
Since build tools is basically "dead", we should probably just do whatever it takes to make this work. We will move to arcade, and all of this will be replaced.

safern:
I agree. Also, if eval works with 2.1 SDK then it should technically be safe to do this change. I don't see anyone using an older SDK and using latest buildtools from master.

PR #2226, opened 2019-02-19
Using eval and explicit variables

Anipik:
Fixes https://github.com/dotnet/buildtools/issues/2224
dotnet cli team changed on how we pass arguments to dotnet commands. The change was introduced in this https://github.com/dotnet/cli/pull/9892

More info is available here https://github.com/dotnet/cli/issues/10802

I tested 4 testcases here
- path without spaces on netcore2.1
- path without spaces on netcore3.0
- path with spaces on netcore2.1
- path with spaces on netcore3.0

@safern @eerhardt any other good way to test these changes

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
@@ -34,9 +34,9 @@ __PORTABLETARGETS_PROJECT_CONTENT="
 
 __PUBLISH_TFM=netcoreapp2.0
 
-__DEFAULT_RESTORE_ARGS="--no-cache --packages \"${__PACKAGES_DIR}\""
-__INIT_TOOLS_RESTORE_ARGS="${__DEFAULT_RESTORE_ARGS} --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json ${__INIT_TOOLS_RESTORE_ARGS:-}"
-__TOOLRUNTIME_RESTORE_ARGS="--source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json ${__INIT_TOOLS_RESTORE_ARGS}"
+__DEFAULT_RESTORE_ARGS="--no-cache --packages"
+__INIT_TOOLS_SOURCES="--source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json"
+__TOOLRUNTIME_SOURCES="--source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"
 
 if [ ! -d "$__PROJECT_DIR" ]; then
     echo "ERROR: Cannot find project root path at '$__PROJECT_DIR'. Please pass in the source directory as the 1st parameter."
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -34,9 +34,9 @@ __PORTABLETARGETS_PROJECT_CONTENT="
 
 __PUBLISH_TFM=netcoreapp2.0
 
-__DEFAULT_RESTORE_ARGS="--no-cache --packages \"${__PACKAGES_DIR}\""
-__INIT_TOOLS_RESTORE_ARGS="${__DEFAULT_RESTORE_ARGS} --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json ${__INIT_TOOLS_RESTORE_ARGS:-}"
-__TOOLRUNTIME_RESTORE_ARGS="--source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json ${__INIT_TOOLS_RESTORE_ARGS}"
+__DEFAULT_RESTORE_ARGS="--no-cache --packages"
```
safern: Do we really need to remove __PACKAGES_DIR from this variable?

Anipik: It was required for the path with spaces subcase. if we don't supply the path variable directly, it was being treated as separate argument, the quotes was getting removed by the eval command and dotnet was treating it as a new argument

eerhardt: Maybe then let's leave `--packages` out of this variable. As it makes no sense to use `__DEFAULT_RESTORE_ARGS` by itself. You always have to append the __PACKAGES_DIR after using it.

So remove `--packages` here, and put it into all the places using this variable.
ex.

```
${__DEFAULT_RESTORE_ARGS} --packages \"${__PACKAGES_DIR}\"
```

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh
```diff
@@ -34,9 +34,9 @@ __PORTABLETARGETS_PROJECT_CONTENT="
 
 __PUBLISH_TFM=netcoreapp2.0
 
-__DEFAULT_RESTORE_ARGS="--no-cache --packages \"${__PACKAGES_DIR}\""
-__INIT_TOOLS_RESTORE_ARGS="${__DEFAULT_RESTORE_ARGS} --source https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json --source https://api.nuget.org/v3/index.json ${__INIT_TOOLS_RESTORE_ARGS:-}"
```
eerhardt: >__INIT_TOOLS_RESTORE_ARGS [](start = 172, length = 25)

It looks like previously someone could have defined `__INIT_TOOLS_RESTORE_ARGS` before this script and the previous value would get used in this variable. But now that behavior looks no longer possible. Is that OK?

safern: Plus, we could have a variable for the sources, that one doesn't need path evaluation does it?

Anipik: No the sources variable doesnot need any path evaluation.

> `__INIT_TOOLS_RESTORE_ARGS` before this script and the previous value would get used in this variable. But now that behavior looks no longer possible

i will add that back

PR #2238, opened 2019-03-28
Updating the toolset compiler to use the cross-platform package and target v3.1.0-beta1-19172-05

tannergooding:
CC. @jaredpar, @jkotas, @safern 

This pulls in the latest compiler (same one as CoreFX was just moved to) in order to get some of the nullable fixes that have recently gone in.

We now have `Microsoft.Net.Compilers.Toolset` that contains both .NET Core and Desktop binaries, rather than having separate/independent packages.

### DIFF
--- a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
+++ b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
@@ -15,13 +15,4 @@
     <DebugType Condition="'$(RunningOnCore)' == 'true'">Portable</DebugType>
   </PropertyGroup>
 
-  <!-- If we're not using the compiler server, set ToolPath/Exe to direct to
-       the exes in this package -->
-  <PropertyGroup Condition="'$(UseRoslynCompilers)' != 'false' and '$(RoslynIncompatibleMsbuildVersion)' == 'true'">
-    <CscToolPath>$(_RoslynTasksDirectory)</CscToolPath>
-    <CscToolExe>csc.exe</CscToolExe>
-    <VbcToolPath>$(_RoslynTasksDirectory)</VbcToolPath>
-    <VbcToolExe>vbc.exe</VbcToolExe>
-  </PropertyGroup>
-
 </Project>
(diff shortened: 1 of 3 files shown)


tannergooding:
I've tested the packages locally against CoreCLR, but it's possible that other changes may be required after this goes in and pumps through.

src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.cmd
```diff
@@ -106,9 +105,6 @@ if not [%RESTORE_PORTABLETARGETS_ERROR_LEVEL%]==[0] (
 )
 Robocopy "%PACKAGES_DIR%\MicroBuild.Core\%MICROBUILD_VERSION%\build\." "%TOOLRUNTIME_DIR%\." /E
 
-:: Copy Roslyn Compilers Over to ToolRuntime
-Robocopy "%PACKAGES_DIR%\Microsoft.Net.Compilers\%ROSLYNCOMPILERS_VERSION%\." "%TOOLRUNTIME_DIR%\net46\roslyn\." /E
```
tannergooding: It didn't look like anything directly depends on this existing anymore and we can just reference the bits directly from the package; much like we were already doing for RunningOnCore==true

src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props
```diff
@@ -1,37 +1,26 @@
...
+    <DebugType Condition="'$(RunningOnCore)' == 'true'">Portable</DebugType>
   </PropertyGroup>
 
-
   <!-- If we're not using the compiler server, set ToolPath/Exe to direct to
        the exes in this package -->
   <PropertyGroup Condition="'$(UseRoslynCompilers)' != 'false' and '$(RoslynIncompatibleMsbuildVersion)' == 'true'">
-    <CscToolPath Condition="'$(RunningOnCore)' == 'true'">$(RoslynPackageDir)tools</CscToolPath>
-    <CscToolPath Condition="'$(RunningOnCore)' != 'true'">$(BuildToolsTaskDir)roslyn/tools</CscToolPath>
+    <CscToolPath>$(_RoslynTasksDirectory)</CscToolPath>
```
tannergooding: I actually couldn't find anything that depends on this, so maybe we can just remove it (especially given we don't look to have an equivalent in Arcade)?

jaredpar: Remove this entire `PropertyGroup`. It's just referring to items we don't need anymore.

tannergooding: Done. Will leave the thread as "unresolved" for visibility, however.

PR #2251, opened 2019-05-15
Python 3 changes for scriptrunner

ulisesh:
Use `open` instead of `file` to read the results
Remove unnecessary semicolons


ulisesh:
cc @MattGal

Issue #2221, opened 2019-02-06
Fix GitHubVersionsRepoUpdater retries for "Update is not a fast forward"


dagood:
`GitHubVersionsRepoUpdater.UpdateBuildInfoAsync` *should* retry when it hits "Update is not a fast forward". That exception simply means some other repo updated dotnet/versions between the beginning of the current update and submission. But it looks like there’s a bug where [GitHubVersionsRepoUpdater.cs#L149](https://github.com/dotnet/buildtools/blob/e4fcde0f2de731312ad10f93428d0c8007e858f6/src/Microsoft.DotNet.VersionTools/Automation/GitHubVersionsRepoUpdater.cs#L149) doesn't catch [`NotFastForwardUpdateException`](https://github.com/dotnet/buildtools/blob/e4fcde0f2de731312ad10f93428d0c8007e858f6/src/Microsoft.DotNet.VersionTools/Automation/GitHubApi/NotFastForwardUpdateException.cs).

This is ironic aftermath of the ProdCon v1 work: I added this exception type so I could add tests that make sure ProdCon v1's build manifest publish will do retries.

ViktorHofer:
// Auto-generated message

Going forward, the .NET team is using https://github.com/dotnet/arcade to develop the code and issues formerly in this repository. Feel free to reopen/move this issue if it still applies to servicing branches in this repository, or source code which now resides in the Arcade repository.

PR #2265, opened 2019-10-23
Retry updating build info on NotFastForwardUpdateException

dibarbet:
See https://github.com/dotnet/buildtools/issues/2221
We hit this issue semi-regularly in our CI pipeline
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3173501

### DIFF
--- a/src/Microsoft.DotNet.VersionTools/Automation/GitHubVersionsRepoUpdater.cs
+++ b/src/Microsoft.DotNet.VersionTools/Automation/GitHubVersionsRepoUpdater.cs
@@ -148,20 +148,38 @@ namespace Microsoft.DotNet.VersionTools.Automation
                     }
                     catch (HttpRequestException ex)
                     {
-                        int nextTry = i + 1;
-                        if (nextTry < MaxTries)
+                        bool retrySucceeded = await TryRetry(i, ex.Message);
+                        if (!retrySucceeded)
                         {
-                            Trace.TraceInformation($"Encountered exception committing build-info update: {ex.Message}");
-                            Trace.TraceInformation($"Trying again in {RetryMillisecondsDelay}ms. {MaxTries - nextTry} tries left.");
-                            await Task.Delay(RetryMillisecondsDelay);
+                            throw;
                         }
-                        else
+                    }
+                    catch (NotFastForwardUpdateException ex)
+                    {
+                        bool retrySucceeded = await TryRetry(i, ex.Message);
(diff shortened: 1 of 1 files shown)


dibarbet:
@dagood Do you have a good way of running the tool locally to test?  I couldn't find anything obvious

dagood:
Well, this is the wrong repo. 😛 This lib is now built over in dotnet/arcade: https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.VersionTools/lib

I generally run through changes with the MSBuild task, [`UpdatePublishedVersions`](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.VersionTools/tasks/src/UpdatePublishedVersions.cs), pointing at my own fork of dotnet/versions and my bot account (to reduce permission set).

I use Fiddler to test things like this where I need to precisely control the order of HTTP calls. (Run the task twice at the same time and manually let HTTP requests through in a way that triggers the problem.)

You could write unit tests by mocking out the `IGitHubClient` interface and adding a place to inject your own client, but I think manual testing is fine to get this merged.

(Link to Roslyn issue for reference: https://github.com/dotnet/roslyn/issues/39308.)

src/Microsoft.DotNet.VersionTools/Automation/GitHubVersionsRepoUpdater.cs
```diff
@@ -148,21 +148,39 @@ public GitHubVersionsRepoUpdater(GitHubAuth gitHubAuth, GitHubProject project)
                     }
                     catch (HttpRequestException ex)
                     {
-                        int nextTry = i + 1;
-                        if (nextTry < MaxTries)
+                        bool retrySucceeded = await TryRetry(i, ex.Message);
```
dagood: This method and the variable are really confusing... after reading it all I can see what it's doing, but these names don't describe it with it as far as I can tell. I see it's reducing duplication--I suggest replacing this loop with this util that handles it to clean it all up:

https://github.com/dotnet/arcade/blob/master/src/Common/ExponentialRetry.cs

(Nit: The method is also missing `Async` suffix.)

dibarbet:
> Well, this is the wrong repo. 😛 This lib is now built over in dotnet/arcade: https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.VersionTools/lib

Ahh!  I'll close this and open a new one to arcade.

PR #2279, opened 2021-01-14
Cleanup feeds

garath:
Cleanup nuget feeds.

To handle bootstrap problems, this also makes potentially temporary changes of:

- Bootstrap from self instead of going to GitHub (to get script with fixed URLs)
- Replace BuildTools dependency with experimental version having updated feed URLs


mmitche (approved):

wtgodbe:
Can we merge this now?

PR #2280, opened 2021-01-14
Cleanup feeds

garath:
Cleanup nuget feeds.

To handle bootstrap problems, this also makes potentially temporary changes of:

- Bootstrap from self instead of going to GitHub (to get script with fixed URLs)
- Replace BuildTools dependency with experimental version having updated feed URLs

### DIFF
--- a/bootstrap/bootstrap.ps1
+++ b/bootstrap/bootstrap.ps1
@@ -92,15 +92,15 @@ $pjContent = $pjContent + "}, `"frameworks`": { `"netcoreapp1.0`": { } } }"
 $pjContent | Out-File $projectJson
 
 # now restore the packages
-$buildToolsSource = "https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json"
-$nugetOrgSource = "https://api.nuget.org/v3/index.json"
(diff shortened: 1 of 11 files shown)


bootstrap/bootstrap.ps1
```diff
@@ -92,15 +92,16 @@ $pjContent = $pjContent + "}, `"frameworks`": { `"netcoreapp1.0`": { } } }"
 $pjContent | Out-File $projectJson
 
 # now restore the packages
-$buildToolsSource = "https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json"
-$nugetOrgSource = "https://api.nuget.org/v3/index.json"
+$buildToolsSource = "https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json"
+$dotnetPublicSource = "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json"
+$tempBuildToolsSource = "https://pkgs.dev.azure.com/dnceng/public/_packaging/general-testing/nuget/v3/index.json"
```
hoyosjs, line 97: Why do we need general testing for an asset used in product construction?

hoyosjs (approved):

